Issue #245, opened 2021-11-02
Oneline'r install script


flipchan:
substrate has a really easy to install process thanks to getsubstrate.io(```sh curl https://getsubstrate.io -sSf | bash -s -- --fast ```)

Would be cool to have a similar install script so that a user can get the dependencies and auto compile and install the node

What's needed:
Multiplatform install script
subdomain to host this script so we can do it all in a one liner thanks to curl and shell.

flipchan:
issue can be closed when #347 is merged

flipchan:
done

PR #246, opened 2021-11-02
install script #245

flipchan:
install script for issue #245


Br1ght0ne (dismissed):

scripts/install_composable.sh
```diff
@@ -0,0 +1,66 @@
...
+	apt update && apt install -y git clang curl libssl-dev llvm libudev-dev
+# Fedora
+elif echo "$myos" | fgrep -q Fedora 2>/dev/null; then
+	dnf update && dnf install clang curl git openssl-devel
+# Arch
+elif echo "$myos" | fgrep -q Arch 2>/dev/null; then
+	pacman -Syu --needed --noconfirm curl git clang
+
+# OpenBSD
+#$ pkg_info -Q rust
```
dzmitry-lahoda, line 20: what is these commented lines about? are these untested or optional? may be just remove it?

scripts/install_composable.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/sh
+echo "Composable Multiplatform installer"
+myos=$(uname -a)
+echo $myos
+
+# Ubuntu:
+if echo "$myos" | fgrep -q Ubuntu 2>/dev/null ; then
```
dzmitry-lahoda, line 7: not sure. is deploy to parachain nodes happens via docker or via nix?  would each developer manually handle deps?

scripts/install_composable.sh
```diff
@@ -0,0 +1,66 @@
...
+else
+	echo "Invalid operating system detected, do you want to proceed with the installation? yes/no"
+	read response
+	if "no" == $response; then
+		echo "Goodbye"
+		exit 1
+
+fi
+
+echo "Getting the latest version of Rust"
```
dzmitry-lahoda, line 50: also risky step, until we agree on some  rust-toolchain.toml we will not be able to handle that well for all

flipchan:
testing the script on debian and bsd, will see how it goes

flipchan:
working on debian!

scripts/install_composable.sh
flipchan: removed them, those were the supported rust packages in openbsd

scripts/install_composable.sh
flipchan: well yeah, there is a docker compose file and nix build, The goal of this script is that a new user is suppose to be able to just spin up a virtual machine(or rent a vps) and be able to install the node with a one liner, so the finished product will look like: curl -sSH get.composable.finance | sh, then that will kick in this script

scripts/install_composable.sh
```diff
@@ -0,0 +1,66 @@
...
+# Fedora
+elif echo "$myos" | fgrep -q Fedora 2>/dev/null; then
+	dnf update && dnf install clang curl git openssl-devel
+# Arch
+elif echo "$myos" | fgrep -q Arch 2>/dev/null; then
+	pacman -Syu --needed --noconfirm curl git clang
+
+# Nix Os
+elif echo "$myos" | fgrep -q NixOS 2>/dev/null; then
+	nix-env -iA nixpkgs.{git,rustup}
```
flipchan, line 21: @Br1ght0ne check this Nix OS

Br1ght0ne: Looks legit

flipchan:
works on MacOS

flipchan:
left to check:
Arch, Fedora and FreeBSD

flipchan:
running it on a fedora vm to test it out

flipchan:
fedora is returning .fc32. and .fc64. and not Fedora with uname -a

flipchan:
working on Fedora 32

scripts/install_composable.sh
```diff
@@ -0,0 +1,67 @@
...
+	source ~/.cargo/env
+else:
+	rustup update
+fi # new
+
+rustup default stable
+rustup update nightly
+rustup target add wasm32-unknown-unknown --toolchain nightly
+echo "Installing composable node"
+#git clone https://github.com/composablefi/composable
```
dzmitry-lahoda, line 63: what this commented out code mean? can we clean it up or write todo if some automation is missing?

scripts/install_composable.sh
dzmitry-lahoda: hm, so instead of installing nix/docker and than run from that, he runs from script. 
additionally we have polkadot launcher to run on dev.

to be able to be our parachain, you have to be pemissioned afaik, so user will not be able just plug his stuff into our consensus? (until it is collator? collators - the more the better).

would you be able to describe target audience of this script?

i will try to run it on my wsl2(sorry, i am windows person) if that works for me later:)

scripts/install_composable.sh
flipchan: Was commented out when I was debugging the script, fixed now, good catch

scripts/install_composable.sh
flipchan: @dzmitry-lahoda, sure thing, so the inspiration comes from how substrate has a oneliner to install and build it(curl -sSH https://getsubstrate.io/ | sh), would be cool to have a one liner a user can use to install and compile the node in order to make it easy for future node operators

dzmitry-lahoda (approved):
i am ok with that, so we do not have yet  way to make sure we use same rust accross all tools (unfortunately cannot add toolchain toml (or i do not know how) , because we use different toolchains for different things (so will need to choose one toolchain into toml and hope to override all other places different  toolchain  used)

Br1ght0ne (approved):

Br1ght0ne (approved):

dzmitry-lahoda (approved):

dzmitry-lahoda (approved):

flipchan:
should be able to mege in a couple of minutes when CI passes

PR #259, opened 2021-11-05
[CU-1rpc2zt] pallet-assets

Br1ght0ne:
[`mint_initialize`](https://github.com/ComposableFi/composable/pull/259/files#diff-c7df28419719d12448616196b77e46d89b69a2caf20ee48ac86e0abccd1b8b17R132-R143) and [`mint_initialize_with_governance`](https://github.com/ComposableFi/composable/pull/259/files#diff-c7df28419719d12448616196b77e46d89b69a2caf20ee48ac86e0abccd1b8b17R145-R164) tests could use a second look.


frame/assets/src/lib.rs
```diff
@@ -179,6 +202,88 @@ pub mod pallet {
...
+		pub fn mint_initialize_with_governance(
+			origin: OriginFor<T>,
+			amount: T::Balance,
+			governance_origin: T::AccountId,
+			dest: T::AccountId,
+		) -> DispatchResultWithPostInfo {
+			ensure_root(origin)?;
+			let id = T::GenerateCurrencyId::create()?;
+			T::GovernanceRegistry::set(id, RawOrigin::Signed(governance_origin))
+				.map_err(|_| Error::<T>::CannotSet)?;
```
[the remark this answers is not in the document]
dzmitry-lahoda: I think that CannotSetNewCurrencyToRegistry or more elaborated error message could be better

dzmitry-lahoda (approved):

dzmitry-lahoda:
i need this for XCMP, multicurrency impl. so that instead of implementing this , just use this pallet

dzmitry-lahoda:
```
	#[pallet::weight(T::WeightInfo::set())]
		pub fn set(
			origin: OriginFor<T>,
			asset: T::AssetId,
			value: T::AccountId,
			ensure_root(origin)?;
			OriginsByAssetId::<T>::insert(asset.clone(), StorageOrigin::Signed(value.clone()));
```
not clear semantics, it check root signature but puts `Signed` into storage without any check of signature. So may be rename to something not related to signature.

dzmitry-lahoda:
Could we check origin of Root or governance?

```
/// Origin to add new price types
		type AddOracle: EnsureOrigin<Self::Origin>
````

type AddOracle = EnsureRootOrHalfCouncil;

Seems operations like these are by non root? @KaiserKarel  ?

**Comment by dzmitry-lahoda** [comment]
my sorrow story is next:
1. i am unable build xcmp on develop because of weird dependency issue (not sure that even runtime is built)
2. all is fin on main where I have simulator of xcmp built and run
3. i need multicurrency from this trait
4. develop is not up to date with main and it is not clear when develop parts will be merged to main?

so could I just take parts of this and PR into main? should try to fix develop?

so I propose next flow

1. runtime is intact until pallet is registered in runtime. 
2. or there is pallet modification on which existing pallets depend.
3. for one - code, but not register
3.1 if need to register - register behind feature = "develop" via procedural macro block of cfg_if
4. for changes to existing pallets, make versioning to changes, as any way need to do that as we have runtime running. 
4.1 make separate PR which hadles v1 and v2 at same time
5. make develop branch build as build for master

cc @KaiserKarel @hussein-aitlahcen @andor0 what do you think?

for now I do not see sane policy for develop and main, for me description i did is sane. 

auditors would safely ignore develop code by rules above. i will raise this topic on monday

File: frame/governance-registry/src/lib.rs [hunk 2]
```diff
@@ -0,0 +1,161 @@
...
+	impl<T: Config> Pallet<T> {
+		/// Sets the value of an asset to the signed account id. Only callable by root.
+		#[pallet::weight(T::WeightInfo::set())]
+		pub fn set(
+			origin: OriginFor<T>,
+			asset: T::AssetId,
+			value: T::AccountId,
+		) -> DispatchResultWithPostInfo {
+			ensure_root(origin)?;
+			OriginsByAssetId::<T>::insert(asset.clone(), StorageOrigin::Signed(value.clone()));
```
dzmitry-lahoda, line 70: Approved

frame/governance-registry/src/lib.rs
```diff
@@ -0,0 +1,161 @@
...
+		pub fn grant_root(origin: OriginFor<T>, asset: T::AssetId) -> DispatchResultWithPostInfo {
+			ensure_root(origin)?;
+			OriginsByAssetId::<T>::insert(asset.clone(), StorageOrigin::Root);
+			Self::deposit_event(Event::<T>::GrantRoot { asset });
+			Ok(().into())
+		}
+
+		/// Removes mapping of an asset_id. Only callable by root.
+		#[pallet::weight(T::WeightInfo::remove())]
+		pub fn remove(origin: OriginFor<T>, asset: T::AssetId) -> DispatchResultWithPostInfo {
```
dzmitry-lahoda, line 86: we have some build in ids (those in CurrencyId). What would it mean if I pass some id which is build in into code here?

KaiserKarel: It removes a governance mapping, which is later used in pallet-democracy.

KaiserKarel:
> ```
> 	#[pallet::weight(T::WeightInfo::set())]
> 		pub fn set(
> 			origin: OriginFor<T>,
> 			asset: T::AssetId,
> 			value: T::AccountId,
> 		) -> DispatchResultWithPostInfo {
> 			ensure_root(origin)?;
> 			OriginsByAssetId::<T>::insert(asset.clone(), StorageOrigin::Signed(value.clone()));
> ```
> 
> not clear semantics, it check root signature but puts `Signed` into storage without any check of signature. So may be rename to something not related to signature.

I'll add a comment explaining GovernanceRegistry a bit more; basically the idea is that a pallet can register it's PalletId.intoi_account(), and then can check it's extrinsics with `ensure_root_or_governance` to allow for token specific governance.

frame/governance-registry/src/lib.rs
dzmitry-lahoda: ok, so we have some build in assets, with 1,2,3,4.. small numbers as const in Rust codes. Will governance be on these either?

dzmitry-lahoda:
>  ensure_root_or_governance to allow for token specific governance.

> If Angular wants pool-specific governance, it registers PalletId("angular").into_sub_account(pool_id), and can then check for that origin. The democracy pallet dispatches by that origin,

dzmitry-lahoda:
i do not see calls to `ensure_admin_or_governance` from `assets` pallet. So it can operate on any tokens regardless of governance, an it is up to some pallet protocol to decide if it wants to check for `ensure_admin_or_governance` or not?

itsbobbyzzz:
Task linked: [CU-1rpc2zt Pallet-Assets](https://app.clickup.com/t/1rpc2zt)

frame/assets/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "pallet-assets"
 version = "0.0.1"
 authors = ["Composable Developers"]
 homepage = "https://composable.finance"
-edition = "2018"
+edition = "2021"
```
andor0, line 6: can add `rust-version = "1.56"` below this.

frame/assets/Cargo.toml
```diff
@@ -29,12 +29,18 @@ features = ["derive"]
 package = "parity-scale-codec"
 version = "2.0.0"
 
+[dev-dependencies]
+governance-registry = { version = "0.0.1", path = "../governance-registry", default-features = false }
+pallet-balances = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.11" }
```
andor0, line 34: we switched to v0.9.12, need to use `branch = "polkadot-v0.9.12"`

frame/assets/Cargo.toml
```diff
@@ -29,12 +29,18 @@ features = ["derive"]
...
+orml-tokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library", rev = "b89b1f98fcf0f5c341fbef1d12bfd869a2f2a9fe" }
```
andor0, line 35: we switched to v0.9.12, need to use `rev = "ee37b76a7a8cf159e6c41851942758746318bb84"`

frame/governance-registry/src/lib.rs
KaiserKarel: Pica will be mapped to Root, rest can be set, but most likely not.

frame/assets/Cargo.toml
dzmitry-lahoda: another reason for not using branch:)

andor0 (approved):

runtime/picasso/Cargo.toml
```diff
@@ -90,84 +92,9 @@ pallet-xcm = { git = "https://github.com/paritytech/polkadot", branch = "release
...
-	"democracy/runtime-benchmarks",
-	"utility/runtime-benchmarks",
-]
\ No newline at end of file
+std = ["codec/std", "sp-api/std", "sp-std/std", "sp-core/std", "sp-runtime/std", "sp-version/std", "sp-offchain/std", "sp-session/std", "sp-block-builder/std", "sp-transaction-pool/std", "sp-inherents/std", "support/std", "executive/std", "frame-system/std", "utility/std", "authorship/std", "balances/std", "randomness-collective-flip/std", "crowdloan-bonus/std", "timestamp/std", "session/std", "sudo/std", "indices/std", "oracle/std", "vault/std", "call-filter/std", "currency-factory/std", "orml-tokens/std", "orml-traits/std", "treasury/std", "democracy/std", "scheduler/std", "common/std", "primitives/std", "collective/std", "transaction-payment/std", "parachain-info/std", "cumulus-pallet-aura-ext/std", "cumulus-pallet-parachain-system/std", "cumulus-pallet-xcmp-queue/std", "cumulus-pallet-xcm/std", "cumulus-primitives-core/std", "cumulus-primitives-timestamp/std", "cumulus-primitives-utility/std", "collator-selection/std", "xcm/std", "xcm-builder/std", "xcm-executor/std", "aura/std", "sp-consensus-aura/std", "scale-info/std", "support/std", "governance-registry/std"]
```
dzmitry-lahoda, line 99: wow. what is it. how it formatted so bad...

dzmitry-lahoda:
seems like it is SPL token in solana. Mints with ability to delegate minting etc. So it seems has some bugs.

Br1ght0ne:
Closed in favor of #293

PR #262, opened 2021-11-08
Feature/mosaic vault v2

ysfkel:


frame/mosaic-vault/Cargo.toml
```diff
@@ -0,0 +1,64 @@
...
+
+[dependencies.codec]
+default-features = false
+features = ["derive"]
+package = "parity-scale-codec"
+version = "2.0.0"
+
+
+[dependencies]
+frame-support = { default-features = false, git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.11"  }
```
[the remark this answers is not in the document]
ysfkel: removed

frame/mosaic-vault/src/tests.rs
```diff
@@ -0,0 +1,106 @@
...
+	{
+		// System: frame_system::{Module, Call, Config, Storage, Event<T>},
+		// TemplateModule: pallet_template::{Module, Call, Storage, Event<T>},
+        Tokens: orml_tokens::{Pallet, Storage, Event<T>, Config<T>},
+        System: system::{ Pallet, Call, Storage, Config, Event<T>},
+       // MosaicVault: mosaic_vault::{ Pallet, Call, Storage, Event<T> }
+	}
+);
+
+// #[test]
```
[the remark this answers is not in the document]
ysfkel: removed

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,915 @@
...
+	#[pallet::storage]
+	#[pallet::getter(fn pause_status)]
+	pub(super) type PauseStatus<T :Config> = StorageValue<_, bool, ValueQuery>;
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+ 	pub enum Event<T: Config> {
+
+		DepositCompleted(
+   			T::AccountId, // sender
```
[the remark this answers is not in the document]
ysfkel: resolved

dzmitry-lahoda:
@ysfkel please tell when we can review again? thanks

dzmitry-lahoda:
it would be nice have rust doc comment on pallet or in readmy some docs, may be reference to clickup docs

hussein-aitlahcen:
@ysfkel any chance we rebase this branch on latest main and try to make it pass the CI :)?

frame/mosaic-vault/src/mocks/runtime.rs
```diff
@@ -0,0 +1,258 @@
...
+	type OnDust = ();
+	type MaxLocks = ();
+	type DustRemovalWhitelist = Everything;
+}
+
+parameter_types! {
+	pub const ExistentialDeposit: u64 = 1000;
+}
+
+// impl pallet_balances::Config for Test {
```
dzmitry-lahoda, line 103: please remove comment or comment why it is commented

frame/mosaic-vault/src/mocks/runtime.rs
```diff
@@ -0,0 +1,258 @@
...
+			.unwrap();
+
+            let mut ext = sp_io::TestExternalities::new(t);
+            ext.execute_with(|| System::set_block_number(1));
+            ext
+
+	}
+}
+
+// pub fn new_test_ext() -> sp_io::TestExternalities {
```
dzmitry-lahoda, line 232: please remove comment or comment why it is commented

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
```
dzmitry-lahoda, line 2: please add //! comment to document this pallet

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+		   Ok(().into())
+		 }
+
+		 #[pallet::weight(10_000)]
+		 pub fn remove_supported_token(origin: OriginFor<T>, asset_id: T::AssetId, remote_network_id: T::RemoteNetworkId) -> DispatchResultWithPostInfo {
+
+			T::AdminOrigin::ensure_origin(origin)?; 
+
+			Self::only_supported_remote_token(remote_network_id.clone(), asset_id.clone())?;
```
dzmitry-lahoda, line 451: i doubt this check is needed here

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+	}
+	#[derive(Encode, Decode, Default, Debug, PartialEq, TypeInfo)]
+	pub struct DepositInfo<AssetId, Balance > {
+        pub asset_id: AssetId,
+		pub amount: Balance,
+	}  
+	
+	#[pallet::storage]
+	#[pallet::getter(fn remote_asset_id)]
+    pub(super) type RemoteAssetId<T: Config> = StorageDoubleMap<_, Blake2_128Concat, T::RemoteNetworkId, Blake2_128Concat, T::AssetId, T::RemoteAssetId, ValueQuery>;
```
dzmitry-lahoda, line 153: may be use OptionQuery? so than option can be get and no need to double call storage to check if asset exists

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+		 }
+
+		 #[pallet::weight(10_000)]
+		 pub fn remove_supported_token(origin: OriginFor<T>, asset_id: T::AssetId, remote_network_id: T::RemoteNetworkId) -> DispatchResultWithPostInfo {
+
+			T::AdminOrigin::ensure_origin(origin)?; 
+
+			Self::only_supported_remote_token(remote_network_id.clone(), asset_id.clone())?;
+          
+ 		    let remote_asset_id = RemoteAssetId::<T>::get(remote_network_id, asset_id);
```
dzmitry-lahoda, line 453: if let Some(remote_asset_id) = ... {
    ...
    .. deposit event
}

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+
+			T::AdminOrigin::ensure_origin(origin.clone())?;
+			
+			let sender = ensure_signed(origin)?;
+
+			 let old_lockup_time = <TransferLockupTime<T>>::get();
+
+			 <TransferLockupTime<T>>::put(lockup_time);
+
+			 let action = "Transfer".as_bytes().to_vec();
```
dzmitry-lahoda, line 501: b"Tranfer".to_vec()

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+
+			 Self::deposit_event(Event::LockupTimeChanged{sender, old_lockup_time, lockup_time, action});
```
dzmitry-lahoda, line 503: not sure, why would you pass this string? I would prever `TransferLockupTimeChanged` event

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+		 }
+
+		 #[pallet::weight(10_000)]
+		 pub fn set_transfer_lockup_time(origin: OriginFor<T>, lockup_time: Timestamp) -> DispatchResultWithPostInfo {
+
+			T::AdminOrigin::ensure_origin(origin.clone())?;
+			
+			let sender = ensure_signed(origin)?;
+
+			 let old_lockup_time = <TransferLockupTime<T>>::get();
```
dzmitry-lahoda, line 497: what Lockup means? we have one LockupTime for all assets? networks are different in timing. should we have time per asset?

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+
+            <MinFee<T>>::put(min_fee);
+
+			Self::deposit_event(Event::MinFeeChanged{min_fee});
+
+			Ok(().into())
+		 }
+
+		 #[pallet::weight(10_000)]
+		 pub fn set_thresh_hold(origin: OriginFor<T>, new_fee_threshold: T::Balance) -> DispatchResultWithPostInfo {
```
dzmitry-lahoda, line 569: set_fee_treshhold

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+		
+			ensure!(transfer_delay <= <MaxTransferDelay<T>>::get(), Error::<T>::TransferDelayAboveMaximum);
+
+			ensure!(amount <= Self::max_asset_transfer_size(asset_id), Error::<T>::AmountAboveMaxAssetTransferSize);
+
+			ensure!(amount >= Self::min_asset_transfer_size(asset_id), Error::<T>::AmountBelowMinAssetTransferSize);
+			// 
+			let pallet_account_id = Self::account_id();            
+            // move funds to pallet amount
+		     T::Currency::burn_from(asset_id, &sender, amount).map_err(|_|Error::<T>::BurnFromFailed)?;
```
dzmitry-lahoda, line 612: move mutating operations to the bottom, cause these can be partially applied and fail other stuff. so check first and mutate later. not sure, could make this `trnsactional` attribute as other option.

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+
+			Self::deposit_event(Event::DepositCompleted{
+				sender,
+				asset_id,
+				remote_asset_id: Self::remote_asset_id(remote_network_id, asset_id),
+				remote_network_id,
+				destination_address,
+				amount, 
+				deposit_id,
+				transfer_delay
```
dzmitry-lahoda, line 634: used only in errors and event, which is informational only. either remove or use it. or commetn why not used

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+			});
+
+			Ok(().into())
+		 }
+
+		 #[pallet::weight(10_000)]
+		 pub fn withdraw(
```
dzmitry-lahoda, line 641: probably this should be transactional

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+
+			  <HasBeenWithdrawn<T>>::insert(deposit_id, true);
+
+			  <LastWithdrawID<T>>::put(deposit_id);
+
+			  let pallet_account_id = Self::account_id(); 
+	
+			  let withdraw_amount = amount.saturating_sub(fee);
+
+			 T::Currency::mint_into(asset_id, &pallet_account_id, amount).map_err(|_|Error::<T>::MintToFailed)?;
```
dzmitry-lahoda, line 671: mint fee only

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+
+			 Self::decrease_total_value_transferred(asset_id, amount)?;
+
+			  T::Currency::transfer(asset_id, &pallet_account_id, &destination_account, withdraw_amount, true).map_err(|_|Error::<T>::TransferFromFailed)?;
```
dzmitry-lahoda, line 675: mint here

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+			
+			 if fee > T::Balance::zero() {  
+			   
+				 T::Currency::transfer(asset_id, &pallet_account_id, &Self::get_fee_address(), fee, true).map_err(|_|Error::<T>::TransferFromFailed)?;
```
dzmitry-lahoda, line 679: so, probably never should mint into pallet

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+			let mut encoded_nonce = Encode::encode(&Self::increment_nonce());
+
+			let mut encoded_data = Vec::new();
+			encoded_data.append(& mut encoded_remote_network_id);
+			encoded_data.append(& mut encoded_block_number);
+			encoded_data.append(& mut encoded_destination_address);
+			encoded_data.append(& mut encoded_pallet_account_id);
+			encoded_data.append(& mut encoded_nonce);
+
+			let deposit_id = keccak_256(&encoded_data);
```
dzmitry-lahoda, line 879: vec![
    remote_network_id.encode(),
    ..
    Self::increment_nonce().encode(),
]

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+
+			// update in_transfer_funds
+			let in_transfer_funds = Self::in_transfer_funds(asset_id);
+			let new_in_transfer_funds = in_transfer_funds.checked_add(&amount).ok_or(Error::<T>::Overflow)?;
+			<InTransferFunds<T>>::insert(asset_id, new_in_transfer_funds);
+           
+			<LastTransfer<T>>::insert(&sender, T::BlockTimestamp::now().as_secs());
+
+			let deposit_id = Self::generate_deposit_id(remote_network_id, &destination_address, pallet_account_id);
+            <Deposits<T>>::insert(asset_id, DepositInfo{asset_id, amount});
```
dzmitry-lahoda, line 624: not sure, did you mean deposit_id ?

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+
+			T::RelayerOrigin::ensure_origin(origin)?;
+
+			ensure!(Self::pause_status() == false, Error::<T>::ContractPaused);
+			
+			ensure!(Self::has_been_completed(deposit_id) == false, Error::<T>::AlreadCompleted);
+
+			ensure!(Self::in_transfer_funds(asset_id) >= amount, Error::<T>::InsufficientFunds);
+
+			let deposit = Self::deposits(asset_id);
```
dzmitry-lahoda, line 719: may be deposit_id?

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+
+			ensure!(deposit.asset_id == asset_id && deposit.amount == amount, Error::<T>::InsufficientFunds);
```
dzmitry-lahoda, line 721: not sure about deposit. is it share for asset? if not, and created each time as its id, than why to pass amount? anyway check for ==

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+				fee,
+				asset_id,
+				deposit_id
+			 });
+
+			 Ok(().into())
+		 }
+
+		 #[pallet::weight(10_000)]
+		 pub fn unlock_in_transfer_funds(
```
dzmitry-lahoda, line 704: please use consistent naming, either in_transfer or intransfer everywhere

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+            
+			 T::RelayerOrigin::ensure_origin(origin.clone())?;
+          
+			 ensure!(Self::has_been_unlocked(deposit_id) == false, Error::<T>::AssetUnlreadyUnlocked);
+
+			 ensure!(Self::total_value_transferred(asset_id) >= amount, Error::<T>::UnlockAmountGreaterThanTotalValueTransferred);
+
+			 <HasBeenUnlocked<T>>::insert(deposit_id, true);
+
+			 <LastUnlockedID<T>>::put(deposit_id);
```
dzmitry-lahoda, line 751: why this is needed?

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+
+			 Self::only_supported_remote_token(remote_network_id.clone(), asset_id.clone())?;
+             
+			ensure!(Self::has_been_withdrawn(deposit_id) == false, Error::<T>::AlreadyWithdrawn);
+
+			ensure!(Self::get_current_token_liquidity(asset_id)? >= amount, Error::<T>::InsufficientAssetBalance); 
+
+			  <HasBeenWithdrawn<T>>::insert(deposit_id, true);
+
+			  <LastWithdrawID<T>>::put(deposit_id);
```
dzmitry-lahoda, line 665: may this should be per network or relayer? so relayer knows its last?

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+			origin: OriginFor<T>,
+			asset_id: T::AssetId,
+			user_account_id: T::AccountId,
+			amount: T::Balance,
+			deposit_id: T::DepositId,
+		 ) ->DispatchResultWithPostInfo {
+            
+			 T::RelayerOrigin::ensure_origin(origin.clone())?;
+          
+			 ensure!(Self::has_been_unlocked(deposit_id) == false, Error::<T>::AssetUnlreadyUnlocked);
```
dzmitry-lahoda, line 745: you should not check for being unlocked, but remove unlocked from state i think. yep, it will be not clear from this call if funds ever been locked or unlocked. but that is performance.

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+
+			 ensure!(Self::total_value_transferred(asset_id) >= amount, Error::<T>::UnlockAmountGreaterThanTotalValueTransferred);
+
+			 <HasBeenUnlocked<T>>::insert(deposit_id, true);
```
dzmitry-lahoda, line 749: remove instead of setting true

frame/mosaic-vault/src/lib.rs
```diff
@@ -0,0 +1,904 @@
...
+		   Self::deposit_event(Event::TransferFundsUnlocked{asset_id, amount, deposit_id});
+	
+			Ok(().into())
+		 }
+
+		 #[pallet::weight(10_000)]
+		 pub fn unlock_funds(
+			origin: OriginFor<T>,
+			asset_id: T::AssetId,
+			user_account_id: T::AccountId,
```
dzmitry-lahoda, line 738: it it same desitnation account which is part of deposit_id? should that be validate

dzmitry-lahoda:
@ysfkel did you ported solidity code? please reference it? did that code contained comments? could we port these either

frame/mosaic-vault/src/mocks/runtime.rs
ysfkel: removed

frame/mosaic-vault/src/lib.rs
ysfkel: removed

frame/mosaic-vault/src/mocks/runtime.rs
ysfkel: removed

frame/mosaic-vault/src/lib.rs
ysfkel: resolved

frame/mosaic-vault/src/lib.rs
ysfkel: just to keep the events same as the solidity code, but what you said makes sense , ill check with them if its okay to omit the string

frame/mosaic-vault/src/lib.rs
ysfkel: The transferLockupTime is a time interval that needs to elapse before thesame address can call deposit, It is also part of the solidity contract https://github.com/ComposableFi/CrosslayerPortal/blob/fc5202334495a177509bdba01ba38bbc70b0f079/contracts/relayer/L2Vault.sol#L362

frame/mosaic-vault/src/lib.rs
ysfkel: not sure what you mean here, can you explain

frame/mosaic-vault/src/lib.rs
ysfkel: not sure what you mean here, can you explain? do you mean to mint each amount separately instead of mint and transfer ?

frame/mosaic-vault/src/lib.rs
ysfkel: thats right,  though i'm  trying to keep the events consistent with the deployed solidity contract , I guess the relayer expects these parameters, but ill ask to enure , removing will be ok

frame/mosaic-vault/src/lib.rs
ysfkel: yes

frame/mosaic-vault/src/lib.rs
ysfkel: yes

frame/mosaic-vault/src/lib.rs
ysfkel: keccak_256 wants a &[u8] , .append moves individual elements into new array, however using vet![] with encode results in Vec<Vec<u8>>

frame/mosaic-vault/src/lib.rs
ysfkel: It keeps track of deposits , once the transfer is completed by the relayer, it is flagged as complete,

frame/mosaic-vault/src/lib.rs
ysfkel: I think it is used by the relayer, it is also available on the equivalent solidity contract , I will confirm that

frame/mosaic-vault/src/lib.rs
ysfkel: It should be per network I believe

frame/mosaic-vault/src/lib.rs
ysfkel: The has_been_unlocked state is used to stop execution for the same id, so if removed, how is the check handled

dzmitry-lahoda:
what is clickup task for this work?

KaiserKarel:
Closing for now, we have another PR being worked on

PR #265, opened 2021-11-08
runtimes simulator setup for future xcm testing

dzmitry-lahoda:


andor0:
Could you run `cargo check -p picasso-runtime` to decrease changes of `Cargo.lock` file?

frame/curve-amm/src/lib.rs
```diff
@@ -227,13 +278,219 @@ pub mod pallet {
...
+					let n_coins = pool.assets.len();
+
+					ensure!(n_coins == pool.balances.len(), Error::<T>::InconsistentStorage);
+
+					ensure!(n_coins == amounts.len(), Error::<T>::IndexOutOfRange);
+					let amp_f = FixedU128::saturating_from_integer(
+						u128::try_from(pool.amplification_coefficient)
+							.ok()
+							.ok_or(Error::<T>::Math)?,
+					);
```
andor0, line 304: looks like it can rewrite so:
```
let amp_f = FixedU128::saturating_from_integer(
  u128::try_from(pool.amplification_coefficient).or(Error::<T>::Math)?,
);
```

andor0 (approved):

frame/curve-amm/src/lib.rs
dzmitry-lahoda: it seems to Vivek PR

dzmitry-lahoda:
have run `cargo check -p picasso-runtime` - seems my lock was clean enough before

hussein-aitlahcen (approved):

PR #270, opened 2021-11-09
Vault benchmarking

hussein-aitlahcen:


frame/vault/src/lib.rs
```diff
@@ -98,6 +103,97 @@ pub mod pallet {
...
+		type MinimumDeposit: Get<Self::Balance>;
+		type MinimumWithdrawal: Get<Self::Balance>;
+		type CreationDeposit: Get<Self::Balance>;
+		type ExistentialDeposit: Get<Self::Balance>;
+		type TombstoneDuration: Get<Self::BlockNumber>;
+		type RentPerBlock: Get<Self::Balance>;
+		type PalletId: Get<PalletId>;
+	}
+
+	#[cfg(not(feature = "runtime-benchmarks"))]
```
andor0, line 196: It needs for an off-chain worker only.

frame/vault/src/mocks/tests.rs
```diff
@@ -136,6 +156,33 @@ impl crate::mocks::strategy::Config for Test {
...
+	fn create_transaction<C: frame_system::offchain::AppCrypto<Self::Public, Self::Signature>>(
+		call: Call,
+		_public: <Signature as Verify>::Signer,
+		_account: AccountId,
+		nonce: u64,
+	) -> Option<(Call, <Extrinsic as ExtrinsicT>::SignaturePayload)> {
+		Some((call, (nonce, ())))
+	}
+}
+
```
andor0, line 185: It needs for an off-chain worker only.

andor0 (changes requested):

frame/vault/src/lib.rs
andor0, line 196: It needs for an off-chain worker only.

andor0 (changes requested):

andor0:
imho, tests and mocks can be without changes if remove changes related to an off-chain worker.

frame/vault/src/lib.rs
hussein-aitlahcen: I think i still need it as I need more constraint such as Into/From for some types...

andor0: Ok, but `crypto` module (109-139 lines) looks redundant.
`CreateSignedTransaction<Call<Self>>` needs for an off-chain worker only.

hussein-aitlahcen: I will remove all the tweaking done for the accountid of course.

hussein-aitlahcen:
@andor0 fixed, any idea on how to make cargofmt pass? I'm using latest rustfmt and it's clashing with the expectation here :thinking: (rustfmt 1.4.38-nightly (baba668 2021-11-03))

andor0:
> @andor0 fixed, any idea on how to make rustfmt pass? I'm using latest rustfmt and it's clashing with the expectation here thinking (rustfmt 1.4.38-nightly (baba668 2021-11-03))

Updated CI config.

frame/vault/src/lib.rs
```diff
@@ -98,6 +105,99 @@ pub mod pallet {
...
+			frame_system::offchain::AppCrypto<
+				<Sr25519Signature as Verify>::Signer,
+				Sr25519Signature,
+			> for TestAuthId
+		{
+			type RuntimeAppPublic = Public;
+			type GenericSignature = sp_core::sr25519::Signature;
+			type GenericPublic = sp_core::sr25519::Public;
+		}
+	}
```
andor0, line 141: compilation succeed without this code

hussein-aitlahcen: Thats my bad I forget it.

andor0 (approved):

PR #272, opened 2021-11-10
Update features of integration-tests

andor0:


dzmitry-lahoda (dismissed):
for some reason this pallet has benchmarks optional. weird

vivekvpandya:
So I assume issue was due to not including society pallet before. Is that correct ?

dzmitry-lahoda:
und check and test things were working. but seems bench somehow references code more directly

dzmitry-lahoda:
but, why would we make ci run longer by running bench mode on non relevant crate for that? :)

andor0:
> So I assume issue was due to not including society pallet before. Is that correct ?

Yes, that is. Needed society pallet with `runtime-benchmarks` feature.

seunlanlege (approved):

andor0:
> but, why would we make ci run longer by running bench mode on non relevant crate for that? :)

`integration-tests` crate excluded from benchmarks build.

dzmitry-lahoda (approved):

PR #273, opened 2021-11-11
Implement fees related code in curve-amm.

vivekvpandya:


frame/composable-traits/src/dex.rs
```diff
@@ -100,6 +100,8 @@ pub trait CurveAmm {
 		who: &Self::AccountId,
 		assets: Vec<Self::AssetId>,
 		amplification_coefficient: Self::Balance,
+		fee: Permill,
+		admin_fee: Permill,
```
dzmitry-lahoda, line 104: is admin fee vault fee? we should have fee afaik to earn on exchanges, but vault will dictate what to do with that extra money, not admin (like in lending)

dzmitry-lahoda: dex calls vault, not vault call dex

frame/curve-amm/src/lib.rs
```diff
@@ -350,21 +378,76 @@ pub mod pallet {
...
 					let d1_b = d1.checked_mul_int(1u64).ok_or(Error::<T>::Math)?.into();
 					let mint_amount;
+					let mut fees = vec![FixedU128::zero(); n_coins];
+					// Only account for fees if we are not the first to deposit
 					if token_supply > zero {
-						// mint_amount = token_supply * (d1 - d0) / d0
+						// Deposit x + withdraw y would chargVe about same
+						// fees as a swap. Otherwise, one could exchange w/o paying fees.
+						// And this formula leads to exactly that equality
+						// fee = pool.fee * n_coins / (4 * (n_coins - 1))
```
dzmitry-lahoda, line 389: in case of 2 coints, we get `fee / 4 `, so fee is divided by 4, is it expected?

frame/composable-traits/src/dex.rs
vivekvpandya: well currently I have not thought how to integrate dex and vault. The fee and admin_fee is added for dex to operate independently. For now it it zero in all tests. But yes I agree it may require some change when we work on vault-dex integration.

frame/curve-amm/src/lib.rs
vivekvpandya: it will be `fee / 2` right? fee * 2 / 4

andor0 (approved):

dzmitry-lahoda (dismissed):
until pallet in runtime it is ok. but probably need to change to vault usage

frame/curve-amm/src/lib.rs
```diff
@@ -600,9 +713,66 @@ pub mod pallet {
...
+							amount,
+							true,
+						)?;
+					}
+
+					for i in 0..n_coins {
+						pool.total_balances[i] = pool.total_balances[i]
+							.checked_sub(&admin_fees[i])
+							.ok_or(Error::<T>::Math)?;
+					}
```
andor0, line 762: Can `Verify first, write last` (https://docs.substrate.io/v3/runtime/storage/#verify-first-write-last) violate here? i.e. there will be a write to the database in the first, then will be an error

andor0 (approved):

dzmitry-lahoda (approved):

PR #274, opened 2021-11-11
remove incomplete, unused strategy liquidation feature from the vault

hussein-aitlahcen:


dzmitry-lahoda (approved):
ok, but so it is not value to have it dispatched as extrincic?

hussein-aitlahcen:
@dzmitry-lahoda yes it will but for the future version, the assertion is wrong and has been yield by the audit, I think it's fine not to have it for the release :)

andor0 (approved):

PR #277, opened 2021-11-12
crowdloan rewards

hussein-aitlahcen:


frame/crowdloan-rewards/src/lib.rs
```diff
@@ -0,0 +1,301 @@
...
+	#[pallet::generate_store(pub(super) trait Store)]
+	pub struct Pallet<T>(_);
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::weight(10_000)]
+		pub fn initialize(
+			origin: OriginFor<T>,
+			rewards: Vec<(RemoteAccountOf<T>, T::Balance, T::BlockNumber)>,
+		) -> DispatchResult {
```
KaiserKarel, line 136: Most likely we will need to repeatedly call this, as the rewards vec is probably too large to fit into a single extrinsic call.

hussein-aitlahcen: As @seunlanlege said, we should use the genesis block.

frame/crowdloan-rewards/src/lib.rs
```diff
@@ -0,0 +1,325 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
dzmitry-lahoda, line 1: may be add some doc about pallet, either into readme or //! module comment or into description attribute in toml, may be just link to wiki

frame/crowdloan-rewards/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+			push_nibble(b / 16);
+			push_nibble(b % 16);
+		}
+		r
+	}
+
+	pub fn ethereum_signable_message(prefix: &[u8], msg: &[u8]) -> Vec<u8> {
+		let mut l = prefix.len() + msg.len();
+		let mut rev = Vec::new();
+		while l > 0 {
```
dzmitry-lahoda, line 302: not sure, it is some eth standard? what if prefix1 + msg1 = prefix2 + msg2 possibl. is it safe? if rev is revision which is version, may be store v1, v2, as enum explicitly?

dzmitry-lahoda: does not feels as being some standard

frame/crowdloan-rewards/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+		v.extend(rev.into_iter().rev());
+		v.extend_from_slice(prefix);
+		v.extend_from_slice(msg);
+		v
+	}
+
+	pub fn ethereum_recover(
+		prefix: &[u8],
+		EcdsaSignature(sig): &EcdsaSignature,
+		what: &[u8],
```
dzmitry-lahoda, line 316: fn(prefix, message, signature) is more redable order of parameters imho. because prefix + message = versioned message

frame/crowdloan-rewards/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+		let wrapped_prefix: &[u8] = b"<Bytes>";
+		let wrapped_postfix: &[u8] = b"</Bytes>";
+		let mut msg = wrapped_prefix.to_vec();
+		msg.append(&mut prefix.to_vec());
+		msg.append(&mut reward_account.using_encoded(to_ascii_hex));
+		msg.append(&mut wrapped_postfix.to_vec());
+		proof.verify(&msg[..], &relay_account.into())
+	}
+
+	pub fn to_ascii_hex(data: &[u8]) -> Vec<u8> {
```
dzmitry-lahoda, line 289: can we avoid writing bit code by using some crates https://docs.rs/hex/0.4.3/hex/ ?

dzmitry-lahoda: if that suits

frame/crowdloan-rewards/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+		}
+
+		#[pallet::weight(10_000)]
+		pub fn claim_from_ethereum(
+			origin: OriginFor<T>,
+			reward_account: T::AccountId,
+			ethereum_address: EthereumAddress,
+			proof: EcdsaSignature,
+		) -> DispatchResult {
+			ensure_signed(origin)?;
```
dzmitry-lahoda, line 186: why would you want this to be signed?

frame/crowdloan-rewards/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+			let data = reward_account.using_encoded(to_ascii_hex);
+			let signer = ethereum_recover(T::Prefix::get(), &proof, &data)
+				.ok_or(Error::<T>::InvalidProof)?;
+			// NOTE(hussein-aitlahcen): ask for the address even if we can extract it
+			// from the signature as we want to differentiate invalid proof from non-existant
+			// contributors
+			ensure!(signer == ethereum_address, Error::<T>::InvalidProof);
```
dzmitry-lahoda, line 193: invalid proof if data is not eth signed versioned msg. if sinature is valid, but if signer not in list, why would on chain code care? it is up to claimer to recheck. so it seems that such distinctions is up to offchain code?

dzmitry-lahoda: just `what can be done offchain should be done off chain`?

frame/crowdloan-rewards/src/lib.rs
hussein-aitlahcen: It's not a standard,  it's just a specific format for claiming, prefix+account_id in this case, I reused the claiming mechanism of  [polkadot](https://github.com/paritytech/polkadot/blob/master/runtime/common/src/claims.rs#L530)

frame/crowdloan-rewards/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+		}
+
+		#[pallet::weight(10_000)]
+		pub fn claim_from_relay(
+			origin: OriginFor<T>,
+			reward_account: T::AccountId,
+			relay_account: T::RelayChainAccountId,
+			proof: MultiSignature,
+		) -> DispatchResult {
+			ensure_signed(origin)?;
```
dzmitry-lahoda, line 207: not sure why this to be signed, as proof is signature

frame/crowdloan-rewards/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+	pub fn verify_relay<AccountId: Encode, RelayChainAccountId: Into<AccountId32>>(
+		prefix: &[u8],
+		reward_account: AccountId,
+		relay_account: RelayChainAccountId,
+		proof: MultiSignature,
+	) -> bool {
+		let wrapped_prefix: &[u8] = b"<Bytes>";
+		let wrapped_postfix: &[u8] = b"</Bytes>";
+		let mut msg = wrapped_prefix.to_vec();
+		msg.append(&mut prefix.to_vec());
```
dzmitry-lahoda, line 283: may be use some thing like <version:u8>:<signature_type:u8>:<msg>? like in libp2p

frame/crowdloan-rewards/src/lib.rs
hussein-aitlahcen: I mean the message header is standard

frame/crowdloan-rewards/src/lib.rs
hussein-aitlahcen: I added it to make the claiming mechanism having a consistent behavior for both eth/relay claims. With the multisignature the proof will be considered invalid but since we extract the account from the eth one,  it will result   in an "NotAContributor" error.

frame/crowdloan-rewards/src/lib.rs
hussein-aitlahcen: I don't really understand what you mean here, this prefix/postfix are the one appended by the polkadotjs framework when signing.

frame/crowdloan-rewards/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+			ensure_signed(origin)?;
+			let data = reward_account.using_encoded(to_ascii_hex);
+			let signer = ethereum_recover(T::Prefix::get(), &proof, &data)
+				.ok_or(Error::<T>::InvalidProof)?;
+			// NOTE(hussein-aitlahcen): ask for the address even if we can extract it
+			// from the signature as we want to differentiate invalid proof from non-existant
+			// contributors
+			ensure!(signer == ethereum_address, Error::<T>::InvalidProof);
+			let remote_account: RemoteAccountOf<T> = RemoteAccount::Ethereum(ethereum_address);
+			let claimed = Self::do_claim(remote_account.clone(), &reward_account)?;
```
dzmitry-lahoda, line 195: could you please write down securit scheme?
i read as this 
1. store contirbutors public keys of any kind in storage 
2. during claim, vefify that a. key in strage b. key sin message c. message contains reward account
3. use that reward account to mint into

as for now as i read code, either scheme is differnt or misses some checks from 2 and has non needed checks like substrate native signatures

hussein-aitlahcen: I am not sure I follow what is missing:
1: we setup the rewards in the genesis block, (PK, Amount, Vesting) with PK either ETH/Relay (kusama)
2: you have to sign a message which consist of an arbitrary prefix and the reward account pk (picasso-account_id in tests)
3: you claim by sending the signed message + reward account
4: if that reward account exists, we mint into the reward account

frame/crowdloan-rewards/src/lib.rs
hussein-aitlahcen: we don't but I have setup the extrinsic as is

frame/crowdloan-rewards/src/lib.rs
hussein-aitlahcen: we definitely don't want this to be signed

frame/crowdloan-rewards/src/lib.rs
dzmitry-lahoda: i see. that is polkadot stuff. thanks.

frame/crowdloan-rewards/src/lib.rs
dzmitry-lahoda: can we reference code of https://github.com/paritytech/polkadot/blob/12df8e0777c0fa2ea92a19ae6d4f1d5bc3917bca/runtime/common/src/claims.rs?_pjax=%23repo-content-pjax-container#L110 instead of claiming ? I reference this for xcmp setup seems anyway.

hussein-aitlahcen: Yes I will reference the original code by documenting everything

frame/crowdloan-rewards/src/lib.rs
dzmitry-lahoda: if feature has internal assymetry and public api has some kind of assymetry, why would implementation in terms of errors to be symmetrical? sounds like OOP:)

frame/crowdloan-rewards/src/lib.rs
dzmitry-lahoda: ok. i see. it does proper flow. thanks.

frame/crowdloan-rewards/src/lib.rs
hussein-aitlahcen: I don't really like this extra "useless" information but as a user it is perhaps better to know if you failed to sign or signed a wrong message vs having a "you are not a contributor" answer.  The API asymmetry is present for the same reason, I feel like it's easier for a user to use a different extrinsic based on what type of contributor it was (relay/eth). I don't have a strong opinion on that, I can remove it :).

dzmitry-lahoda: error could be SignatureOfMessageNotFound in contributor list. Js client can than show list to user and he can match hist metamask to our list. doubt it ever will be backend problem as it should be solved by ui. or you clever to solve it properly by hands, than error is good enough for you.

seunlanlege:
One more consideration for this PR is that we probably shouldn't charge fees for claiming rewards.

dzmitry-lahoda (dismissed):

hussein-aitlahcen:
Well, it look like we want to be able to populate the reward accounts at runtime. I'll revert.

hussein-aitlahcen:
Alright, the new way:
- we allow admin origin to provide the rewards
- the populate extrinsic is idempotent, allowing us to send a batch of transactions and optionally correct wrongly computed rewards
- once the pallet is initialized and users are able to start claiming, we no longer allow reward changes

andor0 (dismissed):

dzmitry-lahoda (approved):
@Br1ght0ne in hist assets PR did awesome job to document hist pallet. If @hussein-aitlahcen you will be able to do so in next PR it would be awesome :)

hussein-aitlahcen:
> @Br1ght0ne in hist assets PR did awesome job to document hist pallet. If @hussein-aitlahcen you will be able to do so in next PR it would be awesome :)

I did push some documentation but it's not enough you are right! 
Sadly, the pallet is not finished yet as we had another "last minute" change...

frame/crowdloan-rewards/src/lib.rs
```diff
@@ -0,0 +1,390 @@
...
+
+		/// Populate pallet by adding more rewards.
+		/// Can be called multiple times. Idempotent.
+		#[pallet::weight(10_000)]
+		pub fn populate(
+			origin: OriginFor<T>,
+			rewards: Vec<(RemoteAccountOf<T>, RewardAmountOf<T>, VestingPeriodOf<T>)>,
+		) -> DispatchResult {
+			T::AdminOrigin::ensure_origin(origin)?;
+			// Make sure we can't populate after any user has claim anything otherwise we are
```
dzmitry-lahoda, line 203: can remove this comment seems. see initialize setup

frame/crowdloan-rewards/src/lib.rs
```diff
@@ -0,0 +1,390 @@
...
+							let current_block = frame_system::Pallet::<T>::block_number();
+							// Current point in time
+							let vesting_point =
+								current_block.saturating_sub(VestingBlockStart::<T>::get());
+							if vesting_point >= reward.vesting_period {
+								// If the user is claiming when the period is over, he should
+								// probably have already claimed everything.
+								reward.total
+							} else {
+								let vesting_period = VestingBlockStart::<T>::get()
```
dzmitry-lahoda, line 317: versing_end_inclusive

frame/crowdloan-rewards/src/lib.rs
```diff
@@ -0,0 +1,390 @@
...
+									.safe_add(&reward.vesting_period)?;
+								let partition = T::VestingPartition::get();
+								// Current window, rounded to previous window.
+								let vesting_window = vesting_point - (vesting_point % partition);
```
dzmitry-lahoda, line 321: vesting_period = vesting_end_block = vesting_start_block + period
vesting_point = (current_block - vesting_start_block)
vesting_window = (current_block - vesting_start_block) - (current_block - vesting_start_block) % partition
if current_block is very big, than `vesting_window = current_block`

available_to_claim = vested_reward * current_block / vesting_end_block = available_to_claim grows with each block. nice

dzmitry-lahoda: not sure if i miss something

frame/crowdloan-rewards/src/lib.rs
```diff
@@ -0,0 +1,390 @@
...
+		/// The relay chain account id.
+		type RelayChainAccountId: Parameter + MaybeSerializeDeserialize + Into<AccountId32> + Ord;
+
+		/// The upfront liquidity unlocked at first claim.
+		#[pallet::constant]
+		type InitialPayment: Get<Perbill>;
+
+		/// The number of blocks a fragment of the reward is vested.
+		#[pallet::constant]
+		type VestingPartition: Get<Self::BlockNumber>;
```
dzmitry-lahoda, line 142: VestingStep. it is like Step in auction

frame/crowdloan-rewards/src/lib.rs
dzmitry-lahoda: missed, upper rule to return total. huh.

dzmitry-lahoda:
```
T::Convert::convert(vesting_window)) /
										T::Convert::convert(vesting_period)
```
so you devide duration by total time, seems but in type system. my brain almost broken

hussein-aitlahcen:
> ```
> T::Convert::convert(vesting_window)) /
> 										T::Convert::convert(vesting_period)
> ```
> 
> so you divide duration by total time, seems bug in type system. my brain almost broken.
> 
> so vesting depends on what block it was started and if vesting period is much larger of start block or not.

I fear I don't understand the concern here? A user has a custom `vesting_period` of **X** blocks divided (lets say 48 weeks) by a `VestingStep` (let's say 1 week).

dzmitry-lahoda:
should = reward * (upfront + (current_vesting_step/vesting_period * (1- upfront))).
but if 
```
let vesting_period = VestingBlockStart::<T>::get()
									.safe_add(&reward.vesting_period)?;

```
than N / (48 + ANY_NUMBER)

hussein-aitlahcen:
> 

Definitely a mistake!

dzmitry-lahoda (approved):

andor0 (approved):

PR #278, opened 2021-11-13
added one more check for low provider edge case

JesseAbram:
Pretty much this ensures to avoid an edge case where stale price never triggers and the aggregation restarts. I think this is a clean and easy way to do it which is not limiting, say setting stale price to 3 and block interval to 5 on most asset IDs is completely acceptable in my eyes

### DIFF
--- a/frame/oracle/src/lib.rs
+++ b/frame/oracle/src/lib.rs
@@ -314,6 +314,8 @@ pub mod pallet {
 		/// Too many weighted averages requested
 		DepthTooLarge,
 		ArithmeticError,
+		/// Block interval is less then stale price
+		BlockIntervalLength,
 	}
 
 	#[pallet::hooks]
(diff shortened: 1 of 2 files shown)


hussein-aitlahcen (approved):
Good catch :)

frame/oracle/src/lib.rs
```diff
@@ -314,6 +314,8 @@ pub mod pallet {
 		/// Too many weighted averages requested
 		DepthTooLarge,
 		ArithmeticError,
+		/// Block interval is less then stale price
```
dzmitry-lahoda, line 317: BlockIntervalLength -> BlockIntervalIsLessThenStalePrice?

frame/oracle/src/lib.rs
```diff
@@ -373,6 +375,7 @@ pub mod pallet {
 			ensure!(max_answers >= min_answers, Error::<T>::MaxAnswersLessThanMinAnswers);
 			ensure!(threshold < Percent::from_percent(100), Error::<T>::ExceedThreshold);
 			ensure!(max_answers <= T::MaxAnswerBound::get(), Error::<T>::ExceedMaxAnswers);
+			ensure!(block_interval > T::StalePrice::get(), Error::<T>::BlockIntervalLength);
```
dzmitry-lahoda, line 378: may be rename StalePrice -> BlockIntervalPriceOutdated?

JesseAbram: So stale price is how long before we remove prices if they are are not aggregated, this check is to make sure that, that is done before a re aggregation

frame/oracle/src/lib.rs
JesseAbram: eh, it seems like a lot of words and the comments get passed with the error

andor0 (approved):

PR #280, opened 2021-11-14
Add a Polkadot cluster initialization script

andor0:
Added a script to create assets mappings in Composable and Basilisk parachains.


dzmitry-lahoda:
is it draft still or i can review? so i have modified assets-registry api to be exactly like in hydra. it uses multilocation. which is not serde.

andor0:
> is it draft still or i can review? so i have modified assets-registry api to be exactly like in hydra. it uses multilocation. which is not serde.

It is draft yet, assets mapping in Hydradx doesn't ready.

dzmitry-lahoda (approved):

dzmitry-lahoda (approved):

hussein-aitlahcen (approved):

PR #286, opened 2021-11-16
[CU-1t7knzg] added registry to runtime with trait impl

dzmitry-lahoda:

### DIFF
--- a/rfcs/0001-feature-gated-runtime.md
+++ b/rfcs/0001-feature-gated-runtime.md
@@ -1,0 +1,61 @@
+# Summary
+
+This proposes a solution to maintain several development streams and runtimes in a single branch.
+
+## Motivation
+
+There is an approach to developing software in branches, which we try to follow.
+However, I argue that this complicates things in the current state and makes our progress slower and more brittle.
+
+I suggest using a structured conditional compilation of runtime to achieve several streams of development without losing the security qualities of the codebase.
+That approach fits very nicely with Rust and Substrate ecosystems.
+
+The suggested approach will make our development easier until we release it onto Polkadot.
+
+## Detailed design
+
+Runtime is statically compiled. It means that if a pallet is coded but and not added to runtime, a pallet cannot influence runtime in any way.
+From this, a pallet under development does not represent a security issue.
+
+### Runtime gating
+
+See usage of 'feature = "develop"` in runtime.
+
+### IDE support
+
+Rust analyzer works well with features behind `rust-analyzer.cargo.features` configuration in VS Code. It is visible what features are on and off.
+
+### Cross pallet functional
+
+Useful functional may consist of several interacting pallets. Can develop and on them one by one behind the gate.
+As soon as the functional is ready, can remove develop gate from related pallets.
+This will make our project tracking software more manageable, too, by avoiding long-running tasks in branches.
+The shorter task is, the more predictable it is, and less friction is added during late integration.
+
+### Shared ownership
+
+As soon as code appears into the main branch, approved by our peer, each of us is the owner of that code.
+Extensive functional development is penalized, and significant refactoring is impossible under the risk of hard-to-merge issues.
+If one has a prominent feature on develop, one can get stuck in the long cycle to keep it up to date.
+Code runaway happens for both in code and in CI/CD in branch-based approach.
+
+Why would others keep up today's feature which is under develop?
+Because company agreed that that feature is valuable and should be developed the way it is.
+If a decision will change, the code could be deleted, which is easy to do.
+
+## Alternatives
+
+Using develop branch is an attempted alternative to go forward. I see next problems with it as of now:
+
+- When I need code in main, which is in develop, and vice versa, I cannot have it instantly as required as by dependency of my task.
+- There are no strict guidelines on what should go to the main and what should go into development.
+- Merging main into develop and develop back is hard and error-prone. Can introduce security bugs.
+- Develop is not up today with the same GitOps and linting practices with main. This leads to broken and hard-to-fix build issues, like the inability to build runtime on current develop.
+
+## Unresolved questions
+
+- Gating new yet audited extrinsic in existing pallets.
+- Versioning conditional blocks of code inside pallets.
+- https://github.com/paritytech/substrate/issues/10286
+
+Two of the above issues should be solved regardless of whether we are using develop branch or feature gate.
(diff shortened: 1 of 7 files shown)


itsbobbyzzz:
Task linked: [CU-1t7knzg XCMP: Impl RemoteAssetRegistry on pallet-assets-registry](https://app.clickup.com/t/1t7knzg)

rfcs/0001-feature-gated-runtime.md
```diff
@@ -0,0 +1,61 @@
...
+Why would others keep up today's feature which is under develop?
+Because company agreed that that feature is valuable and should be developed the way it is.
+If a decision will change, the code could be deleted, which is easy to do.
+
+## Alternatives
+
+Using develop branch is an attempted alternative to go forward. I see next problems with it as of now:
+
+- When I need code in main, which is in develop, and vice versa, I cannot have it instantly as required as by dependency of my task.
+- There are no strict guidelines on what should go to the main and what should go into development.
```
hussein-aitlahcen, line 51: I think is this the main issue we have with the branching strategy right now. Using a single branch will just push the complexity in the code as one would have to determine what the result of the compilation will be by looking at each part of the code with the feature gate. 
Also, this mean that we will somehow push versioning into the code as if you need incompatible changes over a shared piece, you will have to duplicate it? More than just that, when releasing, the release note won't reflect the real changes as the history will contains chunks of both stable/unstable changes, even ongoing not finished work. 

IMHO the current state is not satisfying but pushing struct rules over what should/should not be in main is the way to go. Everything should probably be done in develop (with a second runtime dedicated to be tweaked freely for develop) and then merged to main, one way. If we avoid having main->develop merged this would make it easy for everyone. We could still have to push some very little fixes on main, but thoses should be cherry-pickable back in develop.

dzmitry-lahoda: > I think is this the main issue we have with the branching strategy right now. Using a single branch will just push the complexity in the code as one would have to determine what the result of the compilation will be by looking at each part of the code with the feature gate.

it is up to CI to help with that.

>  single branch will just push the complexity in the code

Complexity still in the code in any case. Multi branch just delays complexity to stage of merge conflict resolution. One still have to think about how hist code would get into main and versa(or somebody else have to think). I would prefer fail fast approach than unpredictable fail later (merge conflicts and need to talk to developer who did some work). 
I can argue that approach I suggests allow more distributed development to happen faster. 
I suggest that mutlibranch will be good when solution will appear on Polkadot. 
Alternative, RFC for handling multibranch (but that RFC is process related - need to settle it on company level)

dzmitry-lahoda: > This will somehow push versioning into the code as if you need incompatible changes over a shared piece, you will have to duplicate it? 

this problem is same for branching. please try to play in head what problem we have with branching. you will find them here, but you delay their resolution. I am fine with delayed resolution when process is defined. right now i have problem. one code was pushed (not by me) into master, other into dev. i need both of them. i started on master because dev runtime was broken. i tried to fix, but failed. where is process to make sure it will not happen each time. 
i would wait fro DevOps for solution..

dzmitry-lahoda: @hussein-aitlahcen so from above list i see some good points, but they blame approach, but really should blame reality about need to handle releases, versioning, backward compatibility, code duplication in any solution we have. 
fun thing, i have similar talk approx 3 years ago about branching were people did not realized that branching = 100% code copy paste. while in code versioning can be lesser evil (specifically with rust support for conditional compilation and substrate support for statically compiled runtime)

dzmitry-lahoda: but yeah, i am ok with more process oriented approch, but without @KaiserKarel and DevOps we cannot solve issue. while with cfg - easy i can do.

seunlanlege: I think it makes sense

KaiserKarel: I think the approach of using cfg makes sense. I agree that auto-gen of release notes is kind of nice, but necessary to be honest.

KaiserKarel: https://softwareengineering.stackexchange.com/questions/270321/appropriate-git-workflow-for-multiple-active-releases-while-handling-hotfixes

hussein-aitlahcen (approved):

andor0 (approved):

rfcs/0001-feature-gated-runtime.md
dzmitry-lahoda: > In general I feel like every single commit in main should be merged back into develop.

So it could be for each main branch PR generate PR into develop? Even more, if it passes gate, it can be auto merged.  This may solve different CI/CD settings in main and develop.  Also will need to make sure CI/CD of developer closer to main.

dzmitry-lahoda: @KaiserKarel 
> https://softwareengineering.stackexchange.com/questions/270321/appropriate-git-workflow-for-multiple-active-releases-while-handling-hotfixes

read some of this, previously seen some GitFlow and you mentioned some other process recently on one of meeting (forget its name). this process needs clickup and biz flows setup, also may be it needs roadmap like long existing products has. i do not have opinion on this and this requires organizational effort to do so. just i have no comments for these:) but if setup some proper rules and merge flows, it should work.

same time. i base approach in rfc on next ideas:
- we are in rust, it is static, runtime is static, mostly typed. we have no (of few) problems java/python/c/js has.
- our runtime is just wasm swap right? so we do not have complex release with huge amount of dependencies. like traditional products have. so this is closed world.
- product is more or less new. and we are not on polkadot yet.
- cfg approach may be used as tool not only amid main <-> develop, but in main only and in develop only. so it is kind intresting tool.
- rust is years ahead of cpp/java/python doing conditional compilation. 
- cfg is supported by IDE, so auditors as being expert should evaluate and undrestand all above, and also have tooling to work with above. also they even may do WASM diff or whatever. so i mean that from audit perspective, formally, it will not be complicated to audit, until auditor somehow proves it other way.

it may appear that git based processes solved issues which are less visible in substrate style rust until runtime on polkadot. it took acala time (half year+?) to reach voting (probably it will win), so can we live with cfg feature based approach for similar period of time?

dzmitry-lahoda:
ok, i have some approves. let me try. if that will not work. will remove it.

PR #289, opened 2021-11-17
CU-1kjehh6 - avoid less than expected bonus by using safe maths

hussein-aitlahcen:


itsbobbyzzz:
Task linked: [CU-1kjehh6 Crowdloan bonus safe math](https://app.clickup.com/t/1kjehh6)

frame/crowdloan-bonus/src/lib.rs
```diff
@@ -142,7 +144,11 @@ pub mod pallet {
 
 			ensure!(pot_balance_value > 0 && token_supply_value > 0, Error::<T>::EmptyPot);
 
-			let to_payout = pot_balance_value.mul(amount).div(token_supply_value);
+			let to_payout = pot_balance_value
+				.checked_mul(amount)
+				.and_then(|x| x.checked_div(token_supply_value))
+				.ok_or(DispatchError::Arithmetic(ArithmeticError::Overflow))?;
+
```
dzmitry-lahoda, line 151: ArithmeticError can be division by zero also. Could we use `safe_div`?

hussein-aitlahcen: Of course!

andor0 (approved):

PR #291, opened 2021-11-18
Implement curve-amm using Vaults.

vivekvpandya:


frame/curve-amm/src/lib.rs
```diff
@@ -269,49 +230,62 @@ pub mod pallet {
...
 			PoolCount::<T>::get()
 		}
 
-		fn pool(id: PoolId) -> Option<PoolInfo<Self::AccountId, Self::AssetId, Self::Balance>> {
+		fn pool(id: PoolId) -> Option<PoolInfo<Self::AccountId, Self::VaultId, Self::Balance>> {
 			Pools::<T>::get(id)
 		}
+
 		fn create_pool(
 			who: &Self::AccountId,
```
dzmitry-lahoda, line 244: may be name variable `manager`

frame/curve-amm/src/lib.rs
```diff
@@ -1011,53 +656,73 @@ pub mod pallet {
...
+								&vault_id,
+								&pool_account_id,
+							)? {
+								FundsAvailability::Depositable(balance) => {
+									let balance = <T as Config>::LpToken::reducible_balance(
+										asset_id,
+										&pool_account_id,
+										false,
+									)
+									.min(balance);
```
dzmitry-lahoda, line 680: i think here you should check that ask from vault does not violates invariant to far OR (better) collect all depositables, and find maximal value of each balance you can deposit back without violation of invariant.

frame/curve-amm/src/lib.rs
```diff
@@ -1011,53 +656,73 @@ pub mod pallet {
...
+									<T::Vault as StrategicVault>::deposit(
+										&vault_id,
+										&pool_account_id,
+										balance,
+									)?
+								},
+								FundsAvailability::Withdrawable(balance) =>
+									<T::Vault as StrategicVault>::withdraw(
```
dzmitry-lahoda, line 688: here may be check withdrwable, and withdraw as much as possible without violating invariant (or may be up to some slippage).

frame/curve-amm/src/lib.rs
```diff
@@ -1011,53 +656,73 @@ pub mod pallet {
...
+										&vault_id,
+										&pool_account_id,
+										balance,
+									)?,
```
hussein-aitlahcen, line 692: Same as previous @dzmitry-lahoda comment, you cannot withdraw everything as you might unbalance yourself, instead, you should try to rebalance and if already balanced, take from each vault an amount which is equal to the minimum you can withdraw from one of the vaults. Also, can we extract this to its own function and add valid/invalid tests for them please?

frame/curve-amm/src/lib.rs
```diff
@@ -1011,53 +656,73 @@ pub mod pallet {
...
+								FundsAvailability::MustLiquidate => {
+									let balance = <T as Config>::LpToken::reducible_balance(
+										asset_id,
+										&pool_account_id,
+										false,
+									);
+									<T::Vault as StrategicVault>::deposit(
```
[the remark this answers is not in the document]
hussein-aitlahcen: I think it is very dangerous to keep the dex running while it has been liquidated from the strategies of one of its vault. We should perhaps just close the dex in this case as you will no longer be able to withdraw from this vault. Or we just remove the coin from the pool coins (I'm not sure of the consequence here)?

frame/curve-amm/src/tests.rs
```diff
@@ -119,111 +122,65 @@ fn get_y_j_greater_than_n() {
...
-		assert_eq!(Tokens::balance(pool_info.pool_asset, &BOB), 0);
-		assert_eq!(Tokens::balance(MockCurrencyId::USDT, &BOB), 200000);
-		assert_eq!(Tokens::balance(MockCurrencyId::BTC, &BOB), 20);
-		assert_eq!(Tokens::balance(MockCurrencyId::BTC, &CurveAmm::account_id(&pool_id)), 0);
-		assert_eq!(Tokens::balance(MockCurrencyId::USDT, &CurveAmm::account_id(&pool_id)), 0);
-	});
-}
-
 #[test]
 fn exchange_test() {
```
hussein-aitlahcen, line 126: Can we have more than 1 test for exchange, can we have some hardcoded values with slippage for instance?

hussein-aitlahcen (changes requested):

frame/curve-amm/src/tests.rs
vivekvpandya: There are more than 1 exchanges in test. I will update test with hardcoded values.

frame/curve-amm/src/lib.rs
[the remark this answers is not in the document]
vivekvpandya: I think closing the DEX is simple to implement, I can keep a boolean variable indicating operation is halted. and in all function we should check that and return Error if it is set to true.

frame/curve-amm/src/lib.rs
vivekvpandya: @hussein-aitlahcen  so in pool creation should I ask fund from vaults? or both values 0 means it is balanced ?

frame/curve-amm/src/lib.rs
```diff
@@ -1011,53 +656,139 @@ pub mod pallet {
...
+                    .map(|vault_id| {
+                        Ok(T::Vault::asset_id(vault_id).ok().ok_or(Error::VaultError)?.into())
+                    })
+                .collect::<Result<Vec<T::AssetId>, Error<T>>>()?;
+                let mut balances = Vec::new();
+                for asset_id in &pool_asset_ids {
+                    let bal = T::LpToken::balance(*asset_id, &pool_account_id);
+                    balances.push(bal);
+                }
+                let mut xp_f: Vec<FixedU128> = balances
```
dzmitry-lahoda, line 686: not sure i get what xp_f means? is it from paper?

dzmitry-lahoda: ah, it is fixed? do we need non fixed? could we remove all _f things?

frame/curve-amm/src/lib.rs
```diff
@@ -1011,53 +656,139 @@ pub mod pallet {
...
+				let results = Pools::<T>::iter()
+					.map(|(pool_id, pool_info)| {
+						let pool_account_id = Self::account_id(&pool_id);
+						for (index, vault_id) in pool_info.assets_vault_ids.iter().enumerate() {
+							let asset_id = <T::Vault>::asset_id(&vault_id)?;
+							match <T::Vault as StrategicVault>::available_funds(
+								&vault_id,
+								&pool_account_id,
+							)? {
+								FundsAvailability::Depositable(balance) => {
```
dzmitry-lahoda, line 737: please write at least one test for this

frame/curve-amm/src/lib.rs
```diff
@@ -1011,53 +656,139 @@ pub mod pallet {
...
 		}
+
+        fn limit_preserving_invariant( 
+			pool_id: PoolId,
+            index : usize,
+            min_balance : T::Balance,
+            max_balance : T::Balance,
+        ) -> Result<T::Balance, DispatchError> {
+            let value = Pools::<T>::try_mutate(pool_id, |pool| -> Result<T::Balance, DispatchError> {
+                let pool = pool.as_mut().ok_or(Error::<T>::PoolNotFound)?;
```
dzmitry-lahoda, line 670: i see, so caller 
1. gets balance of one asset
2. get balance of all assets
3. calculate.
4. repeat for each from one.

could we have got
1. get all balances
2. get all withdrawables/depositbles for each
3. sort from smalles balance to largest
4. withdraw max to smalles at index i = 0
5. check if can withdraw for i+1 to match i
6. if can, add it i+i to match i
7. if not, return back, cut i to be as small as i+1
8. i+i, can added to match, add, cannot <- try cut previous increases.
9. as soon as we hit something in i+n, check if has withdraw to became smaller than previous, make it smaller up to value
10. if it cannot be made as small as previous, make it as small as poossible
11. keep track of all additionas and removals
12. so we maximixed greed to take all possible and make all as equal as possible.
13. next, take a look if can withdraw without violating invariant from each
13. if each withdrawl does not violates invariant, withdraw from each as much as possible.
13. example , 4 + 2 , 7 - 2, 8 - 2 -> 6, 6, 6
14. example,  5 - 2 , 7 - 3, 8 - 5 -> 4, 4, 4
15. example,  5 + 1, 7 + 1, 10 - 1 -> 6, 7, 9
16. example,  5 - 3 , 7 , 8 - 5 -> 5, 7, 7 (incetives to add more to balance, as smalles is most imbalanced, or withdraw)
17. write property test generating random values, ensue that if you drop VERY big imblaances, it converges with smaller randoms to invarint.
18. prove that trying to add big imblances does not leads to inblanaces
19. not sure if above algo works, but it is at least more optimal than looping, may be.
20. i will with about using average or error to find most balanced solution (solution which leads to smalles possible arbitrage ability, i mean that if after balancing will trade in cyrpto, smalles possible diff amid values will be visible)

dzmitry-lahoda:
@vivekvpandya pleas add Clickup number to this work. thanks

frame/curve-amm/src/lib.rs
dzmitry-lahoda: @KaiserKarel can you check if we are not doing here too stupid or complicated. Should we rely on vault admin to balance invariant or like? Or pool should help itself?

dzmitry-lahoda: @vivekvpandya i pretty robust about algorithm. can meet and explain it. but first may be we could ask somebody from financial team about feacibility of such?

frame/curve-amm/src/lib.rs
vivekvpandya: all computations are done in FixedU128 , but in runtime we can;t use it directly so in runtime we use u128 (T::Balance). This is similar to what we have done in lending.

frame/curve-amm/src/lib.rs
dzmitry-lahoda: I am struggling with next scneario right now

9  9 9 11 11 11 12
2  0 0  0   0  0   0      

It is evident that adding to 2 nine is right. That is why binary search and calculate X.

9  9 9 11 11 11 12
10  0 0  0   0  0   0      

In this scenario, algorithm tells, try 9 + 3. If improve -> Add 3. If not, 3/2 and check. For few numbers it can be fast even if n * n . Alternatively, can do best, with up to binary search, and do tune off chain more complex. Again, if violation of target to far from good.

So can run different variants of next, like `12  9 9 11 11 11 12` ``11  9 9 11 11 11 12`

9  9 9 11 11 11 12
10  0 7  0   0  0   0      

binary earch would be
take 7 as max. 
add 7 to 9 and to 9 -> bad.

3.5 -> add. Just like try to increase all currencies to make them equal(thismakes XI good).



And this. And see what is make to make XI = A with up to some 0.95 still making DEX leverage largest. 



image


@vivekvpandya for prove we would use random+property tests. We will run set of randomized experiments and see that algorithm ends up always to best X. 

@hussein-aitlahcen it is intresting, that if USD = 1 and DAI = 1, and ZXC = 100. Still can use ZXC if to divied it by 100. So as soon as some LP token in stable, we can create markets for it with other stable by normalizing.

vivekvpandya:
I have following simple algorithm for re-balancing:
Case1 : All vaults are in withdrwable state. Then withdraw minimum possible amount from each vaults and deposit to DEX.
Case2: At least one vault is in depositable state. Then deposit minimum possible amount from DEX to each vaults.
Case 3: At least one vault is in MustLiquidate state. Then halt the DEX and return all funds to respective vaults.
@dzmitry-lahoda @hussein-aitlahcen

dzmitry-lahoda:
I feel bad about case 1 and case 2. for me take all which is provided and give all requested would be better. with known risks.  so no algorithm better than some which may be bad. 

@vivekvpandya could you write up different scenarios of algorithm how it will lead to balance or now?

vivekvpandya:
closing this PR. In future if required we can use https://github.com/ComposableFi/composable/tree/dex_vault

PR #297, opened 2021-11-22
Fix build

andor0:
* Updated Dockerfile to fix a build


.github/workflows/check.yml
```diff
@@ -107,4 +107,4 @@ jobs:
         run: |
           SKIP_WASM_BUILD=1 cargo test
           SKIP_WASM_BUILD=1 cargo clean -p picasso-runtime
-          SKIP_WASM_BUILD=1 cargo clean -p composable
\ No newline at end of file
+          SKIP_WASM_BUILD=1 cargo clean -p composable
```
hussein-aitlahcen: By the way do we still need the SKIP_WASM_BUILD ???

dzmitry-lahoda (approved):

.github/workflows/check.yml
andor0: imho, can remove `SKIP_WASM_BUILD=1` before `cargo clean ...`.

hussein-aitlahcen (approved):

PR #299, opened 2021-11-23
Test losses when vault strategy withdraws

c410-f3r:
Feel free to indicate any possible misunderstandings.

It is not possible to use `Amount` because, FWICT, no function, method or call makes actual use of it. For example, https://github.com/ComposableFi/composable/blob/main/frame/vault/src/lib.rs does not contain any `Amount` reference and the use of it would probably require a code rewrite that is not in the scope of the current assigned task.

Another thing worth mentioning is that `prop_assert_epsilon ` makes use of `multiply_by_rational(u128, u128, u128)`.

In order to test eventual vault losses a similar structure used on `vault_stock_dilution_k ` will be put in place, i.e., some accounts deposit assets, strategy withdraws (instead of depositing), accounts deposit again and then all shares, tokens and losses will be evaluated accordingly.

This PR is still WIP because I need to know if the described solution above is indeed the right way forward. If yes then I will properly de-duplicate code and make everything green. If not, feel free to indicate any possible misunderstandings (again) or the correct way to test vault losses.

### DIFF
--- a/frame/composable-traits/src/vault.rs
+++ b/frame/composable-traits/src/vault.rs
@@ -55,8 +55,10 @@ where
 	AccountId: core::cmp::Ord,
 {
 	pub asset_id: CurrencyId,
+	/// Idle amount of assets for withdraw
 	pub reserved: Perquintill,
 	pub manager: AccountId,
+	/// Not the vault strategy per si but rather a pool of funds that are used for strategies
 	pub strategies: BTreeMap<AccountId, Perquintill>,
 }
 
--- a/frame/vault/src/models.rs
+++ b/frame/vault/src/models.rs
@@ -14,7 +14,10 @@ pub struct VaultInfo<AccountId, Balance, CurrencyId, BlockNumber> {
 
 #[derive(Copy, Clone, Encode, Decode, Default, Debug, PartialEq, TypeInfo)]
 pub struct StrategyOverview<Balance> {
-	/// The reported balance of the strategy
+	/// The reported balance of the strategy.
+	///
+	/// Added when an account withdraws from a vault and subtracted when an account deposits
+	/// into a vault.
 	pub balance: Balance,
 	/// Sum of all withdrawn funds.
 	pub lifetime_withdrawn: Balance,
--- a/frame/vault/src/tests.rs
+++ b/frame/vault/src/tests.rs
@@ -20,7 +20,7 @@ use frame_support::{
 	traits::fungibles::{Inspect, Mutate},
 };
 use proptest::prelude::*;
-use sp_runtime::{helpers_128bit::multiply_by_rational, FixedPointNumber, Perquintill};
+use sp_runtime::{helpers_128bit::multiply_by_rational, FixedPointNumber, Perbill, Perquintill};
 
 /// Missing macro, equivalent to `assert_ok!`!
 macro_rules! prop_assert_ok {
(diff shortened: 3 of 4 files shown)


hussein-aitlahcen:
@c410-f3r nice first shot :). I think that the `Amount` type was a bit misleading. If you look at the vault, there is a function [`update_strategy_report`](https://github.com/ComposableFi/composable/blob/6b6a2f247e11a415ac749a3065b2117aa19d45ba/frame/vault/src/lib.rs#L882) which is not yet used in tests, it's a function that can be called by a strategy to notify it's current balance w.r.t the vault (a.k.a. the strategy withdraw an amount **X**, it later call `update_strategy_report` with **X'**, possibly >X or <X depending on what the strategy did with the funds). Using this function it should be possible to directly tests strategy loss by using a second `Balance` lower than the **profit**? What do you think?

frame/vault/src/tests.rs
```diff
@@ -522,96 +532,100 @@ proptest! {
...
+			prop_assert_ok!(<Vaults as StrategicVault>::deposit(
+				&vault_id,
+				&strategy_account_id,
+				strategy_profits
+			));
+			prop_assert_ok!(<Vaults as StrategicVault>::withdraw(
+				&vault_id,
+				&strategy_account_id,
+				strategy_losses
+			));
```
c410-f3r, line 582: It is hard for me to understand why `update_strategy_report` should be used because `withdraw` already modifies storage.

https://github.com/ComposableFi/composable/blob/823aa161930734af1a4689facd50b1b128704a51/frame/vault/src/lib.rs#L838-L839

Does the fact that currencies are being transferred makes testing losses impractical? For example, LPs won't be able to withdrawn LP tokens from the subtracted vault balances.

https://github.com/ComposableFi/composable/blob/823aa161930734af1a4689facd50b1b128704a51/frame/vault/src/lib.rs#L845-L846

frame/vault/src/tests.rs
```diff
@@ -522,96 +532,100 @@ proptest! {
...
+						strategy_profits,
+						lp_tokens,
+						lp_tokens_total,
+					)
+					.expect("qed;");
 
-					prop_assert_epsilon!(new_balance, balance + strategy_profit_share);
-				}
-				else {
+					prop_assert_epsilon!(new_native_tokens, initial_native_tokens + strategy);
```
c410-f3r, line 611: Quite didn't understand when withdraw should be called. Within strategy moment? After profits checks? Create another strategy moment for withdraws?

Since the unmodified `vault_stock_dilution_k` test checks profits of a full LP tokens withdraw, I tried to make the strategy deposit and then withdraw with the amount of deposit greater than the amount of withdraw. It would then be possible to assert a small amount of profits that are influenced by the withdrawn amount.

frame/vault/src/tests.rs
hussein-aitlahcen: > 

`StrategicVault::withdraw` does indeed mutate the balance, but `StrategicVault::deposit` subtract the amount transferred back from the strategy balance. Hence, if you `StrategicVault::withdraw` for an amount `X` and then `ReportableStrategicVault::update_strategy_report` with an amount `Y` which is `< X`, you effectively declare a **loss**. Because `https://github.com/ComposableFi/composable/blob/551e705396608de1d469e8836c87bb7e731fed8c/frame/vault/src/lib.rs#L869` will be lowered (the difference `X - Y` has not been transferred back to the vault account).

frame/vault/src/tests.rs
hussein-aitlahcen: There is many scenarios which are untested... Basically the one you describe but many others. For instance, since the vault has a reserve, this `stock_dilution_k` would not work if the strategy was withdrawing, because the vault would not own 100% of the funds. So it would also be nice to check for "incremental" withdrawing from users. Like the strategy would deposit each X blocks (increasing vault funds), allowing users to withdraw.

frame/vault/src/tests.rs
c410-f3r: > `wittdraw(X)` and then `update_strategy_report(Y)` with `X` > `Y`

In a withdraw, the strategy vault balance will have a value `Y` lesser than the transferred amount `X` executed on `withdraw` because of `update_strategy_report`. Therefore, the strategy account balance will be greater than the strategy vault balance meaning a **loss**.

Yes, it makes sense locally but not in an overall architecture picture but such knowledge will eventually come with more development.

frame/vault/src/tests.rs
```diff
@@ -525,93 +525,125 @@ proptest! {
...
 		let asset_id = MockCurrencyId::D;
-		let (strategy_account_id, strategy_profits) = created_accounts[0];
-		let strategy_deposit_moment = random_index;
-		let account_start = 1;
-		let accounts = &created_accounts[account_start..];
+
+		let (strategy_account_id, strategy_deposit) = accounts[0];
+		let strategy_withdraw_pbl = Perbill::from_percent(accounts.len() as u32 % 100);
+		let strategy_withdraw = strategy_withdraw_pbl.mul_floor(strategy_deposit);
+		let strategy_vault_balance = strategy_withdraw - strategy_withdraw_pbl.mul_floor(strategy_withdraw);
```
c410-f3r, line 547: Instead of creating another balance on `valid_amounts_without_overflow_k_with_random_index`,  the strategy withdraw value is calculated multiplying the deposited value by `accounts.len() % 100` and then the strategy balance representing a loss is calculated using the same percentage. That said, `deposit` > `withdraw` > `strategy balance` will be truth with amounts more or less randomized.

frame/vault/src/tests.rs
```diff
@@ -525,93 +525,125 @@ proptest! {
...
-				.map(|(account, _)| Tokens::balance(asset_id, account))
+			let current_sum_of_native_tokens = before_moment_lps()
+				.map(|(account, _)| Tokens::balance(asset_id, &account))
 				.sum::<Balance>();
 
 			prop_assert_epsilon!(
-				current_sum_of_shareholders_balance,
-				initial_sum_of_shareholders_balance + strategy_profits
+				current_sum_of_native_tokens,
+				initial_sum_of_native_tokens / 2 + strategy_deposit / 2 - strategy_withdraw_diff / 2
```
c410-f3r, line 646: Because LPs are taking half of their LP tokens (https://github.com/ComposableFi/composable/pull/299/files#diff-a8b4bcca53a49de6c5fa8f929af30fbecd3773c39c5fdbec31b6b495eb433931R601-R603), this and other divisions by 2 are necessary

frame/vault/src/tests.rs
```diff
@@ -525,93 +525,125 @@ proptest! {
...
+					let strategy_native_tokens_withdraw = multiply_by_rational(
+						strategy_withdraw_diff / 2,
+						lp_tokens,
+						lp_tokens_total,
+					)
+					.expect("qed;");
+
+					let diff = strategy_native_tokens_deposit - strategy_native_tokens_withdraw;
+
+					prop_assert_epsilon!(new_native_tokens, half_initial_native_tokens + diff);
```
c410-f3r, line 629: Works as expected because the simply removal of, for example, withdrawn tokens would result in an checking error

c410-f3r (commented):
@hussein-aitlahcen Thank you for the help.

Taking aside new functional parts of the test, all additional documentation and code were rewritten or added to help me understand the codebase as a whole but since they are optional, it is possible to remove everything if desired.

Is there any other test that also needs to take into consideration losses?

hussein-aitlahcen (approved):
LGTM :)

frame/composable-traits/src/vault.rs
```diff
@@ -55,8 +55,10 @@ where
 	AccountId: core::cmp::Ord,
 {
 	pub asset_id: CurrencyId,
+	/// Idle amount of assets for withdraw
```
dzmitry-lahoda, line 58: enum Parts(reserverd, stragegies) kind of. valuate sum to 1

frame/vault/src/models.rs
```diff
@@ -14,7 +14,12 @@ pub struct VaultInfo<AccountId, Balance, CurrencyId, BlockNumber> {
 
 #[derive(Copy, Clone, Encode, Decode, Default, Debug, PartialEq, TypeInfo)]
 pub struct StrategyOverview<Balance> {
-	/// The reported balance of the strategy
+	/// The reported balance of the strategy.
+	///
+	/// Added when withdrawn and subtracted when deposited more or less like a "borrowed" balance.
```
dzmitry-lahoda, line 19: added by via withdrawals from vault and substrate by depositign back to vault

dzmitry-lahoda:
> In order to test eventual vault losses a similar structure used on vault_stock_dilution_k will be put in place, i.e., some accounts deposit assets, strategy withdraws (instead of depositing), accounts deposit again and then all shares, tokens and losses will be evaluated accordingly.

you mean that in runtime code should do all calculation, than check these sum up to total. and only than apply to vault?

dzmitry-lahoda (approved):

frame/composable-traits/src/vault.rs
c410-f3r: Sorry. Could you elaborate this statement?

frame/vault/src/models.rs
c410-f3r: Changed

c410-f3r:
> > In order to test eventual vault losses a similar structure used on vault_stock_dilution_k will be put in place, i.e., some accounts deposit assets, strategy withdraws (instead of depositing), accounts deposit again and then all shares, tokens and losses will be evaluated accordingly.
> 
> you mean that in runtime code should do all calculation, than check these sum up to total. and only than apply to vault?

For testing purposes (no additional runtime logic), a strategy account withdraws in order to verify if losses are being evaluated correctly but nevertheless, the comment is outdated in the sense that no new `vault_stock_dilution_k`-like test was created but rather the `vault_stock_dilution_k` test itself was modified to account withdraws.

hussein-aitlahcen (approved):

frame/composable-traits/src/vault.rs
c410-f3r: My apologies but I couldn't understand probably because I am still learning the codebase. Feel free to open an issue describing your concerns and I will try to tackle everything afterwards.

dzmitry-lahoda: everything should in sum be 100%. that is type. (idle, (strategy, amount))

dzmitry-lahoda (approved):

PR #322, opened 2021-11-29
[Vault] Add call that signals liquidation of a strategy account

c410-f3r:


frame/vault/src/lib.rs
```diff
@@ -487,6 +490,26 @@ pub mod pallet {
 			Self::deposit_event(Event::VaultStarted { vault });
 			Ok(().into())
 		}
+		#[pallet::weight(10_000)]
+		pub fn liquidate_strategy(
```
c410-f3r, line 494: `liquidate` is a bit misleading IMO, looks like the caller is trying to pay someone's debt

frame/vault/src/lib.rs
```diff
@@ -487,6 +490,26 @@ pub mod pallet {
...
+			origin: OriginFor<T>,
+			vault_idx: VaultIndex,
+			strategy_account: T::AccountId,
+		) -> DispatchResult {
+			let from = ensure_signed(origin)?;
+			ensure!(Self::vault_info(&vault_idx)?.manager == from, Error::<T>::AccountIsNotManager);
```
c410-f3r, line 500: Most parachains use the Collective/Membership pallet to deal with stuff like this

frame/vault/src/lib.rs
```diff
@@ -487,6 +490,26 @@ pub mod pallet {
...
+			let balance = CapitalStructure::<T>::try_get(&vault_idx, &strategy_account)
+				.map_err(|_err| DispatchError::CannotLookup)?
+				.balance;
+			<Self as StrategicVault>::deposit(&vault_idx, &strategy_account, balance)?;
+			Allocations::<T>::remove(&vault_idx, &strategy_account);
+			CapitalStructure::<T>::remove(&vault_idx, &strategy_account);
```
c410-f3r, line 506: The reasoning here is that if a strategy account withdrawn any amount of the vault's balance, then the same account should "repay" the vault before removal

frame/vault/src/lib.rs
```diff
@@ -487,6 +490,26 @@ pub mod pallet {
...
+			Self::deposit_event(Event::StrategyLiquidated {
+				account: strategy_account,
+				amount: balance,
+			});
```
c410-f3r, line 510: Another desired parameter?

frame/vault/src/lib.rs
```diff
@@ -714,6 +734,11 @@ pub mod pallet {
 				.fold(T::Balance::zero(), |sum, item| sum + item.balance);
 			Ok(owned + outstanding)
 		}
+
+		/// Tries to fetch a stored [VaultInfo] through its index.
+		fn vault_info(vault_idx: &VaultIndex) -> Result<VaultInfo<T>, DispatchError> {
+			Ok(Vaults::<T>::try_get(vault_idx).map_err(|_err| Error::<T>::VaultDoesNotExist)?)
+		}
```
c410-f3r, line 741: De-duplicates several occurrences of `Vaults::<T>::try_get(&vault_id).map_err(|_| Error::<T>::VaultDoesNotExist)?;` that were scattered across the crate

frame/vault/src/lib.rs
```diff
@@ -696,7 +716,7 @@ pub mod pallet {
 			*/
 			let lp_shares_value =
 				multiply_by_rational(lp_amount_value, vault_aum_value, lp_total_issuance_value)
-					.map_err(|_| Error::<T>::OverflowError)?;
+					.map_err(|_| ArithmeticError::Overflow)?;
```
c410-f3r, line 719: Substrate already provides a native overflow-like error making `Error::<T>::OverflowError` unnecessary.

c410-f3r (commented):
Still need to document and create unit tests using proptest. Everything should be finished tomorrow.

frame/vault/src/lib.rs
hussein-aitlahcen: I think in this case we don't want to force the strategy to repay right now, as the strategy will call `available_funds` later and notice that it has been liquidated thanks to `FundsAvailability::MustLiquidate`. In this case, the strategy has to repay ASAP, but the funds might be elsewhere and take time to be rapatriated?

frame/vault/src/lib.rs
hussein-aitlahcen: You can rename it if you feel like it is misleading :)

frame/vault/src/lib.rs
dzmitry-lahoda: like EnsureRootOrHalfCouncil

frame/vault/src/lib.rs
dzmitry-lahoda: so in this case liquidate must note remove, but set liquidation state to trigger Must Liquidate?

frame/vault/src/lib.rs
c410-f3r: Yes but I am just saying. Still not sure if it is possible to integrate them with the Vault infrastructure

frame/vault/src/lib.rs
c410-f3r: On a second thought, I think the name should remain as is

frame/vault/src/lib.rs
c410-f3r: According to `available_funds`, `MustLiquidate` will only be triggered if a vault does not exist or an `Allocation` does not exist or an existent vault is marked as stopped. That said, the only available option in the context of this PR is to remove the account `Allocation` but there are some caveats:

* Once removed, there is no way to re-created `Allocation` (within the crate) even if the account repays.
* `Allocation` is tied to `CapitalStructure` so it does not make sense, IMO, to have one without the other.

frame/vault/src/tests.rs
```diff
@@ -142,6 +142,16 @@ prop_compose! {
...
 
+	#[test]
+	fn liquidate_strategy_successfully_liquidates_a_strategy_account(
+		strategy_account_id in strategy_account(),
+		total_funds in valid_amounts_without_overflow_1()
+	) {
+		do_liquidate_strategy_successfully_liquidates_a_strategy_account(
+			strategy_account_id, total_funds
+		)
+	}
```
c410-f3r, line 153: Declarative macros effectively disable `rustfmt` and make interactive debugging harder, thus the reason why the core logic was moved to another function.

c410-f3r:
```
 273 |                         syncing: ExecutionStrategy::Wasm,
     |                                                     ^^^^ variant or associated item not found in `sp_state_machine::ExecutionStrategy`
```

CI is probably broken

hussein-aitlahcen:
> ```
>  273 |                         syncing: ExecutionStrategy::Wasm,
>      |                                                     ^^^^ variant or associated item not found in `sp_state_machine::ExecutionStrategy`
> ```
> 
> CI is probably broken

Have been fixed, should work now :)

frame/vault/src/lib.rs
hussein-aitlahcen: There has been a patch in develop to merged `Allocation` with `CapitalStructure`. Sadly, we were using two branch until we decided to use the feature gate instead. We didn't merged the branches yet so this will result in a conflict while we will merge. In the context of this PR, removing both entry is correct

c410-f3r: OK, the call is now removing both storages but it is worth noting that `<Vaults as StrategicVault>::deposit` only works because of `ValueQuery`.

hussein-aitlahcen (dismissed):
Awesome!

hussein-aitlahcen (approved):

dzmitry-lahoda (approved):

PR #339, opened 2021-12-03
Initial XCM abstraction development

c410-f3r:
Summary of what was understood according to `Abstraction Layer` and spoken conversations. Feel free to point any misunderstandings.

## Library/Pallet

XCM is a protocol that enables **external** communication between different parachains, which means that a library or pallet must dispatch an HTTP or WS call targeting some desired endpoint.

```
# `method` is pallet plus call encoded names
# `params` is the encoded call params

{ id: 1, method: "8ee7418a6531173d60d1f6a82d8f4d51", params: "0x..." }
```

So in order to make an abstraction around different calls of different parachains that third-party users can use, it is possible to (1) Create a client **library** using, for example, `subxt` or (2) create a pallet with the `polkadot` repository primitives.

Since it was already decided that such abstraction will be a pallet, it is important (IMO) to say **pallet** instead of **library** to avoid confusion (even though a pallet is a library)

## Parachain XCM calls

Each parachain is free to define its own set of XCM interfaces but the more-or-less general consensus is to use the `xtokens` implementation (https://github.com/open-web3-stack/open-runtime-module-library/tree/master/xtokens) where common good calls like `transfer` operations are implemented.

Although `xtokens` eases the burden of having multiple different interfaces, the majority of the calls are parachain-unique. Maybe in the future something like EIP interfaces will come to the table or `orml` will provide more common runtime calls.

## Design

There are two possible ways to design the system.

### 1. Abstract individual platform-specific operations into common structures.

```rust
// High level example

// With specific Karura parameters
struct KaruraSwap { ... };

// With generic composable parameters
struct CanonicalSwap { ... };

// Converter
trait IntoXcm<D, F> {
    fn into_xcm(data: D, from: F) -> Self;
}

impl IntoXcm<(), CanonicalSwap> for KaruraSwap { .. }

impl pallet_composable_xcm::PalletApi for Pallet {
    ...
    // After converting `KaruraSwap` to `CanonicalSwap`, callers should be able to use this method
    fn swap(swap: CanonicalSwap) {}
    ...
}
```

Public calls are defined through the `PalletApi` trait, i.e., the `Pallet` structure implements `PalletApi` to allow third-party pallets call methods:

```rust
// Third-party pallet

#[pallet::config]
pub trait Config: frame_system::Config {
    type ComposableXcm: pallet_composable_xcm::PalletApi;
}
```

Another option is to work directly with the Pallet structure or inherit the `Config` trait bound.

### 2. Abstract individual platform-specific operations using traits

The most flexible method in my opinion.

```rust
// High level example

// Generic operations with generic parameters
trait Exchange<P> {
    type Pallet: pallet_composable_xcm::PalletApi;

    fn swap(&self, pallet: &Self::Pallet, currency_in: &Currency, max_price: &Balance, ...);
}

// With specific Karura parameters
struct Karura { ... };

impl Exchange for Karura {
    type Pallet = pallet_composable_xcm::Pallet;

    ...
    fn swap(&self, pallet: &Self::Pallet, currency_in: &Currency,max_price: &Balance, ...) { ... }
    ...
}
```

As seen from above, this second method lets calls use `Pallet` primitives, which is different from the first method where `Pallet` implements calls.

## Implementation

This initial implementation is a **very minimal** attempt that **only demonstrates** the basic architecture idea of a **~3 days** research and development cycle. Nothing is final (nomenclature, arrangements, structures or traits) so any idea, critic or suggestion is welcome for further enhancements and well-defined modifications.

As previously stated before, each individual operation should be abstracted into common. Therefore, the second design method wasn't implemented at all.

## Ethereum compatibility

I am **assuming** that `frontier` won't be used to talk directly with the Ethereum network because, FWIU, this pallet is only about abstracting different parachains. Besides, some networks like Moonbeam already support such feature (They probably charge a fee though, idk).

## Curiosity

Not in the scope of this PR but it is worth mentioning that although interoperability is a thing, atomicity isn't. Thus the reason why sometimes you hear things like DeFi Hubs or composable/specialized parachains.

cc @hussein-aitlahcen @KaiserKarel

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6242,6 +6242,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "pallet-composable_xcm"
+version = "1.0.0"
+dependencies = [
+ "frame-support",
+ "frame-system",
+ "parity-scale-codec",
+ "scale-info",
+ "sp-io",
+ "sp-runtime",
+ "xcm",
+]
+
+[[package]]
 name = "pallet-crowdloan-bonus"
 version = "0.0.1"
 dependencies = [
@@ -6269,7 +6282,6 @@ dependencies = [
 name = "pallet-crowdloan-rewards"
 version = "0.0.1"
 dependencies = [
- "composable-traits",
  "frame-support",
  "frame-system",
  "hex",
@@ -6993,24 +7005,6 @@ dependencies = [
  "scale-info",
  "serde",
  "sp-arithmetic",
- "sp-core",
- "sp-io",
- "sp-runtime",
- "sp-std",
-]
-
-[[package]]
-name = "pallet-vault-group"
-version = "0.1.0"
-dependencies = [
- "composable-traits",
- "frame-support",
- "frame-system",
- "pallet-balances",
- "parity-scale-codec",
- "scale-info",
- "serde",
- "smallvec 1.7.0",
  "sp-core",
  "sp-io",
  "sp-runtime",
--- a/frame/xcm/Cargo.toml
+++ b/frame/xcm/Cargo.toml
@@ -1,0 +1,33 @@
+[dependencies]
+frame-support = { branch = "polkadot-v0.9.12", default-features = false, git = "https://github.com/paritytech/substrate"  }
+frame-system = { branch = "polkadot-v0.9.12", default-features = false, git = "https://github.com/paritytech/substrate" }
+parity-scale-codec = { default-features = false, version = "2.0" }
+scale-info = { default-features = false, features = ["derive"], version = "1.0" }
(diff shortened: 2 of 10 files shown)


KaiserKarel:
Starting to like the look of it. I think you raise a valid point with atomicity. In the instruction set meeting we should discuss how to encode that. At a high level, the approach should most likely be:

```
Params -> InstructionSet -> XCM(karura) | EthAbi | XCM(basilisk)
```

Depending on the target chain, some instructions may be combined into a single atomic operation. For chains where that is not possible, a DAG should be constructed, specifying how rollbacks during failure should be handled.

c410-f3r:
I will close this PR but feel free to ask me anything

PR #341, opened 2021-12-03
[CU-1vqmed2] Oracle tweaks

JesseAbram:
I don't believe this requires any benchmark update. It does an extra loop, however that loop is already accounted for with removed_pre_prices_len. Also I am confident I call remove_price_in_transit in every location a price is removed but double checking that would be great too

### DIFF
--- a/frame/oracle/src/lib.rs
+++ b/frame/oracle/src/lib.rs
@@ -831,14 +831,13 @@ pub mod pallet {
 			let mut to32 = AccountId32::as_ref(&account);
 			let address: T::AccountId = T::AccountId::decode(&mut to32).unwrap_or_default();
 
-			if prices.clone().into_iter().any(|price| price.who == address) {
-				return Err("Tx already submitted")
-			}
-
 			if prices.len() as u32 >= Self::asset_info(price_id).max_answers {
 				return Err("Max answers reached")
 			}
 
+			if prices.into_iter().any(|price| price.who == address) {
+				return Err("Tx already submitted")
+			}
 			// Make an external HTTP request to fetch the current price.
 			// Note this call will block until response is received.
 			let price = Self::fetch_price(price_id).map_err(|_| "Failed to fetch price")?;
--- a/frame/oracle/src/lib.rs
+++ b/frame/oracle/src/lib.rs
@@ -486,7 +486,7 @@ pub mod pallet {
 			DeclaredWithdraws::<T>::remove(&signer);
 			T::Currency::unreserve(&signer, withdrawal.stake);
 			let result = T::Currency::transfer(&signer, &who, withdrawal.stake, AllowDeath);
-			ensure!(!result.is_err(), Error::<T>::TransferError);
+			ensure!(result.is_ok(), Error::<T>::TransferError);
 
 			ControllerToSigner::<T>::remove(&who);
 			SignerToController::<T>::remove(&signer);
--- a/frame/oracle/src/lib.rs
+++ b/frame/oracle/src/lib.rs
@@ -832,10 +832,12 @@ pub mod pallet {
 			let address: T::AccountId = T::AccountId::decode(&mut to32).unwrap_or_default();
 
 			if prices.len() as u32 >= Self::asset_info(price_id).max_answers {
+				log::info!("Max answers reached");
 				return Err("Max answers reached")
 			}
 
 			if prices.into_iter().any(|price| price.who == address) {
+				log::info!("Tx already submitted");
 				return Err("Tx already submitted")
 			}
 			// Make an external HTTP request to fetch the current price.
--- a/frame/oracle/src/lib.rs
+++ b/frame/oracle/src/lib.rs
@@ -322,6 +322,8 @@ pub mod pallet {
 		ArithmeticError,
 		/// Block interval is less then stale price
 		BlockIntervalLength,
+		/// There was an error transferring
+		TransferError
 	}
 
 	#[pallet::hooks]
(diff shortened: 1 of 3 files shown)


frame/oracle/src/lib.rs
```diff
@@ -790,10 +831,14 @@ pub mod pallet {
...
 
-			if prices.into_iter().any(|price| price.who == address) {
+			if prices.clone().into_iter().any(|price| price.who == address) {
 				return Err("Tx already submitted")
 			}
 
+			if prices.len() as u32 >= Self::asset_info(price_id).max_answers {
+				return Err("Max answers reached")
+			}
+
```
hussein-aitlahcen, line 841: If we swap the two checks we can avoid cloning can't we?

frame/oracle/src/lib.rs
```diff
@@ -502,18 +511,25 @@ pub mod pallet {
...
+					),
+				Error::<T>::NotEnoughStake
+			);
 
 			let set_price = PrePrice {
 				price,
 				block: frame_system::Pallet::<T>::block_number(),
 				who: who.clone(),
 			};
+			let asset_info = AssetsInfo::<T>::get(asset_id);
```
hussein-aitlahcen, line 527: It's probably not a big deal here but I think we should avoid using `ValueQuery` on this kind of storage as the `get` will return the `Default` implementation if I am not doing a mistake? This will be backdoor to bug when the code will evolve? This `ValueQuery` really look like the _null_ of substrate... Please correct me if I am wrong.

frame/oracle/src/lib.rs
```diff
@@ -591,13 +613,17 @@ pub mod pallet {
 					let controller = SignerToController::<T>::get(&answer.who)
 						.unwrap_or_else(|| answer.who.clone());
 
-					let _ = T::Currency::deposit_into_existing(&controller, reward_amount);
+					let result = T::Currency::deposit_into_existing(&controller, reward_amount);
+					if result.is_err() {
+						log::warn!("Failed to deposit {:?}", controller);
```
hussein-aitlahcen, line 618: Do we still need this?

hussein-aitlahcen (approved):
LGTM, quick comments

frame/oracle/src/lib.rs
JesseAbram: eh probably not, but it was an item in the audit report, and I can see how this would be a good practice

frame/oracle/src/lib.rs
JesseAbram: ooo ya good catch

frame/oracle/src/lib.rs
JesseAbram: I tend to do valueQuery by default and I believe parity does as well. If I understand correctly, I could be down to change them if you make a good case and have a potential example, but that would be a lot of code changes, as I would have to unwrap every instance and fix the tests and benchamrks too. So I would say down to do it after but not in scope of this PR

frame/oracle/src/lib.rs
hussein-aitlahcen: 👍

frame/oracle/src/lib.rs
```diff
@@ -790,10 +831,13 @@ pub mod pallet {
 			let mut to32 = AccountId32::as_ref(&account);
 			let address: T::AccountId = T::AccountId::decode(&mut to32).unwrap_or_default();
 
+			if prices.len() as u32 >= Self::asset_info(price_id).max_answers {
+				return Err("Max answers reached")
```
dzmitry-lahoda, line 835: how this error is handled?

JesseAbram: the error is generated off chain in an offchain worker, the result of no error is the node prepping and sending in a tx, the error is logged and results in no tx being prepped

dzmitry-lahoda: so you say that if error never handled neither logged it it still saved somewhere ?

image

dzmitry-lahoda:
i see neither readme nor lib docs outline how oracle behaves (like slashing temporary for several answers, stale period of answers, and if there are too many answers - ignore old).  nice, but it is hard to review without simple generics words about these things.

and what is clickup number for this work if that contain some description?

frame/oracle/src/lib.rs
JesseAbram: ahh ya I guess it wont added info loggings

ComposableFiB0t:
Task linked: [CU-1vqmed2 Apply to SubScan](https://app.clickup.com/t/1vqmed2)

frame/oracle/src/lib.rs
hussein-aitlahcen: Let's take the `submit_price` extrinsic, if an oracle is passing a non existing asset, should the extrinsic be successful? Unless am missed something, it look like it will be successful

JesseAbram: yes it will fail with max prices as the max is 0 if an asset is not set, but I see your point, will make a PR after this to fix it

hussein-aitlahcen (approved):

frame/oracle/src/lib.rs
```diff
@@ -477,7 +485,8 @@ pub mod pallet {
 			ensure!(block > withdrawal.unlock_block, Error::<T>::StakeLocked);
 			DeclaredWithdraws::<T>::remove(&signer);
 			T::Currency::unreserve(&signer, withdrawal.stake);
-			let _ = T::Currency::transfer(&signer, &who, withdrawal.stake, AllowDeath);
+			let result = T::Currency::transfer(&signer, &who, withdrawal.stake, AllowDeath);
+			ensure!(!result.is_err(), Error::<T>::TransferError);
```
andor0, line 489: `!result.is_err()` `==` `result.is_ok()`

andor0 (dismissed):

frame/oracle/src/lib.rs
JesseAbram: hahah ya, probably more readable lol

andor0 (approved):

hussein-aitlahcen (approved):

PR #350, opened 2021-12-07
Add Identity pallet

### LINKED ISSUE #331: Add Identity Pallet to Picasso

c410-f3r:
closes #331


runtime/picasso/Cargo.toml
```diff
@@ -37,6 +37,7 @@ balances = { package = "pallet-balances", git = "https://github.com/paritytech/s
 indices = { package = "pallet-indices", git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.12", default-features = false }
 randomness-collective-flip = { package = "pallet-randomness-collective-flip", git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.12", default-features = false }
 collective = { package = "pallet-collective", git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.12", default-features = false }
+identity = { package = "pallet-identity", git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.12", default-features = false }
```
c410-f3r, line 40: It is unusual to have this pattern even across parachain repositories. Auto generated benchmark code also does not play integrate well

runtime/picasso/src/lib.rs
```diff
@@ -199,6 +199,30 @@ impl aura::Config for Runtime {
...
 impl cumulus_pallet_aura_ext::Config for Runtime {}
 
+parameter_types! {
+	pub const BasicDeposit: Balance = 8 * PICA;
+	pub const FieldDeposit: Balance = 256 * MILLI_PICA;
+	pub const MaxAdditionalFields: u32 = 32;
+	pub const MaxRegistrars: u32 = 8;
+	pub const MaxSubAccounts: u32 = 32;
+	pub const SubAccountDeposit: Balance = 2 * PICA;
+}
```
c410-f3r, line 209: Feel free to change any parameter

runtime/picasso/src/lib.rs
```diff
@@ -199,6 +199,30 @@ impl aura::Config for Runtime {
...
+	type Event = Event;
+	type FieldDeposit = FieldDeposit;
+	type ForceOrigin = EnsureRoot<AccountId>;
+	type MaxAdditionalFields = MaxAdditionalFields;
+	type MaxRegistrars = MaxRegistrars;
+	type MaxSubAccounts = MaxSubAccounts;
+	type RegistrarOrigin = EnsureRoot<AccountId>;
+	type Slashed = Treasury;
+	type SubAccountDeposit = SubAccountDeposit;
+	type WeightInfo = weights::identity::WeightInfo<Runtime>;
```
c410-f3r, line 223: These parameters are sorted in alphabetical order

runtime/picasso/src/xcmp.rs
```diff
@@ -1,7 +1,6 @@
 //! Setup of XCMP for parachain to allow cross chain transfers and other operations.
 //! Very similar to https://github.com/galacticcouncil/Basilisk-node/blob/master/runtime/basilisk/src/xcm.rs
 #![allow(unused_imports)] // allow until v2 xcm released (instead creating 2 runtimes)
-#![cfg_attr(not(feature = "std"), no_std)]
```
c410-f3r, line 4: It does not make sense to have much compile configuration flag outside of `lib` or `main`.

runtime/picasso/Cargo.toml
dzmitry-lahoda: fuck. we are so bad.

c410-f3r: ???

It is not bad or good. Just a fact

dzmitry-lahoda (approved):

runtime/picasso/Cargo.toml
dzmitry-lahoda: i mean all 3rd party pallets should be exported without their weights?

dzmitry-lahoda: so we can measure these for our target hardware

c410-f3r: Maybe I wasn't specific enough.

It is not common to see `packages = { ... }` for all substrate/cumulus/polkadot dependencies and this feature makes auto generated code fail to compile because Parity assumed that imported dependencies have their canonical names. For example, I had to rename `frame_support` to `support` in the newly generated weights file of `Identity`.

Therefore, this situation is not related to pallets without weights or import of generated pallet weights

dzmitry-lahoda: oh, i see. weird renames. they hamper my copy pasted from other runtime. and there is another issue with weigths. :(

runtime/picasso/src/lib.rs
seunlanlege: add a TODO: we'll get to it later

seunlanlege (approved):

PR #364, opened 2021-12-09
Composable Runtime

seunlanlege:


runtime/picasso/src/lib.rs
```diff
@@ -90,8 +90,8 @@ pub mod opaque {
 //   https://substrate.dev/docs/en/knowledgebase/runtime/upgrades#runtime-versioning
 #[sp_version::runtime_version]
 pub const VERSION: RuntimeVersion = RuntimeVersion {
-	spec_name: create_runtime_str!("picasso"),
-	impl_name: create_runtime_str!("picasso"),
+	spec_name: create_runtime_str!("composable"),
+	impl_name: create_runtime_str!("composable"),
 	authoring_version: 1,
```
seunlanlege, line 95: 🤦🏾‍♂️

node/src/command.rs
```diff
@@ -144,8 +146,8 @@ macro_rules! construct_async_run {
 		let runner = $cli.create_runner($cmd)?;
 		runner.async_run(|$config| {
 			let $components = new_partial::<
-				RuntimeApi,
-				Executor,
+				picasso_runtime::RuntimeApi,
+				PicassoExecutor,
```
seunlanlege, line 150: we'll need to abstract away the client over any generic runtime, otherwise we can't execute the right sub commands. this is going to be more involved than i thought

node/src/client.rs
```diff
@@ -0,0 +1,269 @@
...
+	}
+}
+
+impl From<Arc<ComposableClient>> for Client {
+	fn from(client: Arc<ComposableClient>) -> Self {
+		Self::Composable(client)
+	}
+}
+
+macro_rules! match_client {
```
dzmitry-lahoda, line 82: there is crate which does this https://docs.rs/enum_dispatch/latest/enum_dispatch/ kind of

node/src/command.rs
```diff
@@ -42,14 +42,16 @@ fn load_spec(id: &str) -> std::result::Result<Box<dyn sc_service::ChainSpec>, St
 		// Must define the default chain here because `export-genesis-state` command
 		// does not support `--chain` and `--parachain-id` arguments simultaneously.
 		"picasso-dev" => Box::new(chain_spec::picasso_dev()),
-		// westend parachain
+		// Dali (Westend Relay)
```
dzmitry-lahoda, line 45: could these to be moved out into enum or constants and documented via /// , kind of clear network topology

node/src/client.rs
seunlanlege: can only use it if our crate defines the trait.

node/src/service.rs
```diff
@@ -56,53 +68,63 @@ impl sc_executor::NativeExecutionDispatch for Executor {
...
+			let components =
+				new_partial::<composable_runtime::RuntimeApi, ComposableExecutor>(config)?;
+			(
+				Arc::new(Client::from(components.client)),
+				components.backend,
+				components.import_queue,
+				components.task_manager,
+			)
+		},
+		_ => {
```
dzmitry-lahoda, line 93: could we encode Composable Picasso to make config for type safe?

runtime/composable/src/lib.rs
```diff
@@ -0,0 +1,930 @@
...
+				relay_chain_slot,
+				sp_std::time::Duration::from_secs(6),
+			)
+			.create_inherent_data()
+			.expect("Could not create the timestamp inherent data");
+
+		inherent_data.check_extrinsics(block)
+	}
+}
+
```
dzmitry-lahoda, line 925: could we have some document on how to maintain 2 runtimes? who will do it? may be reference on how other teams do it. 

could we share some setup code amid 2?

seunlanlege: the [polkadot repo](https://github.com/paritytech/polkadot/tree/master/runtime) maintains 4 runtimes. Best practices are lifted from there.

PR #376, opened 2021-12-13
Implement property test for add/remove_liquidity() and exchange().

vivekvpandya:
Also fix minor bug found with proptesting.

### DIFF
--- a/frame/curve-amm/src/tests.rs
+++ b/frame/curve-amm/src/tests.rs
@@ -3,11 +3,44 @@ use frame_support::assert_ok;
 use crate::mock::*;
 use composable_traits::dex::CurveAmm as CurveAmmTrait;
 use frame_support::traits::fungibles::{Inspect, Mutate};
+use proptest::prelude::*;
 use sp_runtime::{
+	helpers_128bit::multiply_by_rational,
 	traits::{Saturating, Zero},
 	FixedPointNumber, FixedU128, Permill,
 };
 use sp_std::cmp::Ordering;
+
+/// Accepts -2, -1, 0, 1, 2
+macro_rules! prop_assert_zero_epsilon {
+	($x:expr) => {{
+		let epsilon = 2;
+		let upper = 0 + epsilon;
+		let lower = 0;
+		prop_assert!(upper >= $x && $x >= lower, "{} => {} >= {}", upper, $x, lower);
+	}};
+}
+
+/// Accept a 'dust' deviation
+macro_rules! prop_assert_epsilon {
+	($x:expr, $y:expr) => {{
+		let precision = 1000;
+		let epsilon = 5;
+		let upper = precision + epsilon;
+		let lower = precision - epsilon;
+		let q = multiply_by_rational($x, precision, $y).expect("qed;");
+		prop_assert!(
+			upper >= q && q >= lower,
+			"({}) => {} >= {} * {} / {} >= {}",
+			q,
+			upper,
+			$x,
+			precision,
+			$y,
+			lower
+		);
+	}};
+}
 
 #[test]
 fn compute_d_works() {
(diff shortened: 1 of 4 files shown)


frame/curve-amm/src/tests.rs
```diff
@@ -3,12 +3,45 @@ use frame_support::assert_ok;
...
+use proptest::prelude::*;
 use sp_runtime::{
+	helpers_128bit::multiply_by_rational,
 	traits::{Saturating, Zero},
 	FixedPointNumber, FixedU128, Permill,
 };
 use sp_std::cmp::Ordering;
 
+/// Accepts -2, -1, 0, 1, 2
+macro_rules! prop_assert_zero_epsilon {
```
dzmitry-lahoda, line 15: afaik @hussein-aitlahcen moved that to reusable codes

frame/curve-amm/src/tests.rs
```diff
@@ -229,3 +262,126 @@ fn exchange_test() {
 		assert!(65000 - Tokens::balance(MockCurrencyId::USDC, &CHARLIE) < 10);
 	});
 }
+
+proptest! {
+	#![proptest_config(ProptestConfig::with_cases(10000))]
+
+	#[test]
+	fn proptest_exchange(alice_balance in 1..u32::MAX,
```
dzmitry-lahoda, line 270: good,so test to test that curve is balanced would be awesome (kind of if make inbalance, interset grows fast or like)

dzmitry-lahoda (dismissed):

dzmitry-lahoda (approved):

frame/curve-amm/src/tests.rs
hussein-aitlahcen: Not yet in main sadly

frame/curve-amm/src/tests.rs
```diff
@@ -229,3 +262,126 @@ fn exchange_test() {
...
+		assert_eq!(Tokens::balance(MockCurrencyId::USDC, &ALICE), 0);
+		let pool = CurveAmm::pool(pool_id);
+		assert!(pool.is_some());
+		let bob_amounts = vec![bob_balance as u128, bob_balance as u128];
+		assert_ok!(CurveAmm::add_liquidity(&BOB, pool_id, bob_amounts.clone(), 0u128));
+		let bob_lp_balance = Tokens::balance(pool_lp_asset, &BOB);
+		assert_ne!(bob_balance, 0);
+		assert_eq!(Tokens::balance(MockCurrencyId::USDT, &BOB), 0);
+		assert_eq!(Tokens::balance(MockCurrencyId::USDC, &BOB), 0);
+		let min_amt = vec![0u128, 0u128];
```
KaiserKarel, line 369: Only nitpick I have here is that this is very difficult to follow, if you could create groups like this, and explain what you are checking/what stage of the procedure we're in

``
let a = 1;

// Test that arithmetic works
{
   assert_eq!(1, 1);
}

// Test that it works for zero
{ 
   assert_neq!(a, 0);
}
```

**Review by KaiserKarel — CHANGES_REQUESTED** [review]
If you could add some comments in the very big test, perhaps even grouping stages, that'd be great

File: frame/curve-amm/src/tests.rs (hunk 2 above)
Comment (vivekvpandya) [reply frame/curve-amm/src/tests.rs:270]: sure, I have noted this. I will try it in next PR.

File: frame/curve-amm/src/tests.rs [hunk 4]
```diff
@@ -229,3 +262,149 @@ fn exchange_test() {
...
+		assert_ok!(CurveAmm::exchange(&CHARLIE, pool_id, 1, 0, alice_balance as u128, 0));
+		prop_assert_epsilon!(alice_balance as u128, Tokens::balance(MockCurrencyId::USDC, &CHARLIE));
+		Ok(())
+	}).unwrap();
+	}
+
+	#[test]
+	fn proptest_add_remove_liquidity(alice_balance in 0..u32::MAX,
+									 bob_balance in 0..u32::MAX) {
+	new_test_ext().execute_with(|| {
```
KaiserKarel, line 339: weirdly formatted?

KaiserKarel (approved):

frame/curve-amm/src/tests.rs
vivekvpandya: On my system vim and vscode shows this line properly aligned.

dzmitry-lahoda (approved):

PR #389, opened 2021-12-15
Democracy

c410-f3r:
As requested, `democracy-redux` was squashed and re-based from `main` along side the replace of `OriginFor` with `GovernanceRegistry`.

This PR is not ready yet because `main` was two versions of Polkadot/Substrate/Cumulus causing internal type conflicts


c410-f3r:
`cargo build` and `cargo check --tests` are working and now it is only a matter of making tests green.

It seems that `democracy-redux` was mostly a copy and paste of the canonical Substrate `democracy` pallet. If so, it might be worth considering using git modules or git trees to improve future synchronization and maintainability.

hussein-aitlahcen:
> `cargo build` and `cargo check --tests` are working and now it is only a matter of making tests green.
> 
> It seems that `democracy-redux` was mostly a copy and paste of the canonical Substrate `democracy` pallet. If so, it might be worth considering using git modules or git trees to improve future synchronization and maintainability.

This version is implementing multi-assets voting capability while the original is not supporting this feature

c410-f3r:
Tests are passing now

@aettran Internal dispatches were calling `set_balance` but such method requires `ensure_root`, thus the source of all the filtered calls. To fix this problem, `GovernanceRegistry::set` was replaced by `GovernanceRegistry::grant_root`

KaiserKarel (approved):

hussein-aitlahcen (approved):

PR #397, opened 2021-12-16
[CU-1p8545r] auction without dex api

dzmitry-lahoda:


ComposableFiB0t:
Task linked: [CU-1p8545r auctions: auction without dex implementation](https://app.clickup.com/t/1p8545r)

PR #398, opened 2021-12-16
CU-1wef6he - refactor call-filter

hussein-aitlahcen:
We were previously using a map `Entry -> Option<()>` but a StorageMap `Entry -> ()`
is enough as it uses `OptionQuery` by default.

I added a way to hook into enable/disable call of the filter pallet to make it
more flexible. If we want to block a pallet from being disabled, it can now be
done at the Runtime declaration.

I also extracted the whole behavior to its own traits.

### DIFF
--- a/frame/composable-traits/src/call_filter.rs
+++ b/frame/composable-traits/src/call_filter.rs
@@ -1,0 +1,40 @@
+use frame_support::{dispatch::DispatchResult, pallet_prelude::*};
+use sp_std::vec::Vec;
+
+/// An object that is able to tell us whether an entry can or cannot be disabled.
+pub trait CallFilterHook {
+	fn enable_hook(entry: &CallFilterEntry) -> DispatchResult;
+	fn disable_hook(entry: &CallFilterEntry) -> DispatchResult;
+}
+
+impl CallFilterHook for () {
+	#[inline(always)]
+	fn enable_hook(_: &CallFilterEntry) -> DispatchResult {
+		Ok(())
+	}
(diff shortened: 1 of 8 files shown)


ComposableFiB0t:
Task linked: [CU-1wef6he Pallet-Call-Filter minor improvements](https://app.clickup.com/t/1wef6he)

frame/composable-traits/src/call_filter.rs
```diff
@@ -0,0 +1,40 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode, TypeInfo)]
+pub struct CallFilterEntry {
+	pub pallet_name: Vec<u8>,
+	pub function_name: Vec<u8>,
+}
+
+impl CallFilterEntry {
+	pub fn valid(&self) -> bool {
+		sp_std::str::from_utf8(&self.pallet_name).is_ok() &&
+			sp_std::str::from_utf8(&self.function_name).is_ok()
```
hussein-aitlahcen, line 38: We weren't checking if the `function_name` was a valid UTF8 sequence, I'm not sure why so I added it, please double check if this is required.

andor0 (approved):

frame/composable-traits/src/call_filter.rs
KaiserKarel: @seunlanlege

hussein-aitlahcen: It in fact was needed

PR #401, opened 2021-12-16
CU-1wtx9fb Implement Constant Product AMM.

vivekvpandya:


ComposableFiB0t:
Task linked: [CU-1wtx9fb Univ2](https://app.clickup.com/t/1wtx9fb)

frame/composable-traits/src/dex.rs
```diff
@@ -181,7 +181,7 @@ pub type PoolId = u32;
 
 /// Pool type
 #[derive(Encode, Decode, TypeInfo, Clone, Default, PartialEq, Eq, Debug)]
-pub struct PoolInfo<AccountId> {
+pub struct StableSwapPoolInfo<AccountId> {
 	/// Owner of pool
 	pub owner: AccountId,
```
dzmitry-lahoda, line 186: is it viable to unify owner and admin_fee under one struct? so it can be reused?

frame/curve-amm/src/lib.rs
```diff
@@ -824,6 +823,21 @@ pub mod pallet {
...
+		/// Return pool information for given pool_id.
+		pub fn get_pool_info(pool_id: PoolId) -> Option<StableSwapPoolInfo<T::AccountId>> {
+			let pool_data = Pools::<T>::get(pool_id)?;
+			let pool;
+			match pool_data {
+				PoolData::StableSwapPoolData(pool_info) => {
+					pool = pool_info;
+				},
+				_ => {
+					panic!("Only stable swap pool expected.");
```
dzmitry-lahoda, line 835: could we avoid this panic here? it feels not good for stable to be aware of other curves?

frame/uniswap-v2/src/lib.rs
```diff
@@ -0,0 +1,863 @@
...
+			pool_id: PoolId,
+			admin_fee_account: T::AccountId,
+			admin_fees: Vec<T::Balance>,
+		},
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {}
+
+	impl<T: Config> CurveAmm for Pallet<T> {
```
dzmitry-lahoda, line 304: please split curves code

frame/curve-amm/src/lib.rs
vivekvpandya: should it generate Error?

frame/uniswap-v2/src/lib.rs
```diff
@@ -0,0 +1,863 @@
...
+		/// Return pool information for given pool_id.
+		pub fn get_pool_info(pool_id: PoolId) -> Option<ConstantProductPoolInfo<T::AccountId>> {
+			let pool_data = Pools::<T>::get(pool_id)?;
+			let pool;
+			match pool_data {
+				PoolData::ConstantProductPoolData(pool_info) => {
+					pool = pool_info;
+				},
+				_ => {
+					panic!("Only constant product pool expected.");
```
dzmitry-lahoda, line 781: not sure why it is needed here

frame/uniswap-v2/src/lib.rs
vivekvpandya: Don't get this comment. Both models are curve. So CurveAmm is generic traits for DEX based on curve. Although I have planned few refactoring in next PR.

frame/uniswap-v2/src/lib.rs
vivekvpandya: Need to handle all variants of Enum right?

frame/uniswap-v2/src/lib.rs
```diff
@@ -0,0 +1,863 @@
...
+
+	/// Pool's LP asset
+	#[pallet::storage]
+	#[pallet::getter(fn pool_lp_asset)]
+	pub type PoolLPAsset<T: Config> = StorageMap<_, Blake2_128Concat, PoolId, T::AssetId>;
+
+	/// List of assets supported by the pool
+	#[pallet::storage]
+	#[pallet::getter(fn pool_assets)]
+	pub type PoolAssets<T: Config> = StorageMap<_, Blake2_128Concat, PoolId, Vec<T::AssetId>>;
```
dzmitry-lahoda, line 152: should there be 2 assets?

frame/uniswap-v2/src/lib.rs
```diff
@@ -0,0 +1,863 @@
...
+		},
+		FixedPointNumber, FixedPointOperand, FixedU128, KeyTypeId as CryptoKeyTypeId, Permill,
+	};
+	use sp_std::{collections::btree_set::BTreeSet, fmt::Debug, iter::FromIterator};
+
+	pub const PALLET_ID: PalletId = PalletId(*b"Uni-swap");
+	pub const KEY_TYPE: KeyTypeId = KeyTypeId(*b"uswp");
+	pub const CRYPTO_KEY_TYPE: CryptoKeyTypeId = CryptoKeyTypeId(*b"uswp");
+
+	pub mod crypto {
```
dzmitry-lahoda, line 62: not sure, is it used somewhere or will be?

frame/uniswap-v2/src/lib.rs
vivekvpandya: Yes, It will be done in next PR where I do refactoring.

frame/uniswap-v2/src/lib.rs
vivekvpandya: I hope it will be used. The current code is not complete pallet as it still requires extrinsics. It may need OCW. if not used in final version of pallet we can remove these.

frame/curve-amm/src/lib.rs
hussein-aitlahcen: we MUST NEVER panic

hussein-aitlahcen (changes requested):
Can we have a very simple implementation completely separated from the stableswap one? I don't see any benefit in unifying the pool data via an enum as we have two different pallets. Also, you are not even using the same storage to store the enum as you have two pallets. Meaning that you are basically splitting the enum you just created and panicking if you don't have the expected variant. Let's keep it simple and then create a router pallet on top of both later.

vivekvpandya:
> 

Okay, then can you provide sketch of common trait to be shared between two models? If we don't want to share the trait then I can write a separate trait. Also I don;t think the trait can be any simpler than this. i.e add/remove_liquidity and exchange.

hussein-aitlahcen:
> > 
> 
> Okay, then can you provide sketch of common trait to be shared between two models? If we don't want to share the trait then I can write a separate trait. Also I don;t think the trait can be any simpler than this. i.e add/remove_liquidity and exchange.

The trait is fine for now, the issue is that the pallet are sharing the same pooldata enum for instance, which we don't want them to share because they are SEPARATE pallets. So let's keep this simple with each pool having its own variant directly stored rather tha having the enum stored

frame/composable-traits/src/dex.rs
vivekvpandya: I don't see a huge benefit. if required it can be done later.

frame/uniswap-v2/src/lib.rs
```diff
@@ -0,0 +1,847 @@
...
+
+	/// Pool's LP asset
+	#[pallet::storage]
+	#[pallet::getter(fn pool_lp_asset)]
+	pub type PoolLPAsset<T: Config> = StorageMap<_, Blake2_128Concat, PoolId, T::AssetId>;
+
+	/// List of assets supported by the pool
+	#[pallet::storage]
+	#[pallet::getter(fn pool_assets)]
+	pub type PoolAssets<T: Config> = StorageMap<_, Blake2_128Concat, PoolId, Vec<T::AssetId>>;
```
hussein-aitlahcen, line 153: Any chance we use `BoundedVec` here?

frame/uniswap-v2/src/tests.rs
```diff
@@ -0,0 +1,148 @@
...
+		);
+		assert_eq!(
+			Tokens::balance(MockCurrencyId::USDT, &ConstantProductAmm::account_id(&pool_id)),
+			0
+		);
+	});
+}
+
+#[test]
+fn exchange_test() {
```
hussein-aitlahcen, line 97: The tests are too big to be understandable. We should cut this kind of test or comment some part to make it more readable. I personnally think we should cut this into a lot of scenarios.

frame/uniswap-v2/src/lib.rs
hussein-aitlahcen: From what I understand, the uniswap pallet don't need any OCW, we can remove this.

frame/composable-traits/src/dex.rs
```diff
@@ -117,9 +117,6 @@ pub trait CurveAmm {
 	/// Current number of pools (also ID for the next created pool)
 	fn pool_count() -> PoolId;
```
hussein-aitlahcen, line 118: Can we have the poolId generic and provided through the pallet?

hussein-aitlahcen (changes requested):

frame/uniswap-v2/src/lib.rs
vivekvpandya: We don't need it, in next PR I will make it (token-X, token-Y) pair.

vivekvpandya: We want to only support pool with 2 assets.

frame/uniswap-v2/src/tests.rs
vivekvpandya: Added comments

hussein-aitlahcen (approved):

frame/composable-traits/src/dex.rs
```diff
@@ -191,3 +178,11 @@ pub struct PoolInfo<AccountId> {
 	/// Amount of the admin fee pool charges for the exchange
 	pub admin_fee: Permill,
 }
+
+#[derive(Encode, Decode, TypeInfo, Clone, Default, PartialEq, Eq, Debug)]
+pub struct ConstantProductPoolInfo<AccountId> {
```
dzmitry-lahoda, line 183: it is more of Configuraiton kind of. Info may include current exchnage rate.

frame/curve-amm/src/lib.rs
```diff
@@ -124,6 +124,17 @@ pub mod pallet {
 			+ Mutate<Self::AccountId, Balance = Self::Balance, AssetId = <Self as Config>::AssetId>
 			+ Inspect<Self::AccountId, Balance = Self::Balance, AssetId = <Self as Config>::AssetId>;
 		type Precision: Get<FixedU128>;
+		type PoolId: FullCodec
```
dzmitry-lahoda, line 127: I would prefer it is encoded like this 

pub trait AssetIdLike: FullCodec + Copy + Eq + PartialEq + Debug + TypeInfo {}
impl<T: FullCodec + Copy + Eq + PartialEq + Debug + TypeInfo> AssetIdLike for T {}

frame/curve-amm/src/lib.rs
```diff
@@ -133,30 +144,31 @@ pub mod pallet {
...
 	#[pallet::storage]
 	#[pallet::getter(fn pool_lp_asset)]
-	pub type PoolLPAsset<T: Config> = StorageMap<_, Blake2_128Concat, PoolId, T::AssetId>;
+	pub type PoolLPAsset<T: Config> = StorageMap<_, Blake2_128Concat, T::PoolId, T::AssetId>;
 
 	/// List of assets supported by the pool
 	#[pallet::storage]
 	#[pallet::getter(fn pool_assets)]
-	pub type PoolAssets<T: Config> = StorageMap<_, Blake2_128Concat, PoolId, Vec<T::AssetId>>;
+	pub type PoolAssets<T: Config> = StorageMap<_, Blake2_128Concat, T::PoolId, Vec<T::AssetId>>;
```
dzmitry-lahoda, line 163: could you leave comment if other pools on substratre also have Vec, not BoundedVec in list?

frame/curve-amm/src/lib.rs
```diff
@@ -824,6 +769,73 @@ pub mod pallet {
...
 	impl<T: Config> Pallet<T> {
+		pub fn create_pool(
+			who: &T::AccountId,
+			assets: Vec<T::AssetId>,
+			amplification_coefficient: FixedU128,
+			fee: Permill,
+			admin_fee: Permill,
+		) -> Result<T::PoolId, DispatchError> {
+			// Assets related checks
+			ensure!(assets.len() > 1, Error::<T>::NotEnoughAssets);
```
dzmitry-lahoda, line 780: Vec is unbounded. so if we will be success and have as many assets as FTX, it can easy be 1000 of assets. what is plan to have this method secured in terms of weight? seems as count of assets grows, grows price of this method with no limit.

frame/curve-amm/src/lib.rs
```diff
@@ -824,6 +769,73 @@ pub mod pallet {
...
+			let unique_assets = BTreeSet::<T::AssetId>::from_iter(assets.iter().copied());
```
dzmitry-lahoda, line 781: is unique_assets sorted? to avoid address to create same pool with different order?

dzmitry-lahoda: could sort assets and hash. can check if such pools exists from specific wallet. may be.

frame/curve-amm/src/lib.rs
```diff
@@ -824,6 +769,73 @@ pub mod pallet {
 	}
 
 	impl<T: Config> Pallet<T> {
+		pub fn create_pool(
```
dzmitry-lahoda, line 772: need not to forget make it transactional in pallet

frame/curve-amm/src/lib.rs
```diff
@@ -658,20 +604,19 @@ pub mod pallet {
...
-			pool_id: PoolId,
-			i_token: PoolTokenIndex,
-			j_token: PoolTokenIndex,
+			pool_id: T::PoolId,
+			i_token: T::PoolTokenIndex,
+			j_token: T::PoolTokenIndex,
 			dx: Self::Balance,
 			min_dy: Self::Balance,
 		) -> Result<(), DispatchError> {
 			let prec = T::Precision::get();
```
dzmitry-lahoda, line 613: what will be weight of exchange? if it iterats over vec, so with time it can became very big

dzmitry-lahoda: how curve on substrate solves that?

frame/uniswap-v2/src/lib.rs
```diff
@@ -0,0 +1,857 @@
...
+		/// Values in the storage are inconsistent
+		InconsistentStorage,
+		/// Not enough assets provided
+		NotEnoughAssets,
+		/// Some provided assets are not unique
+		DuplicateAssets,
+		/// Pool with specified id is not found
+		PoolNotFound,
+		/// Error occurred while performing math calculations
+		Math,
```
dzmitry-lahoda, line 208: could we be mothe specific using Arithmetic Under/Over-flow or DivisionByZero?

frame/uniswap-v2/src/lib.rs
```diff
@@ -0,0 +1,857 @@
...
+		/// Specified asset amount is wrong
+		WrongAssetAmount,
```
dzmitry-lahoda, line 210: It is not clear what is wrong, like is it not enough to do some operation or not balanced or it does not matches some exact amount of something? like error should be more clear about what is exactly wrong and guide what to do

frame/uniswap-v2/src/lib.rs
```diff
@@ -0,0 +1,857 @@
...
+		}
+
+		fn add_liquidity(
+			who: &Self::AccountId,
+			pool_id: T::PoolId,
+			amounts: Vec<Self::Balance>,
+			min_mint_amount: Self::Balance,
+		) -> Result<(), DispatchError> {
+			let zero = Self::Balance::zero();
+			ensure!(amounts.len() == 2, Error::<T>::IndexOutOfRange);
```
dzmitry-lahoda, line 334: not clear what index out of range means. it is seems like you encode Exact2Values type via array. can we encode it into types than.

most of swaps are in pools of 2, not in many. so having separate interface will be more clear and reusable. also it will fix potential security issuew ith Vec

frame/uniswap-v2/src/lib.rs
```diff
@@ -0,0 +1,857 @@
...
+			let token_supply_u: u128 = token_supply.into();
+			let token_supply_f = FixedU128::saturating_from_integer(token_supply_u);
+			let mint_amount;
+			if token_supply_u == 0 {
+				let mint_amount_f = amount_a.checked_mul(&amount_b).ok_or(Error::<T>::Math)?;
+				let mint_amount_b: T::Balance =
+					mint_amount_f.checked_mul_int(1u64).ok_or(Error::<T>::Math)?.into();
+				mint_amount = mint_amount_b.integer_sqrt();
+			} else {
+				let mint_a = amount_a
```
dzmitry-lahoda, line 397: it is pure math, can we move it into math.rs?

frame/uniswap-v2/README.md
```diff
@@ -0,0 +1 @@
+Uniswap-v2 pallet which has constant-product AMM alogrithm as explained in https://hackmd.io/@HaydenAdams/HJ9jLsfTz?type=view
```
dzmitry-lahoda, line 1: please add here other papers you have read

frame/composable-traits/src/dex.rs
vivekvpandya: Do you want to rename it to Constant..Configuration ?

frame/curve-amm/src/lib.rs
vivekvpandya: I think https://github.com/equilibrium-eosdt/equilibrium-curve-amm/tree/master/pallets/equilibrium-curve-amm/src

frame/uniswap-v2/README.md
vivekvpandya: Uniswap-v2 white paper does not cover much details about implementation. I have refered solidity code of Uniswap-V2. I will add links to that.

frame/uniswap-v2/src/lib.rs
vivekvpandya: Is it okay I do a separate PR for that?

frame/uniswap-v2/src/lib.rs
vivekvpandya: Agreed , will remove Vec in next PR and hopefully it will address this concern.

frame/uniswap-v2/src/lib.rs
vivekvpandya: sure, I prefer a separate PR.

frame/curve-amm/src/lib.rs
vivekvpandya: https://github.com/equilibrium-eosdt/equilibrium-curve-amm/tree/master/pallets/equilibrium-curve-amm/src this may have some answers on that.

vivekvpandya: but we are anyway going to get rid of Vec.

frame/curve-amm/src/lib.rs
vivekvpandya: it is based on BTreeSet so it will prevent case that you are concerned with.

frame/curve-amm/src/lib.rs
vivekvpandya: I will remove Vec in next PR.

frame/uniswap-v2/src/lib.rs
vivekvpandya: Based on its usage , it indicates that asset amounts should be grater than zero.

vivekvpandya: I will update it.

frame/composable-traits/src/dex.rs
vivekvpandya: So @dzmitry-lahoda you prefer it to be renamed as `ConstantProductConfig` ?

frame/curve-amm/src/lib.rs
vivekvpandya: What is benefit? Also if we follow that convention we may want to change every other pallets too.

frame/curve-amm/src/lib.rs
vivekvpandya: Sure, `pallet::call` not implemented yet.

frame/composable-traits/src/dex.rs
vivekvpandya: @hussein-aitlahcen any preferene?

dzmitry-lahoda (approved):

hussein-aitlahcen (approved):

PR #404, opened 2021-12-16
Architecture.md

hussein-aitlahcen:
Please review and add more content if needed!


dzmitry-lahoda:
image can be replaced by copy text from console may be

dzmitry-lahoda:
oh
sorry. pushed thing. was trying to create `suggest edit` from VS Code. Let me fix that

PR #410, opened 2021-12-17
CU-1p8545r  dutch auction

dzmitry-lahoda:


ComposableFiB0t:
Task linked: [CU-1p8545r auctions: auction without dex implementation](https://app.clickup.com/t/1p8545r)

frame/composable-traits/src/loans.rs
```diff
@@ -34,37 +24,8 @@ impl<GroupId, Balance> PriceStructure<GroupId, Balance> {
...
-	type Currency: Transfer<Self::AccountId, Balance = Self::Balance, AssetId = Self::AssetId>
-		+ Mutate<Self::AccountId, Balance = Self::Balance, AssetId = Self::AssetId>
-		// used to check balances before any storage updates allowing acting without rollback
-		+ Inspect<Self::AccountId, Balance = Self::Balance, AssetId = Self::AssetId>;
-}
+// /// bank. vault owned - can transfer, cannot mint
+// type Currency: Transfer<Self::AccountId, Balance = Self::Balance, AssetId = Self::AssetId>
+// 	+ Mutate<Self::AccountId, Balance = Self::Balance, AssetId = Self::AssetId>
+// 	// used to check balances before any storage updates allowing acting without rollback
+// 	+ Inspect<Self::AccountId, Balance = Self::Balance, AssetId = Self::AssetId>;
```
andor0, line 31: did you forget to remove it?

frame/dutch-auction/src/lib.rs
```diff
@@ -1,313 +1,266 @@
...
-						if now > order.started + ONE_HOUR {
-							removed.push(order_id);
-						}
-					},
-					_ => {
-						removed.push(order_id);
-					},
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_finalize(_n: T::BlockNumber) {
```
andor0, line 208: Looks like can use `on_initialize` instead of `on_finalize` here.

frame/liquidations/README.md
```diff
@@ -1,14 +1,12 @@
-## Liquidation
+# Overview
 
 Used to liquidate collateral from `lending` protocol.  Need to liquidate because of bad collateral factor. It should be sold on exchange as soon as possible.
 
-# Overview
-
-- off-chain worker monitors `lending` markets for liquidation opportunities of borrow pairs.
+- off-chain bots monitors `lending` markets for liquidation opportunities of borrow pairs.
```
andor0, line 5: `monitor`, because is plural here.

frame/dutch-auction/src/lib.rs
dzmitry-lahoda: added comment

dzmitry-lahoda:
oh, one hole - need to verify that order about existing assets.  another issue - too many orders. also we do not loop over these, but they are stored. somehow need to kill dead orders - not this PR.

frame/dutch-auction/src/lib.rs
```diff
@@ -150,6 +167,30 @@ pub mod pallet {
...
+				if order.from_to == who {
+					// weights fees are of platform spam protection, so we do not interfere with
+					// this function but on pallet level, we allow "fee less" liquidation by owner
+					// we can later allow liquidate old orders(or orders with some block liquidation
+					// timeout set) using kind of account per order is possible, but may risk to
+					// pollute account system
+					let treasury = &T::PalletId::get().into_account();
+					T::NativeCurrency::transfer(
+						&order.from_to,
+						treasury,
```
hussein-aitlahcen, line 184: If I understand correctly, we allow someone to cancel an order, in this case, shouldn't we swap the two accounts here because we repay (treasury -> from_to)?

frame/dutch-auction/src/lib.rs
```diff
@@ -1,313 +1,311 @@
...
-				_ => Ok(()),
-			})
+				}
+
+				if order.take.amount == T::Balance::zero() {
+					<SellOrders<T>>::remove(order_id);
+					Self::deposit_event(Event::OrderRemoved { order_id });
+				}
+			}
+			<Takes<T>>::remove_all(None);
```
hussein-aitlahcen, line 288: Quick question: why not let the user directly execute the take in the extrinsic transaction without having a hook?

hussein-aitlahcen (dismissed):
Not blocking comments

frame/dutch-auction/src/lib.rs
```diff
@@ -150,6 +167,30 @@ pub mod pallet {
...
+					T::NativeCurrency::transfer(
+						&order.from_to,
+						treasury,
+						T::WeightInfo::liquidate().into(),
+						true,
+					)?;
+					<SellOrders<T>>::remove(order_id);
+					Self::deposit_event(Event::OrderRemoved { order_id });
+				}
+			}
```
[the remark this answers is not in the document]
dzmitry-lahoda: oh yes. bug. will fix

frame/dutch-auction/src/lib.rs
dzmitry-lahoda: oh yes. bug. will fix

frame/dutch-auction/src/lib.rs
```diff
@@ -1,313 +1,268 @@
...
+		StorageMap<_, Twox64Concat, OrderIdOf<T>, SellOf<T>, OptionQuery>;
 
-	impl<T: Config + DeFiComposableConfig> DutchAuction for Pallet<T> {
-		type AccountId = T::AccountId;
+	/// one block storage, users payed N * WEIGHT for this Vec, so will not put bound here (neither
+	/// HydraDX does)
+	#[pallet::storage]
+	#[pallet::getter(fn takes)]
+	pub type Takes<T: Config> =
+		StorageMap<_, Twox64Concat, OrderIdOf<T>, Vec<TakeOf<T>>, OptionQuery>;
```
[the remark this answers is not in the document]
dzmitry-lahoda: what bound to set?

frame/dutch-auction/src/lib.rs
dzmitry-lahoda: - hydra and other books work like that
- you can predict price, block and start sending minimal price as soon as it close. will get minimal amount for collateral.
- but if on some step, several people will decide that it worth it, and they will know that highest bid wins, they will be able to pay little more to overbet. price will be more optimal than predicted by auction.
- if to take directly, than bot in same data center as we host will win bets.
- also i have heard transaction tip to be first, may be it can work too. so it will be tip to composable, not to dutch auction or lending. and need to read more how tip works.

frame/dutch-auction/src/lib.rs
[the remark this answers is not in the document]
dzmitry-lahoda: - hyrda does that
- vec is not stored in blockchain
- it is per order
- so you can spam one order with many tiny payments, but you will buy order in many tiny bits. take price was payed.
- could actually control not number of orders, bart part of order this takes one. either take all or at leas X% of it. but that also question what X% is?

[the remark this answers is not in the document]
dzmitry-lahoda: @hussein-aitlahcen what do you think? I just writing tests, need to decide if there is hole or not.

frame/dutch-auction/src/lib.rs
dzmitry-lahoda: @hussein-aitlahcen what do you think?

frame/dutch-auction/src/lib.rs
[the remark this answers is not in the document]
hussein-aitlahcen: If it's per block I assume it's fine then :)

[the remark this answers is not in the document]
hussein-aitlahcen: Since weight is paid, not sure we need to put a threshold for a minimum %

frame/dutch-auction/src/lib.rs
hussein-aitlahcen: makes sense, thanks for clarifying

frame/dutch-auction/src/lib.rs
KaiserKarel: have we added a unitest to check this bug?

dzmitry-lahoda: i am testing and fixing bug in `dz/weights` branch.

frame/composable-traits/src/auction.rs
```diff
@@ -20,97 +17,17 @@ impl Default for AuctionStepFunction {
...
-	/// cannot retry within current state of system, like assets are not supported
-	FatalFail,
-}
-
-#[derive(Default, Decode, Encode, Clone, TypeInfo)]
+#[derive(Default, Decode, Encode, Clone, TypeInfo, Debug, PartialEq)]
 pub struct LinearDecrease {
-	/// The number of seconds until the price reach zero.
+	/// Seconds after auction start when the price reaches zero
 	pub total: DurationSeconds,
```
KaiserKarel, line 23: Should we express this in blocknumbers, as seconds are less reliable.

frame/dutch-auction/src/lib.rs
dzmitry-lahoda: and another bug with returning not satisfied takes reservers to accounts

KaiserKarel (approved):

frame/dutch-auction/src/lib.rs
KaiserKarel: So we'll merge this, disable angular in cfg(...), and fix bugs in separate PR?

frame/composable-traits/src/auction.rs
dzmitry-lahoda: i do not know. i know that we use seconds in lending extensively. also I assume that seconds are synced in consensus. 

how predictable blocks spans are. if all blocks are exact 6 seconds. then there is no difference. 

if block can 5 or 7, or other kind of random, or consistenly skewed to 6.5 or runtime upgraded changes block time. than seconds are more "reliable" in terms of predictability of total auction duration i believe. so need to look into impl of Time pallet - it may be oracle with voting:) 

seconds are more intuitive, but for people it harder to target specific block number of sell with them. seconds are reliable for change in block time. but if block times are veryyy precies and exact, than second are not better.

frame/dutch-auction/src/lib.rs
dzmitry-lahoda: not sure how better. 

disabling angular from runtime if it was enabled will be in separate pr for sure. for now Dutch auction is not called from Lending

frame/dutch-auction/src/lib.rs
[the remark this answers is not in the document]
dzmitry-lahoda: fixed in branch

frame/composable-traits/src/auction.rs
KaiserKarel: Then we'll keep it in seconds; I don't think it affects security.

frame/dutch-auction/src/lib.rs
dzmitry-lahoda: i will merge branch here.  fixed bugs.

vivekvpandya (dismissed):

dzmitry-lahoda:
ok, my fix of main did not worked. will wait for more clean up from @seunlanlege

vivekvpandya (approved):

KaiserKarel (approved):

PR #420, opened 2021-12-23
CU-1pe15hj: Clippy additions and fixes

benluelo:
This PR has several fixes/ additions:

`clippy::disallowed_method` has been added to every crate, with a top-level clippy.toml file containing the disallowed methods. Currently it only contains Result/Option::unwrap, but it can be easily added to in the future. (`expect` is one that should be added as well; I figured it would be best to start with `unwrap`). I have removed/ audited/ fixed uses of unwrap where I was able to, however for most of them I was unable to figure out what was going on. There are several instances of `some_method_checked().unwrap()`, which doesn't make sense to me; perhaps the not-checked/`unckecked` versions of said functions either don't exist or have different behaviour than the checked functions followed by an `unwrap`? If anyone has any insight into this and could explain it would be greatly appreciated!

`clippy::indexing_slicing` has been added to all crates as well. _Most_ of the time, if explicit slicing/ indexing is being used, there is a better way to go about the problem. I refactored part of the `curve-amm` pallet to not use slicing and use iterators instead. This allowed me to get rid of some loops and mutability, along with some "temporary" variables, improving the flow and reducing the complexity of the code.

Some other miscellaneous clippy warning were fixed, mostly in tests. One of the main ones was changing CONSTANT_CASE enum variants to PascalCase (or UpperCamelCase, however you like to call it). If it is preferred that they be CONSTANT_CASE, they should be annotated with `#[allow(clippy::upper_case_acronyms)]`, ideally with a comment explaining why.
(see https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case)

Other lints fixed were `clippy::excessive_precision` (float literals such as `0.100000000000000_f32` have been trucated to not have the trailing zeros), `clippy::identity_op` (1 * whatever), `clippy::clone_on_copy` (`copy_type.clone()` changed to either `*copy_type` if it's a reference or just `copy_type` if it's by value)

The last few commits were fixing `clippy::unseparated_literal_suffix`, if deemed unnecessary or out of the scope of this PR they can be reverted.

### DIFF
--- a/frame/transaction-fee/src/lib.rs
+++ b/frame/transaction-fee/src/lib.rs
@@ -4,6 +4,7 @@
 //! but with added support for `MultiCurrency` using a `Dex` interface.
 
 #![cfg_attr(not(test), warn(clippy::disallowed_method, clippy::indexing_slicing))] // allow in tests
+#![warn(clippy::unseparated_literal_suffix)]
 #![cfg_attr(not(feature = "std"), no_std)]
 
 use codec::{Decode, Encode};
@@ -201,7 +202,7 @@ where
 		// a very very little potential gain in the future.
 		let dispatch_info = <E as GetDispatchInfo>::get_dispatch_info(&unchecked_extrinsic);
 
-		let partial_fee = Self::compute_fee(len, &dispatch_info, 0u32.into());
+		let partial_fee = Self::compute_fee(len, &dispatch_info, 0_u32.into());
 		let DispatchInfo { weight, class, .. } = dispatch_info;
 
 		RuntimeDispatchInfo { weight, class, partial_fee }
@@ -214,7 +215,7 @@ where
 		E: GetDispatchInfo,
 	{
 		let dispatch_info = <E as GetDispatchInfo>::get_dispatch_info(&unchecked_extrinsic);
-		Self::compute_fee_details(len, &dispatch_info, 0u32.into())
+		Self::compute_fee_details(len, &dispatch_info, 0_u32.into())
 	}
 
 	/// Compute the final fee value for a particular transaction.
--- a/frame/transaction-fee/src/lib.rs
+++ b/frame/transaction-fee/src/lib.rs
@@ -3,6 +3,7 @@
 //! Loosely based on https://github.com/paritytech/substrate/blob/master/frame/transaction-payment/src/lib.rs
 //! but with added support for `MultiCurrency` using a `Dex` interface.
 
+#![cfg_attr(not(test), warn(clippy::disallowed_method))] // allow in tests
 #![cfg_attr(not(feature = "std"), no_std)]
 
 use codec::{Decode, Encode};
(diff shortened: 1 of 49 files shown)


ComposableFiB0t:
Task linked: [CU-1pe15hj Remove the usage of partial functions](https://app.clickup.com/t/1pe15hj)

dzmitry-lahoda:
I vote with 10 TONCOIN for 
```
disallowed-methods = [
    "core::result::Result::unwrap", # { path = "core::result::Result::unwrap", reason = "Errors should be handled properly. If panicking is valid in this context, make sure to write a comment explaining why." },
    "core::option::Option::unwrap", # { path = "core::option::Option::unwrap", reason = "Errors should be handled properly. If panicking is valid in this context, make sure to write a comment explaining why." },
]
```
for forthers not sure

KaiserKarel:
Big thumbs up.

May I suggest also adding ValueQuery (https://docs.substrate.io/rustdocs/latest/frame_support/storage/types/struct.ValueQuery.html) to the disallowed_types, in this PR?

There are only a few cases where it should be used; there it should have an allow attribute and a comment explaining why it is allowed.

benluelo:
> Big thumbs up.
> 
> May I suggest also adding ValueQuery (https://docs.substrate.io/rustdocs/latest/frame_support/storage/types/struct.ValueQuery.html) to the disallowed_types, in this PR?
> 
> There are only a few cases where it should be used; there it should have an allow attribute and a comment explaining why it is allowed.

Added! There are quite a few uses of `ValueQuery` that will have to be audited. If there are any other types that should be disallowed, I can add them as well.

frame/transaction-fee/src/lib.rs
```diff
@@ -3,6 +3,8 @@
 //! Loosely based on https://github.com/paritytech/substrate/blob/master/frame/transaction-payment/src/lib.rs
 //! but with added support for `MultiCurrency` using a `Dex` interface.
 
+#![cfg_attr(not(test), warn(clippy::disallowed_method, clippy::indexing_slicing))] // allow in tests
```
dzmitry-lahoda, line 6: we put this into each pallet in frame. is it possible to put this rule on folder/directory?

benluelo: Unfortunately not, lints need to be specified per crate unless you specify them on the command line (i.e. `cargo clippy -- -W clippy::disallowed_method`)

dzmitry-lahoda:
I see conflicts growing. I could make sense merge part by part? Or rule by rule? (rule by rule seems already not so possible). 

Other rule which may be important, versions. Like we have right now smallvec of 3 versions. Eventually it will fail to compile with weird errors. May be main is broken because of this too.  And many version slow down compilation.

https://embarkstudios.github.io/cargo-deny/checks/bans/index.html

Afaik embark allows to prune 'bad version or crates' which kind of has some critical bugs or know vulnearabilitites.

dzmitry-lahoda (commented):
I am okey overall, but dislike `BTC` -> `Btc` convertion. Currency symbols could be exclusion. Really these are constants, so we represent them as enum.

benluelo:
> I am okey overall, but dislike `BTC` -> `Btc` convertion. Currency symbols could be exclusion. Really these are constants, so we represent them as enum.

That makes sense, they can be reverted with an `#[allow()]` and a comment explaining why

PR #423, opened 2021-12-27
more type safe currency unit calculation

hussein-aitlahcen:
Use 12 decimals for all currencies and move unit/milli etc to the trait definition

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1349,6 +1349,7 @@ checksum = "3d7b894f5411737b7867f4827955924d7c254fc9f4d91a6aad6b097804b1018b"
 name = "common"
 version = "0.8.30"
 dependencies = [
+ "composable-traits",
  "frame-support",
  "frame-system",
  "num-traits",
--- a/frame/composable-traits/src/currency.rs
+++ b/frame/composable-traits/src/currency.rs
@@ -1,3 +1,5 @@
+use core::ops::Div;
+
 use codec::FullCodec;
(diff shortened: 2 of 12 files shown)


dzmitry-lahoda:
what i know about decimals,
- right, in ORML these called same, so they have special trait to know it. so CurrencyId does not knows its decimals.
- decimals are only for currencies with decimals, cannot represent fiat with non decimals increases - ok
- because of benchmarks does not allow using other pallets than tested (cannot use others config), i cannot use CurrencyFactory or NativeCurrencyId traits from other pallets to get currency. so people hacked that by stating that currency id is from 128 on pallet levels, which is false. it is try from. decode is little bit better.
- really we have MayBeCurrency in all extrinsic, they are converted to currency by calling operations (transfer accepts MayBeCurrency) and returns not error if currency exists. like checking by doing. 
- from 2 above type system kind of broken as of now.
- decimals, could we hedge against possible currencies with 15-16 decimals? this number is good binary also (12 has zeroes)
- really, it does not matter how much decimal something have. there are only UI and oracle who cares. we do not do Domain Design here. All calculation should be done in smalles units. Decimals are only for external bounded contexts.

So having currency type safe in substrate is broken in Substrate more or less (at least by benchmarks).

dzmitry-lahoda (approved):

KaiserKarel (dismissed):

KaiserKarel (approved):

dzmitry-lahoda (approved):

PR #430, opened 2021-12-28
Stop using Vec<> in CurveAmm/Uniswap pallet storage.

vivekvpandya:


frame/curve-amm/src/lib.rs
```diff
@@ -834,6 +835,11 @@ pub mod pallet {
 			Ok(pool_id)
 		}
 
+		pub fn get_assets(pool_id: &T::PoolId) -> Option<Vec<T::AssetId>> {
+			let assets_pair = PoolAssets::<T>::get(pool_id)?;
+			Some(vec![assets_pair.0, assets_pair.1])
+		}
```
andor0, line 841: it can rewrite so:
```
pub fn get_assets(pool_id: &T::PoolId) -> Result<Vec<T::AssetId>, DispatchError> {
  let assets_pair = PoolAssets::<T>::get(pool_id).ok_or(Error::<T>::InconsistentStorage)?;
  Ok(vec![assets_pair.0, assets_pair.1])
}
```
then calls of this function will look so:
```
let assets = Self::get_assets(&pool_id)?;
```

vivekvpandya: I don't mind doing that, I just thought to make it similar to other functions surrounding. Note all of them return Option<> and based on that caller of those function returns Error. However it is not required. if @hussein-aitlahcen @dzmitry-lahoda prefer this way then I can update.

frame/curve-amm/src/lib.rs
```diff
@@ -157,10 +157,11 @@ pub mod pallet {
 	#[pallet::getter(fn pool_lp_asset)]
 	pub type PoolLPAsset<T: Config> = StorageMap<_, Blake2_128Concat, T::PoolId, T::AssetId>;
 
-	/// List of assets supported by the pool
+	/// Pair of assets supported by the pool
 	#[pallet::storage]
 	#[pallet::getter(fn pool_assets)]
-	pub type PoolAssets<T: Config> = StorageMap<_, Blake2_128Concat, T::PoolId, Vec<T::AssetId>>;
+	pub type PoolAssets<T: Config> =
```
dzmitry-lahoda, line 163: how about creating and using pair using struct as in dutch auction?

frame/curve-amm/src/lib.rs
dzmitry-lahoda: i am for splitting function into 2.

second, allow to iter over tuples of assets (a,b,c).iter(). you can generalize pair into iterator and than to .collect(). 

for options vs error.  as method in palllet, and pallet has PoolNotFound error, error fits better here.

dzmitry-lahoda: and vec! is allocation, you can create iterator and avoid that. or do not iterate at all without vector.

you can make view onto tuple as iterator (without move) and iterate without allocation. i thin exchange is hot pass and if possible should not allocate.

KaiserKarel: if you need iteration over stack allocated obj, use an array. Recently with const_generics has had loads of improvements.

dzmitry-lahoda: that will work too 

image

It will have nicer serde too.

dzmitry-lahoda: (in my pr)

frame/curve-amm/src/lib.rs
vivekvpandya: I can update code for DEX if your outstanding PR is merged.

KaiserKarel: @vivekvpandya can you check up on this?

dzmitry-lahoda (approved):
so it not stack allocates, but heap. so can improve still.

KaiserKarel (approved):

Issue #422, opened 2021-12-27
How to report security vulnerabilities? 


Noc2:
**Question**

You have a [security policy](https://github.com/ComposableFi/composable/security/policy), but it doesn’t specify how to report potential vulnerabilities. Do you have a specific email, like  security@parity.io or bugbounty@parity.io?

KaiserKarel:
I have updated our policy with the email for reports:

PR #431, opened 2021-12-28
Add email

KaiserKarel:
#422

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -15,6 +15,8 @@ When you report a security vulnerability please include:
 
 The more information you provide the better. We recommend submitting a report where you describe the vulnerability, show us how you found it and provide reproducible code samples.
 Providing mitigation advice is also recommended.
+
+The report should be submitted to security@composable.finance. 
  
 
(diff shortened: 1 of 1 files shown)


hussein-aitlahcen (approved):

dzmitry-lahoda:
As it seems compilation of main broken with new other reason.
Also, should we mention only specific runtime folder and `not(feature ="develop")` pallets? Develop pallets by definition are not secure.

dzmitry-lahoda (approved):

KaiserKarel:
> As it seems compilation of main broken with new other reason. Also, should we mention only specific runtime folder and `not(feature ="develop")` pallets? Develop pallets by definition are not secure.

Even if the crate is actively developed, if someone reports a vulnerability that we did not foresee, we'll reward them appropriately. The bounty will of course be less than a production bounty.

If we are aware of the vulnerability, we should ensure to either add a `todo!(fix vuln)` macro, or a TODO comment as to ensure that we do not forget to fix it (for develop crates).

PR #417, opened 2021-12-22
[CU-1y23865] [CU-1y23865] [CU-1yykqgq] Add a playbook to run polkadot-launch and integration tests

andor0:
ps: I will use build artifacts of composable, polkadot, and basilisk in the next PR.

PR #432, opened 2021-12-28
[CU-1y23865] testnet-release

andor0:
Those build artifacts will be used in the playbook to deploy a local Polkadot cluster (https://github.com/ComposableFi/composable/pull/417) in the future.

### DIFF
--- a/.github/workflows/testnet-release.yml
+++ b/.github/workflows/testnet-release.yml
@@ -3,30 +3,31 @@ name: "Testnet Community Release"
 on:
   push: 
     branches: 
-      - main
+    - main
+
 env:
-  CHAIN: "picasso"
+  CHAIN: "dali"
   RELEASE_VERSION: "latest"
 
 jobs:
   build-and-publish:
     runs-on: 
-        - self-hosted
-        - linux
-        - x64
-        - sre
+    - self-hosted
+    - linux
+    - x64
+    - sre
     steps:
     - uses: actions/checkout@v2
       with: 
         fetch-depth: 0
-        ref: picasso
+        ref: main
 
     - name: Build and Push Artifacts to gcloud
       run: |
-        /home/runner/.cargo/bin/cargo build --release --bins
+        /home/runner/.cargo/bin/cargo build --release --bins --features develop
         git rev-parse HEAD > revision
-        tar -czvf composable-${{ env.RELEASE_VERSION }}.tar.gz target/release/composable
-        tar -czvf picasso_runtime.compact.wasm-${RELEASE_VERSION}.tar.gz target/release/wbuild/picasso-runtime/picasso_runtime.compact.wasm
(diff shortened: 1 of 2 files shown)


ComposableFiB0t:
Task linked: [CU-1y23865 Create a playbook to run polkadot-launch](https://app.clickup.com/t/1y23865)

.github/workflows/testnet-release.yml
```diff
@@ -0,0 +1,32 @@
...
+        - sre
+    steps:
+    - uses: actions/checkout@v2
+      with: 
+        fetch-depth: 0
+        ref: picasso
+
+    - name: Build and Push Artifacts to gcloud
+      run: |
+        /home/runner/.cargo/bin/cargo build --release --bins
```
hussein-aitlahcen, line 26: Since it is triggered for each commit, can we use the `develop` feature in this case?

.github/workflows/testnet-release.yml
```diff
@@ -0,0 +1,32 @@
...
+        git rev-parse HEAD > revision
+        tar -czvf composable-${{ env.RELEASE_VERSION }}.tar.gz target/release/composable
```
hussein-aitlahcen, line 28: Is `RELEASE_VERSION` existing here? Isn't this supposed to be set from the `revision` file?
Also I think devops used the scheme `composable-picasso-{version}` when publishing

hussein-aitlahcen (commented):
Quick comments

.github/workflows/testnet-release.yml
andor0: This is a good idea!
Added.

.github/workflows/testnet-release.yml
andor0: > Is RELEASE_VERSION existing here?

Yes, `RELEASE_VERSION: "latest"`.

> Also I think devops used the scheme composable-picasso-{version} when publishing

Fixed.

> Isn't this supposed to be set from the revision file?

A static URL needs here (e.g. `https://storage.googleapis.com/composable-binaries/testnet-releases/dali/composable-dali-latest.tar.gz`) to use it in https://github.com/ComposableFi/composable/blob/main/.maintain/playbooks/run-integration-tests.yml#L131. Also, we quick full bucket of storage if the version will equal revision.

.github/workflows/testnet-release.yml
```diff
@@ -0,0 +1,33 @@
...
+    - linux
+    - x64
+    - sre
+    steps:
+    - uses: actions/checkout@v2
+      with: 
+        fetch-depth: 0
+        ref: main
+
+    - name: Build and Push Artifacts to gcloud
```
dzmitry-lahoda, line 25: i am not sure, so we may have markdown table of deploymentrs. 
currenc script does not feels right for me

andor0:
This isn't relevant already.

PR #434, opened 2021-12-29
Cleanup ci

haroldsphinx:


dzmitry-lahoda (changes requested):
i think hadolint should be configurable and we should build wasm of runtime on which tests will run (develop runtime). is hadoling config still used? what yml (if not this) will check that wasm can is assembled?

KaiserKarel:
before merging, check my changes regarding the toolchain.toml.

PR #435, opened 2021-12-29
Add rust-toolchain.toml

KaiserKarel:
Stuff was breaking in CI due to automatically updating the Rust toolchain. That also leads to different WASM later on. Adding the rust-toolchain makes it easier to produce reproducible builds.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -10,7 +10,6 @@ on:
       - dali-chachacha
       - picasso
   workflow_dispatch:
-concurrency: ci-check
 jobs:
   check:
     runs-on:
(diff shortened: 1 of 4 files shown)


dzmitry-lahoda (approved):

KaiserKarel:
Seeing how CI handles this; probably can remove some setup stuff from the check.yaml as well with this.

dzmitry-lahoda:
I have setup nightly previously as default toolchain toml , but @seunlanlege deleted it for unknown reason. what was the issue?

dzmitry-lahoda (dismissed):

KaiserKarel:
So actions.rs does not yet support the new toolchain format: https://github.com/actions-rs/toolchain/issues/208. Removing the dependency from actions.rs in CI, as I think that with the toolchain.toml, we have a better solution

dzmitry-lahoda (approved):

.github/workflows/check.yml
```diff
@@ -51,20 +47,17 @@ jobs:
...
         run: |
-          rustup target add wasm32-unknown-unknown
-          rustup component add rustfmt --toolchain nightly-x86_64-unknown-linux-gnu
-          rustup target add wasm32-unknown-unknown --toolchain nightly
+         curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+         echo "$HOME/.cargo/bin:$PATH" >> $GITHUB_PATH
+      - run: |
+          rustup show
+          rustup override unset
+          rustup show
```
KaiserKarel, line 57: I'll leave this in for now; and @seunlanlege and @haroldsphinx can have a look at why it is needed/remove it. The rustup override unset is required atm

hussein-aitlahcen (approved):

PR #436, opened 2021-12-29
CU-1wty1y5 weights for dutch auction

dzmitry-lahoda:
- bench test pass, but not yet part of runtime (will do it in next pr)
- fixed some security holes
- added comment about attempts to refactor currency system

### DIFF
--- a/frame/composable-traits/src/defi.rs
+++ b/frame/composable-traits/src/defi.rs
@@ -44,7 +44,7 @@ pub struct Sell<AssetId, Balance> {
 
 impl<AssetId: PartialEq, Balance: PartialOrd + Zero + SafeArithmetic> Sell<AssetId, Balance> {
 	pub fn is_valid(&self) -> bool {
-		self.take.is_valid() && self.pair.is_valid()
+		self.take.is_valid()
 	}
 	pub fn new(
 		base: AssetId,
@@ -61,6 +61,9 @@ impl<AssetId: PartialEq, Balance: PartialOrd + Zero + SafeArithmetic> Sell<Asset
 
 /// given `base`, how much `quote` needed for unit
 /// see [currency pair](https://www.investopedia.com/terms/c/currencypair.asp)
+/// Pair with same base and quote is considered valid as it allows to have mixer, money laundering
+/// like behavior.
+#[repr(C)]
 #[derive(Encode, Decode, TypeInfo, Debug, Clone, PartialEq)]
 pub struct CurrencyPair<AssetId> {
 	/// See [Base Currency](https://www.investopedia.com/terms/b/basecurrency.asp)
@@ -70,15 +73,37 @@ pub struct CurrencyPair<AssetId> {
 }
 
 impl<AssetId: PartialEq> CurrencyPair<AssetId> {
-	pub fn is_valid(&self) -> bool {
-		self.base != self.quote
+	pub fn new(base: AssetId, quote: AssetId) -> Self {
+		Self { base, quote }
+	}
+
+	///```rust
+	/// let pair = composable_traits::defi::CurrencyPair::<u128>::new(13, 42);
+	/// let slice =  pair.as_slice();
+	/// assert_eq!(slice[0], pair.base);
+	/// assert_eq!(slice[1], pair.quote);
+	/// ```
+	/// ```compile_fail
+	/// # let pair = composable_traits::defi::CurrencyPair::<u128>::new(13, 42);
+	/// # let slice =  pair.as_slice();
+	/// drop(pair);
+	/// let _ = slice[0];
+	/// ```
+	pub fn as_slice(&self) -> &[AssetId] {
+		unsafe { sp_std::slice::from_raw_parts(self as *const Self as *const AssetId, 2) }
+	}
+}
+
+impl<AssetId: PartialEq> AsRef<[AssetId]> for CurrencyPair<AssetId> {
+	fn as_ref(&self) -> &[AssetId] {
+		self.as_slice()
 	}
 }
 
 /// type parameters for traits in pure defi area
 pub trait DeFiEngine {
 	/// The asset ID type
-	type AssetId: AssetIdLike;
+	type MayBeAssetId: AssetIdLike;
 	/// The balance type of an account
 	type Balance: BalanceLike;
 	/// The user account identifier type for the runtime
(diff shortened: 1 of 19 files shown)


ComposableFiB0t:
Task linked: [CU-1wty1y5 auction: weights](https://app.clickup.com/t/1wty1y5)

frame/composable-traits/src/defi.rs
```diff
@@ -70,15 +73,21 @@ pub struct CurrencyPair<AssetId> {
...
-	pub fn is_valid(&self) -> bool {
-		self.base != self.quote
+	pub fn new(base: AssetId, quote: AssetId) -> Self {
+		Self { base, quote }
+	}
+
+	pub fn as_array(&self) -> &[AssetId; 2] {
+		// adding lifetimes explicitly considered by rust as not needed, so lifetime is proper for
+		// return array
+		unsafe { sp_std::mem::transmute(self) }
```
hussein-aitlahcen, line 83: Are we sure about this being safe, shouldn't we introduce a lifetime like `as_array<'a>(&'a self) -> &'a ...`?

dzmitry-lahoda: // adding lifetimes explicitly considered by rust as not needed, so lifetime is proper for
		// return array
I got clippy issue telling it is no needed to be explicit, so unsafe is not so unsafe

hussein-aitlahcen: What was the issue exactly, I'm not sure about the `transmute` here as both unsafe and transmute avoid optimization and hide the dependency between self and the output of the function

hussein-aitlahcen: What is clippy telling in this case?

dzmitry-lahoda: that i can remove explicit lifetimes because implicit are exactly same

dzmitry-lahoda: i was afraid exactly same issue. use after free. added lifetimes. but it told me that there is no need as it is already right.

runtime/primitives/src/currency.rs
```diff
@@ -9,12 +70,13 @@ use sp_runtime::{ArithmeticError, DispatchError, RuntimeDebug};
 use serde::{Deserialize, Serialize};
 use sp_runtime::sp_std::ops::Deref;
 
+/// `MayBe`CurrencyId as not each `u128` is valid id.
```
vivekvpandya, line 73: should we rename struct from `CurrencyId` to `Maybe...` ?

frame/composable-traits/src/defi.rs
```diff
@@ -70,15 +73,21 @@ pub struct CurrencyPair<AssetId> {
...
 	}
 }
 
 /// type parameters for traits in pure defi area
 pub trait DeFiEngine {
 	/// The asset ID type
-	type AssetId: AssetIdLike;
+	type MayBeAssetId: AssetIdLike;
```
vivekvpandya, line 90: will this naming convention apply to all other associated types like `Balance` ?

runtime/primitives/src/currency.rs
dzmitry-lahoda: Not in this PR I think. Too many changes will appear.

frame/composable-traits/src/defi.rs
dzmitry-lahoda: It is only about id. Really this type is never exactly asseit id by always may be id. Tokens type is about using that MayBeId (I assume these must error if ID does not exist)

vivekvpandya (dismissed):
LGTM!

frame/composable-traits/src/defi.rs
hussein-aitlahcen: Does it explicitly ask for an unsafe + transmute here? [As I read the clippy rule](https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/lifetimes.rs#L39) I don't think we should use that.

To avoid introducing the transmute, I suggest that we just use a newtype with two getters such as:
```rust
struct CurrencyPair<Asset>([Asset; 2]);
impl<Asset> CurrencyPair<Asset> {
  pub fn base(&self) -> Asset { .. }
  pub fn quote(&self) -> Asset { .. }
  pub fn as_array(&self) -> &[Asset; 2] { &self.0 }
}
```

dzmitry-lahoda: Having exact binary repr of pair could be as crucial as for asset id cause all trades are based on pairs. Using trait and array will change serde and client types coded (serde). Is scale code can express the exact two items array in API? If unsafe is safe, why not use unsafe if it is safe?
I doubt we should introduce more than exactly one pair type, so that the trait will be one-time use only in most cases.

hussein-aitlahcen: My concern is that I think this code is similar to: `fn f<'a, 'arbitrary>(&'a x: T) -> &'arbitrary T { unsafe { transmute(x) } }`

KaiserKarel: This is UB, as the layout of [T, 2] != Struct { T, T } necessarily. Here we need to use the from_raw_parts API:

```
unsafe { std::slice::from_raw_parts(self as *const Self as *const T, 2) }
```

KaiserKarel: Also I suggest renaming to as_slice

dzmitry-lahoda: cool. i tested that repc(transparent) struct == array. but seems from_raw_parts is way more safe.

KaiserKarel: 99% of the time the transmute will work, but it is not guaranteed and may suddenly break across compiler upgrades etc.

frame/composable-traits/src/defi.rs
```diff
@@ -70,15 +73,34 @@ pub struct CurrencyPair<AssetId> {
...
+
+	///```rust
+	/// let pair = composable_traits::defi::CurrencyPair::<u128>::new(13, 42);
+	/// let slice =  pair.as_slice();
+	/// assert_eq!(slice[0], pair.base);
+	/// assert_eq!(slice[1], pair.quote);
+	/// drop(pair);
+	/// // next line will fail to compile
+	/// // let _ = slice[0];
+	/// ```
```
KaiserKarel, line 88: You can use `compile_fail` to add a doctest that should fail to compile, so that it it still checked.

hussein-aitlahcen (dismissed):

hussein-aitlahcen:
Heh, clippy PR merged, need rebase

hussein-aitlahcen (approved):

KaiserKarel (approved):

KaiserKarel:
@dzmitry-lahoda seems to me it is all signed?

hussein-aitlahcen (approved):

vivekvpandya (approved):

PR #437, opened 2021-12-29
Ben/crowdloan rewards api

benluelo:


frame/crowdloan-rewards/src/lib.rs
```diff
@@ -480,38 +480,41 @@ where
...
 					} else {
-						return InvalidTransaction::Custom(ValidityError::InvalidProof as u8)
-							.into();
+						return InvalidTransaction::Custom(ValidityError::InvalidProof as u8).into()
 					}
-				}
+				},
 			};
 
 			if Rewards::<T>::get(remote_account).is_none() {
```
hussein-aitlahcen, line 516: I think it is cheaper to test this than checking for the proof first, we can perhaps move it before the proof checking?

frame/crowdloan-rewards/src/lib.rs
```diff
@@ -405,3 +417,122 @@ pub mod pallet {
 		Some(addr)
 	}
 }
+
+/// Validate `attest` calls prior to execution. Needed to avoid a DoS attack since they are
```
hussein-aitlahcen, line 421: Validate `associate`

hussein-aitlahcen (commented):
I think we can add some tests for the extension, making sure we effectively deny/allow proper calls, like [parity has been doing for their claiming mechanism](https://github.com/paritytech/polkadot/blob/master/runtime/common/src/claims.rs#L1049)

frame/crowdloan-rewards/src/lib.rs
benluelo: Do you mean just for the relay account?

hussein-aitlahcen: It's not a big deal, since you need to extract it from the signature, the code is currently nice and easy to understand, let's keep it like that :)

hussein-aitlahcen (dismissed):

frame/crowdloan-rewards/src/lib.rs
```diff
@@ -230,7 +240,7 @@ pub mod pallet {
 		/// ```haskell
 		/// proof = sign (concat prefix (hex reward_account))
 		/// ```
-		#[pallet::weight(<T as Config>::WeightInfo::associate(TotalContributors::<T>::get()))]
+		#[pallet::weight((<T as Config>::WeightInfo::associate(TotalContributors::<T>::get()), Pays::No))]
```
dzmitry-lahoda, line 243: Hm, are both Pays::No needed? is this Pays::No males errors also free?

frame/crowdloan-rewards/src/lib.rs
```diff
@@ -423,3 +468,112 @@ pub mod pallet {
...
+
+		if let Some(Call::associate { reward_account, proof }) = IsSubType::is_sub_type(call) {
+			if Associations::<T>::get(reward_account).is_some() {
+				return InvalidTransaction::Custom(ValidityError::AlreadyAssociated as u8).into()
+			}
+
+			let remote_account = match proof {
+				Proof::Ethereum(eth_proof) => {
+					let reward_account_encoded =
+						reward_account.using_encoded(|x| hex::encode(x).as_bytes().to_vec());
```
dzmitry-lahoda, line 540: this hex part is copy pased, can be reused

frame/crowdloan-rewards/src/lib.rs
benluelo: Not sure actually, I will run some tests and check

frame/crowdloan-rewards/src/lib.rs
dzmitry-lahoda: not sure, if you are not going to take weight, than remove weight from assosiate call. amount of compute will be same until error.  i do not get difference how checking here safes from ddos? it is kind of optimization so the call does not gets into queue or something?

frame/crowdloan-rewards/src/tests.rs
```diff
@@ -388,3 +401,91 @@ fn test_valid_eth_hardcoded() {
...
+				TransactionValidity::Err(
+					InvalidTransaction::Custom(ValidityError::AlreadyAssociated as u8).into()
+				)
+			);
+		}
+	});
+}
+
+#[test]
+fn no_reward_associate_transactions_are_recognized() {
```
dzmitry-lahoda, line 466: quick question - if there is test for signed exstrinsic?

frame/crowdloan-rewards/src/lib.rs
benluelo: Removing the Pays::No in the weight annotation makes the valid transactions paid (valid_associate_transactions_are_free fails)

frame/crowdloan-rewards/src/lib.rs
benluelo: As per the [docs](https://github.com/paritytech/substrate/blob/master/primitives/runtime/src/traits.rs#L860-L877), I believe it keeps it out of the queue if it's not validated

frame/crowdloan-rewards/src/tests.rs
benluelo: Not sure what you mean - these are the tests for the SignedExtension

frame/crowdloan-rewards/src/lib.rs
benluelo: See latest commit with deduplicated code

frame/crowdloan-rewards/src/lib.rs
hussein-aitlahcen: I'm unsure but I think that keeping the weight allow to avoid a block being full

frame/crowdloan-rewards/src/lib.rs
hussein-aitlahcen: This should be fine as it is a signed extension. Meaning that if the check does not pass, the extrinsic is not pushed to the mempool, hence not propagated to the network

frame/crowdloan-rewards/src/tests.rs
benluelo: Improved tests in latest commit, if that's what you meant

hussein-aitlahcen (approved):

frame/crowdloan-rewards/Cargo.toml
```diff
@@ -17,32 +17,45 @@ version = "2.0.0"
...
 hex-literal = "0.3"
-libsecp256k1 = { version = "0.7.0", default-features = false, features = [ "hmac", "static-context"] }
+balances = { package = "pallet-balances", git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.13", default-features = false, features = [
+    "std",
+] }
 
 [dependencies]
+# should be a dev-dependency. see: https://github.com/rust-lang/cargo/issues/6915
+# balances = { package = "pallet-balances", git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.13", default-features = false }
+
```
hussein-aitlahcen, line 34: Can we remove that?

frame/crowdloan-rewards/Cargo.toml
```diff
@@ -17,32 +17,45 @@ version = "2.0.0"
 
 [dev-dependencies]
 serde = "1.0.119"
-balances = { package = "pallet-balances", git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.13", default-features = false }
+# orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", rev = "17a791edf431d7d7aee1ea3dfaeeb7bc21944301", default-features = false }
```
hussein-aitlahcen, line 20: Can we remove that?

frame/crowdloan-rewards/Cargo.toml
```diff
@@ -56,18 +69,18 @@ std = [
     "sp-std/std",
     "sp-arithmetic/std",
     "scale-info/std",
-	  "balances/std",
+    # "balances/std",
```
hussein-aitlahcen, line 72: Can we remove that?

frame/crowdloan-rewards/Cargo.toml
benluelo: yep, I will go through and remove comments. Forgot to do that

hussein-aitlahcen (approved):

hussein-aitlahcen:
One thing I forget is that we need to add the extension in https://github.com/ComposableFi/composable/blob/main/runtime/picasso/src/lib.rs#L371 I believe

KaiserKarel (dismissed):

benluelo:
Yes right, I will add it right now. It only needs to be added to the picasso runtime, correct?

KaiserKarel:
> Yes right, I will add it right now. It only needs to be added to the picasso runtime, correct?

yes; but please credit a ticket for us to explicitly audit that as well.

KaiserKarel (approved):

hussein-aitlahcen (approved):

PR #440, opened 2022-01-03
Update outdated Run instructions

cor:
Since Composable is now a parachain, you need to run a full relaychain in addition to running Composable. I've updated the main README.md to reflect this.


hussein-aitlahcen (approved):
:rocket:

PR #441, opened 2022-01-03
CU-1m98wce lending docs and fixes

dzmitry-lahoda:
- https://app.clickup.com/20465559/v/dc/kghwq-20761/kghwq-3701
- fixed double exponent bug (it was not double and was a ladder, not smooth)
- comment  issue with PID (cannot get stable PID with reasonable parameters and UT = Optimal UT with some initial borrow rate, it always goes infinity up - bug, not yet fixed)
- refactoring lending
- all models except double exponent work as before


ComposableFiB0t:
Task linked: [CU-1m98wce Angular interest documentation](https://app.clickup.com/t/1m98wce)

frame/composable-traits/src/lending/math.rs
```diff
@@ -281,54 +275,44 @@ impl DynamicPIDControllerModel {
...
 		// compute derivative term `dt = kd * (et - et_1)`
-		let dt = self.kd.checked_mul(&(et - self.et_1)).ok_or(ArithmeticError::Overflow)?;
-		self.et_1 = et;
+		let dt = self.derivative_parameter.checked_mul(&(et - self.previous_error_value)).ok_or(ArithmeticError::Overflow)?;
+		self.previous_error_value = et;
 
 		// compute u(t), control value `ut = pt + it + dt`
 		let ut = pt + it + dt;
 		// update interest_rate `ir = ir_t_1 + ut`
-		if ut.is_negative() {
```
vivekvpandya: IIRC `et` can be negative and based on that other values like `ut` can be negative @dzmitry-lahoda  right?

frame/composable-traits/src/lending/math.rs
```diff
@@ -281,54 +275,44 @@ impl DynamicPIDControllerModel {
...
-			.checked_add(&self.ki.checked_mul(&et).ok_or(ArithmeticError::Overflow)?)
+			.previous_integral_term
+			.checked_add(&self.integral_parameter.checked_mul(&et).ok_or(ArithmeticError::Overflow)?)
 			.ok_or(ArithmeticError::Overflow)?;
-		self.it_1 = it;
+		self.previous_integral_term = it;
 		// compute derivative term `dt = kd * (et - et_1)`
-		let dt = self.kd.checked_mul(&(et - self.et_1)).ok_or(ArithmeticError::Overflow)?;
-		self.et_1 = et;
+		let dt = self.derivative_parameter.checked_mul(&(et - self.previous_error_value)).ok_or(ArithmeticError::Overflow)?;
```
vivekvpandya: if we want to use full names like `integral_parameter` then should we use same convention for local variables as well?

frame/composable-traits/src/lending/math.rs
```diff
@@ -356,19 +358,15 @@ impl DoubleExponentModel {
...
-				.iter()
-				.try_fold((result_0, polynomial_0), |(res, poly), coeff| {
-					let coefficient = FixedU128::from_inner(u128::from(*coeff));
-					let result = res.checked_add(&coefficient.checked_mul(&poly)?)?;
-					let polynomial = poly.checked_mul(&utilization.into())?;
-					Some((result, polynomial))
-				});
-		res.map(|(r, _p)| r.checked_div(&FixedU128::from_inner(EXPECTED_COEFFICIENTS_SUM.into())))
-			.unwrap()
+		// as per model
```
vivekvpandya: Is there any difference between previous implementation vs this new code? or is it just to make it easy to read? @dzmitry-lahoda

vivekvpandya: Thanks @dzmitry-lahoda for explaining the diff offline on Slack. Marking this resolved.

vivekvpandya (dismissed):
LGTM!

hussein-aitlahcen:
Good for me, need rebase

vivekvpandya (approved):

KaiserKarel (approved):

PR #442, opened 2022-01-04
CU-1wtx9hm Add proptest for Uniswap-v2

vivekvpandya:


frame/uniswap-v2/src/tests.rs
```diff
@@ -174,3 +206,144 @@ fn exchange_test() {
 		assert!(65000 >= Tokens::balance(MockCurrencyId::USDC, &CHARLIE));
 	});
 }
+
+proptest! {
+	#![proptest_config(ProptestConfig::with_cases(10000))]
+	#[test]
+	fn proptest_add_remove_liquidity(alice_balance in 0..u32::MAX,
+									 bob_balance in 0..u32::MAX) {
```
KaiserKarel, line 214:
```suggestion
	fn proptest_add_remove_liquidity(
	    alice_balance in 0..u32::MAX,
	    bob_balance in 0..u32::MAX
	) {
```

frame/uniswap-v2/src/tests.rs
```diff
@@ -3,7 +3,39 @@ use frame_support::assert_ok;
...
+			"({}) => {} >= {} * {} / {} >= {}",
+			q,
+			upper,
+			$x,
+			precision,
+			$y,
+			lower
+		);
+	}};
+}
```
hussein-aitlahcen, line 38: Some functions have been factored out in `composable-test-helpers` https://github.com/ComposableFi/composable/tree/main/frame/composable-tests-helpers.

Please use this lib so that they are reusable :)

hussein-aitlahcen (changes requested):

frame/uniswap-v2/src/tests.rs
```diff
@@ -174,3 +206,144 @@ fn exchange_test() {
...
+		prop_assert_zero_epsilon!(Tokens::balance(MockCurrencyId::USDC, &ConstantProductAmm::account_id(&pool_id)));
+		prop_assert_zero_epsilon!(Tokens::balance(MockCurrencyId::USDT, &ConstantProductAmm::account_id(&pool_id)));
+		prop_assert_epsilon!(Tokens::balance(MockCurrencyId::USDT, &BOB), bob_balance as u128);
+		prop_assert_epsilon!(Tokens::balance(MockCurrencyId::USDC, &BOB), bob_balance as u128);
+		Ok(())
+	}).unwrap();
+	}
+	#[test]
+	fn proptest_exchange(alice_balance in 1..u32::MAX,
+						 bob_balance in 1..u32::MAX) {
```
KaiserKarel, line 287:
```suggestion
	fn proptest_exchange(
	    alice_balance in 1..u32::MAX,
	    bob_balance in 1..u32::MAX
	) {
```

KaiserKarel (changes requested):
Just the formatting related stuff

frame/composable-tests-helpers/src/test/proptest.rs
```diff
@@ -36,3 +36,36 @@ macro_rules! prop_assert_acceptable_computation_error {
...
+			"({}) => {} >= {} * {} / {} >= {}",
+			q,
+			upper,
+			$x,
+			precision,
+			$y,
+			lower
+		);
+	}};
+}
```
hussein-aitlahcen, line 71: We don't need this function anymore?

frame/curve-amm/src/tests.rs
```diff
@@ -392,18 +368,18 @@ proptest! {
 		// ALICE removes liquidity from DEX pool.
 		assert_ok!(CurveAmm::remove_liquidity(&ALICE, pool_id, alice_lp_balance, min_amt.clone()));
 		prop_assert_zero_epsilon!(Tokens::balance(pool_lp_asset, &ALICE));
-		prop_assert_epsilon!(Tokens::balance(MockCurrencyId::USDT, &ALICE), alice_balance as u128);
-		prop_assert_epsilon!(Tokens::balance(MockCurrencyId::USDC, &ALICE), alice_balance as u128);
-		prop_assert_epsilon!(Tokens::balance(MockCurrencyId::USDC, &CurveAmm::account_id(&pool_id)), bob_balance as u128);
-		prop_assert_epsilon!(Tokens::balance(MockCurrencyId::USDT, &CurveAmm::account_id(&pool_id)), bob_balance as u128);
+		prop_assert_acceptable_computation_error!(Tokens::balance(MockCurrencyId::USDT, &ALICE), alice_balance as u128, 1000);
```
hussein-aitlahcen, line 371: The `precision` is by default `1000` so no need to manually add it?

hussein-aitlahcen (commented):
Can we have more static tests that reflect the slippage etc? Like small, isolated tests with some fixed ratio over two assets and a swap execution?

frame/composable-tests-helpers/src/test/proptest.rs
vivekvpandya: I require it ,as other functions does not allow `q` in range 0.995 to 1.005. For exchange with high slippage i.e almost emptying the pool I need this.

vivekvpandya:
> Can we have more static tests that reflect the slippage etc? Like small, isolated tests with some fixed ratio over two assets and a swap execution?

Okay, what percentage of slippage do we want to allow? Also should test use large numbers ?

KaiserKarel:
> Okay, what percentage of slippage do we want to allow? Also should test use large numbers ?

I will get research to provide the bounds. If you can add the unit tests with a configurable threshold, that'd be good

frame/composable-tests-helpers/src/test/proptest.rs
```diff
@@ -36,3 +36,36 @@ macro_rules! prop_assert_acceptable_computation_error {
 		);
 	}};
 }
+
+/// Accepts -2, -1, 0, 1, 2
+#[macro_export]
+macro_rules! prop_assert_zero_epsilon {
```
dzmitry-lahoda, line 42: very unclear naming

dzmitry-lahoda: proper_assert_around_zero(x, 2);

vivekvpandya:
> > Okay, what percentage of slippage do we want to allow? Also should test use large numbers ?
> 
> I will get research to provide the bounds. If you can add the unit tests with a configurable threshold, that'd be good

@hussein-aitlahcen @KaiserKarel do we have any updates or ideas for this? otherwise I can address pending comments and complete review for this PR.

frame/composable-tests-helpers/src/test/proptest.rs
vivekvpandya: @dzmitry-lahoda proper_assert_around_zero() is also not very clear :sweat_smile:  . And comment above definition helps understanding the expected behavior.

dzmitry-lahoda (dismissed):

ComposableFiB0t:
Task linked: [CU-1wtx9hm proptest univ2](https://app.clickup.com/t/1wtx9hm)

KaiserKarel:
> > > Okay, what percentage of slippage do we want to allow? Also should test use large numbers ?
> > 
> > 
> > I will get research to provide the bounds. If you can add the unit tests with a configurable threshold, that'd be good
> 
> @hussein-aitlahcen @KaiserKarel do we have any updates or ideas for this? otherwise I can address pending comments and complete review for this PR.

We'll follow that up with QA and merge this.

frame/composable-tests-helpers/src/test/proptest.rs
hussein-aitlahcen: I was meaning that there is already a function in the crate that is doing that

hussein-aitlahcen (approved):
Quick comment and I think we need more tests but it's a good start

frame/composable-tests-helpers/src/test/proptest.rs
vivekvpandya: both are slightly different than each other.

dzmitry-lahoda (approved):

dzmitry-lahoda:
it seems i cannot handle such tests
```
		assert_eq!(Tokens::balance(CurrencyId::BTC, &CHARLIE), 0);
		assert_ok!(Tokens::mint_into(CurrencyId::BTC, &CHARLIE, borrow_asset_deposit));
		assert_eq!(Tokens::balance(CurrencyId::BTC, &CHARLIE), borrow_asset_deposit);
```
as I mentioned, these can be single lined. another problem. when I start protecting API with rents/stake, first equality failed in all such tests. another issue with that it tests property of x = 0 + x each time, which is dumb. it could test x = (d + x) - x - than it would be more relevant.

dzmitry-lahoda (dismissed):

dzmitry-lahoda:
@vivekvpandya please add configuration parameter into sell/buy/exchange. slippage may be controlled from that. not this pr. either in next or todo or clickup.

vivekvpandya:
> it seems i cannot handle such tests
> 
> ```
> 		assert_eq!(Tokens::balance(CurrencyId::BTC, &CHARLIE), 0);
> 		assert_ok!(Tokens::mint_into(CurrencyId::BTC, &CHARLIE, borrow_asset_deposit));
> 		assert_eq!(Tokens::balance(CurrencyId::BTC, &CHARLIE), borrow_asset_deposit);
> ```
> 
> as I mentioned, these can be single lined. another problem. when I start protecting API with rents/stake, first equality failed in all such tests. another issue with that it tests property of x = 0 + x each time, which is dumb. it could test x = (d + x) - x - than it would be more relevant.

Those checks are removed in #550

vivekvpandya:
> 

Added click-up [#21fd7kg](https://app.clickup.com/t/21fd7kg)

dzmitry-lahoda (approved):

PR #446, opened 2022-01-05
CU-1nh9erq Implement basic DEX Router pallet.

vivekvpandya:
I am thinking to add function `get_route(x_asset_id : AssetId, y_asset_id : AssetId) -> Vec<PoolId>` .
Also @KaiserKarel @hussein-aitlahcen  is this pallet responsible to execute `exchange()` functions on pools found on route?

### DIFF
--- a/frame/composable-traits/src/dex.rs
+++ b/frame/composable-traits/src/dex.rs
@@ -131,14 +131,14 @@ pub enum DexRoute<PoolId, MaxHops: Get<u32>> {
 	Direct(BoundedVec<DexRouteNode<PoolId>, MaxHops>),
 }
 
-pub trait DexRouter<AccountId, AssetId, PoolId, Balance> {
+pub trait DexRouter<AccountId, AssetId, PoolId, Balance, MaxHops> {
 	/// If route is `None` then delete existing entry for `asset_pair`
 	/// If route is `Some` and no entry exist for `asset_pair` then add new entry
 	/// else update existing entry.
 	fn update_route(
 		who: &AccountId,
 		asset_pair: CurrencyPair<AssetId>,
-		route: Option<Vec<DexRouteNode<PoolId>>>,
+		route: Option<BoundedVec<DexRouteNode<PoolId>, MaxHops>>,
 	) -> Result<(), DispatchError>;
 	/// If route exist return `Some(Vec<PoolId>)`, else `None`.
 	fn get_route(asset_pair: CurrencyPair<AssetId>) -> Option<Vec<DexRouteNode<PoolId>>>;
--- a/frame/composable-traits/src/dex.rs
+++ b/frame/composable-traits/src/dex.rs
@@ -1,7 +1,10 @@
 use codec::{Decode, Encode};
+use frame_support::{traits::Get, BoundedVec, RuntimeDebug};
 use scale_info::TypeInfo;
 use sp_runtime::{DispatchError, FixedU128, Permill};
 use sp_std::vec::Vec;
+
+use crate::defi::CurrencyPair;
 
 /// Implement AMM curve from "StableSwap - efficient mechanism for Stablecoin liquidity by Micheal
 /// Egorov" Also blog at https://miguelmota.com/blog/understanding-stableswap-curve/ has very good explanation.
@@ -18,6 +21,9 @@ pub trait CurveAmm {
 
 	/// Type that represents pool id
 	type PoolId;
+
+	/// Check pool with given id exists.
+	fn pool_exists(pool_id: Self::PoolId) -> bool;
 
 	/// Current number of pools (also ID for the next created pool)
 	fn pool_count() -> Self::PoolId;
@@ -55,7 +61,7 @@ pub trait CurveAmm {
 		j: Self::PoolTokenIndex,
 		dx: Self::Balance,
 		min_dy: Self::Balance,
-	) -> Result<(), DispatchError>;
+	) -> Result<Self::Balance, DispatchError>;
 
 	/// Withdraw admin fees
 	fn withdraw_admin_fees(
@@ -66,7 +72,7 @@ pub trait CurveAmm {
 }
 
 /// Pool type
-#[derive(Encode, Decode, TypeInfo, Clone, Default, PartialEq, Eq, Debug)]
+#[derive(Encode, Decode, TypeInfo, Clone, Default, PartialEq, Eq, RuntimeDebug)]
 pub struct StableSwapPoolInfo<AccountId, AssetId> {
 	/// Owner of pool
 	pub owner: AccountId,
@@ -102,7 +108,7 @@ pub trait SimpleExchange {
 	) -> Result<Self::Balance, DispatchError>;
 }
 
-#[derive(Encode, Decode, TypeInfo, Clone, Default, PartialEq, Eq, Debug)]
+#[derive(Encode, Decode, TypeInfo, Clone, Default, PartialEq, Eq, RuntimeDebug)]
 pub struct ConstantProductPoolInfo<AccountId, AssetId> {
 	/// Owner of pool
 	pub owner: AccountId,
@@ -110,4 +116,36 @@ pub struct ConstantProductPoolInfo<AccountId, AssetId> {
 	pub lp_token: AssetId,
 	/// Amount of the fee pool charges for the exchange
 	pub fee: Permill,
+}
+
+#[derive(Encode, Decode, TypeInfo, Clone, PartialEq, Eq, RuntimeDebug)]
+pub enum DexRouteNode<PoolId> {
+	Curve(PoolId),
+	Uniswap(PoolId),
+}
+
+/// Describes route for DEX.
+/// `Direct` gives vector of pool_id to use as router.
+#[derive(Encode, Decode, TypeInfo, Clone, PartialEq, Eq, RuntimeDebug)]
+pub enum DexRoute<PoolId, MaxHops: Get<u32>> {
+	Direct(BoundedVec<DexRouteNode<PoolId>, MaxHops>),
+}
+
+pub trait DexRouter<AccountId, AssetId, PoolId, Balance> {
+	/// If route is `None` then delete existing entry for `asset_pair`
+	/// If route is `Some` and no entry exist for `asset_pair` then add new entry
+	/// else update existing entry.
+	fn update_route(
+		who: &AccountId,
+		asset_pair: CurrencyPair<AssetId>,
+		route: Option<Vec<DexRouteNode<PoolId>>>,
+	) -> Result<(), DispatchError>;
+	/// If route exist return `Some(Vec<PoolId>)`, else `None`.
+	fn get_route(asset_pair: CurrencyPair<AssetId>) -> Option<Vec<DexRouteNode<PoolId>>>;
+	/// Exchange `dx` of given `asset_pair` to get `dy`.
+	fn exchange(
+		who: &AccountId,
+		asset_pair: CurrencyPair<AssetId>,
+		dx: Balance,
+	) -> Result<Balance, DispatchError>;
 }
--- a/frame/dex-router/src/lib.rs
+++ b/frame/dex-router/src/lib.rs
@@ -230,6 +230,7 @@ pub mod pallet {
 				.map(|DexRoute::Direct(route)| route.into_inner())
 		}
 
+		// TODO: expected minimum value can be provided from input parameter.
 		fn exchange(
 			who: &T::AccountId,
 			asset_pair: CurrencyPair<T::AssetId>,
--- a/frame/dex-router/src/lib.rs
+++ b/frame/dex-router/src/lib.rs
@@ -145,83 +145,91 @@ pub mod pallet {
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {}
 
-	impl<T: Config> Pallet<T> {}
-
-	impl<T: Config> DexRouter<T::AccountId, T::AssetId, T::PoolId, T::Balance> for Pallet<T> {
-		fn update_route(
+	impl<T: Config> Pallet<T> {
+		fn do_update_route(
 			who: &T::AccountId,
 			asset_pair: CurrencyPair<T::AssetId>,
-			route: Option<Vec<DexRouteNode<T::PoolId>>>,
+			route: BoundedVec<DexRouteNode<T::PoolId>, T::MaxHopsInRoute>,
 		) -> Result<(), DispatchError> {
 			let k1 = asset_pair.base;
 			let k2 = asset_pair.quote;
-			match route {
-				Some(route_vec) => {
-					for r in &route_vec {
-						match r {
-							DexRouteNode::Curve(pool_id) => {
-								if !T::StableSwapDex::pool_exists(*pool_id) {
-									return Err(Error::<T>::PoolDoesNotExist.into())
-								}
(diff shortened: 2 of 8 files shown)


frame/dex-router/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+
+	impl<T: Config> Pallet<T> {
+	}
+
+    impl<T: Config> DexRouter<T::AssetId, T::PoolId, T::RouteId> for Pallet<T> {
+        fn update(asset_pair : (T::AssetId, T::AssetId), route : Option<DexRoute<T::PoolId, T::RouteId>>) {
+            match route {
+                Some(r) => {
+                    let mut update = false;
+                    if DexRoutes::<T>::contains_key(asset_pair.0, asset_pair.1) {
```
dzmitry-lahoda, line 74: would you make TODO to use CurrenchPair?

frame/dex-router/src/lib.rs
```diff
@@ -0,0 +1,105 @@
+//! # DEX Router Pallet
+//!
+//! Is used to add route to DEX for given asset_id's pair.
```
dzmitry-lahoda, line 3: please add reason why we have this pallet - to have permissions approved pools

frame/dex-router/src/lib.rs
vivekvpandya: once your PR will be merged I will use that.

frame/dex-router/src/lib.rs
vivekvpandya: Is it permission approved pools or permission approved routes?

frame/composable-traits/src/dex.rs
```diff
@@ -111,3 +112,28 @@ pub struct ConstantProductPoolInfo<AccountId, AssetId> {
 	/// Amount of the fee pool charges for the exchange
 	pub fee: Permill,
 }
+
+#[derive(Encode, Decode, TypeInfo, Clone, PartialEq, Eq, RuntimeDebug)]
+pub enum DexRouteNode<PoolId> {
+	Curve(PoolId),
```
KaiserKarel, line 118: This would force both DEXes to use the same PoolId, better to make it take two generics

frame/dex-router/src/lib.rs
KaiserKarel: permissioned approval of routes, pools are still freely creatable

frame/composable-traits/src/dex.rs
```diff
@@ -111,3 +112,28 @@ pub struct ConstantProductPoolInfo<AccountId, AssetId> {
...
+}
+
+/// Describes route for DEX.
+/// `Direct` gives vector of pool_id to use as router.
+#[derive(Encode, Decode, TypeInfo, Clone, PartialEq, Eq, RuntimeDebug)]
+pub enum DexRoute<PoolId, MaxHops: Get<u32>> {
+	Direct(BoundedVec<DexRouteNode<PoolId>, MaxHops>),
+}
+
+pub trait DexRouter<AssetId, PoolId> {
```
KaiserKarel, line 129: Should add `buy` and `sell` functions, which route and execute.

frame/composable-traits/src/dex.rs
vivekvpandya: what if in future we may have one more AMM? @KaiserKarel

frame/composable-traits/src/dex.rs
vivekvpandya: can we have a single `exchange` method instead of buy/sell ? if not could you please highlight differences?

frame/composable-traits/src/dex.rs
dzmitry-lahoda: yeah, having poolid seems resonable tradeoff. if new pallet to be integrated, than it should be refactoed to be generic on poolid and support same poold id type.  If pallet will use different model of poolid thing, than may add <PoolId, SomeOtherPoolId> later

frame/composable-traits/src/dex.rs
dzmitry-lahoda: there are 2 desings - have buy+sell vs order(direction).

First can be modelled as trait (see Sell trait) and can support only onsided defi for pair. 

Direction is kind of more serde friendly. For

Afaik Hydra have buy/sell in API traits, but stores all these in single collection with direction.  

Solana Serum (whole ecosystem is binary oriented and less abstact) has order(direction).

So it seems buy/sell is accepted in Polkadot as interface.

frame/composable-traits/src/dex.rs
```diff
@@ -111,3 +112,28 @@ pub struct ConstantProductPoolInfo<AccountId, AssetId> {
...
+	/// If route is `None` then delete existing entry for `asset_pair`
+	/// If route is `Some` and no entry exist for `asset_pair` then add new entry
+	/// else update existing entry.
+	fn update_route(
```
dzmitry-lahoda, line 133: May be use CurrencyPair?

frame/dex-router/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+		type AssetId: FullCodec
+			+ Eq
+			+ PartialEq
+			+ Copy
+			+ MaybeSerializeDeserialize
+			+ Debug
+			+ Default
+			+ TypeInfo
+			+ Ord;
+		type Balance: Default
```
dzmitry-lahoda, line 45: consider using MathBalance as trait bound for part of bounds here

frame/dex-router/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+			+ From<u64> // at least 64 bit
+			+ Zero
+			+ One
+			+ IntegerSquareRoot
+			+ FixedPointOperand
+			+ Into<LiftedFixedBalance>
+			+ Into<u128>; // cannot do From<u128>, until LiftedFixedBalance integer part is larger than 128
+			  // bit
+		/// The maximum hops in the route.
+		type MaxHopsInRoute: Get<u32> + TypeInfo;
```
dzmitry-lahoda, line 63: do we really can ever have u32 hopes? i think u8 would be better. Also, please consider making it #[pallet::constat] so we know it does not chages during runtime. 

i would even make it u4 (via Validatable trat) to make more predictable wights and more resonable amounts of hops. (16 hops is already very big loop)

frame/dex-router/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {}
+
+	impl<T: Config> Pallet<T> {}
+
+	impl<T: Config> DexRouter<T::AssetId, T::PoolId> for Pallet<T> {
+		fn update_route(
+			asset_pair: (T::AssetId, T::AssetId),
+			route: Option<Vec<DexRouteNode<T::PoolId>>>,
```
dzmitry-lahoda, line 141: Consider using bounded vector. Does pallet::constat get provides constant to use as bound?

frame/dex-router/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+		) -> DispatchResult {
+			let k1 = asset_pair.0;
+			let k2 = asset_pair.1;
+			match route {
+				Some(route_vec) => {
+					let bounded_route =
+						route_vec.clone().try_into().map_err(|_| Error::<T>::MaxHopsExceeded)?;
+					if DexRoutes::<T>::contains_key(k1, k2) {
+						DexRoutes::<T>::insert(k1, k2, DexRoute::Direct(bounded_route));
```
dzmitry-lahoda, line 150: let exisitng = ...;
insert();
let event = if exising { update } else { add };
deposit(event)

this males flow more linear and less error prone

dzmitry-lahoda:
add at leas one test or remove file

frame/composable-traits/src/dex.rs
vivekvpandya: Currently I think `buy` can not be implemented. It requires `CurveAmm` trait to expose `buy`, `sell` like interface. And possibly `get_dx` like function in pallets. if we want `buy` `sell` then perhaps its batter to have separate PR enhancing AMMs for above functions.

frame/dex-router/src/lib.rs
vivekvpandya: IIUC `BoundedVec` requires `Get<u32>`

frame/dex-router/src/lib.rs
vivekvpandya: AFAIU `BoundedVec` is only beneficial for `Substrate::Storage` , no need to use it over `Vec` in regular code.

frame/composable-traits/src/dex.rs
vivekvpandya: I think with adding `get_dy` in `CurveAmm` should enable writing `buy`, `sell`. 
But I wanted to know why we need `buy`, `sell` instead exchange? To me it seems that UI can make RPC call to pallet invoking `get_dy()` function and get amount of token that can be received for given input amount. And flipping the direction just requires to change in `i` and `j` in `get_dy()` Hence exchange should be sufficient.

frame/composable-traits/src/dex.rs
vivekvpandya: `CurrencyPair` has `base` and `quote` but usually these should be able to use symmetrically.

frame/dex-router/src/lib.rs
vivekvpandya: Not in this PR. I think that also applied to Dex pallets as well. So I will do it later.

frame/composable-traits/src/dex.rs
dzmitry-lahoda: afaik limit os sell is this price or larger. limit of buy is this price or smaller. 
can rename into from to?
user may not care what swap he uses, stable or uniswap, so pallets seems should have api for pairs regardless of amm, for pairs.

vivekvpandya: Sorry that you got confused, but  `CurveAmm` is name for generic trait which has `exchange` function. It is not specific to Curve.Finance . We may consider to rename it in separate PR.
`exchange` function does not require price to be passed. It just needs balance i,e how many token X you want to swap, based on that amount you get token Y.

vivekvpandya: @KaiserKarel @dzmitry-lahoda for buy/sell I have separate PR which adds functionality to CurveAmm trait first. Progess can be tracked at https://github.com/ComposableFi/composable/pull/550

frame/dex-router/src/lib.rs
dzmitry-lahoda: u8 as u32. so in runtime setup it would be clearly visible the limit. @hussein-aitlahcen would it be helpful to setup runtime if constants will tell what preferred bounds they have? like u8 or u32? i know there are already u4 imitations?

frame/dex-router/src/lib.rs
dzmitry-lahoda: not in this pr, but BoundedVec documents limits and avoids more code like in codebase

``
let bounded_route =
						route_vec.clone().try_into().map_err(|_| Error::<T>::MaxHopsExceeded)?;
```
afaik we have support crate to help with that

File: frame/dex-router/src/lib.rs (hunk 6 above)
Comment (dzmitry-lahoda) [reply frame/dex-router/src/lib.rs:46]: thanks

**Review by dzmitry-lahoda — DISMISSED** [review]

File: frame/dex-router/src/lib.rs (hunk 8 above)
Comment (hussein-aitlahcen) [reply frame/dex-router/src/lib.rs:154]: It is a good thing to avoid arbitrary Vec over anything we are writing as it will enforce the usage of this type up to the pallet and up to the extrinsic

File: frame/dex-router/src/lib.rs [hunk 10]
```diff
@@ -0,0 +1,264 @@
...
+							},
+							DexRouteNode::Uniswap(pool_id) => {
+								if !T::ConstantProductDex::pool_exists(*pool_id) {
+									return Err(Error::<T>::PoolDoesNotExist.into())
+								}
+							},
+						}
+					}
+					let bounded_route =
+						route_vec.clone().try_into().map_err(|_| Error::<T>::MaxHopsExceeded)?;
```
hussein-aitlahcen, line 175: Please use BoundedVec as input as here you check if routes are valid even if you exceed the max hops.

frame/dex-router/src/lib.rs
```diff
@@ -0,0 +1,272 @@
...
+			dx: T::Balance,
+		) -> Result<T::Balance, DispatchError> {
+			let route = Self::get_route(asset_pair).ok_or(Error::<T>::NoRouteFound)?;
+			let mut dx_t = dx;
+			let mut dy_t = T::Balance::zero();
+			for route_node in &route {
+				match route_node {
+					DexRouteNode::Curve(pool_id) => {
+						dy_t = T::StableSwapDex::exchange(
+							who,
```
dzmitry-lahoda, line 245: let x = match dex { a => T::StableSwapDex::exchange, b => DexRouteNode::Uniswap}.

x(...all parameters here..).

dzmitry-lahoda: are you ok doing this in next PR?

dzmitry-lahoda: if yes, add TODO here

frame/dex-router/src/lib.rs
```diff
@@ -0,0 +1,272 @@
...
+						dx_t = dy_t;
+					},
+					DexRouteNode::Uniswap(pool_id) => {
+						dy_t = T::ConstantProductDex::exchange(
+							who,
+							*pool_id,
+							0_u8.into(),
+							1_u8.into(),
+							dx_t,
+							T::Balance::zero(),
```
dzmitry-lahoda, line 262: what is balance::Zero? is it slippage? please add TODO to make it be passed from input

dzmitry-lahoda (commented):
i am mostly ok, so need to make sure we ok with slippage (do not forget)

frame/dex-router/src/lib.rs
vivekvpandya: Kind of, that last parameter indicates minimum value expected from exchange. it that is not satisfied exchange will return error. Even if we want to take it from input , it will be applicable to last swap only and that is not very important for now. In future if required we may add it.

frame/dex-router/src/lib.rs
vivekvpandya: I tried doing but I see compilation errors. Please do the changes after this PR is done.

dzmitry-lahoda (approved):

hussein-aitlahcen (dismissed):

seunlanlege:
Benchmarks would be nice

vivekvpandya:
> Benchmarks would be nice

Noted. I or @andor0  will add it in next PR.

hussein-aitlahcen (approved):

seunlanlege:
Actually we should add it to this pr. So the benchmarks pr isn't accidentally omitted when cherry picking to the release branch

vivekvpandya:
> Actually we should add it to this pr. So the benchmarks pr isn't accidentally omitted when cherry picking to the release branch

That is nice, but since DEX are not ready yet, I don't know if benchmarks can be added such that they are stable and does not require any follow-up PRs.

dzmitry-lahoda (approved):

PR #447, opened 2022-01-06
One release

haroldsphinx:
- use one workflow to manage all releases


.github/workflows/client-release.yml
```diff
@@ -2,11 +2,8 @@ name: "Dali-chachacha Community Release"
 
 on:
   push: 
-    tags: 
-      - 'dali-*'
-
-env:
-  CHAIN: "dali"
+    branches: 
+      - 'one-release'
```
seunlanlege, line 6: Why not just `release`?

seunlanlege (changes requested):
I also don't see the check runtime workflow

.github/workflows/client-release.yml
haroldsphinx: it's still WIP...thats my test branch

haroldsphinx:
> I also don't see the check runtime workflow

its on a separate workflow https://github.com/ComposableFi/composable/blob/main/.github/workflows/runtime-check.yml

seunlanlege (approved):

nimboya (approved):

PR #450, opened 2022-01-06
Ben/rustfmt pre commit hook

benluelo:
This PR adds a simple pre-commit hook to run `make style`. It formats the whole repository (as `make style` does) and then re-stages previously staged files.

### DIFF
--- a/.tools/README.md
+++ b/.tools/README.md
@@ -1,0 +1,5 @@
+# Tools
+
+This folder contains various useful script that may make development easier or more streamlined.
+
+Each folder should contain a `README.md` describing what each of the contained scripts are, and how to use them.
--- a/.tools/hooks/README.md
+++ b/.tools/hooks/README.md
@@ -1,0 +1,15 @@
+# Hooks
+
+This folder contains scripts that are designed to be run as [git hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks).
+
+Each file name should follow the format of `<hook name>.<descriptive name>`; for example `pre-commit.style` for a pre-commit hook that runs `make style`. This allows for multiple scripts for the same hook to exist.
+
+To install a script, simply copy the contents of the script you want to install to `.git/hooks/<hook name>` (or append it if you're already using said hook). There will probably be various `<hook name>.sample` files in `.git/hooks/`; they are populated by git and serve as examples for each of the possible hooks.
+
+***NOTE:*** **Hooks will only be run if the file name matches the hook name exactly!**
+
+## `pre-commit.style`
+
+The style `pre-commit` hook runs `make style` to format all files and re-stage all originally changed files. This can be useful to help avoid "rustfmt" commits to get CI to pass.
+
+**NOTE**: You probably want to disable or  turn this off when merging and/ or rebasing.
--- a/.tools/hooks/pre-commit.style
+++ b/.tools/hooks/pre-commit.style
@@ -1,0 +1,31 @@
+#!/bin/sh
+
+if git rev-parse --verify HEAD >/dev/null 2>&1; then
+	against=HEAD
+else
+	# Initial commit: diff against an empty tree object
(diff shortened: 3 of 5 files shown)


hussein-aitlahcen:
Is there an easy way for someone to disable this? If not, I don't think it's a good thing to enforce this kind of stuff when we know the CI will do.

benluelo:
> Is there an easy way for someone to disable this? If not, I don't think it's a good thing to enforce this kind of stuff when we know the CI will do.

I honestly don't see any reason why one would want to disable this, since any non-formatted code will fail CI... however it is fairly trivial to bypass with [--no-verify](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---no-verify)

KaiserKarel:
> My main reason for this PR is that my rustfmt is not on nightly, and as such can't use the configurations in the rustfmt.toml; any file I edit has to be re-formatted with the nightly formatter. Perhaps I'm alone in this?

Cargo should pick up the toolchain.toml and use nightly for everything.

hussein-aitlahcen:
> > Is there an easy way for someone to disable this? If not, I don't think it's a good thing to enforce this kind of stuff when we know the CI will do.
> 
> I honestly don't see any reason why one would want to disable this, since any non-formatted code will fail CI... however it is fairly trivial to bypass with [--no-verify](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---no-verify)
> 
> My main reason for this PR is that my rustfmt is not on nightly, and as such can't use the configurations in the `rustfmt.toml`; any file I edit has to be re-formatted with the nightly formatter. Perhaps I'm alone in this?

I just fear of commit pre-hooks being run when rebasing and manually fixing conflicts for instance, something I don't want to happen ^_^

benluelo:
> > My main reason for this PR is that my rustfmt is not on nightly, and as such can't use the configurations in the rustfmt.toml; any file I edit has to be re-formatted with the nightly formatter. Perhaps I'm alone in this?
> 
> Cargo should pick up the toolchain.toml and use nightly for everything.

I just checked and mine definitely doesn't; I'll review my setup and see if there's something wrong.

> I just fear of commit pre-hooks being run when rebasing and manually fixing conflicts for instance, something I don't want to happen ^_^.
> More than just that, I tend to havoid having implicit side effect when executing a git command.

Totally fair - I spoke with Karel, and he said we could put it in a `.tools` directory so people can enable it as they see fit (with instructions on how to do so in said directory)
He also told me that `.git` is not synced, which I didn't know, hence the 0 files changed (hah!) I'll move the script I wrote to a new `.tools` directory and add instructions on how to use it :+1:

hussein-aitlahcen:
> > > My main reason for this PR is that my rustfmt is not on nightly, and as such can't use the configurations in the rustfmt.toml; any file I edit has to be re-formatted with the nightly formatter. Perhaps I'm alone in this?
> > 
> > 
> > Cargo should pick up the toolchain.toml and use nightly for everything.
> 
> I just checked and mine definitely doesn't; I'll review my setup and see if there's something wrong.
> 
> > I just fear of commit pre-hooks being run when rebasing and manually fixing conflicts for instance, something I don't want to happen ^_^.
> > More than just that, I tend to havoid having implicit side effect when executing a git command.
> 
> Totally fair - I spoke with Karel, and he said we could put it in a `.tools` directory so people can enable it as they see fit (with instructions on how to do so in said directory) He also told me that `.git` is not synced, which I didn't know, hence the 0 files changed (hah!) I'll move the script I wrote to a new `.tools` directory and add instructions on how to use it 👍

Alright perfect, thanks for considering this :)

benluelo:
I took some liberty with the structure of the directory and documentation, if anything should be done differently please let me know!

KaiserKarel (approved):
Nice! structure is good as well.

hussein-aitlahcen (approved):

PR #460, opened 2022-01-09
Karel/clippy forbid panics

KaiserKarel:


KaiserKarel:
This should trigger loads of failures right now, mainly related to slice indexing as well as usage of some other stuff.

seunlanlege (approved):

frame/assets/src/lib.rs
```diff
@@ -49,8 +49,18 @@
...
 
-#![cfg_attr(not(test), warn(clippy::disallowed_method, clippy::indexing_slicing))] // allow in tests
-#![warn(clippy::unseparated_literal_suffix, clippy::disallowed_type)]
+#![cfg_attr(
+	not(test),
+	warn(
+		clippy::disallowed_method,
+		clippy::disallowed_type,
+		clippy::indexing_slicing,
+		clippy::todo,
```
dzmitry-lahoda, line 58: how we should mark to do items? todo was nice way to do it

seunlanlege: I would say that we shouldn't mark todo items, and prs should either finish the implementation or not be merged

KaiserKarel: I can remove the todo macro from the crates, and enable it in CI for releases if that makes more sense. But I think for PR's to main, `todo` should be handled by then to be honest.

dzmitry-lahoda (dismissed):

dzmitry-lahoda (dismissed):

dzmitry-lahoda (approved):

benluelo (approved):

PR #461, opened 2022-01-09
Feature/simnode workflow

haroldsphinx:
- Add backup script to CI
- Add workflow step to check simonode test


.github/workflows/check.yml
```diff
@@ -100,6 +100,26 @@ jobs:
...
+          args: --release -p simnode
+
+      - name: Simnode Test
+        run: |
+          YDATE=$(date -d yesterday +'%m-%d-%Y')
+          FILENAME=cl-1-$YDATE.zip
+          GS_BUCKET="composable-picasso-data-sync"
+          sudo gsutil cp gs://$GS_BUCKET/$FILENAME .
+          sudo unzip $FILENAME -d  /tmp/db
+          ./target/release/simnode --chain=picasso --base-path=/tmp/db --wasm-runtime-overrides=./target/release/picasso_runtime.compact.wasm --pruning=archive --execution=wasm
```
seunlanlege, line 121: I think this should refer to a folder such that this is the only wasm file in that folder.

.github/workflows/picashot.yml
```diff
@@ -0,0 +1,18 @@
+name: Check Runtime
```
seunlanlege, line 1:
```suggestion
name: Nightly backup
```

.github/workflows/check.yml
```diff
@@ -95,6 +95,26 @@ jobs:
       - x64
       - sre
     steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
```
KaiserKarel, line 101: We've added a toolchain.toml, which pins the rust toolchain we're using. actions-rs does not support that yet, so you'll have to manually install rustup + cargo.

.github/workflows/check.yml
```diff
@@ -21,23 +21,18 @@ jobs:
...
         with:
           fetch-depth: 0     
-      - name: Setup toolchain
-        run: |
-         curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
-         echo "$HOME/.cargo/bin:$PATH" >> $GITHUB_PATH
       - name: Check
         run: |
-          SKIP_WASM_BUILD=1 cargo check --all --benches
+          SKIP_WASM_BUILD=1 /home/runner/.cargo/bin/cargo check --all --benches
```
KaiserKarel, line 29: Can we add cargo to the path, to abstract how the actual runner is configured?

.github/workflows/check.yml
```diff
@@ -96,19 +86,20 @@ jobs:
       - sre
     steps:
       - uses: actions/checkout@v2
-      - name: Install toolchain for cargo-llvm-cov
-        run: |
-         curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
-         echo "$HOME/.cargo/bin:$PATH" >> $GITHUB_PATH
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
```
KaiserKarel, line 91: So we still need to remove this right?

.github/workflows/client-release.yml
```diff
@@ -27,9 +32,20 @@ jobs:
...
+
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+
+    - name: 🔨 Build
+      uses: actions-rs/cargo@v1
+      with:
+        command: build
+        args: --release --bins
```
KaiserKarel, line 44: This also won't respect the toolchain.toml, meaning that tests are not run vs the same version of rustc as the final build

KaiserKarel (changes requested):
Would be good to add cargo to the path, as well as removing actions-rs

KaiserKarel (approved):
After discussing offline, we decided to add the feedback in a later PR

seunlanlege (approved):

nimboya (approved):

PR #476, opened 2022-01-11
[draft] currency 

dzmitry-lahoda:


rfcs/0003-currency.md
```diff
@@ -0,0 +1,116 @@
...
+ - will make serde setup complicated (need to write and consistently apply static singletons
+   to all places with asset id)
+ - validate will involve at minimum in memory cache call (in worth case db call) during
+   extrinsic invocation
+ - will need to disable this during calls when it is really no need for validation (new
+   currency mapping)
+ - normal path will pay price (validate each time), in instead when fail pays only (like
+   trying to transfer non existing asset id)
+ - we cannot guarantee existence of asset as it may be removed during transaction (so we
+   should make removal exclusive case)
```
KaiserKarel, line 17: We could always create a wrapper type: `Exists<T: QueryExists>` and if your API needs asset_ids to be valid, it could ask for a `Exists<CurrencyId>`, where `impl QueryExists for CurrencyId` would perform the queries needed. However if the inner function does actions such as transfer etc, it is a waste of storage calls.

dzmitry-lahoda: Yes. So all apis will work with may be exists as they do now. Just need to make it clear. in some cases maybe would be cast into AssetIdForSure to act on that. But i am not yeat aware of this case. So in most cases Iam about zero waste.  Assuming that eventually and module checks currncy exists and we do not check it on entry..Non checking at all is attack with spam and garbage. This way may have well known ids to avoid storage check.

rfcs/0003-currency.md
```diff
@@ -0,0 +1,161 @@
...
+- `Remote Symbols`. 
+
+All currencies are derived, so some currencies can be derived in local consensus. So:
+- `Dillution`. If currency was direcly derived from protocol in local consensus than can tell exact ration of one amount needed to swap for other amount. With time currency amounts can change (locked, unlocked, minted), which will changed dillution factor. So we cannot tell nothing about external currencies dillution factors, only Oracles can tell. In this case we can observer whole chain of of tokens as they wrap each other and dillute. Process is not nessecary bidirectional.
+- `Local Simple Dillution` - when for each amount there is mint of other amount, may be allowing different reserver/hold propreties. May eventually be not one to one as other can be burnt. But because of can witness total supply, can say ratio. So what to do if curreny transfered from and to system?
+
```
dzmitry-lahoda, line 34: still in progress, as i saw karel did some improvments; i did to testing currencies. acala is good, but seems we are not compatible with their approach

PR #477, opened 2022-01-11
remove unnecessary cargo clean calls

KaiserKarel:


hussein-aitlahcen (approved):
Are we sure this is fine? We experienced issues when upgrading Rust if the cache was not wiped

KaiserKarel:
> Are we sure this is fine? We experienced issues when upgrading Rust if the cache was not wiped

I'll refer to @haroldsphinx on this. The two existing calls do not actually wipe the full cache I think, besides that seems to be CI independent logic to me, and if possible should remain outside of the check.yml file

dzmitry-lahoda (approved):

haroldsphinx (approved):

PR #481, opened 2022-01-12
fix asset count to change an existing asset

JesseAbram:


KaiserKarel:
What are we fixing here exactly?

JesseAbram:
> What are we fixing here exactly?

pretty much if you want to make a change on an asset, like the slash amount, or min answers etc, it will increment the asset count no matter what, now it will only do it when the asset is a new asset

KaiserKarel (dismissed):

dzmitry-lahoda (commented):
what happens if getting default? may be use if let pattern match?

JesseAbram:
> what happens if getting default? may be use if let pattern match?

hmmm ya took that into account, handled one better with ok or error, the others I made sure the default which would be 0 is ok

KaiserKarel (approved):

frame/oracle/src/lib.rs
```diff
@@ -612,15 +615,15 @@ pub mod pallet {
 			price: T::PriceValue,
 			asset_id: T::AssetId,
 		) {
-			let asset_info = Self::asset_info(asset_id);
+			let asset_info = Self::asset_info(asset_id).unwrap_or_default();
```
hussein-aitlahcen, line 618: I think it would be nice to make this consistent with `.ok_or(Error::<T>::InvalidAssetId)?`, such as tweaking the function `Self::asset_info` to directly return a result and never rely on the default implementation, which is basically a null struct?

JesseAbram: concerned about panics and state changes I thought the best way to handle this would be to pass it down from the initial iteration, also should save state reads too so that is just better

hussein-aitlahcen (approved):

frame/oracle/src/lib.rs
hussein-aitlahcen: :+1:

KaiserKarel (approved):

PR #483, opened 2022-01-13
CU-1yykjpw Fix calc_utilization_ratio() and add proptest for it.

vivekvpandya:


ComposableFiB0t:
Task linked: [CU-1yykjpw Add tests for composable_traits/rate_model.rs](https://app.clickup.com/t/1yykjpw)

frame/composable-traits/src/lending/math.rs
```diff
@@ -39,12 +39,10 @@ pub fn calc_utilization_ratio(
 	}
 
 	let total = cash.safe_add(&borrows)?;
-	// also max value is 1.00000000000000000, it still fails with u8, so mul by u16 and cast to u8
-	// safely
 	let utilization_ratio = borrows
 		.checked_div(&total)
 		.expect("above checks prove it cannot error")
-		.checked_mul_int(100_u16)
+		.checked_mul_int(100_u8)
```
dzmitry-lahoda, line 45: how that fixed issue?

frame/composable-traits/src/rate_model.rs
```diff
@@ -47,28 +47,6 @@ pub type Ratio = FixedU128;
 /// accounts to length of year)
 pub const SECONDS_PER_YEAR: DurationSeconds = 365 * 24 * ONE_HOUR;
 
-/// utilization_ratio = total_borrows / (total_cash + total_borrows)
```
dzmitry-lahoda, line 50: assume copy paste

frame/composable-traits/src/lending/math.rs
vivekvpandya: `Percent::from_percent()` requires `u8` input. `checked_mut_int()` can convert `FixedU128` to `u8` but earlier for some reason `u16->u8` conversion is done. However with new change it works and proptest also works fine. So for now should it be enough?

frame/composable-traits/src/rate_model.rs
vivekvpandya: yes, after removing it no compile-time complaints nor any tests broke.

dzmitry-lahoda (dismissed):

dzmitry-lahoda (dismissed):

andor0 (dismissed):

dzmitry-lahoda (approved):

andor0 (approved):

PR #484, opened 2022-01-13
CU-1wty1zv fixed lending

dzmitry-lahoda:
- removed fixed point things from public api
- consistently lift up precision in math,  instead of lowering it (hence cannot accrue on small amounts)
- called liquidation engine (drafted next step for xcm or dex engines)
- fixed auction test


ComposableFiB0t:
Task linked: [CU-1wty1zv liquidation: initialize engine and during sale via api](https://app.clickup.com/t/1wty1zv)

frame/lending/src/lib.rs
```diff
@@ -521,54 +486,44 @@ pub mod pallet {
...
+			origin: OriginFor<T>,
+			_input: UpdateInput<T::LiquidationStrategyId>,
+		) -> DispatchResultWithPostInfo {
+			let _who = ensure_signed(origin)?;
+			// 1. validate owner
+			// 2. update configuration
+			// 3. what is owner updates and forces liquidations? ok.
+			Err(DispatchError::Other("not implemented").into())
+		}
+
```
hussein-aitlahcen, line 526: I think it would be nice to introduce `TODO: small-description - clickup-ticket` when we just deploy a stub.

frame/liquidations/src/lib.rs
```diff
@@ -58,47 +66,182 @@ pub mod pallet {
...
-			_total_amount: Self::Balance,
-		) -> Result<Self::LiquidationId, DispatchError> {
-			Self::deposit_event(Event::<T>::PositionWasSentToLiquidation {});
-			Err(DispatchError::Other("todo"))
+			from_to: &Self::AccountId,
+			order: Sell<Self::MayBeAssetId, Self::Balance>,
+			configuration: Vec<Self::LiquidationStrategyId>,
+		) -> Result<T::OrderId, DispatchError> {
+			if configuration.is_empty() {
+				let configuration = Strategies::<T>::get(DefaultStrategyIndex::<T>::get())
```
hussein-aitlahcen, line 217: To avoid having an `expect`, we could perhaps decouple the storage `Strategies` with a second `DefaultStrategy` storage?

hussein-aitlahcen, line 217: To avoid having an `expect`, we could perhaps decouple the storage `Strategies` with a second `DefaultStrategy` storage?

frame/lending/src/lib.rs
hussein-aitlahcen, line 526: I think it would be nice to introduce TODO: small-description - clickup-ticket when we just deploy a stub.

hussein-aitlahcen (approved):
Quick comments

hussein-aitlahcen (approved):
Big PR but LGTM

vivekvpandya (approved):

PR #486, opened 2022-01-13
CU-1uq4a7x - ValidateUnsigned instead of SignedExtension in Crowdloan

hussein-aitlahcen:
From what I have been reading online and based on the substrate source code, a SignedExtension cannot be used to allow an unsigned extrinsic to be dispatched. In fact, the online documentation is stating that it is usable but that is not true. Even on the claiming mechanism of Polkadot, the implementation is using a `ValidateUnsigned` impl for the pallet. What is strange is that there is a `validate_unsigned` [function exposed by SignedExtension](https://docs.substrate.io/rustdocs/latest/sp_runtime/traits/trait.SignedExtension.html#method.validate_unsigned).

I refactored the tests to reflect the change, deployed the devnet and ran some integrations tests successfully with this change. I was not able  to submit an unsigned extrinsic using the `SignedExtension` mechanism. And in fact, if no pallet is implementing `ValidateUnsigned` in the runtime, the default implementation expanded by `construct_runtime` is refusing any unsigned extrinsic with a `NoUnsignedValidator`.

### DIFF
--- a/frame/crowdloan-rewards/src/lib.rs
+++ b/frame/crowdloan-rewards/src/lib.rs
@@ -47,15 +47,7 @@ Reference for proof mechanism: https://github.com/paritytech/polkadot/blob/maste
 	unused_extern_crates
 )]
 
-use codec::{Decode, Encode};
-use frame_support::{
-	pallet_prelude::{InvalidTransaction, ValidTransaction},
-	traits::IsSubType,
-	unsigned::{TransactionValidity, TransactionValidityError},
-};
 pub use pallet::*;
-use scale_info::TypeInfo;
-use sp_runtime::traits::{DispatchInfoOf, SignedExtension, Zero};
 
 pub mod models;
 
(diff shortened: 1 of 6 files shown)


ComposableFiB0t:
Task linked: [CU-1uq4a7x Move crowdloan from SignedExtension to ValidateUnsigned](https://app.clickup.com/t/1uq4a7x)

dzmitry-lahoda (dismissed):
just as note, seesome tx number is made 1, not 2

hussein-aitlahcen:
> just as note, seesome tx number is made 1, not 2

I reverted the transaction version update because it was incremented when adding the SignedExtension. Since I reverted the change, I decremented the version as it won't introduce an upgrade in the transaction signature.

benluelo (approved):

KaiserKarel (approved):

PR #472, opened 2022-01-11
Feature/simnode workflow

haroldsphinx:
- Cleanup cargo path

PR #489, opened 2022-01-14
Check including wasm build

KaiserKarel:
Changes the check step to exhaustively check every packages, including the WASM build and benchmarks. integration tests are still skipped as they are incompatible with `runtime-benchmarks` at the moment. Will add that in a second PR

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -26,14 +26,9 @@ jobs:
       - sre
     steps:
       - uses: actions/checkout@v2
-      - run: |
-          SKIP_WASM_BUILD=1 /home/runner/.cargo/bin/cargo check --all --benches
-      - name: Benchmark
+      - name: Compile Check
         run: |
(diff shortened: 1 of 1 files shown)


hussein-aitlahcen (dismissed):

.github/workflows/check.yml
```diff
@@ -26,14 +26,9 @@ jobs:
       - sre
     steps:
       - uses: actions/checkout@v2
-      - run: |
-          SKIP_WASM_BUILD=1 /home/runner/.cargo/bin/cargo check --all --benches
```
seunlanlege, line 30: I still think we need this, otherwise we're missing std check

seunlanlege: `runtime-benchmarks` and `std` are mutually exclusive features, personally of the opinion we don't need this #472 will run the actual runtime benchmarks if runtime changes are detected

KaiserKarel: I've split the check for wasm benchmarks + std check up. As we have enough runnersm this shouldn't affect CI times and gives early feedback of incompatibility.

hussein-aitlahcen (approved):

andor0 (approved):

PR #492, opened 2022-01-14
CU-1ubvcrq - introduce `Validated` and `Validate` for custom codec validation

hussein-aitlahcen:
This PR introduce a new `composable-support` crate that will allow us to add extra pallet support functionality.

I started to implement a validation trait that would allow us to ensure that complex input structs provided by external sources through extrinsics are validated when doing the actual decoding. In fact, we do have some pallets such as the `bonded-finance` and `lending` that are doing this somehow manually. Such as ensuring that the domain of a type is restricted to some values or clamped within an arbitrary range.

### DIFF
--- a/frame/composable-support/src/validation.rs
+++ b/frame/composable-support/src/validation.rs
@@ -1,21 +1,62 @@
-use core::marker::PhantomData;
+use core::{marker::PhantomData, ops::Deref};
 use scale_info::TypeInfo;
 
 /// Black box that embbed the validated value.
-#[derive(Eq, PartialEq, Debug, TypeInfo, codec::Encode)]
+#[derive(Default, Copy, Clone, PartialEq, Eq, Debug, TypeInfo)]
 pub struct Validated<T, U> {
-	pub value: T,
+	value: T,
 	_marker: PhantomData<U>,
 }
 
-pub trait Validate<U> {
-	fn validate(&self) -> Result<(), &'static str>;
+impl<T: Copy, U> Validated<T, U> {
+	#[inline(always)]
+	pub fn value(&self) -> T {
+		self.value
+	}
 }
 
-impl<T: codec::Decode + Validate<U>, U> codec::Decode for Validated<T, U> {
+impl<T, U> Deref for Validated<T, U> {
+	type Target = T;
+	#[inline(always)]
+	fn deref(&self) -> &Self::Target {
+		&self.value
+	}
+}
+
+impl<T, U> AsRef<T> for Validated<T, U> {
+	#[inline(always)]
+	fn as_ref(&self) -> &T {
+		&self.value
+	}
+}
+
+pub trait Validate<U>: Sized {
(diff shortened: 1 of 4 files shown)


ComposableFiB0t:
Task linked: [CU-1ubvcrq Introduce `Validated` newtype over SCALE](https://app.clickup.com/t/1ubvcrq)

frame/composable-support/src/validation.rs
```diff
@@ -0,0 +1,98 @@
...
+/// Black box that embbed the validated value.
+#[derive(Eq, PartialEq, Debug, TypeInfo, codec::Encode)]
+pub struct Validated<T, U> {
+	pub value: T,
+	_marker: PhantomData<U>,
+}
+
+pub trait Validate<U> {
+	fn validate(&self) -> Result<(), &'static str>;
+}
```
KaiserKarel, line 13: Changing this to `fn validate(self) -> Result<Self, ...>;` will allow implementors to modify as well, such as performing sorting operations etc.

frame/composable-support/src/validation.rs
```diff
@@ -0,0 +1,98 @@
+use core::marker::PhantomData;
+use scale_info::TypeInfo;
+
+/// Black box that embbed the validated value.
+#[derive(Eq, PartialEq, Debug, TypeInfo, codec::Encode)]
+pub struct Validated<T, U> {
+	pub value: T,
+	_marker: PhantomData<U>,
+}
```
KaiserKarel, line 9: Should also implement AsRef, Into, Etc. I'd suggest removing the pub modifier, so that constructing an invalid Validated becomes impossible.

frame/composable-support/src/validation.rs
KaiserKarel: Perhaps also have it return a DispatchError for better substrate compatibility.

KaiserKarel: Or look into how errors are produced, as I think they have an intermediate trait for the covnersion step

dzmitry-lahoda:
/// A marker trait for types that wrap other encodable type.
///
/// Such types should not carry any additional information
/// that would require to be encoded, because the encoding
/// is assumed to be the same as the wrapped type.
///
/// The wrapped type that is referred to is the [`Deref::Target`].
pub trait WrapperTypeEncode: Deref {}

frame/composable-support/src/validation.rs
hussein-aitlahcen: done

hussein-aitlahcen:
> /// A marker trait for types that wrap other encodable type. /// /// Such types should not carry any additional information /// that would require to be encoded, because the encoding /// is assumed to be the same as the wrapped type. /// /// The wrapped type that is referred to is the [`Deref::Target`]. pub trait WrapperTypeEncode: Deref {}

Used it, thanks for pointing :)

dzmitry-lahoda (approved):

andor0 (approved):

PR #500, opened 2022-01-17
Releasenotes

nimboya:
- one release file with all release info


.maintain/test.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+chains=(
+  "./runtime/composable/src/weights,composable-dev"
```
haroldsphinx, line 4: lets comment out the composable part for now

PR #511, opened 2022-01-19
cleanup some CI cruft

KaiserKarel:


rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+[toolchain]
+channel = "nightly-2021-11-29"
+components = [ "rustfmt", "rustc", "rust-std", "cargo", "clippy", "llvm-tools-preview"]
+targets = [ "wasm32-unknown-unknown" ]
+profile = "minimal"
```
seunlanlege, line 5: we removed this because we shouldn't be building the project against nightly. Nightly has been added to the tools that need it

KaiserKarel: Then here we should set it to stable, or use a different toolchain.toml for different sections of the build.

KaiserKarel: Same goes for having one place which define what components are necessary to build the project. For developers, their normal flow of using nightly is fine. In CI and for releases we should instead check if it is buildable by stable.

KaiserKarel: In the Makefile.toml, we can change the build command to use stable for example.

hussein-aitlahcen: Yeah I think we should pin it, both for devs and in the CI, the best would probably be to have a single file that would install both stable and nightly. I know we can do that with nix and everyone (dev + ci) would have the same pinned versions for stable/nightly.

KaiserKarel: For pinning to stable, that can be done in Cargo.toml as well using the new rust version field: https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field. I'd prefer using the toolchain, and having nix use that to figure out what version to pick.

dzmitry-lahoda:
not sure if `.yarn` files added are needed

KaiserKarel:
> not sure if `.yarn` files added are needed

Crap, those should indeed have been ignored

hussein-aitlahcen (approved):

benluelo (approved):

PR #514, opened 2022-01-19
Add <Bytes> tag for polkadotjs signature in crowdloan

hussein-aitlahcen:
[The public polkadotjs api signRaw is wrapping the message in this tag.](https://github.com/polkadot-js/extension/blob/master/packages/extension-base/src/background/RequestBytesSign.ts#L24)

### DIFF
--- a/frame/crowdloan-rewards/src/lib.rs
+++ b/frame/crowdloan-rewards/src/lib.rs
@@ -411,8 +411,12 @@ pub mod pallet {
 		relay_account: RelayChainAccountId,
 		proof: &MultiSignature,
 	) -> bool {
-		let mut msg = prefix.to_vec();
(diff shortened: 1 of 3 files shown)


Dom-Roth (approved):

frame/crowdloan-rewards/src/lib.rs
```diff
@@ -411,8 +411,12 @@ pub mod pallet {
 		relay_account: RelayChainAccountId,
 		proof: &MultiSignature,
 	) -> bool {
-		let mut msg = prefix.to_vec();
+		let wrapped_prefix: &[u8] = b"<Bytes>";
```
dzmitry-lahoda, line 414: const

dzmitry-lahoda (approved):

KaiserKarel (approved):

PR #515, opened 2022-01-20
Add cleanup step

haroldsphinx:


.github/workflows/check.yml
```diff
@@ -25,6 +25,12 @@ jobs:
       - x64
       - sre
     steps:
+      - name: Clean up
+        continue-on-error: true
+        run: |
+          sudo chown -R $USER:$USER $GITHUB_WORKSPACE
+          docker system prune --force --all --volumes
```
KaiserKarel, line 32: Won't this remove our caching?

haroldsphinx: No it won't, cached files are saved to a specified path

seunlanlege (approved):

PR #529, opened 2022-01-20
[1u5ykup] Oracle Runtime Tests, Runtime Tests deps update, Bugfixes & Cleanup

Dom-Roth:
**Greetings**, this PR for the runtime integration-tests contains the following:

1. **Oracle** Extrinsics Runtime Integration Tests
2. **Dependency** reduction & update.
3. **Bugfix**: When polkadotjs.ts received specific error returns, it forwarded them into .isInBlock state & .isFinalized state for successful answers. Leading to undefined objects.
4. **Bugfix**: Mochawesome reporter didn't work, now does. :) The application now reports its results in JSON and HTML. (Example see below)
5. **Bugfix**: The failed tests had no descriptions after the finishing.
6. The tests.ts file now wraps all calls within describe() functions, this way we get nice formatting together with mochawesome.
7. Removed most lambda functions from the code. According to the mocha documentation are lambda (arrow) functions overwriting "this" within their context, which is normally occupied by a special mocha-crafted object. And can therefore cause problems. Reference: https://mochajs.org/#arrow-functions

# In Depth:
## 1. Oracle Extrinsics Runtime Integration Tests
~~As of right now, 67% (4/6) success tests for the extrinsics are integrated. All of them are working, but some might require more result checks.~~ All success tests are implemented, see updates below.

## 2. Dependency reduction & update.
tsconfig-paths: Added!
bluebird: Removed!
Updated:
- ts-node
- typescript
- ramda & @types/ramda
- web3
- @types/node
- @polkadotapi & types
~~I tried to update mocha to v.9 but it stopped finding tests. I found a thread referring to this, but it'd require some restructuring, so I'll come back later to this.~~ See comment below!

## 4. We now have awesome-looking test summaries.
Attached is a screenshot and a link for a live preview.
https://report-example.roth.systems/
Edit: The screenshot can be found here: 
https://report-example.roth.systems/Screenshot%202022-01-20%20at%2016-46-19%20Mochawesome%20Report.png
The image was so big it made the thread hard to read.

Up to date test results, with our latest changes can be found below.

## 5. New Bugs / Problems

~~Since the latest update to the **crowdloan** pallet, the **associate** tests fail. I'll figure out why this happens and provide an update to this thread.
Since the error was so vague (See test report above), I ran the type generator. This didn't solve the problem but leads us to the second point.~~

~~With the newest changes to the **dutchAuction** pallet, the type generator generates a faulty line in **augment-api-tx.ts** for the **ask()** extrinsic. In this context it doesn't know what **ComposableTraitsTimeTimeReleaseFunction** is.
For debug purposes I replaced it with any and the code works. It is just about this item it doesn't know.~~
```typescript
ask: AugmentedSubmittable<(order: ComposableTraitsDefiSell | { pair?: any; take?: any } | string | Uint8Array, configuration: ComposableTraitsTimeTimeReleaseFunction | { LinearDecrease: any } | { StairstepExponentialDecrease: any } | string | Uint8Array) => SubmittableExtrinsic<ApiType>, [ComposableTraitsDefiSell, ComposableTraitsTimeTimeReleaseFunction]>;
```

### DIFF
--- a/integration-tests/runtime-tests/src/utils/polkadotjs.ts
+++ b/integration-tests/runtime-tests/src/utils/polkadotjs.ts
@@ -35,9 +35,9 @@ export function sendUnsignedAndWaitFor<T extends AnyTuple>(
   call: SubmittableExtrinsic<"promise">,
   intendedToFail:boolean
 ): Promise<IEvent<T>> {
-  return new Promise<IEvent<T>>((resolve, reject) => {
+  return new Promise<IEvent<T>>(function (resolve, reject) {
     call
-      .send(res => {
+      .send(function (res) {
         const { dispatchError, status } = res;
         if (dispatchError) {
           if (dispatchError.isModule) {
@@ -49,6 +49,8 @@ export function sendUnsignedAndWaitFor<T extends AnyTuple>(
           }
         }
         if (status.isInBlock || status.isFinalized) {
+          if (res.events.find(e => filter(e.event)) == undefined)
+            return reject(status.toString());
           const event = res.events.find(e => filter(e.event)).event;
           if (filter(event)) {
             resolve(event);
@@ -57,8 +59,8 @@ export function sendUnsignedAndWaitFor<T extends AnyTuple>(
           }
         }
       })
-      .catch((e) => {
-        reject(Error(e.message));
+      .catch(function (e) {
+        reject(Error(e.stack));
       });
   });
 }
@@ -79,9 +81,9 @@ export function sendAndWaitFor<T extends AnyTuple>(
   call: SubmittableExtrinsic<"promise">,
   intendedToFail:boolean
 ): Promise<IEvent<T>> {
-  return new Promise<IEvent<T>>((resolve, reject) => {
+  return new Promise<IEvent<T>>(function (resolve, reject) {
     call
-      .signAndSend(sender, { nonce: -1 }, res => {
+      .signAndSend(sender, { nonce: -1 }, function (res) {
         const { dispatchError, status } = res;
         if (dispatchError) {
           if (dispatchError.isModule) {
@@ -104,6 +106,8 @@ export function sendAndWaitFor<T extends AnyTuple>(
           }
         }
         if (status.isInBlock || status.isFinalized) {
+          if (res.events.find(e => filter(e.event)) == undefined)
+            return reject(status.toString());
           const event = res.events.find(e => filter(e.event)).event;
           if (filter(event)) {
             if (intendedToFail) {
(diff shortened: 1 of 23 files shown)


integration-tests/runtime-tests/src/utils/polkadotjs.ts
```diff
@@ -35,9 +35,9 @@ export function sendUnsignedAndWaitFor<T extends AnyTuple>(
   call: SubmittableExtrinsic<"promise">,
   intendedToFail:boolean
 ): Promise<IEvent<T>> {
-  return new Promise<IEvent<T>>((resolve, reject) => {
+  return new Promise<IEvent<T>>(function (resolve, reject) {
```
hussein-aitlahcen, line 38: The mocha documentation state that `If you do not need to use Mocha’s context, lambdas should work`. All the functions exept the top level one where we do the Mocha `this.settimeout` does not require the context? If so, why refactoring this functions as well?

Dom-Roth: This is mainly an artifact, where I tried to call `this.skip()` within that function, to debug the errors mentioned in 3.
I just didn't revert it afterward.

hussein-aitlahcen: Makes sense :)!

Dom-Roth:
I've finished implementing all success tests, and while adding more checks, I found problems with the call to .setsigner(). So there's an unexpected delay to fix this.

Also couldn't find the solution to the current problems of the crowdloan reward test, but I'll keep trying. I'll add the stack trace and more information about this here later.

Dom-Roth:
**Finally!** All of the extrinsic success tests for the oracle pallet are implemented.
And thanks to the help of @hussein-aitlahcen the crowdloan reward associate() bug is now fixed.

We now have **21 Tests** implemented and working. The results of the current test can be seen below.

With the oracle .reclaimStake() test, we now have a test that needs to wait for a specific block to be reached. This makes the test quite slow in our current setup. (~ 11 minutes)
For the next update, I'll do some structural changes to make our test suites work in parallel. Luckily some/most of these changes overlap with the changes required to update to Mocha version 9.

Dom-Roth:
image

benluelo (dismissed):

Dom-Roth:
There are some docstrings wrong and variables misslabled. I'm fixing them right now.

hussein-aitlahcen (approved):
Well done:)

dzmitry-lahoda (approved):
i have 2 years of experience in ts, and all is just running cli tools and tests:) this looks ok:)

PR #532, opened 2022-01-21
Add BoundedSortedVec and SortedVec

KaiserKarel:
In need of these data types for a different pallet. Mainly modeled after BoundedVec and SortedVec.

### DIFF
--- a/frame/composable-support/src/collections/vec/bounded/sorted_vec.rs
+++ b/frame/composable-support/src/collections/vec/bounded/sorted_vec.rs
@@ -1,0 +1,344 @@
+use crate::collections::vec::sorted::SortedVec;
+use codec::{Decode, Encode, EncodeLike, MaxEncodedLen};
+use core::{
+	ops::{Deref, Index},
+	slice::SliceIndex,
+};
+use frame_support::{traits::Get, BoundedVec};
+use sp_std::{convert::TryFrom, fmt, marker::PhantomData, prelude::*};
+
+/// A bounded, sorted vector.
+///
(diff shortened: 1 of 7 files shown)


frame/composable-support/src/collections/vec/bounded/sorted_vec.rs
```diff
@@ -0,0 +1,344 @@
...
+pub struct BoundedSortedVec<T: Ord, S>(SortedVec<T>, PhantomData<S>);
+
+impl<T: Decode + Ord, S: Get<u32>> Decode for BoundedSortedVec<T, S> {
+	#[inline]
+	fn decode<I: codec::Input>(input: &mut I) -> Result<Self, codec::Error> {
+		let inner = Vec::<T>::decode(input)?;
+		if inner.len() > S::get() as usize {
+			return Err("BoundedSortedVec exceeds its limit".into())
+		}
+		Ok(Self(SortedVec::from_unsorted(inner), PhantomData))
```
dzmitry-lahoda, line 27: hm, feels risky to add O(n log N) into serde.

dzmitry-lahoda: how to express weight in such case?  seem need to write custom weight depending on length some how not only linearly. example could be use full If n log n is case

KaiserKarel: Oh yeah correct, i'll instead check if sorted and otherwise error out

dzmitry-lahoda (approved):

hussein-aitlahcen (approved):

PR #533, opened 2022-01-21
Update picashot

haroldsphinx:


.github/workflows/check.yml
```diff
@@ -32,9 +34,12 @@ jobs:
           docker system prune --force --all --volumes
 
       - uses: actions/checkout@v2
+      - name: Clean up
+        continue-on-error: true
```
dzmitry-lahoda, line 38: why would we continue on error? 

next is very simple and should never error

 cargo check --features=runtime-benchmarks --workspace --exclude integration-tests

.github/workflows/release-drafter.yml
```diff
@@ -4,13 +4,58 @@ on:
     branches:
       - releases
 jobs:
-  update_release_draft:
-    runs-on: ubuntu-latest
+  build-and-publish:
```
dzmitry-lahoda, line 7: is this pr still actual?

PR #547, opened 2022-01-24
pallet mosaic

cor:


frame/mosaic/src/decay.rs
```diff
@@ -0,0 +1,99 @@
...
+
+pub trait Decayer<Balance, BlockNumber> {
+	fn checked_decay(
+		&self,
+		amount: Balance,
+		last_decay_block: BlockNumber,
+		current_block: BlockNumber,
+	) -> Option<Balance>;
+
+	fn full_recovery_period(&self, amount: Balance) -> Option<BlockNumber>;
```
dzmitry-lahoda, line 12: what None return means here?

frame/composable-tests-helpers/src/test/proptest.rs
```diff
@@ -1,8 +1,9 @@
 /// Equivalent of assert_ok when inside a proptest context.
+
```
dzmitry-lahoda, line 2: white space:)

frame/mosaic/Cargo.toml
```diff
@@ -0,0 +1,73 @@
+[package]
+name = "pallet-mosaic"
+version = "0.1.0"
+authors = ["Composable Developers"]
+homepage = "https://composable.finance"
+edition = "2021"
+rust-version = "1.56"
```
dzmitry-lahoda, line 7: not sure if we need this version here, but ok

frame/mosaic/Cargo.toml
```diff
@@ -0,0 +1,73 @@
...
+
+[[bin]]
+name = "plot"
+path = "src/plots.rs"
```
dzmitry-lahoda, line 11: you can make this in test. why? one of variants is to produce docs from test docs

frame/mosaic/README.md
```diff
@@ -0,0 +1,10 @@
+# Mosaic
+
+Pallet implementing an interface for the Mosaic Relayer. As opposed to the EVM-EVM bridge, this pallet takes a different approach and uses `mint` and `burn` operations. 
+Because of that it also limits the mintable amount by the relayer using a decaying penalty.
+
+## Decaying Penalty
+
+At moment N, the relayer has a maximum budget `B`. Minting a token adds a penalty `P` to the relayer. The penalty decreases each block according to decay function `d`, 
+which depends on the penalty, current_block, and last_minted_block. The current maximum amount that the relayer can mint is given by `B - d(P, current_block, last_minted_block)`.
```
dzmitry-lahoda, line 9: `BudgedAmount - decay(PenaltyAmount, blocks_since_last_mint`)

frame/mosaic/src/decay.rs
```diff
@@ -0,0 +1,99 @@
+use frame_support::pallet_prelude::*;
+use num_traits::{CheckedAdd, CheckedDiv, CheckedMul, CheckedSub, One, Saturating, Zero};
+
+pub trait Decayer<Balance, BlockNumber> {
+	fn checked_decay(
+		&self,
+		amount: Balance,
+		last_decay_block: BlockNumber,
```
dzmitry-lahoda, line 8: seems different order of params `d(P, current_block, last_minted_block)`

dzmitry-lahoda: please name these exactly like in readme).

frame/mosaic/src/decay.rs
```diff
@@ -0,0 +1,99 @@
...
+		match self {
+			BudgetPenaltyDecayer::Linear(lin) => lin.full_recovery_period(amount),
+		}
+	}
+}
+
+#[derive(Decode, Encode, TypeInfo, Default, Debug, PartialEq, Clone)]
+pub struct LinearDecay<Balance, BlockNumber> {
+	/// Factor by which we decay every block.
+	factor: Balance,
```
dzmitry-lahoda, line 56: hm, this is not relative to block? why?

dzmitry-lahoda: please check time.rs for examples of decays https://github.com/ComposableFi/composable/blob/ff39c92f7eabfa3218adfb252cef89ba7b3593f3/frame/composable-traits/src/time.rs#L21 too

frame/mosaic/src/decay.rs
```diff
@@ -0,0 +1,99 @@
...
+	BlockNumber: CheckedSub + Saturating + Into<Balance> + TryFrom<Balance> + One + CheckedAdd,
+	Balance: CheckedMul + CheckedDiv + Saturating + Zero,
+{
+	fn checked_decay(
+		&self,
+		amount: Balance,
+		last: BlockNumber,
+		current: BlockNumber,
+	) -> Option<Balance> {
+		let diff = current.saturating_sub(last);
```
dzmitry-lahoda, line 71: safe_sub , do not hide bug behind 0 value

dzmitry-lahoda: please document return value of this trait that it is remaining penalty

frame/mosaic/src/decay.rs
```diff
@@ -0,0 +1,99 @@
...
+		let full_period = amount.checked_div(&self.factor)?;
+		let block_full_period: BlockNumber = TryFrom::<Balance>::try_from(full_period).ok()?;
+		let block_full_period_plus_one: BlockNumber = block_full_period.checked_add(&One::one())?;
+		Some(block_full_period_plus_one)
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
```
dzmitry-lahoda, line 86: please write test which shows call to decay - recovery - decay showing these are kind of identity/quine/fixed point.

may be rename factor to frequency (amount per block - as comment)

frame/mosaic/src/plots.rs
```diff
@@ -0,0 +1,37 @@
+mod decay;
+
+fn main() {
+	#[cfg(feature = "visualization")]
+	{
+		plot_linear_decay(5);
+	}
+}
+
+#[cfg(feature = "visualization")]
```
dzmitry-lahoda, line 10: +++

frame/mosaic/src/decay.rs
dzmitry-lahoda: speed

frame/mosaic/src/lib.rs
```diff
@@ -0,0 +1,676 @@
...
+
+	impl<T: Config> SubAccount<T> {
+		pub fn to_id(&self) -> impl Encode {
+			let prefix = match self.transaction_type {
+				TransactionType::Incoming => b"incoming________",
+				TransactionType::Outgoing => b"outgoing________",
+			};
+			[prefix.to_vec(), self.account_id.encode()]
+		}
+		pub fn outgoing(account_id: AccountIdOf<T>) -> Self {
```
dzmitry-lahoda, line 108: new_...

frame/mosaic/src/lib.rs
```diff
@@ -0,0 +1,676 @@
...
+	}
+
+	/// User incoming/outgoing accounts, that hold the funds for transactions to happen.
+	pub struct SubAccount<T: Config> {
+		transaction_type: TransactionType,
+		account_id: AccountIdOf<T>,
+	}
+
+	impl<T: Config> SubAccount<T> {
+		pub fn to_id(&self) -> impl Encode {
```
dzmitry-lahoda, line 101: consider impl AccountIdConversion

frame/mosaic/src/lib.rs
```diff
@@ -0,0 +1,676 @@
...
+		}
+
+		/// Sets supported networks and maximum transaction sizes accepted by the relayer.
+		#[pallet::weight(10_000)]
+		pub fn set_network(
+			origin: OriginFor<T>,
+			network_id: NetworkIdOf<T>,
+			network_info: NetworkInfo<BalanceOf<T>>,
+		) -> DispatchResultWithPostInfo {
+			ensure_relayer::<T>(origin)?;
```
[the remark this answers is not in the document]
hussein-aitlahcen: Penalize here refer to the fact that we don't allow the relayer to inifintely mint within a period of time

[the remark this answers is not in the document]
hussein-aitlahcen: The penalty is decaying each block

frame/mosaic/src/lib.rs
```diff
@@ -0,0 +1,676 @@
...
+		/// Sets the relayer budget for _incoming_ transactions for specific assets. Does not reset
+		/// the current `penalty`.
+		///
+		/// # Restrictions
+		/// - Only callable by root
+		#[pallet::weight(10_000)]
+		#[transactional]
+		pub fn set_budget(
+			origin: OriginFor<T>,
+			asset_id: AssetIdOf<T>,
```
dzmitry-lahoda, line 295: could split assets which are XCMP assets and not? for XCMP assets we do not need transfers - must forbid to mint them.
what do you think?

dzmitry-lahoda: this will ensure we do not hack other parachain

frame/mosaic/src/lib.rs
```diff
@@ -0,0 +1,676 @@
...
+				keep_alive,
+			)?;
+			let now = <frame_system::Pallet<T>>::block_number();
+			let lock_until = now.safe_add(&TimeLockPeriod::<T>::get())?;
+
+			OutgoingTransactions::<T>::try_mutate(
+				caller.clone(),
+				asset_id,
+				|prev| -> Result<(), DispatchError> {
+					match prev.as_mut() {
```
dzmitry-lahoda, line 365: *prev is not always previous. variable names seems bad

frame/mosaic/src/lib.rs
```diff
@@ -0,0 +1,676 @@
...
+						// the storage item.
+						if amount == balance {
+							*maybe_tx = None
+						} else {
+							let new_balance =
+								balance.checked_sub(&amount).ok_or(Error::<T>::AmountMismatch)?;
+							*maybe_tx = Some((new_balance, lock_period));
+						}
+
+						Self::deposit_event(Event::<T>::TransferAccepted {
```
dzmitry-lahoda, line 426: accepted transfer can be partially accepted transfer

frame/mosaic/src/lib.rs
```diff
@@ -0,0 +1,676 @@
...
+				let new_penalty = penalty_decayer
+					.checked_decay(penalty, last_mint_block, current_block)
+					.unwrap_or_else(Zero::zero);
+
+				let penalised_budget = budget.saturating_sub(new_penalty);
+
+				// Check if the relayer has a sufficient budget to mint the requested amount.
+				ensure!(amount <= penalised_budget, Error::<T>::InsufficientBudget);
+
+				T::Assets::mint_into(
```
dzmitry-lahoda, line 504: same question, are we sure relayer prevents double mint?

frame/mosaic/src/lib.rs
```diff
@@ -0,0 +1,676 @@
...
+			let current_block = <frame_system::Pallet<T>>::block_number();
+			Relayer::<T>::get().update(current_block).account_id().cloned()
+		}
+	}
+
+	/// Convenience identifiers emitted by the pallet for relayer bookkeeping.
+	pub type Id = H256;
+
+	/// Raw ethereum addresses.
+	pub type EthereumAddress = [u8; 20];
```
dzmitry-lahoda, line 656: we have pallet who has that may be. could share typedef

frame/mosaic/src/plots.rs
dzmitry-lahoda: consider moving thist to test

frame/mosaic/README.md
```diff
@@ -0,0 +1,10 @@
+# Mosaic
```
dzmitry-lahoda, line 1: please add referene to mosic docs so we can check if properly handles txs - avoids double mints and double burns.

frame/mosaic/src/decay.rs
hussein-aitlahcen: It mean that it cannot compute the recovery period, A.K.A. that the recovery < 1 block or > max_size(block)

dzmitry-lahoda: so it is kind of no recovery period if <1 , so it is 0 recovery period?

dzmitry-lahoda: i am clear about how decay works and per semantics. so it is not clear from API description.
if amount > than possible, than can mint so much so no possible recovery

dzmitry-lahoda:
it seems like this pallet should have kind of hard stop button:) may be by disabling all relayers. so it would be awesome to have as simple API as possible to call.

hussein-aitlahcen:
> it seems like this pallet should have kind of hard stop button:) may be by disabling all relayers. so it would be awesome to have as simple API as possible to call.

This can be done by root, using the `set_relayer` to a new account_id

frame/mosaic/src/decay.rs
hussein-aitlahcen: We have a more high level tests that does that in the tests of mosaic, `penalty = 0, budget = x - penatly, penatly = y, budget' = budget - penalty, penatly = decay(penalty, full_recovery) = 0, budget'' = budget -  penalty = budget`

frame/mosaic/src/decay.rs
hussein-aitlahcen: CU-1wana2u

frame/mosaic/src/lib.rs
hussein-aitlahcen: The trait has many methods we don't want to use, would be overkill to implement them and make the feature more obscure

frame/mosaic/src/lib.rs
hussein-aitlahcen: I've reflected this in the events.

frame/mosaic/src/lib.rs
```diff
@@ -0,0 +1,676 @@
...
+							amount,
+						)?;
+
+						// No remaing funds need to be transferred for this asset, so we can delete
+						// the storage item.
+						if amount == balance {
+							*maybe_tx = None
+						} else {
+							let new_balance =
+								balance.checked_sub(&amount).ok_or(Error::<T>::AmountMismatch)?;
```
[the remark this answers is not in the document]
hussein-aitlahcen: This is up to the relayer to ensure this

frame/mosaic/src/lib.rs
hussein-aitlahcen: this is a relayer concern

frame/mosaic/README.md
hussein-aitlahcen: CU-1wanc6n

frame/mosaic/src/lib.rs
hussein-aitlahcen: CU-1wanc6n

frame/mosaic/src/lib.rs
hussein-aitlahcen: This is done by root

hussein-aitlahcen (approved):

dzmitry-lahoda (approved):

frame/mosaic/src/decay.rs
dzmitry-lahoda: so if you pass last which is more than current, it would be better to return error to prove error in program, not hid and making diff = 0 always:(

hussein-aitlahcen: CU-1waqf8d

PR #548, opened 2022-01-24
Added 'std' features to Balances

PoisonPhang:
Small fix to allow for the crowdloan-rewards pallet to pass tests independently from the rest of the project.
Previously, running `cargo test` from inside the `./frame/crowdloan-rewards` directory would fail and say that Balances needs the `std` feature. Running `cargo test` from the project root would pass because the import of Balances in the main `Cargo.toml` file includes the `std` feature.

### DIFF
--- a/frame/crowdloan-rewards/Cargo.toml
+++ b/frame/crowdloan-rewards/Cargo.toml
@@ -23,7 +23,7 @@ libsecp256k1 = { version = "0.7.0", default-features = false, features = [
     "static-context",
 ] }
 hex-literal = "0.3"
-balances = { package = "pallet-balances", git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.13", default-features = false }
+balances = { package = "pallet-balances", git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.13", default-features = false, features = [ "std" ]}
 
 [dependencies]
 ### Benchmarking
@@ -32,7 +32,7 @@ libsecp256k1 = { version = "0.7.0", default-features = false, optional = true, f
     "hmac",
     "static-context",
 ] }
-balances = { package = "pallet-balances", git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.13", default-features = false, optional = true }
+balances = { package = "pallet-balances", git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.13", default-features = false, features = [ "std" ], optional = true }
 frame-benchmarking = { default-features = false, optional = true, git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.13" }
 sp-application-crypto = { default-features = false, optional = true, git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.13" }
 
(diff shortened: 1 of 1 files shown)


hussein-aitlahcen (approved):

frame/crowdloan-rewards/Cargo.toml
```diff
@@ -32,7 +32,7 @@ libsecp256k1 = { version = "0.7.0", default-features = false, optional = true, f
     "hmac",
     "static-context",
 ] }
-balances = { package = "pallet-balances", git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.13", default-features = false, optional = true }
+balances = { package = "pallet-balances", git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.13", default-features = false, features = [ "std" ], optional = true }
```
dzmitry-lahoda, line 35: remove this line i guess. we do not make direct deps on this in other pallets. use NativeCurrency dep

frame/crowdloan-rewards/Cargo.toml
```diff
@@ -23,7 +23,7 @@ libsecp256k1 = { version = "0.7.0", default-features = false, features = [
     "static-context",
 ] }
 hex-literal = "0.3"
-balances = { package = "pallet-balances", git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.13", default-features = false }
+balances = { package = "pallet-balances", git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.13", default-features = false, features = [ "std" ]}
```
dzmitry-lahoda, line 26: can remove std, like in other traits

seunlanlege, line 26: this will break wasm builds

PoisonPhang: I seem to be able to build wasm successfully while including the std feature in Balances. Is there something I'm missing?
Running `cargo build -p picasso-runtime` in both the project root and inside `./frame/crowdload-rewards` builds successfully.

frame/crowdloan-rewards/Cargo.toml
PoisonPhang: Are you suggesting we remove balances entirely?

seunlanlege: ahh its opptional

seunlanlege: if its used only by benchmarks, best to have it only in dev-dependencies

PoisonPhang: This should be resolved in ed0e6f51bb61fdc2388878ec8634796d98432221. Let me know if it's not. Everything seems okay locally and CI is going well for now.

seunlanlege (approved):

frame/crowdloan-rewards/Cargo.toml
hussein-aitlahcen: Try enabling benchmarking using `--features runtime-benchmarks`

dzmitry-lahoda (approved):

PR #550, opened 2022-01-25
Implement get_token_index() and get_price() function in DEXs.

vivekvpandya:


frame/composable-traits/src/dex.rs
```diff
@@ -22,6 +22,35 @@ pub trait CurveAmm {
 	/// Current number of pools (also ID for the next created pool)
 	fn pool_count() -> Self::PoolId;
 
+	/// Get pool token index for given asset_id.
+	fn get_token_index(
```
dzmitry-lahoda, line 26: i am concerned that we have impl for pair and api for multi pool.

did we talked about to have dexes only for pairs? cc @hussein-aitlahcen

If yes, than we can have all api for pairs. but later can have impl for multi (router, balancer, curve multi) are multi dexes.

frame/composable-traits/src/dex.rs
```diff
@@ -22,6 +22,35 @@ pub trait CurveAmm {
...
+
+	/// Get price for given units of given asset.
+	fn get_price(
+		pool_id: Self::PoolId,
+		asset_id: Self::AssetId,
+		balance: Self::Balance,
+	) -> Result<Self::Balance, DispatchError>;
+
+	/// Buy given units of given asset from the pool.
+	fn buy(
```
dzmitry-lahoda, line 39: please add comment what is different of buy vs sell

frame/curve-amm/src/lib.rs
```diff
@@ -281,6 +279,109 @@ pub mod pallet {
...
+			pool_id: T::PoolId,
+			asset_id: Self::AssetId,
+		) -> Result<Self::PoolTokenIndex, DispatchError> {
+			let assets_pair = PoolAssets::<T>::get(pool_id).ok_or(Error::<T>::PoolNotFound)?;
+			if assets_pair.base == asset_id {
+				Ok(0.into())
+			} else if assets_pair.quote == asset_id {
+				Ok(1.into())
+			} else {
+				Err(Error::<T>::AssetNotFound.into())
```
dzmitry-lahoda, line 292: i am bothered with this - we have pair, but have runtime code to handle non pairs..

frame/curve-amm/src/lib.rs
```diff
@@ -281,6 +279,109 @@ pub mod pallet {
...
+			let assets_pair = PoolAssets::<T>::get(pool_id).ok_or(Error::<T>::PoolNotFound)?;
+			let assets = assets_pair.as_slice();
+			let mut balances = Vec::new();
+			for asset_id in assets {
+				balances.push(PoolAssetBalance::<T>::get(pool_id, asset_id));
+			}
+			let n_coins = assets.len();
+
+			let i = token_index.into() as usize;
+			let j = if i == 0 { 1 } else { 0 };
```
dzmitry-lahoda, line 314: if i == 2?

frame/curve-amm/src/lib.rs
```diff
@@ -107,7 +107,7 @@ pub mod pallet {
 			+ Debug
 			+ CheckedAdd
 			+ One;
-		type PoolTokenIndex: Copy + Debug + Eq + Into<u32>;
+		type PoolTokenIndex: Copy + Debug + Eq + Into<u32> + From<u32>;
```
dzmitry-lahoda, line 110: i would prefer u8 or even u4

frame/curve-amm/src/tests.rs
```diff
@@ -152,6 +171,26 @@ fn get_y_j_greater_than_n() {
 	assert_eq!(result, None);
 }
 
+#[test]
+fn test_get_price() {
+	new_test_ext().execute_with(|| {
+		let assets = vec![MockCurrencyId::USDC, MockCurrencyId::USDT];
+		let amp_coeff = FixedU128::saturating_from_rational(1_000_i128, 1_i128);
```
dzmitry-lahoda, line 178: can do from integer, no need second parameter than

frame/curve-amm/src/tests.rs
```diff
@@ -152,6 +171,26 @@ fn get_y_j_greater_than_n() {
 	assert_eq!(result, None);
 }
 
+#[test]
+fn test_get_price() {
+	new_test_ext().execute_with(|| {
+		let assets = vec![MockCurrencyId::USDC, MockCurrencyId::USDT];
```
dzmitry-lahoda, line 177: can remove Mock form Currency, than easy to work with code in real runtme and in tests

frame/curve-amm/src/tests.rs
```diff
@@ -152,6 +171,26 @@ fn get_y_j_greater_than_n() {
...
+		let amp_coeff = FixedU128::saturating_from_rational(1_000_i128, 1_i128);
+		let fee = Permill::zero();
+		let admin_fee = Permill::zero();
+		let amounts = vec![999999999u128, 999999999u128];
```
dzmitry-lahoda, line 181: seems will not pass clippy

frame/curve-amm/src/lib.rs
andor0: it can rewrite so:
```
let j = i ^ 1;
```
(https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e4875ba8f5e80ac4cc7c109735b5c174)

frame/curve-amm/src/lib.rs
```diff
@@ -281,6 +279,109 @@ pub mod pallet {
...
+
+		fn buy(
+			who: &Self::AccountId,
+			pool_id: Self::PoolId,
+			asset_id: Self::AssetId,
+			balance: Self::Balance,
+		) -> Result<Self::Balance, DispatchError> {
+			let token_index = Self::get_token_index(pool_id, asset_id)?;
+			let i: u32 = token_index.into();
+			let j: u32 = if i == 0 { 1 } else { 0 };
```
andor0, line 361: This is often used. Can create a function for this?

frame/composable-traits/src/dex.rs
KaiserKarel: An option would be to split up the trait into multiple ones, as is done with Mutate, Transfer etc.

frame/curve-amm/src/lib.rs
hussein-aitlahcen: See my previous comment

frame/curve-amm/src/lib.rs
hussein-aitlahcen: This should probably be removed, see previous comment.

hussein-aitlahcen (changes requested):

frame/curve-amm/src/tests.rs
vivekvpandya: both CI and local clippy passes.

frame/curve-amm/src/tests.rs
vivekvpandya: I will do it in separate PR.

frame/composable-traits/src/dex.rs
```diff
@@ -12,16 +12,37 @@ pub trait CurveAmm {
...
+		asset_id: Self::AssetId,
+		amount: Self::Balance,
+	) -> Result<Self::Balance, DispatchError>;
+
+	/// Sell given units of given asset to the pool.
+	fn sell(
+		who: &Self::AccountId,
+		pool_id: Self::PoolId,
+		asset_id: Self::AssetId,
+		amount: Self::Balance,
```
dzmitry-lahoda, line 43: please comment what amount means. still not clear from comments what is difference of sell vs buy.

dzmitry-lahoda:
please add at least TODO comment to make test which shows off buy vs sell difference

frame/composable-traits/src/dex.rs
```diff
@@ -12,16 +12,37 @@ pub trait CurveAmm {
...
 
+	/// Get pure exchange value for given units of given asset. (Note this does not include fees.)
+	fn get_exchange_value(
+		pool_id: Self::PoolId,
+		asset_id: Self::AssetId,
+		amount: Self::Balance,
+	) -> Result<Self::Balance, DispatchError>;
+
+	/// Buy given units of given asset from the pool.
+	/// In buy user does not know how much assets he/she has to exchange to get desired amount.
```
dzmitry-lahoda, line 30: /// `who` desires to buy `asset_id` of amount from `pool_id` withing given slippage `configuration`

so it lacks configuration parameter as for me.

dzmitry-lahoda:
you may consider using MathBalance trait as constant and also may add sell test to tests

frame/composable-traits/src/dex.rs
vivekvpandya:
```
fn buy(
			who: &Self::AccountId,
			pool_id: Self::PoolId,
			asset_id: Self::AssetId,
			amount: Self::Balance,
			let assets_pair = PoolAssets::<T>::get(pool_id).ok_or(Error::<T>::PoolNotFound)?;
			let mut exchange_asset = assets_pair.quote;
			if asset_id == assets_pair.quote {
				exchange_asset = assets_pair.base;
			} else if asset_id != assets_pair.base {
				return Err(Error::<T>::AssetNotFound.into())
			}
			// Since when buying asset user can't executed exchange as he don't know how much
			// amount of token he has to trade-in to get expected buy tokens.
			// So we compute price assuming user wants to sell instead of buy.
			// And then do exchange computed amount with token indices flipped.
			let dx = Self::get_exchange_value(pool_id, asset_id, amount)?;
			Self::update_balance(who, &pool_id, &exchange_asset, dx, &asset_id, amount)?;
			Ok(amount)
		}
```
The implementation of `buy` is such that it asks `dx` amount from the user which is required to buy `amount` tokens so its not required to have slippage.

dzmitry-lahoda: nice documentation. could we have that on pallet call level or other means in /// code?

hussein-aitlahcen:
I'm currently refactoring this pallet along the interface

hussein-aitlahcen:
Closing this for a clean restart.

PR #552, opened 2022-01-25
[CU-23t6rux] Add decimals for foreign assets in assets-registry

andor0:


ComposableFiB0t:
Task linked: [CU-23t6rux Add an opportunity to set decimals for foreign assets in assets-registry](https://app.clickup.com/t/23t6rux)

frame/assets-registry/src/lib.rs
```diff
@@ -73,6 +73,11 @@ pub mod pallet {
 		ForeignAdminApproved,
 	}
 
+	#[derive(Debug, Clone, Copy, PartialEq, Eq, Encode, Decode, TypeInfo)]
+	pub struct Metadata {
```
dzmitry-lahoda, line 77: can we renamte it to `RemoteMetadata`? Because locally we have 12 for all currencies. Can you comment on what decimals are used on Hydra/Akala/Moonriver? Just to make sure these are different from ours 12?

frame/assets-registry/src/lib.rs
```diff
@@ -226,6 +239,24 @@ pub mod pallet {
...
 			Ok(().into())
 		}
+
+		#[pallet::weight(10_000)]
+		pub fn set_metadata(
+			origin: OriginFor<T>,
+			local_asset_id: T::LocalAssetId,
+			metadata: Metadata,
+		) -> DispatchResultWithPostInfo {
+			let _who = ensure_signed(origin.clone())?;
```
dzmitry-lahoda, line 249: let _ =  ? :)

hussein-aitlahcen (dismissed):
Nothing more than Dmitry comments.

frame/assets-registry/src/lib.rs
andor0: fixed

frame/assets-registry/src/lib.rs
andor0: Renamed to `Foreign Metadata`, because we already use `ForeignAdmin`, `ForeignAssetId` and etc.

Moonriver uses 18 decimals of precision (https://github.com/PureStake/moonbeam/blob/master/runtime/moonriver/src/lib.rs#L117).
Acala uses 12 decimals of precision (https://github.com/AcalaNetwork/Acala/blob/master/modules/evm/src/lib.rs#L1117).
Acala, HydraDX, Moonriver support assets with various precision in their assets-registry.

hussein-aitlahcen (approved):

dzmitry-lahoda (approved):

PR #493, opened 2022-01-14
Ben/assets rpc

benluelo:

PR #560, opened 2022-01-28
[1u5ykup] Runtime integration tests parallelization & Mocha v.9

Dom-Roth:
Greetings, here we have the next update for the runtime integration tests.
~~It's built upon #529 and Waits for #493, because I'll convert @benluelo 's tests to the new structure as soon as they're done.~~

This update contains the following:

1. Parallelization Integration
2. Update to Mocha v.9
3. Making tests keep working on a running chain.
4. Updates checks for existing tests.
5. Added configuration file for tests.

~~I'll add more information, about the new structure and other details, here later.~~
# More Information
## 1. & 2. 
The new structure enabled us to upgrade mocha to version 9, and make our tests run in parallel.
I had to get rid of our test.ts file which did call our tests, as well as the run function within the classes of the tests.
Mocha now searches recursively through the folders for the tests, and the describe() block needs to be on the root level of each test file.

When multiple extrinsic requests are sent at the same time, it can be that 2 are sent with the same nonce, even though `nonce=-1` is used.
To avoid this I recursively retry the request after a few seconds, if this specific error occurs.
```typescript
.catch(async function (e) {
  if (e.message.contains("1014: Priority is too low:")) {
    // This happens when we send 2 transaction from the same wallet, at the same time.
    // We solve it by waiting 2 seconds and retrying it.
    await sleep(2000);
    const {data: [result],} = await sendAndWaitFor(api, sender, filter, call, intendedToFail).catch(function (exc) {
      reject(exc);
      return {data:[exc]};
    });
    expect(result).to.not.be.an('Error');
    resolve(result);
  }
  else
    reject(Error(e.stack));
});
```
## 3. 
The tests can now be called multiple times without resetting the chain.

## 4.
I've added a couple more result checks to different test.

## 5.
Each test now has a configuration file, to for example deactivate specific tests, or a whole block.


Dom-Roth:
Parallelization works fine now.
Fixed the generators not working, and added a start balance generator.

Dom-Roth:
Screenshot 2022-02-01 at 11-39-24 Mochawesome Report

hussein-aitlahcen (approved):

benluelo (approved):

PR #572, opened 2022-02-01
Inverted Validate trait

PoisonPhang:
These changes invert to `Validate` trait so that instead of `impl Validate<ValidationType> for ExtrinsicInputType` we do `impl Validate<ExtrinsicInputType, ValidationType> for ValidationType.
This will allow us to create validations for types we don't have direct ownership of.

I've also gone ahead and updated existing implementations of `Validate` so that this branch won't fail. @dzmitry-lahoda, please make sure I didn't miss anything in converting these. Thanks.

### DIFF
--- a/frame/composable-support/src/validation.rs
+++ b/frame/composable-support/src/validation.rs
@@ -11,10 +11,16 @@ pub struct Validated<T, U> {
 
 impl<T, U> Validated<T, U>
 where
-	Validated<T, U>: Validate<U>,
+	Validated<T, U>: Validate<T, U>,
+	U: Validate<T, U>,
 {
 	pub fn new(value: T, _validator_tag: U) -> Result<Self, &'static str> {
-		Validate::<U>::validate(Self { value, _marker: PhantomData })
+		// Validate::<U>::validate(Self { value, _marker: PhantomData })
+		// Ok(Self { value: Validate::<T, U>::validate(value), _marker: PhantomData })
+		match <U as Validate<T, U>>::validate(value) {
+			Ok(value) => Ok(Self { value, _marker: PhantomData }),
+			Err(e) => Err(e),
+		}
 	}
 }
 
@@ -22,9 +28,9 @@ pub trait ValidateDispatch<U>: Sized {
 	fn validate(self) -> Result<Self, DispatchError>;
 }
 
-pub trait Validate<U>: Sized {
+pub trait Validate<T, U> {
 	// use string here because in serde layer there is not dispatch
-	fn validate(self) -> Result<Self, &'static str>;
+	fn validate(extrinsic_input: T) -> Result<T, &'static str>;
 }
 
 #[derive(Debug, Eq, PartialEq, Default)]
@@ -33,25 +39,29 @@ pub struct Valid;
 #[derive(Debug, Eq, PartialEq, Default)]
 pub struct Invalid;
 
-impl<T> Validate<Invalid> for T {
+impl<T> Validate<T, Invalid> for Invalid {
 	#[inline(always)]
-	fn validate(self) -> Result<Self, &'static str> {
+	fn validate(_extinsic_input: T) -> Result<T, &'static str> {
 		Err("not valid")
 	}
 }
 
-impl<T> Validate<Valid> for T {
+impl<T> Validate<T, Valid> for Valid {
 	#[inline(always)]
-	fn validate(self) -> Result<Self, &'static str> {
-		Ok(self)
+	fn validate(extrinsic_input: T) -> Result<T, &'static str> {
+		Ok(extrinsic_input)
 	}
 }
 
-impl<T: Validate<U> + Validate<V>, U, V> Validate<(U, V)> for T {
+impl<T, U, V> Validate<T, (U, V)> for U
+where
+	U: Validate<T, U>,
+	V: Validate<T, V>,
+{
 	#[inline(always)]
-	fn validate(self) -> Result<Self, &'static str> {
-		let value = Validate::<U>::validate(self)?;
-		let value = Validate::<V>::validate(value)?;
+	fn validate(extrinsic_input: T) -> Result<T, &'static str> {
+		let value = <U as Validate<T, U>>::validate(extrinsic_input)?;
+		let value = <V as Validate<T, V>>::validate(value)?;
 		Ok(value)
 	}
 }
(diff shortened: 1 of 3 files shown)


frame/composable-support/src/validation.rs
```diff
@@ -11,20 +67,26 @@ pub struct Validated<T, U> {
 
 impl<T, U> Validated<T, U>
 where
-	Validated<T, U>: Validate<U>,
+	Validated<T, U>: Validate<T, U>,
+	U: Validate<T, U>,
 {
 	pub fn new(value: T, _validator_tag: U) -> Result<Self, &'static str> {
-		Validate::<U>::validate(Self { value, _marker: PhantomData })
+		// Validate::<U>::validate(Self { value, _marker: PhantomData })
```
dzmitry-lahoda, line 74: please remove commented items

frame/composable-support/src/validation.rs
```diff
@@ -11,20 +67,26 @@ pub struct Validated<T, U> {
...
 
 pub trait ValidateDispatch<U>: Sized {
 	fn validate(self) -> Result<Self, DispatchError>;
 }
 
-pub trait Validate<U>: Sized {
+pub trait Validate<T, U> {
 	// use string here because in serde layer there is not dispatch
-	fn validate(self) -> Result<Self, &'static str>;
+	fn validate(extrinsic_input: T) -> Result<T, &'static str>;
```
dzmitry-lahoda, line 89: may be in can work work for Validate<T>?

frame/composable-support/src/validation.rs
```diff
@@ -33,64 +95,77 @@ pub struct Valid;
 #[derive(Debug, Eq, PartialEq, Default)]
 pub struct Invalid;
 
-impl<T> Validate<Invalid> for T {
+impl<T> Validate<T, Invalid> for Invalid {
```
dzmitry-lahoda, line 98: would this work `impl<T> Validate<T> for Invalid`?

frame/composable-support/src/validation.rs
```diff
@@ -33,64 +95,77 @@ pub struct Valid;
...
+	Z: Validate<T, Z>,
 {
 	#[inline(always)]
-	fn validate(self) -> Result<Self, &'static str> {
-		let value = Validate::<U>::validate(self)?;
-		let value = Validate::<V>::validate(value)?;
-		let value = Validate::<W>::validate(value)?;
-		let value = Validate::<Z>::validate(value)?;
+	fn validate(extrinsic_input: T) -> Result<T, &'static str> {
+		let value = <U as Validate<T, U>>::validate(extrinsic_input)?;
```
dzmitry-lahoda, line 152: `U::validate(extrinsic_input)` may work

frame/composable-support/src/validation.rs
```diff
@@ -113,11 +188,22 @@ pub(crate) mod private {
 	}
 }
 
-impl<T: codec::Encode + codec::Decode + Validate<U>, U> codec::WrapperTypeEncode
+impl<T: codec::Encode + codec::Decode, U: Validate<T, U>> codec::WrapperTypeEncode
 	for Validated<T, U>
 {
 }
 
+// impl<T: codec::Encode + codec::Decode, U: Validate<T, U>> codec::WrapperTypeDecode
```
dzmitry-lahoda, line 196: A marker trait for types that can be created solely from other decodable types.

The decoding of such type is assumed to be the same as the wrapped type.

If we comment that out, we may fail to have goo TS scale API

frame/composable-support/src/validation.rs
```diff
@@ -132,6 +218,9 @@ mod test {
 	type CheckARangeTag = (ValidARange, Valid);
 	type CheckBRangeTag = (ValidBRange, Valid);
 	type CheckABRangeTag = (ValidARange, (ValidBRange, Valid));
+	// type ManyValidatorsTagsNested = (ValidARange, (ValidBRange, (Invalid, Valid)));
```
dzmitry-lahoda, line 221: @hussein-aitlahcen suggested that nested is option to choose. will it still work?

frame/composable-support/src/validation.rs
```diff
@@ -144,43 +233,62 @@ mod test {
...
-				Ok(self)
+				Ok(extrinsic_input)
 			}
 		}
 	}
 
-	#[test]
-	fn nested_validator() {
-		let valid = X { a: 10, b: 0xCAFEBABE };
+	// #[test]
```
dzmitry-lahoda, line 256: is possible to restore nested?

frame/composable-traits/src/lending/math.rs
```diff
@@ -386,7 +390,7 @@ impl DoubleExponentModel {
 	pub fn new(coefficients: [u8; 16]) -> Option<Self> {
 		let sum_of_coefficients = coefficients.iter().fold(0_u16, |acc, &c| acc + c as u16);
 		if sum_of_coefficients == EXPECTED_COEFFICIENTS_SUM {
-			return Some(DoubleExponentModel { coefficients })
+			return Some(DoubleExponentModel { coefficients });
```
dzmitry-lahoda, line 393: cargo +nightly fmt

frame/composable-support/src/validation.rs
hussein-aitlahcen: I think this will still work as long as we keep the impl for (T, U) & Valid, such that we can do (T, Valid), (T, (U, Valid)), (T, (U, (V, Valid)))... So perhaps we should uncomment?

frame/composable-support/src/validation.rs
```diff
@@ -144,43 +233,62 @@ mod test {
 		b: u32,
 	}
 
-	impl Validate<ValidARange> for X {
-		fn validate(self) -> Result<X, &'static str> {
-			if self.a > 10 {
+	impl Validate<X, ValidARange> for ValidARange {
+		fn validate(extrinsic_input: X) -> Result<X, &'static str> {
```
hussein-aitlahcen, line 237: BTW it's an abstraction to validate a structure, not specifically an extrinsic input, since it's fully generic, the name is meaningless isn't it :P?

frame/composable-support/src/validation.rs
hussein-aitlahcen: :+1:

hussein-aitlahcen (commented):
Well done, nitpick comments :). Also, did you tried to implement it for primitive types?

frame/composable-support/src/validation.rs
PoisonPhang: This wasn't removed from the existing code. This is a leftover from me trying to fix an issue that I resolved in a different way. This has been removed here 005472a15e3a48707e9a81b975df0cf6f9ae0a8d

frame/composable-support/src/validation.rs
PoisonPhang: Nested restored here 7ff004b29650795dc5fbb23fb96035061d07ee8c

frame/composable-traits/src/lending/math.rs
PoisonPhang: Resolved here 6e777b19f7be5adbf3c10d49111d528672e85510

frame/composable-support/src/validation.rs
PoisonPhang: No, we need to supply the `Validate` trait with two generic arguments. Otherwise, we couldn't support multiple flat/nested validations

frame/composable-support/src/validation.rs
PoisonPhang: That's a good point. I've cleaned up the name here 120020d33c59a904453d831923a20b7c1064e5c3

frame/composable-support/src/validation.rs
PoisonPhang: Good call, this will clean up the calls quite a bit. Fixed this here 0c38ae60ceefc9a143e3b85cffc314c94ec7c58f

dzmitry-lahoda (approved):

hussein-aitlahcen (approved):

PR #573, opened 2022-02-01
mosaic benchmarks

cor:
Adds benchmarks to mosaic and uses the generated weights for dali


cor:
Wanted to check the draft PR behavior

seunlanlege:
you need to fix conflicts for actions to run

dzmitry-lahoda (approved):

seunlanlege:
still has conflicts

frame/mosaic/src/weights.rs
```diff
@@ -0,0 +1,68 @@
...
+	fn timelocked_mint() -> Weight;
+	fn set_timelock_duration() -> Weight;
+	fn rescind_timelocked_mint() -> Weight;
+	fn claim_to() -> Weight;
+}
+
+// For backwards compatibility and tests
+impl WeightInfo for () {
+    fn set_relayer() -> Weight {
+        0 as Weight
```
dzmitry-lahoda, line 26: if would prefer even in tests that is not 0

runtime/dali/src/weights/mosaic.rs
```diff
@@ -0,0 +1,110 @@
+
+//! Autogenerated weights for `mosaic`
+//!
+//! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
```
dzmitry-lahoda, line 4: how it was generated? machine generated that must be same spec as our runtime running i guess

dzmitry-lahoda (dismissed):

frame/mosaic/src/weights.rs
cor: Not really sure what you mean by this. These weights are just there for backwards compatibility and tests? What would you like to change it to?

cor:
I've just rebased `cor/mosaic-benchmarks` on top of `main` again in order to trigger the CI

frame/mosaic/src/weights.rs
KaiserKarel: I think Dzmitry would prefer them to be larger, arbitrary numbers. In case they're ever insecurely used, the damage of the vulnerability would be limited.

runtime/dali/src/weights/mosaic.rs
hussein-aitlahcen: this will be overwritten by the benchmark ran in the CI I believe, it's more like a placeholder to make the CI happy

frame/mosaic/src/weights.rs
hussein-aitlahcen: :+1:

dzmitry-lahoda: so yeah, if we spammed, we still earn. 

also weights are used for add/remove. when store happens on behalf of users, than remove (after some conditions are met), is removed. than it is easy to ensure our tests are properly set to non-zero weights. 

also they ensure our tests are more like e2e tests. which I find useful in simulator.

runtime/dali/src/weights/mosaic.rs
dzmitry-lahoda: i see. so can put here any weight even like 100_000 for each value?  so that we have not need to run it locally

dzmitry-lahoda (approved):

frame/mosaic/src/weights.rs
cor: Ah I see what you mean now

hussein-aitlahcen (approved):

KaiserKarel (approved):

seunlanlege (approved):

PR #579, opened 2022-02-02
bonded-finance benchmarking

hussein-aitlahcen:
- refactor benchmarking
- add `keep_alive` parameter to bonded finance extrinsics
- add weights to dali/picasso


frame/bonded-finance/src/weights.rs
```diff
@@ -0,0 +1,24 @@
...
+
+pub trait WeightInfo {
+	fn offer() -> Weight;
+	fn bond() -> Weight;
+	fn cancel() -> Weight;
+}
+
+impl WeightInfo for () {
+	fn offer() -> Weight {
+    0 as Weight
```
dzmitry-lahoda, line 16: i usually put here something. so that any test will consider that - so our tests become more real:)

runtime/dali/src/weights/bonded_finance.rs
```diff
@@ -0,0 +1,57 @@
...
+// --execution=wasm
+// --wasm-execution=compiled
+// --pallet=bonded-finance
+// --extrinsic=*
+// --steps=10
+// --repeat=5
+// --raw
+// --output=./runtime/dali/src/weights
+
+#![cfg_attr(rustfmt, rustfmt_skip)]
```
dzmitry-lahoda, line 21: is this from your laptop or ci? just askin - cause only if CI has same spec as runtme machines than these will be good fit.

frame/bonded-finance/src/lib.rs
```diff
@@ -204,10 +212,14 @@ pub mod pallet {
 		/// appropriate funds.
 		///
 		/// Emits a `NewOffer`.
-		#[pallet::weight(10_000)]
-		pub fn offer(origin: OriginFor<T>, offer: BondOfferOf<T>) -> DispatchResult {
+		#[pallet::weight(T::WeightInfo::offer())]
```
dzmitry-lahoda, line 215: if API generator generates doc too, than better to document what is keep_alive :)

dzmitry-lahoda (dismissed):

runtime/dali/src/weights/bonded_finance.rs
hussein-aitlahcen: CI will generate this file when adding the `needs-benchmarks` label :)

frame/bonded-finance/src/weights.rs
hussein-aitlahcen: :+1:

KaiserKarel (dismissed):

dzmitry-lahoda (approved):

PR #584, opened 2022-02-03
Karel/currency ranges

KaiserKarel:
Overhauls currency-factory, allowing for dynamic ranges. Associated with that, it cleans up loads of mocks, since bounds are added to CurrencyId.


frame/currency-factory/src/ranges.rs
```diff
@@ -0,0 +1,191 @@
...
+pub use composable_traits::currency::RangeId;
+use frame_support::{traits::Get, BoundedVec};
+use scale_info::TypeInfo;
+use sp_runtime::{
+	traits::{CheckedAdd, Saturating},
+	ArithmeticError, DispatchError,
+};
+pub struct MaxRanges;
+
+impl Get<u32> for MaxRanges {
```
dzmitry-lahoda, line 11: very dynamic and extensible.

dzmitry-lahoda: ok. so i always think that as of risk.

dzmitry-lahoda (approved):

hussein-aitlahcen (approved):
Perhaps we could have something like `CurrencyAdapter` for pallets that are meant to create tokens from a well known range? Such that we don't have to pass the range when creating?

frame/curve-amm/src/lib.rs
```diff
@@ -788,7 +788,7 @@ pub mod pallet {
 						// We expect that PoolInfos have sequential keys.
 						// No PoolInfo can have key greater or equal to PoolCount
 						ensure!(maybe_pool_info.is_none(), Error::<T>::InconsistentStorage);
-						let lp_asset = T::CurrencyFactory::create()?;
+						let lp_asset = T::CurrencyFactory::reserve_lp_token_id()?;
```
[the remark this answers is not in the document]
KaiserKarel: I wanted to make the distinction that only an id is reserved, not actually put into use.

frame/currency-factory/src/ranges.rs
KaiserKarel: Not sure if I fully understand the comment. It can definitely be passed down from the config trait of the pallet, although I do not see a usecase for more that ~5 ranges to be honest. In this case we just put a reasonable upper bound on it. If we wanted to upgrade that bound, we'd need a runtime upgrade anyways.

KaiserKarel: I'd also prefer not to pollute the signatures too much. With the introduction of MaxEncodedLen, many storage items previously using Vec will need to decide on an upper bound.

.github/workflows/benchmark.yml
```diff
@@ -51,7 +51,7 @@ jobs:
           BASE_BRANCH: ${{ github.event.pull_request.base.ref }}
           GITHUB_BRANCH_NAME: ${{ steps.branch-name.outputs.current_branch }}
 
-        if: env.RUNTIME_CHECK == 1
+        # if: env.RUNTIME_CHECK == 1
```
hussein-aitlahcen, line 54: we should remove this line?

hussein-aitlahcen (approved):

dzmitry-lahoda (approved):

PR #585, opened 2022-02-03
simplify benchmarking

KaiserKarel:


.maintain/run_benchmarks.sh
```diff
@@ -71,8 +52,8 @@ run_benchmarks() {
 
 for i in "${VERSIONS_FILES[@]}"; do
   while IFS=',' read -r output chain folder; do
-    if has_runtime_changes "${BASE_BRANCH}" "${GITHUB_BRANCH_NAME}" "$folder"; then
-      run_benchmarks $output $chain $folder
-    fi
+    # if has_runtime_changes "${BASE_BRANCH}" "${GITHUB_BRANCH_NAME}" "$folder"; then
+    run_benchmarks $output $chain $folder
+    # fi
```
seunlanlege, line 57: 🥲

KaiserKarel: just tinkering around, trying to figure out why CI is so much slower than my local PC

KaiserKarel: fml, it uses the current workflow config instead of this PRs it seems >.<

PR #603, opened 2022-02-07
CU-1kjdbd0 Oracle proptests

cor:
# Oracle proptests

https://app.clickup.com/t/1kjdbd0

Please note that there will be a separate ticket for the reminder of the tests, as noted by Rob

Please also note that the following tests are removed since the start of this PR, as they are no longer needed because of our use of `Validated` types:
- `max_answers_cannot_be_more_than_max_answer_bound`
- `threshold_cannot_be_100`
- `max_answers_cannot_be_less_than_min_answers`
- `min_answers_cannot_be_0`
- `block_interval_cannot_be_less_than_stale_price`

I'm happy that this is the case, because it is better to solve these kind of issue at the typesystem level.


ComposableFiB0t:
Task linked: [CU-1kjdbd0 Oracle proptests](https://app.clickup.com/t/1kjdbd0)

docs/proptest.md
```diff
@@ -102,3 +102,25 @@ prop_compose! {
 ```
 
 This small introduction is not meant to fully cover proptest but rather to give a quick and practical overview of its feature. To learn more about it, you should probably read [the proptest documentation](https://altsysrq.github.io/rustdoc/proptest/1.0.0/proptest/index.html).
+
+## Setting up a pallet for proptests
+First, add the `proptest` and `composable-tests-helpers` dev-dependencies to the pallet's `Cargo.toml`:
+```toml
+[dev-dependencies]
+proptest = "0.9.6" # or a newer version
```
dzmitry-lahoda, line 110: 1.0 please. we have both.

frame/oracle/src/tests.rs
```diff
@@ -20,205 +20,462 @@ use sp_runtime::{
...
+    fn asset_info()
+        (
+            min_answers in 1..MaxAnswerBound::get(),
+            max_answers in 1..MaxAnswerBound::get(),
+            block_interval in (StalePrice::get()+1)..(BlockNumber::MAX/16),
+            threshold in 0..100u8,
+            reward in 0..u64::MAX,
+            slash in 0..u64::MAX,
+        ) -> AssetInfo<Percent, BlockNumber, Balance> {
+            let min_answers = max_answers.saturating_sub(min_answers) + 1;
```
dzmitry-lahoda, line 38: i used prop_filter to filter out cases when max < min. As i see now you test exactly same case of 1 way to many times.

frame/oracle/src/tests.rs
```diff
@@ -20,205 +20,462 @@ use sp_runtime::{
...
+        #[test]
+        fn fails_when_non_root_account(
+            asset_id in asset_id(),
+            asset_info in asset_info(),
+            account_id in account_id(),
+        ) {
+            // very small chance of happening, but for correctness' sake ;)
+            prop_assume!(account_id != get_root_account());
+
+            new_test_ext().execute_with(|| {
```
dzmitry-lahoda, line 143: no sure if should be prop test so. it is easy to spot if somebody changes root to non root in review. probably we will have that non root in future. and non prop test is enough i think. prop test cost - time to maintain and time to run in ci

frame/oracle/src/tests.rs
```diff
@@ -20,205 +20,462 @@ use sp_runtime::{
...
+
+
+        #[test]
+        fn threshold_cannot_be_100(
+            asset_id in asset_id(),
+            asset_info in asset_info(),
+        ) {
+            let root_account = get_root_account();
+
+            new_test_ext().execute_with(|| {
```
dzmitry-lahoda, line 239: probably not prop test too. OR proptest but should be on different variable - on precentage :)

frame/oracle/src/tests.rs
```diff
@@ -20,205 +20,462 @@ use sp_runtime::{
...
+                let root_account = get_root_account();
+                // let account_3 = get_account_3();
+                // let account_4 = get_account_4();
+                // let account_5 = get_account_5();
+
+                prop_assert_ok!(Oracle::set_signer(Origin::signed(root_account), signer_account));
+                prop_assert_eq!(Oracle::controller_to_signer(root_account), Some(signer_account));
+                prop_assert_eq!(Oracle::signer_to_controller(signer_account), Some(root_account));
+
+                // prop_assert_ok!(Oracle::set_signer(Origin::signed(account_1), account_5));
```
dzmitry-lahoda, line 427: not sure. may be delete?

frame/oracle/src/tests.rs
cor: @dzmitry-lahoda  Where did you use prop_filter like this? Could you link to it?

frame/oracle/src/tests.rs
cor: I think that makes sense. The 100% one was because it was migrated from an older test. Makes sense to make the percentage a prop

frame/oracle/src/mock.rs
```diff
@@ -16,6 +16,7 @@ use sp_runtime::{
 type UncheckedExtrinsic = frame_system::mocking::MockUncheckedExtrinsic<Test>;
 type Block = frame_system::mocking::MockBlock<Test>;
 
+
```
dzmitry-lahoda, line 19: new line

frame/oracle/src/mock.rs
```diff
@@ -34,6 +35,9 @@ parameter_types! {
 	pub const SS58Prefix: u8 = 42;
 }
 
+
```
dzmitry-lahoda, line 38: new line

frame/oracle/src/tests.rs
dzmitry-lahoda: https://github.com/search?q=org%3AComposableFi+prop_filter&type=code

dzmitry-lahoda (approved):

frame/oracle/src/tests.rs
```diff
@@ -22,228 +22,716 @@ use std::sync::Arc;
...
+		// 		System::set_block_number(last_update + asset_info.block_interval + 1);
+
+		// 		prop_assert_noop!(
+		// 			Oracle::submit_price(Origin::signed(submitter_account), price_value, asset_id),
+		// 			Error::<Test>::NotEnoughStake
+		// 		);
+
+		// 		Ok(())
+		// 	})?;
+		// }
```
andor0, line 724: Should add TODO or remove this.

andor0 (approved):

dzmitry-lahoda (approved):

PR #604, opened 2022-02-07
Use tarpaulin for code coverage

hussein-aitlahcen:


.github/workflows/check.yml
```diff
@@ -63,19 +63,15 @@ jobs:
...
+          /home/runner/.cargo/bin/cargo install cargo-tarpaulin
+          SKIP_WASM_BUILD=1 /home/runner/.cargo/bin/cargo tarpaulin --verbose --features=runtime-benchmarks --workspace --locked --out Xml
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v1
+        uses: codecov/codecov-action@v2
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
-          files: lcov.info
-          fail_ci_if_error: false
+          fail_ci_if_error: true
```
dzmitry-lahoda, line 74: could we make 2 things separately? intoroduce new cover and than start fail:) so we sure no have false positives.

please comment why tarpauling better than llvm cover i guess because rust native works better:)

hussein-aitlahcen: I'll give a second shot to this PR, if the result is not good I'll just drop it

hussein-aitlahcen:
Closing as the investigation was not concluding.

PR #608, opened 2022-02-08
CU-1wty1h0 liquidations test/bench/runtime - required by lending

dzmitry-lahoda:


ComposableFiB0t:
Task linked: [CU-1wty1h0 liqudation: write runtime tests](https://app.clickup.com/t/1wty1h0)

runtime/dali/Cargo.toml
```diff
@@ -174,12 +175,14 @@ std = [
...
 	"bonded-finance/std",
 	"dutch-auction/std",
+	"liquidations/std",
 	"vesting/std",
 	"mosaic/std",
 	"preimage/std",
 	"membership/std",
 	"system-rpc-runtime-api/std",
 	"transaction-payment-rpc-runtime-api/std",
+	"liquidations/std",
```
vivekvpandya, line 185: seems redundant

frame/liquidations/src/mock/runtime.rs
```diff
@@ -0,0 +1,266 @@
...
+}
+
+parameter_types! {
+	pub const SS58Prefix: u8 = 42;
+	pub const BlockHashCount: u64 = 250;
+}
+
+impl frame_system::Config for Runtime {
+	type BaseCallFilter = Everything;
+
```
vivekvpandya, line 62: extra white spaces.

runtime/dali/Cargo.toml
dzmitry-lahoda: why? all other pallets set their std here?

vivekvpandya: yes but line number 178 and 185 are identical , I don't understand why we need both of them?

dzmitry-lahoda: ahaa:)  thanks
image

dzmitry-lahoda: just was not visible in uI

vivekvpandya (approved):

frame/liquidations/src/mock/runtime.rs
dzmitry-lahoda: fixed in next pr to unclude lending

PR #617, opened 2022-02-10
CU-25cey0g Update benchmarks settings

andor0:
We already use `define_benchmarks!` macro (https://github.com/paritytech/substrate/blob/polkadot-v0.9.16/frame/benchmarking/src/lib.rs#L1839-L1869), so, we can use `list_benchmarks` and `add_benchmarks` macros instead of `list_benchmark` and `add_benchmark` macros accordingly.


ComposableFiB0t:
Task linked: [CU-25cey0g Update benchmarks settings](https://app.clickup.com/t/25cey0g)

dzmitry-lahoda (approved):

hussein-aitlahcen (approved):

PR #619, opened 2022-02-11
Refactored Validated Traits

PoisonPhang:
Explicitly implemented `TypeInfo`, `PartialEq`, `Eq`, and `Debug` for `Validated` and redirected them to `T`'s implementations.

This allows us to create `Validated` structs where `U` and its types do not implement the traits. `derive` would otherwise require that both `T` and `U` would implement these traits.

This has been tested on the CreationDeposit amount in the Vault pallet here https://github.com/ComposableFi/composable/tree/connor/validate-vault-extrinsics/frame/vault/src
See:
validation.rs
lib.rs
tests.rs

### DIFF
--- a/frame/composable-support/src/validation.rs
+++ b/frame/composable-support/src/validation.rs
@@ -65,15 +65,43 @@
 //! }
 //! ```
 
-use core::marker::PhantomData;
+use core::{marker::PhantomData, fmt};
 use scale_info::TypeInfo;
 use sp_runtime::DispatchError;
 
 /// Black box that embbed the validated value.
-#[derive(Default, Copy, Clone, PartialEq, Eq, Debug, TypeInfo)]
+#[derive(Default, Copy, Clone)]
 pub struct Validated<T, U> {
 	value: T,
 	_marker: PhantomData<U>,
+}
+
+impl<T, U> TypeInfo for Validated<T, U>
+where T: TypeInfo,
+{
+    type Identity = <T as TypeInfo>::Identity;
+
+    fn type_info() -> scale_info::Type {
+        T::type_info()
+    }
+}
+
+impl<T, U> PartialEq for Validated<T, U>
+where T: PartialEq,
+{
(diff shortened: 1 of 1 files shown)


dzmitry-lahoda (approved):

frame/composable-support/src/validation.rs
```diff
@@ -65,17 +65,48 @@
...
 //! ```
 
-use core::marker::PhantomData;
+use core::{fmt, marker::PhantomData};
 use scale_info::TypeInfo;
 use sp_runtime::DispatchError;
 
 /// Black box that embbed the validated value.
-#[derive(Default, Copy, Clone, PartialEq, Eq, Debug, TypeInfo)]
+#[derive(Default, Copy, Clone)]
```
dzmitry-lahoda, line 73: you can do that to default where T: Default . So Validated<T,_> can be default iff T is Default

hussein-aitlahcen (dismissed):

dzmitry-lahoda (approved):

frame/composable-support/src/validation.rs
PoisonPhang: Take a look at 74095f779fe4b4221bbe24061e86d8249ec73354, I think I implemented this as you described

dzmitry-lahoda (approved):

frame/composable-support/src/validation.rs
dzmitry-lahoda: hm, may be u said  something stupid. may be derive and derive_more do all these things. if traget is traited, than trait is for wrapper.  and probably infernce will work like with hashmap, but we have no example

PoisonPhang: Reverted changes here 00187d3786d302f87f11c3415823c1ede0a943b0

hussein-aitlahcen (approved):

dzmitry-lahoda (approved):

PR #640, opened 2022-02-15
add initialize_at

KaiserKarel:
Allows the VestingBlockstart to be set at a specific block, which plays a lot nicer with some other pallets in conjunction with `batch` calls.

### DIFF
--- a/frame/crowdloan-rewards/src/tests.rs
+++ b/frame/crowdloan-rewards/src/tests.rs
@@ -96,6 +96,15 @@ fn test_initialize_ok() {
 }
 
 #[test]
+fn test_initialize_at_ok() {
+	ExtBuilder::default().build().execute_with(|| {
+		assert_ok!(CrowdloanRewards::initialize_at(Origin::root(), 10));
+		assert_eq!(CrowdloanRewards::total_rewards(), 0);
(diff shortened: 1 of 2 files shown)


frame/crowdloan-rewards/src/tests.rs
```diff
@@ -95,6 +95,15 @@ fn test_initialize_ok() {
...
 
+#[test]
+fn test_initialize_at_ok() {
+	ExtBuilder::default().build().execute_with(|| {
+		assert_ok!(CrowdloanRewards::initialize_at(Origin::root(), 10));
+		assert_eq!(CrowdloanRewards::total_rewards(), 0);
+		assert_eq!(CrowdloanRewards::claimed_rewards(), 0);
+	});
+}
+
```
hussein-aitlahcen, line 106: Please add a test to ensure you cannot claim if you initialize_at a future block

KaiserKarel: yessir

hussein-aitlahcen:
I think we need to add the NotClaimableYet check in the ValidateUnsigned impl as well as the associate call is free. It's better to avoid including them if they can't associate yet

benluelo (approved):

hussein-aitlahcen (approved):

PR #642, opened 2022-02-15
Fix amount_available_to_claim_for RPC

benluelo:
Fixes the `crowdloan_rewards` `amount_available_to_claim_for` RPC to return the amount available to claim, not the amount that should have been claimed.

### DIFF
--- a/frame/crowdloan-rewards/src/lib.rs
+++ b/frame/crowdloan-rewards/src/lib.rs
@@ -372,6 +372,17 @@ pub mod pallet {
 		}
 	}
 
+	/// Returns the amount available to claim for the specified account.
+	pub fn amount_available_to_claim_for<T: Config>(
+		account_id: <T as frame_system::Config>::AccountId,
+	) -> Result<T::Balance, DispatchError> {
+		let association = Associations::<T>::get(account_id).ok_or(Error::<T>::NotAssociated)?;
+		let reward = Rewards::<T>::get(association).ok_or(Error::<T>::NothingToClaim)?;
+		let should_have_claimed = should_have_claimed::<T>(&reward)?;
+		let available_to_claim = should_have_claimed - reward.claimed;
+		Ok(available_to_claim)
+	}
+
 	pub fn get_remote_account<T: Config>(
 		proof: Proof<<T as Config>::RelayChainAccountId>,
 		reward_account: &<T as frame_system::Config>::AccountId,
--- a/runtime/composable/src/lib.rs
+++ b/runtime/composable/src/lib.rs
@@ -843,13 +843,8 @@ impl_runtime_apis! {
 	impl crowdloan_rewards_runtime_api::CrowdloanRewardsRuntimeApi<Block, AccountId, Balance> for Runtime {
 		fn amount_available_to_claim_for(account_id: AccountId) -> SafeRpcWrapper<Balance> {
 			SafeRpcWrapper(
-			crowdloan_rewards::Associations::<Runtime>::get(account_id)
-				.map(crowdloan_rewards::Rewards::<Runtime>::get)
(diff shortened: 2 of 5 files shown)


hussein-aitlahcen (approved):

frame/crowdloan-rewards/src/lib.rs
```diff
@@ -372,6 +372,17 @@ pub mod pallet {
 		}
 	}
 
+	/// Returns the amount available to claim for the specified account.
+	pub fn amount_available_to_claim_for<T: Config>(
+		account_id: <T as frame_system::Config>::AccountId,
+	) -> Result<T::Balance, DispatchError> {
+		let association = Associations::<T>::get(account_id).ok_or(Error::<T>::NotAssociated)?;
+		let reward = Rewards::<T>::get(association).ok_or(Error::<T>::NothingToClaim)?;
+		let should_have_claimed = should_have_claimed::<T>(&reward)?;
```
dzmitry-lahoda, line 381: may could have claimed? because she is not should do it, but could.

dzmitry-lahoda, line 381: possible claim up to now

benluelo: I just named the variable after the function, I do agree that it could do with a better name (function and variable)

runtime/composable/src/lib.rs
```diff
@@ -843,13 +843,8 @@ impl_runtime_apis! {
...
 		fn amount_available_to_claim_for(account_id: AccountId) -> SafeRpcWrapper<Balance> {
 			SafeRpcWrapper(
-			crowdloan_rewards::Associations::<Runtime>::get(account_id)
-				.map(crowdloan_rewards::Rewards::<Runtime>::get)
-				.flatten()
-				.as_ref()
-				.map(crowdloan_rewards::should_have_claimed::<Runtime>)
-				.unwrap_or_else(|| Ok(Balance::zero()))
-				.unwrap_or_else(|_| Balance::zero())
+				crowdloan_rewards::amount_available_to_claim_for::<Runtime>(account_id)
```
dzmitry-lahoda, line 846: oh, we have to define rpc 3 times in each runtime,  not like once define extrincic?

benluelo: Yes, the each runtime much implement the RPC

frame/crowdloan-rewards/src/lib.rs
dzmitry-lahoda: [😞](https://emojikeyboard.org/copy/Disappointed_Face_Emoji_%F0%9F%98%9E?utm_source=extlink)

hussein-aitlahcen: The term initially refer to the fact that the use was currently claiming (do_claim). In the RPC the user is not claiming but rather checking the current available claimable amount so you can change it. I don't have a strong opinion it's up to you

hussein-aitlahcen: As a contributor, you 'should' claim this amount, as this reward is the only consequence of being a contributor

hussein-aitlahcen: I think the wording is accurate

benluelo: I think if it is to be changed, it's outside the scope of this PR and can be reviewed later

andor0 (approved):

frame/crowdloan-rewards/src/lib.rs
benluelo: Yeah, I can open a new PR regarding that. This PR was just for a hotfix for the RPC

PR #646, opened 2022-02-16
Update make version

haroldsphinx:


Makefile
```diff
@@ -50,7 +51,7 @@ dev:
 .PHONY: version
 version:
 	@if [ ${RELEASE_VERSION} ]; then \
-	sed -i "s|^version =.*|version = '"${RELEASE_VERSION}"'|" node/Cargo.toml; \
+	sed -i "s|^version =.*|version = '"${CARGO_VERSION}"'|" Cargo.toml; \
```
seunlanlege, line 54:
```suggestion
	sed -i "s|^version =.*|version = '"${RELEASE_VERSION}"'|" Cargo.toml; \
```

Makefile
```diff
@@ -4,7 +4,8 @@ REPO=composablefi
 SERVICE_NAME=composable
 INSTALL_DIR=docker/
 IMAGE_URL:=${REPO}/${SERVICE_NAME}
-CARGO_VERSION:=$(shell git tag --sort=committerdate | grep -E '^v[0-9]' | tail -1 )
+RELEASE_VERSION:=$(shell git tag --sort=committerdate | grep -E '^v[0-9]' | tail -1)
```
seunlanlege, line 7:
```suggestion
```

.github/workflows/docker-release.yml
```diff
@@ -37,7 +36,7 @@ jobs:
   
     - name: Set env
       run: |
-        echo "RELEASE_VERSION=$(git tag --sort=committerdate | grep -E '^v[0-9]' | tail -1 | cut -c2- )" >> $GITHUB_ENV
+        echo "RELEASE_VERSION=$(git tag --sort=committerdate | grep -E '^v[0-9]' | tail -1)" >> $GITHUB_ENV
```
seunlanlege, line 39:
```suggestion
        echo "RELEASE_VERSION=$(git tag --sort=committerdate | grep -E '^v[0-9]' | tail -1 | cut -c2- )" >> $GITHUB_ENV
```

seunlanlege (changes requested):

Makefile
haroldsphinx: We don't need to remove this...the way the docker release it's set, the release version needs to be set in the makefile and it doesnt affect the way release-pipeline works

.github/workflows/docker-release.yml
haroldsphinx: Same applies here

PR #655, opened 2022-02-17
[CU-1ux39qh] Validation/oracle

ysfkel:


frame/oracle/src/benchmarking.rs
```diff
@@ -41,10 +38,10 @@ benchmarks! {
 	add_asset_and_info {
 		let caller = T::AddOracle::successful_origin();
 		let asset_id = 1;
-		Validated::new(Percent::from_percent(80)).unwrap(),
-		Validated::new(3).unwrap(),
-		Validated::new(5).unwrap(),
-		Validated::<T::BlockNumber, ValidBlockInterval<T::StalePrice>>::new(5).unwrap(),
+		Validated::new(Percent::from_percent(80)).unwrap();
```
dzmitry-lahoda, line 41: in lending benches are part of tests and run as part, so these will never break

frame/oracle/src/lib.rs
```diff
@@ -457,23 +461,27 @@ pub mod pallet {
...
 			origin: OriginFor<T>,
 			asset_id: T::AssetId,
-			threshold: Percent,
-			min_answers: u32,
-			max_answers: u32,
-			block_interval: T::BlockNumber,
+			valid_threshold: Validated<Percent, ValidThreshhold>,
+			valid_min_answers: Validated<u32, ValidMinAnswers>,
+			valid_max_answers: Validated<u32, ValidMaxAnswer<T::MaxAnswerBound>>,
+			valid_block_interval: Validated<T::BlockNumber, ValidBlockInterval<T::StalePrice>>,
```
KaiserKarel, line 467: I don't think there is a need to name these valid_foo to be honest. The type reflects what it is.

frame/oracle/src/lib.rs
```diff
@@ -457,23 +461,27 @@ pub mod pallet {
...
 			reward: BalanceOf<T>,
 			slash: BalanceOf<T>,
 		) -> DispatchResultWithPostInfo {
 			T::AddOracle::ensure_origin(origin)?;
-			ensure!(min_answers > 0, Error::<T>::InvalidMinAnswers);
+
+			let threshold = valid_threshold.value();
+			let min_answers = valid_min_answers.value();
+			let max_answers = valid_max_answers.value();
+			let block_interval = valid_block_interval.value();
```
KaiserKarel, line 476: Can we make the Deref impl public to avoid the need to remap names?

dzmitry-lahoda: deref would eventually lead to clones if we use move:
```
let (x,y,z) = (z.into(), y.into(), z.into());
```
or as in lending - if there are so much parameters (and new may be added for v2) - create struct.

dzmitry-lahoda: they are thinking about DerefMove for several years:)

dzmitry-lahoda: oh, in this case it is all copy values. will work.

frame/oracle/src/benchmarking.rs
cor: What do you mean by this?

frame/oracle/src/lib.rs
cor: Fixed in 628e2577c713fb871ca3aae8c98926b0bf9383b0

frame/oracle/src/lib.rs
cor: fixed in 43da2998d2db1e50149a6882807743a7684de31e

cor (dismissed):
Hmm, `cargo test -p pallet-oracle` was succesfull, but CI is failing

KaiserKarel:
Still some Ci errors :*(

cor (approved):
LGTM

PR #656, opened 2022-02-18
xcmp: treasury, trap, weights, tests, fees, ed, common

dzmitry-lahoda:
- remaining test will run in next PR
- started to handling ED
- started to move common parts to reuse in runtimes
- identified issues with assets registry issues (created task)
- started abstracting XCMP parts to allow to test them out of XCMP (generic conversion, treasury, et

### DIFF
--- a/frame/curve-amm/src/benchmarking.rs
+++ b/frame/curve-amm/src/benchmarking.rs
@@ -22,153 +22,120 @@ benchmarks! {
 	  let protocol_fee = Permill::from_percent(1);
   } : _(RawOrigin::Signed(owner), pair, amplification_factor, fee, protocol_fee)
 
-  add_liquidity {
-	  let usdc: T::AssetId = 0.into();
-	  let usdt: T::AssetId = 1.into();
-	  let owner = whitelisted_caller();
-	  let pool_id = StableSwap::<T>::do_create_pool(
-		  &owner,
-		  CurrencyPair::new(usdc, usdt),
-		  1000_u16,
-		  Permill::from_percent(1),
-		  Permill::from_percent(1),
-	  ) .expect("impossible; qed;");
-	  let unit = 1_000_000_000_000;
-	  // 100_000_000 USDC , 100_000_000 USDT
-	  let initial_usdc: T::Balance = (100_000_000_u128 * unit).into();
-	  let initial_usdt: T::Balance = (100_000_000_u128 * unit).into();
-	  // Mint the tokens
-	  assert_ok!(T::Assets::mint_into(usdc, &owner, initial_usdc));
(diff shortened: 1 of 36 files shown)


andor0 (approved):

frame/curve-amm/src/benchmarking.rs
```diff
@@ -22,153 +22,120 @@ benchmarks! {
...
-			  initial_usdc,
-			  initial_usdt,
-			  0.into(),
-			  false
-	  ));
-	  let user = account("user", 0, 0);
-	  assert_ok!(T::Assets::mint_into(usdt, &user, (1000_u128 * unit).into()));
-	  // buy 1000 USDC
-  }: _(RawOrigin::Signed(user), pool_id, usdc, (1000_u128 * unit).into(), false)
+// TODO: Fix this benchmarks
```
hussein-aitlahcen, line 25: Since we have multideposit zero for unknown currency, can we reenable ll those benchmarkings?

runtime/primitives/src/currency.rs
```diff
@@ -10,6 +10,15 @@ use composable_support::rpc_helpers::FromHexStr;
 use serde::{Deserialize, Serialize};
 use sp_runtime::sp_std::ops::Deref;
 
+/// Trait used to write generalized code over well know currencies
+/// We use const to allow for match on these
+pub trait WellKnownCurrency<AssetId: Copy> {
```
hussein-aitlahcen, line 15: Do we have a reason to be generic here? Perhaps `pub trait WellKnownCurrency { const NATIVE: Self... }` is enough?

hussein-aitlahcen (commented):
Small comments

frame/curve-amm/src/benchmarking.rs
dzmitry-lahoda: i think we must fix the issue to prevent possible security holes to spam us with an infinite amount of accounts. overall we have other currencies defined. created task https://app.clickup.com/t/25ya821

runtime/primitives/src/currency.rs
dzmitry-lahoda: i will check it.  so I doubt that currency will be also provider of well know CONST (consts kind of needed) in all cases. example, DOT vs KSM providers.  so will look - may be impl WellKnownCurrency<Self> will work.

dzmitry-lahoda:
@hussein-aitlahcen created issue to hande non zero weight. i doubt it is minor issue as it allows to spam network. 
for currency, I also doubt will check during making XCMP cross relay support (DOT, WND, KSM)

andor0 (approved):

PR #647, opened 2022-02-17
[CU-1ux39mq] Validate vault extrinsics

PoisonPhang:
Added validation for Creation Deposit and Max Strategies.

Validating any input that was checked against `origin` or anything that was derived from `origin` proved to be more cumbersome than any existing validation.

Be sure to check that I didn't miss any other extrinsic inputs.

PR #666, opened 2022-02-18
[CU-1vzab3b] Added TryIntoValidated

PoisonPhang:
This does type inference just fine but should be tested with #647 before leaving draft.

### DIFF
--- a/frame/composable-support/src/validation.rs
+++ b/frame/composable-support/src/validation.rs
@@ -130,15 +130,16 @@ pub trait Validate<T, U> {
 }
 
 pub trait TryIntoValidated<T, U> {
-    fn try_into_validated(self) -> Result<Validated<T, U>, &'static str>;
+	fn try_into_validated(self) -> Result<Validated<T, U>, &'static str>;
 }
 
-impl<T, U> TryIntoValidated<T, U> for T 
-where U: Validate<T, U>,
+impl<T, U> TryIntoValidated<T, U> for T
+where
+	U: Validate<T, U>,
 {
-    fn try_into_validated(self) -> Result<Validated<T, U>, &'static str> {
-        Validated::new(self)
-    }
+	fn try_into_validated(self) -> Result<Validated<T, U>, &'static str> {
+		Validated::new(self)
+	}
 }
 
 #[derive(Debug, Eq, PartialEq, Default)]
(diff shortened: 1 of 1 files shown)


ComposableFiB0t:
Task linked: [CU-1vzab3b Introduce a new Validated helper](https://app.clickup.com/t/1vzab3b)

benluelo (dismissed):

dzmitry-lahoda (commented):
if do just try into and prelude will rust solve types without explicit casts than?

benluelo:
> if do just try into and prelude will rust solve types without explicit casts than?

As long as the validated type is used somewhere with a concrete type (i.e. a function or method) it should be fine. If you're just setting a variable, you'll have to supply type annotations for the `U` type.

```rust
let t: Validated<_, SomeType> = value.try_into_validated().unwrap();
```

Perhaps we could put the `U` parameter on the method? That way it could be specified like this:

```rust
let t = value.try_into_validated::<SomeType>().unwrap();
```

hussein-aitlahcen:
> > if do just try into and prelude will rust solve types without explicit casts than?
> 
> As long as the validated type is used somewhere with a concrete type (i.e. a function or method) it should be fine. If you're just setting a variable, you'll have to supply type annotations for the `U` type.
> 
> ```rust
> let t: Validated<_, SomeType> = value.try_into_validated().unwrap();
> ```
> 
> Perhaps we could put the `U` parameter on the method? That way it could be specified like this:
> 
> ```rust
> let t = value.try_into_validated::<SomeType>().unwrap();
> ```

Adding U as parameter would be nifty 👍

benluelo:
> > > if do just try into and prelude will rust solve types without explicit casts than?
> > 
> > 
> > As long as the validated type is used somewhere with a concrete type (i.e. a function or method) it should be fine. If you're just setting a variable, you'll have to supply type annotations for the `U` type.
> > ```rust
> > let t: Validated<_, SomeType> = value.try_into_validated().unwrap();
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Perhaps we could put the `U` parameter on the method? That way it could be specified like this:
> > ```rust
> > let t = value.try_into_validated::<SomeType>().unwrap();
> > ```
> 
> Adding U as parameter would be nifty +1

I believe this would work, haven't tested it:

```rust
pub trait TryIntoValidated<T> {
	fn try_into_validated<U: Validate<T, U>>(self) -> Result<Validated<T, U>, &'static str>;
}

impl<T> TryIntoValidated<T> for T
{
	fn try_into_validated<U: Validate<T, U>>(self) -> Result<Validated<T, U>, &'static str> {
		Validated::new(self)
	}
}
```

benluelo (approved):

PoisonPhang:
> > > > if do just try into and prelude will rust solve types without explicit casts than?
> > > 
> > > 
> > > As long as the validated type is used somewhere with a concrete type (i.e. a function or method) it should be fine. If you're just setting a variable, you'll have to supply type annotations for the `U` type.
> > > ```rust
> > > let t: Validated<_, SomeType> = value.try_into_validated().unwrap();
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Perhaps we could put the `U` parameter on the method? That way it could be specified like this:
> > > ```rust
> > > let t = value.try_into_validated::<SomeType>().unwrap();
> > > ```
> > 
> > 
> > Adding U as parameter would be nifty +1
> 
> I believe this would work, haven't tested it:
> 
> ```rust
> pub trait TryIntoValidated<T> {
> 	fn try_into_validated<U: Validate<T, U>>(self) -> Result<Validated<T, U>, &'static str>;
> }
> 
> impl<T> TryIntoValidated<T> for T {
> 	fn try_into_validated<U: Validate<T, U>>(self) -> Result<Validated<T, U>, &'static str> {
> 		Validated::new(self)
> 	}
> }
> ```
> 
> I think this would also make more sense since we implement the trait for all types, and then the method itself is generic over the validator

Yep, that works and is quite a bit cleaner. Thanks

hussein-aitlahcen (approved):

PR #673, opened 2022-02-21
lending market creation event is good

dzmitry-lahoda:


frame/lending/src/tests.rs
```diff
@@ -314,16 +319,40 @@ fn can_create_valid_market() {
...
 
 		Tokens::mint_into(borrow_asset, &manager, INITIAL_BORROW_ASSET_AMOUNT).unwrap();
-		let created = <Lending as composable_traits::lending::Lending>::create(manager, config);
+		let created = Lending::create_market(
+			Origin::signed(manager),
+			Validated::new(config.clone()).unwrap(),
+		);
 		assert_ok!(created);
+
+		let (market_id, borrow_vault_id) = System::events()
```
benluelo, line 333: You should be able to match on `System::events().last()` since the latest event should be the `MarketCreated` event; something like the following?

```rust
match System::events().last() {
			Some(EventRecord {
				topics: event_topics,
				phase: Phase::Initialization,
				event: Event::Lending(crate::Event::MarketCreated {
					input: event_input,
					market_id: MarketIndex(1),
					vault_id: 1,
					manager: event_manager,
				}),
			}) => {}
		}
```

dzmitry-lahoda: i do not want match or last actually. i want what ever market created, but created.

frame/lending/src/tests.rs
```diff
@@ -314,16 +319,40 @@ fn can_create_valid_market() {
 			reserved_factor: DEFAULT_MARKET_VAULT_RESERVE,
 			currency_pair: CurrencyPair::new(collateral_asset, borrow_asset),
 		};
-		let created =
+		let failed =
 			<Lending as composable_traits::lending::Lending>::create(manager, config.clone());
-		assert!(!created.is_ok());
+		assert!(!failed.is_ok());
```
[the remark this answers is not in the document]
benluelo: ^

frame/lending/src/tests.rs
benluelo: So you want all the market creation events, and then the created market id?

dzmitry-lahoda: i want some market-created event and get its id.  i do not want to be more specific

Dom-Roth (approved):
LGTM

andor0 (approved):

PR #674, opened 2022-02-21
Karel/add substrace

KaiserKarel:
Adds a custom linter which checks if pallets have the proper clippy attributes to avoid panics.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -239,8 +239,8 @@ jobs:
           
       - name: Cargo clippy
         run: |
-          SKIP_WASM_BUILD=1 env -u RUSTFLAGS cargo +nightly clippy --all-targets --tests -- -D warnings
-
+          SKIP_WASM_BUILD=1 env -u RUSTFLAGS cargo +nightly clippy --workspace --all-targets --tests -- -D warnings
+      
       - name: Cargo Single Depedency
         # description: Speed up builds and reduced security check surface
         run: |
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -234,9 +234,9 @@ jobs:
 
       - name: Substrace
         run: |
-          cargo install dylint dylint-link
-          RUSTFLAGS="" SKIP_WASM_BUILD=1 cargo dylint substrace --workspace
-          
+          cargo install cargo-dylint dylint-link
+          RUSTC_WRAPPER="" SKIP_WASM_BUILD=1 cargo dylint substrace --workspace    
+      
       - name: Cargo clippy
         run: |
           SKIP_WASM_BUILD=1 env -u RUSTFLAGS cargo +nightly clippy --workspace --all-targets --tests -- -D warnings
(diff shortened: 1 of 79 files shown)


cor:
The linter looks very nice, but why are there changes to `runtime/composable/src/lib.rs`?

Cargo.toml
```diff
@@ -207,3 +207,8 @@ pallet-beefy-mmr = { git = "https://github.com/composableFi/substrate", branch =
 beefy-merkle-tree = { git = "https://github.com/composableFi/substrate", branch = "polkadot-v0.9.16" }
 pallet-nicks = { git = "https://github.com/composableFi/substrate", branch = "polkadot-v0.9.16" }
 pallet-staking-reward-fn = { git = "https://github.com/composableFi/substrate", branch = "polkadot-v0.9.16" }
+
+[workspace.metadata.dylint]
+libraries = [
+    { git = "https://github.com/kaiserkarel/substrace" },
```
dzmitry-lahoda, line 213: ok. so would prefer this to be part of composable repo for easy maintenance until tool will become stable amid several upgrade of substrate.

dzmitry-lahoda (approved):

seunlanlege (approved):

Cargo.toml
KaiserKarel: Yeah I can fork the repo to composable, do dev on my branch then sync regularly for example, or just migrate it over entirely.

KaiserKarel:
Just need to wait for sccache to be removed from the CI-pipeline for this to pass CI I think.

KaiserKarel:
> Substrace looks very nice, but why are there changes to `runtime/composable/src/lib.rs`?

Due to how some of the substrate macros expand, `impl_runtime_apis` adds `panic!` without putting the expanded code in module with a tag indicating it came from a macro (at least seems like it). This caused the `clippy::panic` lint to fail as `panics` were present in the expanded codebase. 

Originally we just disabled that warn, as the module generated did not export `RUNTIME_API_VERSIONS`, so you could not put the macro specifically in it's own module and allow `clippy::panic` there.

Now with some slight code reorgs, it is actually possible, meaning that we only allow `panic` for the macro expanded code, and lint for it in other sections. Before this someone could do:


```
ExistentialDeposit = || { panic!("think of a deposit") } 
```

In the runtime and actually have the code pass checks.

.github/workflows/check.yml
```diff
@@ -232,10 +232,15 @@ jobs:
...
+        run: |
+          cargo install cargo-dylint dylint-link
+          RUSTC_WRAPPER="" SKIP_WASM_BUILD=1 cargo dylint substrace --workspace    
+      
       - name: Cargo clippy
         run: |
-          SKIP_WASM_BUILD=1 env -u RUSTFLAGS cargo +nightly clippy --all-targets --tests -- -D warnings
-
+          SKIP_WASM_BUILD=1 env -u RUSTFLAGS cargo +nightly clippy --workspace --all-targets --tests -- -D warnings
+
```
dzmitry-lahoda, line 243: spaces, may be make style

PR #686, opened 2022-02-26
fix(ci): mandatory single-pass benchmarking

hussein-aitlahcen:


hussein-aitlahcen:
Awaiting #684

seunlanlege:
Nice work.

But I think we should run this in the paritytech:ci-linux container, and only run it when there are runtime changes.

hussein-aitlahcen:
> Nice work.
> 
> But I think we should run this in the paritytech:ci-linux container, and only run it when there are runtime changes.

Yep, will make the change.

PoisonPhang (approved):

PR #688, opened 2022-02-28
polkadot v0.9.17

seunlanlege:


seunlanlege:
no oooo

dzmitry-lahoda (approved):

seunlanlege:
removing needs simnode because the latest update broke simnode, we'll test the upgrade on dali instead

PR #690, opened 2022-02-28
Add few more test for dex.

vivekvpandya:
Add few more test for dex.
Add extrinsics for add/remove_liquidity().
Update benchmarks.

### DIFF
--- a/frame/curve-amm/src/benchmarking.rs
+++ b/frame/curve-amm/src/benchmarking.rs
@@ -2,7 +2,10 @@ use super::*;
 use crate::Pallet as StableSwap;
 use composable_traits::{defi::CurrencyPair, dex::CurveAmm};
 use frame_benchmarking::{account, benchmarks, impl_benchmark_test_suite, whitelisted_caller};
-use frame_support::{assert_ok, traits::fungibles::Mutate};
+use frame_support::{
+	assert_ok,
+	traits::fungibles::{Inspect, Mutate},
+};
 use frame_system::RawOrigin;
 use sp_arithmetic::Permill;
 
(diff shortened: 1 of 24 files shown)


frame/curve-amm/src/tests.rs
```diff
@@ -211,6 +211,64 @@ fn add_lp_imbalanced() {
...
+			&BOB,
+			pool_id,
+			bob_usdc,
+			bob_usdt,
+			expected_min_value,
+			false
+		));
+		let lp = Tokens::balance(pool.lp_token, &BOB);
+		// must have received some lp tokens
+		assert!(lp > 0_u128);
```
KaiserKarel, line 250: Can we assert that we received the correct amount?

frame/uniswap-v2/src/lib.rs
```diff
@@ -302,6 +302,46 @@ pub mod pallet {
...
+		pub fn add_liquidity(
+			origin: OriginFor<T>,
+			pool_id: T::PoolId,
+			base_amount: T::Balance,
+			quote_amount: T::Balance,
+			min_mint_amount: T::Balance,
+			keep_alive: bool,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			let _ = <Self as CurveAmm>::add_liquidity(
```
KaiserKarel, line 316: Why CurveAmm? Should we rename the trait to just AMM?

KaiserKarel (changes requested):
Would be good to check that the obtained LP amount is correct

frame/uniswap-v2/src/lib.rs
KaiserKarel: Probably good for a second PR

hussein-aitlahcen (dismissed):
Thanks for adding a bunch of tests, next step is to add some proptests.

frame/curve-amm/src/benchmarking.rs
```diff
@@ -19,6 +22,57 @@ benchmarks! {
...
+			  &owner,
+			  pool_id,
+			  initial_usdc,
+			  initial_usdt,
+			  0.into(),
+			  false
+	  ));
+	  let pool_info = StableSwap::<T>::get_pool(pool_id).expect("impossible; qed;");
+	  let lp_amount = T::Assets::balance(pool_info.lp_token, &owner);
+  }: _(RawOrigin::Signed(owner), pool_id, lp_amount, (0_u128).into(), (0_u128).into())
```
dzmitry-lahoda, line 74: does this test handles error to remove? so if remove will fail, benchmark with fail?

frame/curve-amm/src/benchmarking.rs
```diff
@@ -19,6 +22,57 @@ benchmarks! {
...
+		  Permill::from_percent(1),
+	  ) .expect("impossible; qed;");
+	  let unit = 1_000_000_000_000;
+	  // 100_000_000 USDC , 100_000_000 USDT
+	  let initial_usdc: T::Balance = (100_000_000_u128 * unit).into();
+	  let initial_usdt: T::Balance = (100_000_000_u128 * unit).into();
+	  // Mint the tokens
+	  assert_ok!(T::Assets::mint_into(usdc, &owner, initial_usdc));
+	  assert_ok!(T::Assets::mint_into(usdt, &owner, initial_usdt));
+  }: _(RawOrigin::Signed(owner), pool_id, initial_usdc, initial_usdt, 0.into(), false)
```
dzmitry-lahoda, line 43: does this test handles error case?

frame/curve-amm/src/tests.rs
```diff
@@ -217,6 +232,64 @@ fn add_lp_imbalanced() {
...
+			bob_usdc,
+			bob_usdt,
+			expected_min_value,
+			false
+		));
+		let lp = Tokens::balance(pool.lp_token, &BOB);
+		// must have received some lp tokens
+		assert!(lp == bob_usdc + bob_usdt);
+
+		let bob_usdc = 1000 * unit;
```
dzmitry-lahoda, line 273: may be split into 2 tests?

frame/curve-amm/src/tests.rs
```diff
@@ -242,14 +315,80 @@ fn remove_lp_failure() {
...
+				lp,
+				min_expected_usdc,
+				min_expected_usdt
+			),
+			crate::Error::<Test>::CannotRespectMinimumRequested
+		);
+	});
+}
+
+//
```
dzmitry-lahoda, line 346: // non needed
// -> ///

frame/uniswap-v2/src/lib.rs
```diff
@@ -302,9 +302,55 @@ pub mod pallet {
 			)?;
 			Ok(())
 		}
+
+		/// Add liquidity to a stable-swap pool.
```
dzmitry-lahoda, line 306: wrong comment

KaiserKarel (approved):

dzmitry-lahoda (approved):

hussein-aitlahcen (approved):

PR #692, opened 2022-02-28
CU-1yyr5hk Update integration tests of assets-registry 

andor0:


ComposableFiB0t:
Task linked: [CU-1yyr5hk xcmp: assets registry : fix acala tests for assets registy](https://app.clickup.com/t/1yyr5hk)

KaiserKarel (approved):

frame/assets-registry/src/lib.rs
```diff
@@ -250,23 +280,23 @@ pub mod pallet {
 	impl<T: Config> RemoteAssetRegistry for Pallet<T> {
 		type AssetId = T::LocalAssetId;
 
-		type AssetNativeLocation = T::ForeignAssetId;
+		type AssetNativeLocation = T::Location;
```
dzmitry-lahoda, line 283: Native has notion of native to some chain or protocol, so all assets are in some sense native. may be call AssetOriginLocation `assets originated from location`

dzmitry-lahoda (approved):

PR #700, opened 2022-03-02
feat(runtimes): filter Tokens and unfilter Balance

hussein-aitlahcen:
- do not filter balance as filtering this pallet does not result in any benefit (some services might be already based on balances)
- filter tokens as we rely on assets for operations
- faster crowdloan on dali
- use assets instead of tokens for dex and fix benchmarking


cor (approved):
LGTM

andor0 (approved):

PR #701, opened 2022-03-02
[cu-20185wq] stableSwapDex Runtime Integration Tests

Dom-Roth:
**Greetings**, this update implements runtime integration tests for the "stableSwapDex" Pallet.

Waits for #690 and #706

Everything went quite straightforward, except 2 issues with the `.sell` and the `.swap` functions.
The extrinsic failed with an **arithmetic overflow** or **underflow** error.
At first, this only happened on the `.sell` function, and randomly later appeared on the `.swap` function. While investigating it solved itself, though I'm currently investigating this a bit more.

**Update**: Just realized this was due to the `amplificationCoefficient` variable when creating a pool being 0. After changing to 1 it worked, though the error message is still quite unclear and could use some improvements. Or directly forbidding 0, if this isn't intended to be set to this.

Most extrinsic return a whole bunch of BigNumbers, of which I couldn't identify all yet, and are currently just defined as `resultNumberX`.

The testing report can be found here: https://app.clickup.com/20465559/docs/kghwq-32881/kghwq-20942

While running the full test suite, I realized the crowdloanRewards test is broken with the latest changes. @hussein-aitlahcen Are there changes required or is this a temporary thing? Asking because of the error message, check the mochawesome summary below.

Screenshot 2022-03-02 at 13-31-01 Mochawesome Report


vivekvpandya:
On Wed, Mar 2, 2022 at 6:06 PM Dominik Roth ***@***.***>
wrote:

> *Greetings*, this update implements runtime integration tests for the
> "stableSwapDex" Pallet.
>
> Everything went quite straightforward, except 2 issues with the .sell and
> the .swap functions.
> The extrinsic failed with an *arithmetic overflow* or *underflow* error.
> At first, this only happened on the .sell function, and randomly later
> appeared on the .swap function. While investigating it solved itself,
> though I'm currently investigating this a bit more.
>
> *Update*: Just realized this was due to the amplificationCoefficient
> variable when creating a pool being 0. After changing to 1 it worked,
> though the error message is still quite unclear and could use some
> improvements. Or directly forbidding 0, if this isn't intended to be set to
> this.
>
Thanks Dominik for these tests. According to
https://curve.fi/files/stableswap-paper.pdf A can be 0. I am trying to
reproduce the same error on my end.

vivekvpandya:
@Dom-Roth amplification_factor can be 0 after https://github.com/ComposableFi/composable/pull/706 .  It is marked as draft as it will be rebased after https://github.com/ComposableFi/composable/pull/690/

integration-tests/runtime-tests/src/types/interfaces/augment-api-consts.ts
```diff
@@ -198,8 +205,49 @@ declare module '@polkadot/api-base/types/consts' {
        **/
       [key: string]: Codec;
     };
+    lending: {
```
vivekvpandya, line 208: Does this PR combines changes related to lending pallets too? if not please remove these changes.

vivekvpandya:
It feels like without `add_liquidity` extrinsic `buy`,  `sell`, `swap` will not do anything real as pool's amounts are 0.
https://github.com/ComposableFi/composable/pull/690/ add those extrinsic so after that better to add test with having some balances in pool.  Please correct me if this is wrong.

Dom-Roth:
> @Dom-Roth amplification_factor can be 0 after #706 . It is marked as draft as it will be rebased after #690

Awesome! Thanks for taking care of this. :)


> It feels like without `add_liquidity` extrinsic `buy`, `sell`, `swap` will not do anything real as pool's amounts are 0. #690 add those extrinsic so after that better to add test with having some balances in pool. Please correct me if this is wrong. Also I apologize for missing them. pray

You're right, the liquidity functions still need to be added, haven't seen #690 before I was working on the tests. 
Thanks for showing, I'll update the tests asap.

integration-tests/runtime-tests/src/types/interfaces/augment-api-consts.ts
Dom-Roth: Just ignore this file. It's part of the typescript types for our chain and gets automatically generated.

hussein-aitlahcen:
@Dom-Roth the crowloan pallet has been updated and require the funds to be moved to the pallet prior to populating the list of contributors

ComposableFiB0t:
Task linked: [CU-20185wq stableSwapDex Pallet Tests](https://app.clickup.com/t/20185wq)

Dom-Roth:
Greetings, there are currently 2 issues with the stableSwapDex.

# Issues
## 1. Buying, Selling, Swapping, or removing liquidity without existing liquidity succeeds.
When no liquidity was provided to a pool before trying to `.buy`, `.sell`, `.swap`, or `.removeLiquidity`, it should fail with an **arithmetic overflow** or **underflow** error. The unit tests are already checking for this, but in the runtime tests, the extrinsics are succeeding.

## 2. Couldn't get `.removeLiquidity` working (yet).
When calling `.removeLiquidity` the call fails with:
```
Error: stableSwapDex.CannotRespectMinimumRequested: 
```
I'm still looking into this one since there are some variables I haven't checked yet.


# Other notes:

## Pica Asset ID
It is not possible to use PICA as an asset ID (ID 1). It of course makes sense since it isn't a stable coin but @kollegian experienced the same issue while working on the **constantProductDex**. Which afaik should allow trading with PICA.
The issue appears on `.addLiquidity` and is the following:
```
Error: balances.ExistentialDeposit: Value too low to create account due to existential deposit
```

## Wallet Refactor
In the beginning, I had all transactions be made by the same wallet. To make sure the issues were not due to this, I'm now using different wallets to trade, provide liquidity, and create the pool. This is making the tests quite slow since I need to mint a lot of assets, but it's safer to be more accurate.


## `Priority too low`
I observed sometimes problems with my "fix" for transacting many extrinsics at the same time by the same wallet.
This was implemented when the parallelization was implemented.
It generally works, but sometimes the transaction which was blocked due to the same nonce gets pushed many, many blocks into the future. Which could lead to timeouts.


# Mochawesome Report
image

Dom-Roth:
Waits for the Pablo refactoring!

PR #702, opened 2022-03-02
feat(tests): custom precision for proptests to avoid edge cases

hussein-aitlahcen:
The issue with proptest is that because of the fact that we generate random
values, we might have some issue because of division. The issue rely on the fact
that for instane, integer division might result in less/more value than expected (odd number
for instance). We can circumvent thoses cases (pretty rare but still making the
CI fail) by lowering the precision of the expected value for some tests.


andor0 (approved):

cor (approved):
LGTM

PR #703, opened 2022-03-02
feat(ci): deploy devnet from parity image

hussein-aitlahcen:


.github/workflows/devnet-release.yml
```diff
@@ -3,56 +3,56 @@ name: "Devnet Community Release"
...
-    - uses: actions/setup-python@v2
-
-    - uses: cachix/install-nix-action@v16
+    - name: Install Dependencies
+      run: |
+        nix-channel --add https://nixos.org/channels/nixpkgs-unstable nixpkgs
+        nix-channel --update
+        nix-env -iA nixpkgs.python3 nixpkgs.rustup nixpkgs.sccache nixpkgs.jq nixpkgs.google-cloud-sdk nixpkgs.nix nixpkgs.neofetch
+        nix --version
+        rustup install stable
```
dzmitry-lahoda, line 29: there are nix packages supporting stable and nightly as it seems. why not nix? ok. just may be leave comment why not.

hussein-aitlahcen: Because I don't want to pin stable

PR #705, opened 2022-03-03
[cu-1ubuznb] Vesting Benchmarks

vimukthi-git:
This PR adds benchmark tests to Vesting pallet according to the [guide](https://github.com/ComposableFi/composable/blob/main/docs/benchmarking.md).


ComposableFiB0t:
Task linked: [CU-1ubuznb Vesting pallet benchmarking](https://app.clickup.com/t/1ubuznb)

PoisonPhang:
You'll want to format with nightly to pass linter checks
`cargo +nightly fmt`

hussein-aitlahcen:
> You'll want to format with nightly to pass linter checks `cargo +nightly fmt`

We have the `make style` command for this :)

hussein-aitlahcen:
We probably need to add this to the `dali` runtime as well as the vesting pallet is already used here :)

vimukthi-git:
@hussein-aitlahcen Sure I will try that. Right now I'm trying to fix an incompatibility between the mock `Runtime` and the benchmark tests for `VestedTransferOrigin` type.

frame/vesting/src/lib.rs
```diff
@@ -56,6 +56,8 @@ use sp_runtime::{
 };
 use sp_std::{convert::TryInto, vec::Vec};
 
+#[cfg(feature = "runtime-benchmarks")]
```
dzmitry-lahoda, line 59: i would prefer bench run as tests too

frame/vesting/src/mock.rs
```diff
@@ -75,21 +78,33 @@ impl frame_system::Config for Runtime {
 	type MaxConsumers = frame_support::traits::ConstU32<16>;
 }
 
+fn benchmark_vested_transfer_acc() -> AccountId {
```
dzmitry-lahoda, line 81: i would prefer _account

hussein-aitlahcen:
@vimukthi-git, I've added the label `needs-benchmarks` to regenerate the benchmarks from the canonical CI machine, we probably need to rebase to keep the new file only and skip the other so that you don't have to deal with conflicts.

hussein-aitlahcen (approved):

dzmitry-lahoda (dismissed):

dzmitry-lahoda (approved):

hussein-aitlahcen (approved):

PR #706, opened 2022-03-03
Amp can be zero

vivekvpandya:


frame/curve-amm/src/tests.rs
```diff
@@ -42,81 +44,86 @@ fn create_pool(
...
-		.expect("impossible; qed;");
+fn test_dex_demo(amplification_factor: u16) {
+	let pool_id = StableSwap::do_create_pool(
+		&ALICE,
+		CurrencyPair::new(USDC, USDT),
+		amplification_factor,
+		Permill::zero(),
+		Permill::zero(),
+	)
+	.expect("impossible; qed;");
```
KaiserKarel, line 55: please make this more descriptive, in case of test failure

KaiserKarel (approved):

dzmitry-lahoda:
compute_base - also it is clear from papers what it returns, but please document what it returns.

frame/curve-amm/src/maths.rs
```diff
@@ -97,30 +100,29 @@ pub fn compute_base(new_quote: u128, amp_coeff: u128, d: u128) -> Result<u128, D
...
-	let b = s.add(&d_ann); // substract d later
-	let mut c = d_ann.mul(&d_n).mul(&d_n);
-	let c = safe_div(&mut c, &mut p)?;
+	let mut c = d_n.clone().mul(&d_n).mul(&d);
+	let term1 = safe_div(&mut c, &mut p)?;
 
 	let mut y = d.clone();
 
-	// y = (y^2 + c) / (2y + b)
+	// y = (y^2 * ann + term1) / (ann * term2 + d)
```
dzmitry-lahoda, line 118: as per paper, can calculate terms once, and use in iteration. as terms are constants. why not here?

frame/curve-amm/src/maths.rs
```diff
@@ -61,7 +61,10 @@ pub fn compute_d(
 
 		// d = (ann * sum + d_p * n) * d / ((ann - 1) * d + (n + 1) * d_p)
 		let mut term1 = ann.clone().mul(&sum).add(&d_p.clone().mul(&n)).mul(&d);
```
dzmitry-lahoda, line 63: numerator

frame/curve-amm/src/maths.rs
```diff
@@ -61,7 +61,10 @@ pub fn compute_d(
...
-		let mut term2 = d.mul(&ann_one).add(&n.clone().add(&one).mul(&d_p));
+		let mut term2 = ann_d
```
dzmitry-lahoda, line 64: denumerator

frame/curve-amm/src/maths.rs
```diff
@@ -61,7 +61,10 @@ pub fn compute_d(
...
+			.add(&n.clone().add(&one).mul(&d_p))
```
dzmitry-lahoda, line 65: where is -1 from formula?

dzmitry-lahoda:
@vivekvpandya oh, sorry. i have accidentally pushed into branch (wanted into branch of branch), force pushed old commit, not sure if that is mergeable now

frame/curve-amm/src/maths.rs
vivekvpandya: need to update comment.if `ann` is 0 `ann - 1` will underflow.

frame/curve-amm/src/maths.rs
vivekvpandya: In paper there is division by `ann` which prevents `ann` to be 0. So I have rewrite formulas so that things work fine even if `ann` is 0. As per that `term1` can be reused but `term2` requires `y` so it must be recomputed. 
See comments above.

dzmitry-lahoda: is model valid is ann = 0? or it degenrates to some constant product? does viper also handles ann zero as we do? could it overflow if ann is 0.0000000000000000001 (close to zero)?

vivekvpandya: As per paper if `amp` is 0 it is equal to constant product. I think viper does not handle this. 
for last question I don't have very confident answer for following equation but to have underflow it needs to satisfy following
`(n + 1) * d_p < d`
it can be simplified for `n = 2` and by substituting equations
` 3 * (x + dx)^2 + 2 (x + dx) (y + dy) + 3 * (y + dy)^2  < 0`
x and y current pool balances ,  dx = changes in X and dy = changes in y.
So it seems it can only be negative if either dx or dy is negative but I think that is not possible.
I have also added proptest for pool with amp = 0. and they passes.
Please let me know if there is issues.

dzmitry-lahoda: non handling zero is feature imho, not a bug

vivekvpandya: sure I agree.

dzmitry-lahoda (approved):

KaiserKarel (approved):

vivekvpandya:
Closing this PR in favor of https://github.com/ComposableFi/composable/pull/725

PR #707, opened 2022-03-03
Validation/call filter

ysfkel:


frame/call-filter/src/lib.rs
```diff
@@ -98,7 +100,7 @@ pub mod pallet {
 		/// Possibly emits a `Disabled` event.
 		#[pallet::weight(T::WeightInfo::disable())]
 		#[transactional]
-		pub fn disable(origin: OriginFor<T>, entry: CallFilterEntryOf<T>) -> DispatchResult {
+		pub fn disable(origin: OriginFor<T>, entry: Validated<CallFilterEntryOf<T>, ValidEntry<T>>) -> DispatchResult {
 			T::UpdateOrigin::ensure_origin(origin)?;
 			ensure!(entry.valid(), Error::<T>::InvalidString);
```
hussein-aitlahcen, line 105: We don't need this anymore right?

frame/call-filter/src/validation.rs
```diff
@@ -0,0 +1,23 @@
...
+use crate::pallet::{Config, CallFilterEntryOf};
+
+#[derive(Debug, Decode, Copy, Clone)]
+pub struct ValidEntry<T>{
+    _marker: PhantomData<T>,
+}
+
+impl<T: Config> Validate<CallFilterEntryOf<T>, ValidEntry<T>> for ValidEntry<T> {
+
+    fn validate(input: CallFilterEntryOf<T>) -> Result<CallFilterEntryOf<T>, &'static str>{
```
dzmitry-lahoda, line 12: no sure why separate rs file is needed for this? may be retain in lib.rs or move into types.rs this and CallFilterEntryOf

dzmitry-lahoda (approved):

PR #723, opened 2022-03-05
Updated bonded-finance docs

PoisonPhang:
Expanded the readme and clarified the language used in the source docs.


hussein-aitlahcen (dismissed):

frame/bonded-finance/README.md
```diff
@@ -1,3 +1,68 @@
-# Overview
+# Bonded Finance
+
+A pallet providing means of submitting and maintaining bond offers.
+
+## Bond Offer Workflow
```
dzmitry-lahoda, line 5: should we have docs in rs and in md? I think right is one place. i do not care what.
also, I should that explanatory should go first - what is value of pallet and what is mechanic. example is good, but if you can do summary - ok. i do not think that details should go first. high level introductory should guide user reading later docs.

PoisonPhang: As for your first suggestion, let me know what you think of these new changes.
As for the second, while I do see where you're coming from, I think there is value to having the readme and top-level cargo docs mirror each other in the case of pallets.

hussein-aitlahcen (dismissed):

hussein-aitlahcen (dismissed):

dzmitry-lahoda:
so we copy docs into readme.md and into lib.rs?

dzmitry-lahoda (approved):

PoisonPhang:
Rebased to implement newer documentation and amended README.md, please re-review

hussein-aitlahcen (dismissed):

frame/bonded-finance/README.md
```diff
@@ -1,3 +1,100 @@
...
+A pallet providing means of submitting and maintaining bond offers.
+
+## Overview
+
+The bonded finance pallet enables users to create bond offers, buy bonds from 
+other users, and cancel existing bond offers via admin intervention. These 
+bonds can be listed with various assets and can offer a different asset as the 
+reward.
+
+One bound offer may contain multiple identically priced bounds. Each bound will
```
benluelo, line 12: bound or bond?

frame/bonded-finance/README.md
```diff
@@ -1,3 +1,100 @@
...
+
+Bond offers can be canceled with the `cancel` extrinsic. This can only be 
+successfully called by the `AdminOrigin`.
+
+Once canceled, the stake and liquidity will be returned to the offer creator. 
+However, this will not cancel currently vested rewards.
+
+## Technical Notes
+
+* This pallet implements the `composable_traits::bonded_finance::BondedFinance`.
```
benluelo, line 99: implements the [...] trait

frame/bonded-finance/src/lib.rs
```diff
@@ -30,26 +30,22 @@
 	unused_extern_crates
 )]
 
-//! Bonded Finance pallet
+//! # Bonded Finance
```
benluelo, line 33: why not just use `#![doc = include_str!("../README.md")]`?

benluelo (changes requested):

frame/bonded-finance/README.md
PoisonPhang: ad593807111cc66cffec3fe50cf921ebedc21179

frame/bonded-finance/README.md
PoisonPhang: ff16c9a96662adeffcae3b528d3098240aa7e5ea

frame/bonded-finance/src/lib.rs
PoisonPhang: Didn't realize this was possible, thanks. 

09c51a49f669938a6716b531fa2e2f71109747af

benluelo (approved):

dzmitry-lahoda (approved):

PR #734, opened 2022-03-08
Refactor/safe arithmentic

ysfkel:


frame/composable-traits/src/math.rs
```diff
@@ -13,32 +13,48 @@ pub fn safe_multiply_by_rational(a: u128, b: u128, c: u128) -> Result<u128, Arit
...
+
+pub trait SafeMul: Sized {
 	fn safe_mul(&self, rhs: &Self) -> Result<Self, ArithmeticError>;
+}
+
+pub trait SafeSub: Sized {
 	fn safe_sub(&self, rhs: &Self) -> Result<Self, ArithmeticError>;
 }
 
-impl<T: CheckedAdd + CheckedMul + CheckedDiv + CheckedSub + Zero> SafeArithmetic for T {
```
hussein-aitlahcen, line 23: Can we still have `SafeArithmetic`? But instead of asking for `CheckedAdd+CheckedSub..`, relying on `SafeAdd+SafeSub...`?

hussein-aitlahcen (changes requested):

frame/composable-traits/src/math.rs
dzmitry-lahoda: @ysfkel can look into MathBalance definition - now it is hack, but on nightly they have type binding to have type A = C + B + D;

dzmitry-lahoda (dismissed):

hussein-aitlahcen (approved):

dzmitry-lahoda (approved):

PR #743, opened 2022-03-09
Exchange in reverse direction in dex_router

vivekvpandya:


frame/dex-router/src/benchmarking.rs
```diff
@@ -0,0 +1,135 @@
...
+	let dex_route = vec![
+			DexRouteNode::Uniswap(0_u32.into()),
+			DexRouteNode::Curve(0_u32.into()),
+	];
+	let eth: <T as pallet_dex_router::Config>::AssetId = 102_u128.into();
+	let usdt: <T as pallet_dex_router::Config>::AssetId = 100_u128.into();
+	let currency_pair = CurrencyPair::new(eth, usdt);
+  } : _(RawOrigin::Signed(owner), currency_pair, Some(dex_route.clone().try_into().unwrap()))
+
+  exchange {
```
vimukthi-git, line 62: Just for my understanding, is there a reason why this test does not have any execution range parameters eg: like here https://github.com/paritytech/substrate/blob/master/frame/examples/basic/src/benchmarking.rs#L40?

vivekvpandya: I am not sure if that is required in general for benchmarks.
```
 14 // Executed Command:                                                                                                                                                       │  process didn't exit successfully: `/home/vivek/dev/        │
 13 // ./target/release/composable                                                                                                                                             │composable/target/debug/build/dali-runtime-acfdbd85040f8d9e/ │
 12 // benchmark                                                                                                                                                               │build-script-build` (exit status: 1)                         │
 11 // --chain=dali-dev                                                                                                                                                        │  --- stdout                                                 │
 10 // --execution=wasm                                                                                                                                                        │  ^[[1;33mInformation that should be included in a bug   @@@ │
  9 // --wasm-execution=compiled                                                                                                                                               └─────────────────────────────────────────────────────────────┘
  8 // --pallet=*                                                                                                                                                                                                                             
  7 // --extrinsic=*                                                                                                                                                           ┌─────────────────────────────────────────────────────────────┐
  6 // --steps=50                                                                                                                                                              │rust-analyzer failed to run build scripts:                   │
  5 // --repeat=20                                                                                                                                                             │error: failed to run custom build command for `dali-runtime  │
  4 // --raw                                                                                                                                                                   │v0.1.0 (/home/vivek/dev/composable/runtime/dali)`            │
  3 // --output=runtime/dali/src/weights                                                                                                                                       │                                                             │
  2 // --log                                                                                                                                                                   │Caused by:                                                   │
  1 // error  
  ```
In above command we specify how many times we want to repeat , and that should be enough right?

Comment (vimukthi-git) [reply frame/dex-router/src/benchmarking.rs:62]: According to my understanding the repeat serves a different purpose than what the execution range does. Execution range allows varying of complexity in designated steps for the extrinsic to estimate the correct weights as the range value increases. For example if there is complexity arising from the storage increasing with the number of calls to the function(even if the call is not taking the execution range as input), then the benchmark needs to simulate that as the goal is to estimate the weights taking into account the worst case. There are some examples of this in democracy pallet (https://github.com/ComposableFi/composable/blob/main/frame/democracy/src/benchmarking.rs#L177).

Comment (vivekvpandya) [reply frame/dex-router/src/benchmarking.rs:58]: I see, thanks for explaining this. IIUC this depends on values of parameters for extrinsic being tested for benchmark.
For dex-router benchmarks IMHO it does not matter much because
1) update_route complex case is to create new route so that is being tested for other operations like delete_route and update details of existing route complexity can be not be worst than inserting the new route.
2) for exchange and buy complexity should not change based on what amount is being exchanged.

Comment (vimukthi-git) [reply frame/dex-router/src/benchmarking.rs:58]: I see, thanks for the context
For 1. It seems the complexity also depends on the parameter passed in as the `dex_route` vector? i.e when it increases the function would take more time to execute(in reference to for loop in `do_update_route`). May be it does not affect much as the `T::MaxHopsInRoute` can't be that large anyway 
For 2. The complexity seems to depend on the number of `DexRouteNode`s returned by `get_route` for the given asset pair, which again seems to depend on `T::MaxHopsInRoute`.  But as you said it does not depend on the amount exchanged.

File: frame/composable-traits/src/dex.rs [hunk 2]
```diff
@@ -168,22 +168,25 @@ pub trait DexRouter<AccountId, AssetId, PoolId, Balance, MaxHops> {
...
+	/// Exchange `amount` of `quote` asset of `asset_pair` with associated route.
 	fn exchange(
 		who: &AccountId,
 		asset_pair: CurrencyPair<AssetId>,
-		dx: Balance,
+		amount: Balance,
+		min_receive: Balance,
 	) -> Result<Balance, DispatchError>;
-	/// Sell `amount` of `base` asset of asset_pair with associated route.
+	/// Sell `amount` of `quote` asset of asset_pair with associated route.
```
dzmitry-lahoda, line 178: it was right. base usually afaik is something like unsalbe BTC. while quote usually USD. so you sell 100 BTC for >=10000 USD 
buy 100 BTC for <= 10000 USD in quite

frame/composable-traits/src/dex.rs
```diff
@@ -168,22 +168,25 @@ pub trait DexRouter<AccountId, AssetId, PoolId, Balance, MaxHops> {
 	) -> Result<(), DispatchError>;
 	/// If route exist return `Some(Vec<PoolId>)`, else `None`.
 	fn get_route(asset_pair: CurrencyPair<AssetId>) -> Option<Vec<DexRouteNode<PoolId>>>;
-	/// Exchange `dx` of `base` asset of `asset_pair` with associated route.
+	/// Exchange `amount` of `quote` asset of `asset_pair` with associated route.
 	fn exchange(
```
dzmitry-lahoda, line 172: will check why it is needed

frame/dex-router/src/lib.rs
```diff
@@ -127,7 +146,86 @@ pub mod pallet {
 	}
 
 	#[pallet::call]
-	impl<T: Config> Pallet<T> {}
+	impl<T: Config> Pallet<T> {
+		/// Create, update or remove route.
+		/// On successful emits one of `RouteAdded`, `RouteUpdated` or `RouteDeleted`.
+		#[pallet::weight(T::WeightInfo::update_route())]
+		pub fn update_route(
```
dzmitry-lahoda, line 153: why not 3 different method like add update delete?

frame/dex-router/src/lib.rs
```diff
@@ -214,19 +313,44 @@ pub mod pallet {
...
-			dx: T::Balance,
+			amount: T::Balance,
+			min_receive: T::Balance,
 		) -> Result<T::Balance, DispatchError> {
-			let route = Self::get_route(asset_pair).ok_or(Error::<T>::NoRouteFound)?;
-			let mut dx_t = dx;
+			let route;
+			let reverse;
+			let res = Self::get_route(asset_pair);
+			if res.is_some() {
```
dzmitry-lahoda, line 326: if let some pattern ?

frame/dex-router/src/lib.rs
```diff
@@ -251,29 +378,63 @@ pub mod pallet {
...
+			min_receive: T::Balance,
 		) -> Result<T::Balance, DispatchError> {
-			let route = Self::get_route(asset_pair).ok_or(Error::<T>::NoRouteFound)?;
+			// let route = Self::get_route(asset_pair).ok_or(Error::<T>::NoRouteFound)?;
+			let route;
+			let reverse;
+			let res = Self::get_route(asset_pair);
+			if res.is_some() {
+				reverse = false;
+				route = Self::get_route(asset_pair).ok_or(Error::<T>::NoRouteFound)?;
```
dzmitry-lahoda, line 414: copy paste?

frame/dex-router/src/lib.rs
```diff
@@ -214,19 +313,44 @@ pub mod pallet {
...
+			} else {
+				route = Self::get_route(asset_pair.swap()).ok_or(Error::<T>::NoRouteFound)?;
+				reverse = true;
+			}
+			// let route =
```
dzmitry-lahoda, line 333: remove commented out text or comment why it is commented out

frame/dex-router/src/lib.rs
dzmitry-lahoda: so that 3 different events are raised

PR #746, opened 2022-03-09
[CH] Add margin to user acc

0xangelo:
This PR implements the ability for a user to add margin to its account


kevin-composable:
The logic looks great. Will the user be able to call add_margin where the asset is the parachain's native token (e.g. PICA)?

frame/clearing-house/src/lib.rs
```diff
@@ -153,7 +186,21 @@ pub mod pallet {
...
-	impl<T: Config> Pallet<T> {}
+	impl<T: Config> Pallet<T> {
+		#[allow(unused_variables)]
+		pub fn do_add_margin(
+			acc: &T::AccountId,
+			asset: AssetIdOf<T>,
+			amount: T::Balance,
+		) -> Result<(), DispatchError> {
+			if !CollateralTypes::<T>::contains_key(asset) {
+				return Err(Error::<T>::UnsupportedCollateralType.into())
```
Cardosaum, line 197: What do you think sending more informative messages in case of errors?

Something similar to Drift's [`print_error` macro](https://github.com/drift-labs/protocol-v1/blob/master/programs/clearing_house/src/error.rs).

The idea behind `ErrorCode` enum would be good too, I guess.
Thoughs?

0xangelo:
> The logic looks great. Will the user be able to call add_margin where the asset is the parachain's native token (e.g. PICA)?

@kevin-composable for now we'll restrict it to stablecoins, but future support for non-stable assets is possible using an oracle and a risk factor (weighting).
I'm gonna use a more general trait nevertheless to allow transfers of any token (like pallet-asset does)

kevin-composable:
@0xangelo Ok nice that is what I was going to suggest. Looking back I think you can make two changes:

change 
    if !CollateralTypes::<T>::contains_key(asset) {
        return Err(Error::<T>::UnsupportedCollateralType.into())
    }
to
    ensure!(CollateralTypes::<T>::contains_key(asset), Error::<T>::UnsupportedCollateralType)
I think this is a little more readable. 

Also, you might be planning to do this already, but you should have a trait (probably called clearing-house) that your pallet implements and provide a method for each of your extrinsics inside the trait.

frame/clearing-house/src/lib.rs
0xangelo: That sounds good, although I don't know if there's something analogous to that in Substrate. The [docs](https://docs.substrate.io/v3/runtime/events-and-errors/#errors) don't say much. But thanks for the input

0xangelo:
> > @kevin-composable for now we'll restrict it to stablecoins, but future support for non-stable assets is possible using an oracle and a risk factor (weighting).
> > I'm gonna use a more general trait nevertheless to allow transfers of any token (like pallet-asset does)
> 
> @0xangelo Ok nice that is what I was going to suggest. Looking back I think you can make two changes:
> 
> change
> 
> ```
>     if !CollateralTypes::<T>::contains_key(asset) {
>         return Err(Error::<T>::UnsupportedCollateralType.into())
>     }
> ```
> 
> to
> 
> ```
>     ensure!(CollateralTypes::<T>::contains_key(asset), Error::<T>::UnsupportedCollateralType)
> ```
> 
> I think this is a little more readable.
> 
> Also, you might be planning to do this already, but you should have a trait (probably called clearing-house) that your pallet implements and provide a method for each of your extrinsics inside the trait.

Yeah, good catch with the `ensure!` macro. I was just reading the [docs](https://docs.substrate.io/v3/runtime/events-and-errors/#errors) and realized I should have used that.

About the trait, should that be added under `frame/composable-traits/clearing_house.rs` for instance?

kevin-composable:
> > > @kevin-composable for now we'll restrict it to stablecoins, but future support for non-stable assets is possible using an oracle and a risk factor (weighting).
> > > I'm gonna use a more general trait nevertheless to allow transfers of any token (like pallet-asset does)
> > 
> > 
> > @0xangelo Ok nice that is what I was going to suggest. Looking back I think you can make two changes:
> > change
> > ```
> >     if !CollateralTypes::<T>::contains_key(asset) {
> >         return Err(Error::<T>::UnsupportedCollateralType.into())
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to
> > ```
> >     ensure!(CollateralTypes::<T>::contains_key(asset), Error::<T>::UnsupportedCollateralType)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think this is a little more readable.
> > Also, you might be planning to do this already, but you should have a trait (probably called clearing-house) that your pallet implements and provide a method for each of your extrinsics inside the trait.
> 
> Yeah, good catch with the `ensure!` macro. I was just reading the [docs](https://docs.substrate.io/v3/runtime/events-and-errors/#errors) and realized I should have used that.
> 
> About the trait, should that be added under `frame/composable-traits/clearing_house.rs` for instance?

Yeah I would keep it in `frame/composable-traits/clearing_house.rs`. Just make sure to include 'pub mod clearing_house;' inside of `frame/composable-traits/lib.rs`

frame/clearing-house/Cargo.toml
```diff
@@ -30,6 +30,13 @@ scale-info = { version = "1.0", default-features = false, features = [
 ] }
 serde = { version = '1', optional = true }
 
+[dev-dependencies]
+orml-tokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library", rev = "2b1c9fb367ccb8e13601b2da43d1c5d9737b93c6", default-features = true }
```
0xangelo, line 34: I turned `default-features = true` here since without it I was running into the following error via `cargo check -p clearing-house --all-targets`
```zsh
error: `Tokens` does not have the std feature enabled, this will cause the `orml_tokens::GenesisConfig` type to be undefined.
  --> frame/clearing-house/src/mock/runtime.rs:22:1
   |
22 | / frame_support::construct_runtime!(
23 | |     pub enum Runtime where
24 | |         Block = Block,
25 | |         NodeBlock = Block,
...  |
34 | |     }
35 | | );
   | |_^
   |
   = note: this error originates in the macro `orml_tokens::__substrate_genesis_config_check::is_std_enabled_for_genesis` (in Nightly builds, run with -Z macro-backtrace for more info)
```
Not sure this is the best solution though, as I've seem other pallets use the `orml_tokens` pallet in their mocks without turning this on (e.g., the mocks for `frame/dutch-auction`)

frame/clearing-house/src/lib.rs
```diff
@@ -101,6 +115,10 @@ pub mod pallet {
 	/// Minimum margin ratio, below which liquidations can occur
 	type MaintenanceMarginRatio<T: Config> = StorageValue<_, T::Decimal, ValueQuery>;
 
+	#[pallet::storage]
+	/// Supported collateral asset ids
+	type CollateralTypes<T: Config> = StorageMap<_, Twox64Concat, AssetIdOf<T>, ()>;
+
```
0xangelo, line 121: Btw, this 'hack' for having a set of allowable collateral types resulted from [this](https://substrate.stackexchange.com/q/685/587) discussion

frame/clearing-house/src/lib.rs
```diff
@@ -126,28 +144,84 @@ pub mod pallet {
...
+			asset: Self::AssetId,
+			amount: Self::Balance,
+		) -> Result<(), DispatchError> {
+			ensure!(
+				CollateralTypes::<T>::contains_key(asset),
+				Error::<T>::UnsupportedCollateralType
+			);
+
+			// Assuming stablecoin collateral and all markets quoted in dollars
+			T::Assets::transfer(asset, account, &T::PalletId::get().into_account(), amount, true)?;
```
0xangelo, line 215: Open to suggestions about whether or not to set the `keep_alive` option to `true`

frame/clearing-house/src/mock/governance_registry.rs
```diff
@@ -0,0 +1,17 @@
+use composable_traits::governance::SignedRawOrigin;
```
0xangelo, line 1: Basically copied from [dutch-auction](https://github.com/ComposableFi/composable/blob/48742cf69b2d73425e5ff45a3f63c4ae37f2d9b9/frame/dutch-auction/src/mock/governance_registry.rs)

frame/clearing-house/src/lib.rs
```diff
@@ -52,6 +56,16 @@ pub mod pallet {
 		/// The virtual AMM ID type for this pallet. `pallet-virtual-amm` should implement a trait
 		/// VAMM with an associated type 'VAMMId' compatible with this one.
 		type VAMMId: FullCodec + MaxEncodedLen + TypeInfo;
+		/// Pallet implementation of asset transfers.
+		type Assets: Transfer<
+			Self::AccountId,
+			Balance = Self::Balance,
+			AssetId = Self::MayBeAssetId,
+		>;
```
0xangelo, line 64: This trait bound is confusing by I've seen it implemented like this elsewhere. `Transfer` itself is actually bounded by the `Inspect` trait from the same crate, but the compiler doesn't complain that I don't make that explicit here

0xangelo (commented):
Proper tests and docs shall be coming after this

frame/clearing-house/src/mock/accounts.rs
```diff
@@ -0,0 +1,4 @@
+pub type AccountId = u64;
+
+pub const ADMIN: AccountId = 0;
```
dzmitry-lahoda, line 3: ALICE = 0
BOB = 1
etc
usually

frame/clearing-house/src/lib.rs
```diff
@@ -93,14 +113,20 @@ pub mod pallet {
...
 
 	#[pallet::storage]
 	#[pallet::getter(fn get_maintenance_margin_ratio)]
+	#[allow(clippy::disallowed_types)]
 	/// Minimum margin ratio, below which liquidations can occur
 	type MaintenanceMarginRatio<T: Config> = StorageValue<_, T::Decimal, ValueQuery>;
 
+	#[pallet::storage]
+	/// Supported collateral asset ids
+	pub type CollateralTypes<T: Config> = StorageMap<_, Twox64Concat, AssetIdOf<T>, ()>;
```
dzmitry-lahoda, line 128: you could build permissioned core, and build on top of that permissioned list. not sure why doing permissions as part of this pallet. same of our Pablo(dexes). Core can always be permissionless, but router on top can give permissions. cc @hussein-aitlahcen

frame/clearing-house/Cargo.toml
dzmitry-lahoda: add std and runtime-benchamarks for each pallet you add and default features false

frame/clearing-house/src/lib.rs
```diff
@@ -8,6 +8,8 @@ mod mock;
 #[cfg(test)]
 mod tests;
 
+mod weights;
```
dzmitry-lahoda, line 11: i doubt i will be able to review this work until lib.rs or README.md added with documentation or reference to it.

frame/clearing-house/src/lib.rs
0xangelo: You mean routing the transfer of assets from trader to clearing house through another pallet that checks if the collateral type can be deposited? That sounds interesting.

Cardosaum (approved):
Really good! :rocket:

frame/clearing-house/diagrams/add-margin.plantuml
```diff
@@ -0,0 +1,19 @@
...
+CH -> CH : assert collateral is supported
+
+CH -> Acc : transfer collateral to pallet account
+
+CH -> CHS : get/create margin account
+return
+
+CH -> CHS : update margin amount
+
+CH ->o? : emit event
```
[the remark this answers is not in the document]
0xangelo: Yeah, adding margin using an unsupported collateral type would result in an `Error::<T>::UnsupportedCollateral`. Not sure how to represent that in PlantUML right now, as that would 'cancel' the rest of the sequence diagram.

The response in the successful case would be the emission of the `MarginAdded` event. I suppose I can specify that on line 18

frame/composable-traits/src/clearing_house.rs
```diff
@@ -0,0 +1,23 @@
...
+	type AssetId;
+	/// The balance type for an account
+	type Balance;
+
+	/// Add margin to a user's account
+	///
+	/// Assumes margin account is unique to each wallet address, i.e., there's only one margin
+	/// account per user.
+	fn add_margin(
+		acc: &Self::AccountId,
```
kevin-composable, line 19: This entirely comes down to personal opinion, but I would use the full name 'account' over 'acc'.

kevin-composable (approved):

kevin-composable:
Everything looked great

frame/clearing-house/diagrams/add-margin.plantuml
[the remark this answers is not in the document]
0xangelo: I'll leave diagram changes to be made directly in the `escher` branch, instead of clogging up the PRs

PR #751, opened 2022-03-11
[CH] Add commit-msg hook

0xangelo:
This adds a commit linter via a `pre-commit` hook at the `commit-msg` stage. Install it with
```zsh
pre-commit install -t commit-msg
```
The [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) commit message standard is used.

### DIFF
--- a/frame/clearing-house/.pre-commit-config.yaml
+++ b/frame/clearing-house/.pre-commit-config.yaml
@@ -6,3 +6,9 @@ repos:
     entry: rustfmt --config skip_children=true --check
     language: system
     types: [rust]
+- repo: https://github.com/alessandrojcm/commitlint-pre-commit-hook
+  rev: v8.0.0
+  hooks:
+  - id: commitlint
+    stages: [commit-msg]
+    additional_dependencies: ['@commitlint/config-conventional']


frame/clearing-house/.pre-commit-config.yaml
```diff
@@ -6,3 +6,9 @@ repos:
     entry: rustfmt --config skip_children=true --check
     language: system
     types: [rust]
+- repo: https://github.com/alessandrojcm/commitlint-pre-commit-hook
+  rev: v8.0.0
+  hooks:
+  - id: commitlint
+    stages: [commit-msg]
+    additional_dependencies: ['@commitlint/config-conventional']
```
Cardosaum, line 14: Is it required to have internet access every time one would commit?
(How does this remote repo reference works?)

Cardosaum:
I'd like to suggest to add a new entry on Development Guidelines for Escher explaining that we are supposed to use Conventional Commits. Doing that we assure every new contributor knows the expected dev workflow.

0xangelo:
> I'd like to suggest to add a new entry on Development Guidelines for Escher explaining that we are supposed to use Conventional Commits. Doing that we assure every new contributor knows the expected dev workflow.

Sure thing! I'll update it in ClickUp after this

Cardosaum (approved):
The idea looks really good, thanks!

frame/clearing-house/.pre-commit-config.yaml
Cardosaum: Neat. Thanks for the info.

PR #752, opened 2022-03-11
Cleanup `Cargo.toml` files & add taplo step in CI

hussein-aitlahcen:
- cleanup `dev-dependencies` of most pallets, they don't require `default-features = false`
- format `.toml` files using [taplo](https://taplo.tamasfe.dev/)
- introduce a new taplo step in CI linters


benluelo (dismissed):

.taplo.toml
```diff
@@ -0,0 +1,30 @@
+include = ["Cargo.toml", "**/*.toml"]
+
+[formatting]
+# Align consecutive entries vertically.
+align_entries = true
```
cor, line 5: I'm not sure if I like this, as the reformating in order to preserve this property when adding new entries affects a lot of lines, making git diffs larger (so harder to see what changed)

benluelo: Personally, I think it's an OK trade-off for the readability benefits

hussein-aitlahcen: I think the giant diff is a bigger issue than having some part of the files aligned, I just used a default config and reverted it not to apply aligning.

cor (approved):
LGTM. Let's merge this, and if we observe that a formatting setting causes unwieldy diffs, then we can always revert that specific setting

benluelo (approved):

PR #767, opened 2022-03-15
CU-23v2xj9 Add code and tests for StableSwap pallet in Pablo.

vivekvpandya:
Add new PoolInitConfiguration enum in Pablo to be used as input for create() extrinsic.
Remove redundant pallet_balances from curve_amm and uniswap_v2 mocks.


hussein-aitlahcen:
Can we move the graphs under a `plot` directory? Probably `plot/stable, plot/constantproduct...`

frame/pablo/src/stable_swap.rs
```diff
@@ -0,0 +1,240 @@
...
+	traits::{CheckedAdd, Convert, One, Zero},
+	ArithmeticError, DispatchError, Permill,
+};
+
+use sp_std::ops::Mul;
+
+pub(crate) struct StableSwap<T>(PhantomData<T>);
+
+impl<T: Config> StableSwap<T> {
+	pub fn do_create_pool(
```
vimukthi-git, line 25: Should these functions be
```suggestion
        [#transactional]
	pub fn do_create_pool(
```
? Just asking, because I'm not sure about the behavior of this annotation in case of chained calls to functions that modify storage

frame/pablo/src/stable_swap.rs
```diff
@@ -0,0 +1,240 @@
...
+						protocol_fee,
+					}),
+				);
+				*pool_count = pool_id.safe_add(&T::PoolId::one())?;
+				Ok(pool_id)
+			})?;
+
+		Ok(pool_id)
+	}
+	fn get_invariant(
```
vimukthi-git, line 61: Nit
```suggestion

	fn get_invariant(
```
Is it not something part of the style?

frame/pablo/src/lib.rs
```diff
@@ -299,40 +346,302 @@ pub mod pallet {
...
+				(lp_fee, protocol_fee)
+			} else {
+				(T::Balance::zero(), T::Balance::zero())
+			};
+
+			let base_amount_excluding_fees = base_amount.safe_sub(&lp_fee)?;
+			Ok((base_amount_excluding_fees, quote_amount, lp_fee, protocol_fee))
+		}
+	}
+	impl<T: Config> Amm for Pallet<T> {
```
vimukthi-git, line 430: Nit
```suggestion

	impl<T: Config> Amm for Pallet<T> {
```

frame/pablo/src/stable_swap.rs
vimukthi-git: :+1: for the rename though

frame/pablo/src/lib.rs
```diff
@@ -299,40 +346,302 @@ pub mod pallet {
...
+			pool_id: T::PoolId,
+		) -> Result<PoolConfigurationOf<T>, DispatchError> {
 			Pools::<T>::get(pool_id).ok_or_else(|| Error::<T>::PoolNotFound.into())
 		}
 
 		pub(crate) fn account_id(pool_id: &T::PoolId) -> T::AccountId {
 			T::PalletId::get().into_sub_account(pool_id)
 		}
+
+		fn do_compute_swap(
```
vimukthi-git, line 404: Seems this belongs in the `stableswap.rs`?

frame/pablo/src/stable_swap.rs
vivekvpandya: I run cargo fmt, but I think that does not introduce new lines.

frame/pablo/src/lib.rs
vivekvpandya: it can be used across all DEXs, but can be specialized if all DEXs does not want to follow same maths for fees. Like in this case protocol_fee is % of lp_fee. (same as curve-amm). If we want something different in uniswap-v2 (like protocol_fee + lp_fee) is total fee then in future PR we can refactor this to have different specialized versions.

frame/pablo/src/stable_swap.rs
hussein-aitlahcen: As stated, it is transactional, meaning that it save the current storage state, execute the function and rollback if Err is returned

vivekvpandya: do_create_pool() in pablo/src/lib.rs has #[transactional] which calls this function so I think suggested change is not required.

vimukthi-git: @hussein-aitlahcen storage transaction part was clear, just that I was curious what the behavior was when there is a chained call. 
@vivekvpandya seems like the way it should work, but I wasn't sure.

vimukthi-git (approved):
LGTM

frame/pablo/src/stable_swap.rs
vimukthi-git: I see. May be you can also try `make style` to see if it works.

hussein-aitlahcen (approved):

PR #770, opened 2022-03-15
[cu-23v2xkb][Pablo] ConstantProductPool Integration

vimukthi-git:
- Integrate constant product pool dex(Uniswap) in to pallet-pablo

### DIFF
--- a/frame/pablo/src/lib.rs
+++ b/frame/pablo/src/lib.rs
@@ -50,7 +50,7 @@ pub mod pallet {
 	use crate::{stable_swap::StableSwap, uniswap::Uniswap, PoolConfiguration::ConstantProduct};
 	use codec::{Codec, FullCodec};
 	use composable_traits::{
-		currency::{CurrencyFactory, LocalAssets},
+		currency::CurrencyFactory,
 		defi::CurrencyPair,
 		dex::{Amm, ConstantProductPoolInfo, StableSwapPoolInfo},
 		math::{SafeAdd, SafeSub},
@@ -93,7 +93,7 @@ pub mod pallet {
 	type BalanceOf<T> = <T as Config>::Balance;
 	type AccountIdOf<T> = <T as frame_system::Config>::AccountId;
 	type PoolIdOf<T> = <T as Config>::PoolId;
-	pub type PoolConfigurationOf<T> =
+	type PoolConfigurationOf<T> =
 		PoolConfiguration<<T as frame_system::Config>::AccountId, <T as Config>::AssetId>;
 	type PoolInitConfigurationOf<T> = PoolInitConfiguration<<T as Config>::AssetId>;
 
@@ -450,10 +450,6 @@ pub mod pallet {
 			let base_amount = Self::get_exchange_value(pool_id, pair.base, quote_amount)?;
 			let base_amount_u: u128 = T::Convert::convert(base_amount);
 
-			// https://uniswap.org/whitepaper.pdf
-			// 3.2.1
-			// we do not inflate the lp for the owner fees
-			// cut is done before enforcing the invariant
 			let (lp_fee, protocol_fee) = if apply_fees {
 				let lp_fee = fee.mul_floor(base_amount_u);
 				// protocol_fee is computed based on lp_fee
@@ -679,16 +675,17 @@ pub mod pallet {
 					Ok(base_amount_excluding_fees)
 				},
 				ConstantProduct(constant_product_pool_info) => {
-					let (base_amount, quote_amount, lp_fees, owner_fees) = Self::do_compute_swap(
-						pool_id,
-						pair,
-						quote_amount,
-						true,
-						constant_product_pool_info.fee,
-						constant_product_pool_info.owner_fee,
-					)?;
+					let (base_amount, quote_amount_excluding_fees, lp_fees, owner_fees) =
+						Uniswap::<T>::do_compute_swap(
+							&constant_product_pool_info,
+							pool_account,
+							pair,
+							quote_amount,
+							true,
+						)?;
 					let total_fees = lp_fees.safe_add(&owner_fees)?;
-					let quote_amount_including_fees = quote_amount.safe_add(&total_fees)?;
+					let quote_amount_including_fees =
+						quote_amount_excluding_fees.safe_add(&total_fees)?;
 
 					ensure!(base_amount >= min_receive, Error::<T>::CannotRespectMinimumRequested);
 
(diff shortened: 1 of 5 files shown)


frame/pablo/src/uniswap.rs
```diff
@@ -0,0 +1,340 @@
...
+#[cfg(test)]
+mod tests {
+	use crate::mock::{new_test_ext, ALICE, BOB, BTC, USDT};
+	use composable_traits::{defi::CurrencyPair, dex::Amm};
+	use frame_support::assert_ok;
+	use sp_arithmetic::Permill;
+
+	#[test]
+	fn test() {
+		new_test_ext().execute_with(|| {});
```
vimukthi-git, line 338: TODO

frame/pablo/src/lib.rs
```diff
@@ -56,13 +60,14 @@ pub mod pallet {
...
 		StableSwap(StableSwapPoolInfo<AccountId, AssetId>),
+		ConstantProduct(ConstantProductPoolInfo<AccountId, AssetId>),
 	}
 
 	type AssetIdOf<T> = <T as Config>::AssetId;
 	type BalanceOf<T> = <T as Config>::Balance;
 	type AccountIdOf<T> = <T as frame_system::Config>::AccountId;
 	type PoolIdOf<T> = <T as Config>::PoolId;
-	type PoolOf<T> =
+	pub type PoolConfigurationOf<T> =
```
vimukthi-git, line 70: Minor rename @vivekvpandya

frame/pablo/src/lib.rs
```diff
@@ -324,10 +360,19 @@ pub mod pallet {
 					Self::deposit_event(Event::PoolCreated { pool_id, owner: pool.owner.clone() });
 					Ok(pool_id)
 				},
+				PoolConfiguration::ConstantProduct(constant_product_pool_info) =>
+					Uniswap::<T>::do_create_pool(
```
vimukthi-git, line 364: Uniswap module can be called in the relevant places including the Pallet AMM impl functions

vimukthi-git:
What's left: Need to fix some tests still

frame/pablo/src/lib.rs
```diff
@@ -122,6 +130,28 @@ pub mod pallet {
 			/// Amount of minted lp tokens.
 			mint_amount: T::Balance,
 		},
+		/// Liquidity added into the pool `T::PoolId`.
```
vivekvpandya, line 133: Do we need this change?

frame/pablo/src/lib.rs
```diff
@@ -412,6 +450,10 @@ pub mod pallet {
 			let base_amount = Self::get_exchange_value(pool_id, pair.base, quote_amount)?;
 			let base_amount_u: u128 = T::Convert::convert(base_amount);
 
+			// https://uniswap.org/whitepaper.pdf
```
vivekvpandya, line 453: I think reference to uniswa-paper is not required here. Please also note that if we want to compute fees differently for `Curve` and `Uniswap` we can specialize that part with a new function like `compute_fee()`. If current code is fine then `Uniswap-v2` 's test related to fees should fail and expected values will be required to be adjusted.

frame/pablo/src/lib.rs
```diff
@@ -122,6 +130,28 @@ pub mod pallet {
...
+		LiquidityAddedToLiquidityBootstrappingPool {
+			/// Pool id to which liquidity added.
+			pool_id: T::PoolId,
+			/// Amount of base asset deposited.
+			base_amount: T::Balance,
+			/// Amount of quote asset deposited.
+			quote_amount: T::Balance,
+		},
+		/// Liquidity added into the pool `T::PoolId`.
+		LiquidityAddedToConstantProductPool {
```
vivekvpandya, line 143: LiquidityAdded events for ConstantProduct and StableSwap are same. And LBP can also use that same event without any issue.

frame/pablo/src/lib.rs
```diff
@@ -592,7 +678,53 @@ pub mod pallet {
...
+				ConstantProduct(constant_product_pool_info) => {
+					let (base_amount, quote_amount, lp_fees, owner_fees) = Self::do_compute_swap(
+						pool_id,
+						pair,
+						quote_amount,
+						true,
+						constant_product_pool_info.fee,
+						constant_product_pool_info.owner_fee,
+					)?;
+					let total_fees = lp_fees.safe_add(&owner_fees)?;
```
vivekvpandya, line 690: this is contradicting to do_compute_swap(), as noted if we want we can specialize fee computation method.

frame/pablo/src/lib.rs
```diff
@@ -75,21 +76,28 @@ pub mod pallet {
...
 		StableSwap(StableSwapPoolInfo<AccountId, AssetId>),
+		ConstantProduct(ConstantProductPoolInfo<AccountId, AssetId>),
 	}
 
 	type AssetIdOf<T> = <T as Config>::AssetId;
 	type BalanceOf<T> = <T as Config>::Balance;
 	type AccountIdOf<T> = <T as frame_system::Config>::AccountId;
 	type PoolIdOf<T> = <T as Config>::PoolId;
-	type PoolConfigurationOf<T> =
+	pub type PoolConfigurationOf<T> =
```
vivekvpandya, line 96: Is this `pub` required ?

vimukthi-git: Good catch, it was from an earlier version

frame/pablo/src/lib.rs
vimukthi-git: Idea is to keep it this way till reach cu-23v2y3n

frame/pablo/src/lib.rs
vimukthi-git: I was using this while trying to understand this function if it's the same as what uniswap has, IMO we need to move this function to `stable_swap.rs` as the logic is specific to that.

frame/pablo/src/lib.rs
vimukthi-git: Yes, got a bit confused by `do_compute_swap` function in this file.

frame/pablo/src/lib.rs
vimukthi-git: fixed

frame/pablo/src/lib.rs
vimukthi-git: fixed

vivekvpandya: so basically we want to specialize dp_compute_swap(). Okay then in next PR I will create it in stable_swap.

frame/pablo/src/lib.rs
vivekvpandya: This change does not bring in LBP to pablo. So perhaps we can remove `LiquidityAddedToLiquidityBootstrappingPool` and just have `LiquidityAdded` which is used for both StableSwap and Uniswap

vivekvpandya (dismissed):

vivekvpandya (approved):

Issue #772, opened 2022-03-16
bonded-finance: Dali and Picasso runtime configuration mismatch with technical specification


CodeanIO:
**Rating: None**

# Description
Several types are differently configured then specified in the technical specifications for both the `dali` and `picasso` runtime.

- `BondOfferId` has type `u64` while the specifications allow `u64` or `u128`
- `AdminOrigin` has type `EnsureRoot<AccountId>` instead of `RootOrHalfCouncil`
- `MinReward` has type 10 or 100 times PICA balance instead of `Get<Balance>`
- `Stake` has type 10 PICA balance instead of `Get<Balance>`
- `Duration` is not configurable but is specified in the technical specifications

https://github.com/ComposableFi/composable/blob/35ec6680d07c08c207f8b3b0dc349cc8163000f1/runtime/dali/src/lib.rs#L830-L830

https://github.com/ComposableFi/composable/blob/35ec6680d07c08c207f8b3b0dc349cc8163000f1/runtime/picasso/src/lib.rs#L745-L745

https://github.com/ComposableFi/composable/blob/35ec6680d07c08c207f8b3b0dc349cc8163000f1/runtime/dali/src/lib.rs#L825-L825

https://github.com/ComposableFi/composable/blob/35ec6680d07c08c207f8b3b0dc349cc8163000f1/runtime/dali/src/lib.rs#L831-L831

https://github.com/ComposableFi/composable/blob/35ec6680d07c08c207f8b3b0dc349cc8163000f1/runtime/dali/src/lib.rs#L826-L826

https://github.com/ComposableFi/composable/blob/35ec6680d07c08c207f8b3b0dc349cc8163000f1/runtime/picasso/src/lib.rs#L740-L740

https://github.com/ComposableFi/composable/blob/35ec6680d07c08c207f8b3b0dc349cc8163000f1/runtime/picasso/src/lib.rs#L741-L741

https://github.com/ComposableFi/composable/blob/35ec6680d07c08c207f8b3b0dc349cc8163000f1/runtime/picasso/src/lib.rs#L746-L746

# Impact
There does not appear to be a concrete security-related impact, and some of the types are conceptually the same. Nevertheless, assumptions based on these types might be violated.

# Recommendation
We recommend making sure that the type definitions and technical specification match, in order to avoid confusion.

---

**Disclaimer:**
[...]

Issue #771, opened 2022-03-16
bonded-finance: Missing check for wrapping of BondOfferId


CodeanIO:
**Rating: Low**

# Description
The `do_offer` function in `frame/bonded-finance/src/lib.rs` uses the `.next()` method of `BondOfferId` in order to increment the stored `BondOfferCount`, thereby generating a new, unique ID.

The implementation of `.next()` as defined by the `WrappingNext` trait, specifies that the caller must ensure that the value does not *wrap*. Specifically, `saturating_add` is used, which will repeat the maximum value upon pending overflow.

However, in this case, the result is not checked, at the call-site of `next`.

https://github.com/ComposableFi/composable/blob/35ec6680d07c08c207f8b3b0dc349cc8163000f1/frame/bonded-finance/src/lib.rs#L288-L291

https://github.com/ComposableFi/composable/blob/35ec6680d07c08c207f8b3b0dc349cc8163000f1/frame/composable-traits/src/math.rs#L46-L47

https://github.com/ComposableFi/composable/blob/35ec6680d07c08c207f8b3b0dc349cc8163000f1/frame/bonded-finance/src/lib.rs#L147-L147

# Impact
After generating 2^64 bond offers, this will result in the same `offer_id` value being used for multiple bond offers. This violates some implicit assumptions, and causes assets for multiple bond offers to be combined into one bond offer account.

In practice however, this is an infeasible limit, so the impact is minimal. In case the underlying type is ever changed to a smaller one, this could however become a practical issue.

# Recommendation
We recommend adding a wrapping check at the call-site of `offer_id.next()`.

---

**Disclaimer:**
This vulnerability was found using a combination of code-review and limited penetration testing. The final impact may depend on the larger (production) environment. The above recommendation is included to help mitigate or resolve this issue, but may not necessarily be a complete or final solution.

hussein-aitlahcen:
Fixed by #775

PR #775, opened 2022-03-16
Rework vesting origin & audit fixes

### LINKED ISSUE #773: bonded-finance: Comment (assumption) does not match reality
**Rating: None**

# Description
A comment inside the `do_bond` function in `frame/bonded-finance/src/lib.rs` contains an incorrect description of a previously applied constraint.

The comment indicates that `nb_of_bonds < offer.nb_of_bonds` holds, while in fact, `nb_of_bonds <= offer.nb_of_bonds` holds.

https://github.com/ComposableFi/composable/blob/35ec6680d07c08c207f8b3b0dc349cc8163000f1/frame/bonded-finance/src/lib.rs#L369-L370

https://github.com/ComposableFi/composable/blob/35ec6680d07c08c207f8b3b0dc349cc8163000f1/frame/bonded-finance/src/lib.rs#L322-L322

# Impact
This does not have a security impact, as the error is in a comment.

Nevertheless, the comment could mislead a future developer about the applied constraint.

# Recommendation
We recommend changing the comment to match the previously applied constraint, in order to prevent future confusion.

---

**Disclaimer:**
This vulnerability was found using a combination of code-review and limited penetration testing. The final impact may depend on the larger (production) environment. The above recommendation is included to help mitigate or resolve this issue, but may not necessarily be a complete or final solution.

hussein-aitlahcen:
This PR must fix #773 #772 #771 

I also modified the Vesting pallet not to use the `origin` account as source account for the transfer.
This allow us to define the `RootOrHalfCouncil` as `VesterTransferOrigin` in the runtimes.
Cross-pallet call is still permisionless.

This PR also remove the usage of `self` in `CurrencyId` to allow us to represent amounts without confusion (A unit of any currency was previously defined as `CurrencyId.PICA.unit()` while it is now represented as `CurrencyId::unit()` as all currencies are normalized to the same number of decimals within our system.

### DIFF
--- a/frame/bonded-finance/src/lib.rs
+++ b/frame/bonded-finance/src/lib.rs
@@ -72,7 +72,7 @@ pub mod pallet {
 	use composable_support::validation::Validated;
 	use composable_traits::{
 		bonded_finance::{BondDuration, BondOffer, BondedFinance, ValidBondOffer},
-		math::{SafeAdd},
+		math::SafeAdd,
 		vesting::{VestedTransfer, VestingSchedule, VestingWindow::BlockNumberBased},
 	};
 	use frame_support::{
@@ -87,7 +87,7 @@ pub mod pallet {
 	use scale_info::TypeInfo;
 	use sp_runtime::{
 		helpers_128bit::multiply_by_rational,
-		traits::{AccountIdConversion, BlockNumberProvider, Convert, Zero, One},
+		traits::{AccountIdConversion, BlockNumberProvider, Convert, One, Zero},
 		ArithmeticError,
 	};
 	use sp_std::fmt::Debug;
@@ -156,8 +156,8 @@ pub mod pallet {
 			+ Eq
 			+ Debug
 			+ Zero
-      + SafeAdd
-      + One
+			+ SafeAdd
+			+ One
 			+ FullCodec
 			+ MaxEncodedLen
 			+ TypeInfo;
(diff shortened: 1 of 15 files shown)


frame/crowdloan-rewards/src/lib.rs
```diff
@@ -114,7 +114,7 @@ pub mod pallet {
 	pub enum Error<T> {
 		NotInitialized,
 		AlreadyInitialized,
-		InvalidInitializationBlock,
+		BackToTheFuture,
```
benluelo, line 117: nice

benluelo (approved):

frame/bonded-finance/src/lib.rs
```diff
@@ -312,10 +312,12 @@ pub mod pallet {
...
 		) -> Result<T::BondOfferId, DispatchError> {
-			let offer_id = BondOfferCount::<T>::try_mutate(|offer_id| -> Result<T::BondOfferId, DispatchError> {
-				*offer_id = offer_id.safe_add(&T::BondOfferId::one())?;
-				Ok(*offer_id)
-			})?;
+			let offer_id = BondOfferCount::<T>::try_mutate(
+				|offer_id| -> Result<T::BondOfferId, DispatchError> {
+					*offer_id = offer_id.safe_add(&T::BondOfferId::one())?;
+					Ok(*offer_id)
+				},
```
benluelo, line 319: Perhaps somewhat unrelated to this PR, but we use this "try increment storage" pattern a lot, maybe we could extract it to a helper function in `composable-helpers`?

hussein-aitlahcen: I think it's pretty fair to extract that, yes

vimukthi-git (approved):
LGTM

PR #779, opened 2022-03-17
Fortify mosaic & vesting pallets

hussein-aitlahcen:
In this PR we introduce a new `min_transfer_size` for networks in Mosaic.
We also check that `from != to` when doing a `vested_transfer` in the Vesting pallet.


vimukthi-git (approved):
LGTM

PR #785, opened 2022-03-20
CU-28yy29e Simplify multi_existential_deposits function

andor0:


ComposableFiB0t:
Task linked: [CU-28yy29e Simplify multi_existential_deposits function](https://app.clickup.com/t/28yy29e)

runtime/common/src/lib.rs
```diff
@@ -180,11 +180,8 @@ pub fn multi_existential_deposits(_currency_id: &CurrencyId) -> Balance {
 
 #[cfg(not(feature = "runtime-benchmarks"))]
 pub fn multi_existential_deposits(currency_id: &CurrencyId) -> Balance {
-	match *currency_id {
-		CurrencyId::PICA => NativeExistentialDeposit::get(),
-		id => PriceConverter::get_price_inverse(id, NativeExistentialDeposit::get())
-			.unwrap_or(Balance::MAX), // TODO: here DEX call to pemissioned markets should come
-	}
+	PriceConverter::get_price_inverse(*currency_id, NativeExistentialDeposit::get())
```
hussein-aitlahcen, line 183: This mean that a non priceable asset have max for ED?

dzmitry-lahoda: this means that unknown currency requires infinie ED - same as in acala (actually current API is broken, it should return Option<Balance>, Not balance. 

we talked with @andor0 and he may consider working on what was discussed about having zero native and multi in benches.

hussein-aitlahcen: Sounds good :+1:

dzmitry-lahoda: see max if not known for any currency https://github.com/AcalaNetwork/Acala/blob/0916abbcc520df66239642692b6843faf9107896/runtime/karura/src/lib.rs#L796 
any non pricebale currency is potentially priceable, it is matter of time. but if price ability is not possible now, it should not allow to spam

hussein-aitlahcen (approved):

dzmitry-lahoda (approved):

PR #790, opened 2022-03-21
[cu-23v2y3n] Event publishing refactoring

vimukthi-git:
- Reusing the event publishing code for all three dexes
- Removing TODOs
- Verify some of the events published

I tried to write tests for event publishing before making this change, but I think we need to make a change to the AMM interface to expose things like calculated fees to verify the events

### DIFF
--- a/frame/pablo/src/lib.rs
+++ b/frame/pablo/src/lib.rs
@@ -445,7 +445,7 @@ pub mod pallet {
 					LiquidityBootstrapping::<T>::do_create_pool(validated_pool_config)?
 				},
 			};
-			Self::deposit_event(Event::PoolCreated { owner: who.clone(), pool_id });
+			Self::deposit_event(Event::<T>::PoolCreated { owner: who.clone(), pool_id });
 			Ok(pool_id)
 		}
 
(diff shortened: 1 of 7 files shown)


ComposableFiB0t:
Task linked: [CU-23v2y3n Abstract the events to be agnostic w.r.t the pool type](https://app.clickup.com/t/23v2y3n)

frame/pablo/src/lib.rs
```diff
@@ -583,16 +555,16 @@ pub mod pallet {
...
 						keep_alive,
 					)?;
-					Self::deposit_event(Event::<T>::LiquidityAdded {
-						who: who.clone(),
-						pool_id,
-						base_amount,
-						quote_amount,
-						minted_lp: T::Balance::zero(),
-					});
+					T::Balance::zero()
```
vivekvpandya, line 558: I think it is better to return zero() from LBP's add_liquidity() . It is better to keep logic of LBP hidden from pablo pallet.

vivekvpandya (approved):

hussein-aitlahcen (dismissed):

hussein-aitlahcen (approved):

vivekvpandya (approved):

hussein-aitlahcen (approved):

PR #792, opened 2022-03-22
review runtime versions

seunlanlege:


runtime/picasso/src/lib.rs
```diff
@@ -104,9 +104,9 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	//   runtime in substitute for the on-chain Wasm runtime unless all of `spec_name`,
 	//   `spec_version`, and `authoring_version` are the same between Wasm and native.
 	spec_version: 1000,
```
hussein-aitlahcen, line 106: Don't we increase spec version here?

seunlanlege: picasso spec version is currently 200 on kusama

PR #814, opened 2022-03-24
Escher feature/vAMM Pallet Initialization

Cardosaum:


frame/vamm/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+	// ----------------------------------------------------------------------------------------------------
+	//                                            Runtime Events
+	// ----------------------------------------------------------------------------------------------------
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
+		Test { issuer: T::AccountId },
+
+		LiquidityAdded {},
```
0xangelo, line 42: Can probably remove these two, no?

frame/vamm/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+	// ----------------------------------------------------------------------------------------------------
+	//                                             Pallet Types
+	// ----------------------------------------------------------------------------------------------------
+
+	// ----------------------------------------------------------------------------------------------------
+	//                                            Runtime Events
+	// ----------------------------------------------------------------------------------------------------
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
```
0xangelo, line 38: Doesn't `clippy` complain about `deposit_event` being unused?

frame/vamm/src/mock.rs
```diff
@@ -0,0 +1,125 @@
...
+use sp_core::H256;
+
+pub type BlockNumber = u64;
+pub type AccountId = u128;
+pub type Balance = u128;
+
+pub const ALICE: AccountId = 0;
+
+// ----------------------------------------------------------------------------------------------------
+//                                                Config
```
0xangelo, line 32: Nitpick: since you commented on other config sections with the name of the pallet, wouldn't it make sense to do the same here? E.g., 'FRAME system'. 

I think I'll do this on mine as well

frame/vamm/src/tests.rs
```diff
@@ -0,0 +1,28 @@
...
+
+use crate::mock::{
+    ALICE, Event, ExtBuilder, Vamm, Origin, System
+};
+use crate::pallet;
+
+use frame_support::assert_ok;
+
+#[test]
+fn call_test_extrinsic() {
```
0xangelo, line 11: I'm probably stating the obvious, but this won't work cause there are no extrinsics. Same for `test_extrinsic_emits_event()`

frame/vamm/src/tests.rs
```diff
@@ -0,0 +1,28 @@
...
+    ExtBuilder::default().build().execute_with(|| {
+        assert_ok!(Vamm::test(Origin::signed(ALICE)));
+    });
+}
+
+#[test]
+fn test_extrinsic_emits_event() {
+    ExtBuilder::default().build().execute_with(|| {
+        System::set_block_number(1);
```
0xangelo, line 20: If you want, you can set the block number in `ExtBuilder.build()`

0xangelo (commented):
🚀

frame/vamm/src/lib.rs
Cardosaum: Correct. 
Removed it :+1:

frame/vamm/src/lib.rs
Cardosaum: It does. 
but since I was running `pre-commit` and it didn't show the error I thought it was okay. Running `cargo clippy` as a stand alone command does show errors. 
Fixed, thanks for spotting this!

frame/vamm/src/tests.rs
Cardosaum: Thanks!

frame/vamm/src/tests.rs
Cardosaum: Oh, nice to know.
TY

frame/vamm/src/mock.rs
Cardosaum: Yeah, seems better this way. 
Thanks!

frame/vamm/src/mock.rs
```diff
@@ -0,0 +1,125 @@
+// use super::currency_factory::MockCurrencyId;
```
dzmitry-lahoda, line 1: why this code is commented out?

frame/vamm/src/lib.rs
```diff
@@ -0,0 +1,68 @@
+#[frame_support::pallet]
+pub mod pallet {
+	// ----------------------------------------------------------------------------------------------------
```
dzmitry-lahoda, line 3: @KaiserKarel @hussein-aitlahcen are we ok to diverge from what is accepted in terms of such comments in polkadot ecosystem?

frame/vamm/src/mock.rs
Cardosaum: I commented out just because I wasn't using any of these yet (but I will in the next days).

Is it preferable to just let all the imports there uncommented, regardless if they are being used or not?

Cardosaum:
I'll close this PR since it's outdated and no longer needed, but feel free to comment either way.

PR #827, opened 2022-03-25
[cu-23v352q][Pablo] Benchmarking and Integration to Dali

vimukthi-git:
- Benchmarking pablo pallet
- Integrating the pablo pallet into Dali runtime
- Minor modification to `CreationOrigin`


ComposableFiB0t:
Task linked: [CU-23v352q Add Pablo in Dali once everything is done](https://app.clickup.com/t/23v352q)

frame/pablo/src/benchmarking.rs
```diff
@@ -0,0 +1,275 @@
...
+	}
+}
+
+benchmarks! {
+  where_clause { where T::BlockNumber: From<u32>, T::Balance: From<u128>, T::AssetId: From<u128> }
+	create {
+		  let usdc: T::AssetId = 100.into();
+		  let usdt: T::AssetId = 101.into();
+		  let owner: T::AccountId = whitelisted_caller();
+		  let pair = CurrencyPair::new(usdc, usdt);
```
vivekvpandya, line 57: indentation seems wrong, also interesting that CI has not been triggered.

frame/pablo/src/benchmarking.rs
```diff
@@ -0,0 +1,275 @@
...
+			  initial_usdt,
+			  0.into(),
+			  false
+		  ));
+		  let user = account("user", 0, 0);
+		  assert_ok!(T::Assets::mint_into(usdt, &user, (1000_u128 * unit).into()));
+		  // swap 1000 USDC
+	 }: _(RawOrigin::Signed(user), pool_id, CurrencyPair::new(usdc, usdt), (1000_u128 * unit).into(), 0.into(), false)
+}
+
```
vivekvpandya, line 274: should we add buy/sell/swap for LBP or uniswap?

vimukthi-git: All of them are of `O(1)` complexity in terms of code, Then considering the maximum number of database operations I added stableSwap benchmarks only. For LBP I also added create/add_liquidity/remove_liquidity benchmarks as the code is some operations that are different from the other two dexes.

seunlanlege:
You'll need to bump `impl_version` for picasso and composable too

runtime/dali/src/lib.rs
```diff
@@ -111,7 +111,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	//   `spec_version`, and `authoring_version` are the same between Wasm and native.
 	// This value is set to 100 to notify Polkadot-JS App (https://polkadot.js.org/apps) to use
 	//   the compatible custom types.
-	spec_version: 2000,
+	spec_version: 3000,
```
seunlanlege, line 114: should just bump impl version instead

vimukthi-git: As we are including a new pallet, doesn't that mean we are changing the spec?

vimukthi-git:
@seunlanlege Just for my understanding, since we only including this new pallet for dali runtime, why does the picasso and composable versions need to be updated?

seunlanlege:
> @seunlanlege Just for my understanding, since we are only including this new pallet for dali runtime, why do picasso and composable versions need to be updated?

the reason we check if any changes have been made to the `frame/*` folder whether or not that pallet is included in the runtime, reason is because we can't figure put which runtimes are linked to each pallet.

runtime/dali/src/lib.rs
seunlanlege: well true

Dom-Roth (approved):
LGTM!

vivekvpandya (approved):

Dom-Roth (dismissed):

Dom-Roth (approved):

vivekvpandya (approved):
Simnode is failing due to new spec version but I think it should be fixed saperate.

vimukthi-git:
Awaiting @seunlanlege to fix the simnode before merging this cc @hussein-aitlahcen

dzmitry-lahoda (approved):

hussein-aitlahcen (dismissed):

hussein-aitlahcen:
Removed simnode as still not updated.

hussein-aitlahcen (approved):

vivekvpandya (approved):

PR #829, opened 2022-03-28
IBC onchain and offchain components

Wizdave97:
Closes https://github.com/ComposableFi/composable/issues/1060


bridges/pallet-ibc/src/benchmarking.rs
```diff
@@ -0,0 +1,25 @@
...
+		let mut v1: Vec<u8> = Vec::new();
+		let mut v2: Vec<u8> = Vec::new();
+		for i in 0..m {
+			v1.push(b)
+			v2.push(b)
+		}
+		let any = Any {type_url: v1, value: v2};
+		let caller = whitelisted_caller();
+	}: deliver(RawOrigin::Signed(caller), vec![any], b)
+}
```
seunlanlege, line 23: hmm, not sure this is how we should benchmark this pallet

seunlanlege: we definitely need some sophisticated mock testing

bridges/pallet-ibc/src/channel.rs
```diff
@@ -0,0 +1,744 @@
...
+	timestamp::Timestamp,
+	Height,
+};
+use tendermint_proto::Protobuf;
+
+impl<T: Config> ChannelReader for Context<T> {
+	/// Returns the ChannelEnd for the given `port_id` and `chan_id`.
+	fn channel_end(&self, port_channel_id: &(PortId, ChannelId)) -> Result<ChannelEnd, ICS04Error> {
+		// Todo: Confirm if all the errors are accurate
+		log::info!("in channel : [channel_end]");
```
seunlanlege, line 28: more like `log::trace` if you want to leave it in

bridges/pallet-ibc/src/channel.rs
```diff
@@ -0,0 +1,744 @@
...
+		log::info!(
+			"in channel : [client_consensus_state] >> client_id = {:?}, height = {:?}",
+			client_id,
+			height
+		);
+
+		let height = height.encode_vec().map_err(|_| ICS04Error::implementation_specific())?;
+		let value = <ConsensusStates<T>>::get(client_id.as_bytes());
+
+		for item in value.iter() {
```
seunlanlege, line 135: I feel this is a sign to make `ConsensusStates` a double-map

bridges/pallet-ibc/src/impls.rs
```diff
@@ -0,0 +1,728 @@
...
+}
+
+fn client_type_from_bytes<T: Config>(client_type: Vec<u8>) -> Result<ClientType, Error<T>> {
+	ClientType::from_str(&String::from_utf8(client_type).map_err(|_| Error::<T>::DecodingError)?)
+		.map_err(|_| Error::<T>::DecodingError)
+}
+
+pub fn host_height<T: Config>() -> Result<u64, Error<T>> {
+	let block_number = format!("{:?}", <frame_system::Pallet<T>>::block_number());
+	let current_height = block_number.parse().map_err(|_| Error::<T>::DecodingError)?;
```
seunlanlege, line 726: this looks roundabout, why not cast from u32 to u64?

bridges/pallet-ibc/src/impls.rs
```diff
@@ -0,0 +1,728 @@
...
+				.map_err(|_| Error::<T>::Other)?;
+		ctx.store_packet_result(send_packet_result.result)
+			.map_err(|_| Error::<T>::Other)?;
+
+		// store packet offchain
+		let key = Pallet::<T>::offchain_key(channel_id, port_id);
+		let mut offchain_packets: BTreeMap<u64, OffchainPacketType> =
+			sp_io::offchain::local_storage_get(sp_core::offchain::StorageKind::PERSISTENT, &key)
+				.and_then(|v| codec::Decode::decode(&mut &*v).ok())
+				.unwrap_or_default();
```
seunlanlege, line 691: very sad about lack of complex db apis for offchain worker, where is this function called from? can't find its references anywhere

bridges/pallet-ibc/src/benchmarking.rs
Wizdave97: Yeah, this is just a placeholder benchmark

bridges/pallet-ibc/src/port.rs
```diff
@@ -0,0 +1,43 @@
...
+use ibc::core::{
+	ics05_port::{
+		capabilities::{Capability, CapabilityName},
+		context::{CapabilityReader, PortReader},
+		error::Error as ICS05Error,
+	},
+	ics24_host::identifier::PortId,
+};
+
+// TODO: Revisit when port binding is implemented
```
seunlanlege, line 13: is this an ibc-rs problem?

bridges/pallet-ibc/src/routing.rs
```diff
@@ -0,0 +1,20 @@
...
+	pub _pd: PhantomData<T>,
+}
+
+impl<T: Config> Context<T> {
+	pub fn new() -> Self {
+		Self { _pd: PhantomData::default() }
+	}
+}
+
+impl<T: Config> Ics26Context for Context<T> {}
```
seunlanlege, line 18: seems we're also waiting on ibc-rs here

bridges/pallet-ibc/src/impls.rs
Wizdave97: This is not used yet, it should be theoretically be called by other modules that need to initiate a packet transfer.

bridges/pallet-ibc/src/routing.rs
Wizdave97: Yeah, that's the initial router trait.

bridges/pallet-ibc/src/port.rs
Wizdave97: It's a bit dependent on how we implement the router, reason why I left it out initially.
I think I can do the capability key storage now.

bridges/pallet-ibc/src/impls.rs
seunlanlege: I'm imagining a situation where we have a lot of packets that have to be loaded into memory before we can send packets.

frame/ibc/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+	{
+		fn on_finalize(_n: BlockNumberFor<T>) {
+			let root = Pallet::<T>::extract_ibc_state_root();
+			if let Ok(root) = root {
+				let height = crate::impls::host_height::<T>();
+				let height = ibc::Height::new(0, height);
+				let height = height.encode_vec().unwrap_or_default();
+				let timestamp = T::TimeProvider::now().as_nanos().saturated_into::<u64>();
+				let ibc_cs = IbcConsensusState { timestamp, root: root.clone() };
+				CommitmentRoot::<T>::insert(height, ibc_cs.clone());
```
seunlanlege, line 273: don't think we use this value anywhere

Wizdave97: We'll need it here https://github.com/ComposableFi/composable/pull/829/files#diff-15d5711034ee5ceacf10a76d9f210d95de421a30f7fec36497517888be809135R151 after the consensus state  for  a beefy light client makes its way into ibc-rs, I need a way to prune it though, we only need recent ones, probably dating back a 100 blocks at most.

frame/ibc/src/impls.rs
```diff
@@ -0,0 +1,752 @@
...
+	}
+
+	pub fn generate_raw_proof(keys: Vec<Vec<u8>>) -> Result<Vec<u8>, Error<T>> {
+		let keys = keys.iter().collect::<Vec<_>>();
+		let root = Self::build_ibc_state_root()?;
+		let db = sp_trie::MemoryDB::<BlakeTwo256>::default();
+		sp_trie::generate_trie_proof::<sp_trie::LayoutV0<BlakeTwo256>, _, _, _>(&db, root, keys)
+			.map(|proof| proof.encode())
+			.map_err(|_| Error::<T>::ProofGenerationError)
+	}
```
seunlanlege, line 202: This will unfortunately generate an empty proof, seeing as the db is empty. Anyways, Since this will be called in rpc, i think its fine to have a method that simply returns all the data that was committed at a specific height and do proof generation in native, instead of the runtime.

frame/ibc/src/weights.rs
```diff
@@ -134,4 +134,38 @@ impl<T: frame_system::Config> crate::weight::WeightInfo for WeightInfo<T> {
...
+	// Storage: Ibc Channels (r:1 w:0)
+	// Storage: Ibc Connections (r:1 w:0)
+	// Storage: Ibc PacketCommitment (r:1 w:1)
+	// Storage: Ibc ClientStates (r:1 w:0)
+	// Storage: Ibc ConsensusStates (r:1 w:0)
+	// Storage: Timestamp Now (r:1 w:0)
+	// Storage: Ibc ClientUpdateTime (r:1 w:0)
+	// Storage: Ibc ClientUpdateHeight (r:1 w:0)
+	fn ack_packet(i: u32, j: u32, ) -> Weight {
+		(113_242_000 as Weight)
```
seunlanlege, line 163: goddamn, those in wasm hash functions are showing their weight

frame/ibc/src/weights.rs
```diff
@@ -55,7 +55,7 @@ impl<T: frame_system::Config> crate::weight::WeightInfo for WeightInfo<T> {
 	// Storage: Ibc ClientUpdateTime (r:0 w:1)
 	// Storage: Ibc ClientUpdateHeight (r:0 w:1)
 	fn update_client() -> Weight {
-		(473_000_000 as Weight)
+		(498_000_000 as Weight)
```
seunlanlege, line 58: jesus

Wizdave97: Must be the signature verification in tendermint-rs, and it’s still just two signatures that were used.

seunlanlege:
also please add code comments to the benchmarking and wherever you can outlining the hows and whys

frame/ibc/ibc-runtime-api/src/lib.rs
```diff
@@ -8,6 +8,8 @@ use sp_std::vec::Vec;
 sp_api::decl_runtime_apis! {
 	/// IBC Runtime Apis
 	pub trait IbcRuntimeApi {
+		/// Get parachain id
+		fn para_id() -> u32;
```
seunlanlege, line 12: `cumulus-pallet-parachain-info` already provides this

runtime/composable/src/lib.rs
```diff
@@ -1004,6 +1004,10 @@ impl_runtime_apis! {
 	}
 
 	impl ibc_runtime_api::IbcRuntimeApi<Block> for Runtime {
+		fn para_id() -> u32 {
+			<Runtime as cumulus_pallet_parachain_system::Config>::SelfParaId::get().into()
```
seunlanlege, line 1008: 👍🏿

frame/ibc-ping/src/lib.rs
```diff
@@ -0,0 +1,345 @@
...
+	#[pallet::generate_store(pub(super) trait Store)]
+	#[pallet::without_storage_info]
+	pub struct Pallet<T>(_);
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::weight(0)]
+		pub fn open_channel(origin: OriginFor<T>, params: OpenChannelParams) -> DispatchResult {
+			ensure_root(origin)?;
+			let state = match params.state {
```
blasrodri, line 85: can we define `TryFrom`?

frame/ibc/ibc-trait/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+	/// UTF8 string bytes
+	pub version: Vec<u8>,
+}
+
+pub trait IbcTrait {
+	fn send_packet(data: SendPacketData) -> Result<(), Error>;
+	fn open_channel(port_id: PortId, channel_end: ChannelEnd) -> Result<ChannelId, Error>;
+}
+
+pub trait CallbackWeight {
```
blasrodri, line 37: can we have a bit more doc on this trait and its methods?

frame/ibc/ibc-primitives/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+};
+use scale_info::prelude::string::{String, ToString};
+use sp_std::str::FromStr;
+#[cfg(not(feature = "std"))]
+use sp_std::vec::Vec;
+
+pub struct SendPacketData {
+	pub data: Vec<u8>,
+	pub timeout_height_offset: u64,
+	/// This value should be represent nano seconds
```
Wizdave97, line 18:
```suggestion
	/// This value should represent nano seconds
```

node/src/runtime.rs
```diff
@@ -224,4 +223,22 @@ define_trait! {
...
+		impl for composable_runtime {}
+
+		impl for picasso_runtime {}
+
+		#[cfg(feature = "dali")]
+		impl for dali_runtime {
+			fn (io, deps) {
+				io.merge(IbcRpcHandler::new(deps.client, deps.chain_props).into_rpc())
+			}
+		}
```
seunlanlege, line 242: Is this a new syntax feature?

Wizdave97: Yeah, for adding rpc handlers

frame/ibc/ibc-runtime-api/src/lib.rs
seunlanlege: @Wizdave97

frame/ibc/src/port.rs
```diff
@@ -0,0 +1,33 @@
...
+		error::Error as ICS05Error,
+	},
+	ics24_host::identifier::PortId,
+	ics26_routing::context::ModuleId,
+};
+
+impl<T: Config + Sync + Send> PortReader for Context<T> {
+	fn lookup_module_by_port(&self, port_id: &PortId) -> Result<ModuleId, ICS05Error> {
+		match port_id.as_str() {
+			val if val == pallet_ibc_ping::PORT_ID =>
```
seunlanlege, line 16: why not just match directly to `pallet_ibc_ping::PORT_ID` ?

frame/ibc/ibc-runtime-api/src/lib.rs
Wizdave97: Yeah, but there's no runtime api to get the parachain id, so I'm providing it to the rpc handler through this runtime api

frame/oracle/src/lib.rs
```diff
@@ -73,7 +73,9 @@ pub mod pallet {
 		AccountId32, ArithmeticError, FixedPointNumber, FixedU128, KeyTypeId as CryptoKeyTypeId,
 		PerThing, Percent, RuntimeDebug,
 	};
-	use sp_std::{borrow::ToOwned, fmt::Debug, str, vec, vec::Vec};
+	#[cfg(not(feature = "std"))]
+	use sp_std::vec::Vec;
+	use sp_std::{borrow::ToOwned, fmt::Debug, str, vec};
```
seunlanlege, line 78: Is this change necessary?

Wizdave97: Added it earlier when I was debugging the `no-std` issue, I'll revert them.

seunlanlege:
left some todos, also we have this error type everywhere which is basically abused: `ICS02Error::implementation_specific`, lets swap it out for more relevant errors

Wizdave97:
> @Wizdave97 Is there any doc that I can use to understand the protocol before reviewing? Thanks!

The high level doc for the core  ibc protocol is specified here https://github.com/cosmos/ibc/tree/master/spec/core
This is a pallet is a wrapper around the implementation of the protocol in `ibc-rs` https://github.com/informalsystems/ibc-rs/tree/master/modules 
The major difference is that we are using `parity-trie` for proof verification.

blasrodri (approved):

blasrodri:
good job @Wizdave97

PR #830, opened 2022-03-28
CU-24qmg40 - fNFT/Staking rewards protocol

hussein-aitlahcen:


ComposableFiB0t:
Task linked: [CU-24qmg40 Draft skeleton of Chaos pallet implementation](https://app.clickup.com/t/24qmg40)

frame/chaos/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+			origin: OriginFor<T>,
+			staking_presets: BoundedBTreeSet<Timestamp, MaxStakingPresetsOf<T>>,
+		) -> DispatchResultWithPostInfo {
+			let _ = T::AdminOrigin::ensure_origin(origin)?;
+			StakingPresets::<T>::set(staking_presets);
+			Ok(().into())
+		}
+
+		/// Overwrite the unlock penalty.
+		/// The penalty will be cut off the stake asset, Chaos in this case.
```
dzmitry-lahoda, line 241: can remove Chaos word here

frame/chaos/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+		/// Arguments
+		///
+		/// * `origin` the origin that signed this extrinsic, must be `T::AdminOrigin`.
+		/// * `penalty` the penalty applied to users unstaking early.
+		#[pallet::weight(10_000)]
+		pub fn set_unlock_penalty(
+			origin: OriginFor<T>,
+			penalty: Permill,
+		) -> DispatchResultWithPostInfo {
+			let _ = T::AdminOrigin::ensure_origin(origin)?;
```
dzmitry-lahoda, line 252: better name GovernanceOrigin as I read it was dicussed that it will have some own governance

frame/chaos/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+	impl<T: Config> ChaosProtocol for Pallet<T> {
+		type AccountId = AccountIdOf<T>;
+		type AssetId = AssetIdOf<T>;
+		type Balance = BalanceOf<T>;
+		type InstanceId = T::InstanceId;
+
+		fn stake(
+			from: &Self::AccountId,
+			amount: Self::Balance,
+			duration: Timestamp,
```
dzmitry-lahoda, line 348: better use type Duration so it is easy to read code

frame/chaos/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+
+			// Transfer back the (possibly penalized) staked amount.
+			let chaos_account = Self::account_id();
+			// NOTE(hussein-aitlahcen): no need to keep protocol account alive.
+			T::ChaosAsset::transfer(&chaos_account, &to, penalized_stake, false)?;
+
+			// Decrement total shares.
+			// TODO: decrement by penalized only????? the protocol would own this fraction
+			// indefinitely
+			TotalShares::<T>::try_mutate(|x| {
```
dzmitry-lahoda, line 405: decremente by original value, you brought amount and you took amount. simmetrical. you burn - so to make others rich faster.  if you burn not whole amount you make inflation to happen faster. it feels that burning is good.

frame/chaos/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+		) -> DispatchResult {
+			// Make sure the asset has been registered as reward, otherwise we don't track it.
+			ensure!(RewardAssets::<T>::get().contains(&asset), Error::<T>::RewardAssetDisabled);
+
+			// Transfer the reward locally.
+			let chaos_account = Self::account_id();
+			T::Assets::transfer(asset, from, &chaos_account, amount, keep_alive)?;
+
+			// Increment the reward index, used to compute user rewards.
+			RewardIndexes::<T>::try_mutate(asset, |x| {
```
dzmitry-lahoda, line 438: not sure, it feels inex not good name. more like amount/total/accrued/collected rewardamounts?

frame/composable-traits/src/chaos.rs
```diff
@@ -0,0 +1,110 @@
...
+	/// Transfer a reward to the Chaos protocol.
+	///
+	/// Arguments
+	///
+	/// * `from` the account to transfer the reward from.
+	/// * `asset` the reward asset to transfer.
+	/// * `amount` the amount of reward to transfer.
+	/// * `keep_alive` whether to keep alive or not the `from` account while transferring the
+	///   reward.
+	fn transfer_reward(
```
dzmitry-lahoda, line 104: can be separate trait to show it should be used by protocols, not user facing like

frame/chaos/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+						})
+				};
+
+				let rewards = nft
+					.reward_indexes
+					.iter()
+					.map(|(asset, index)| -> Result<(AssetIdOf<T>, BalanceOf<T>), DispatchError> {
+						match RewardIndexes::<T>::get(&asset) {
+							Some(current_index) => {
+								let delta_index = current_index.saturating_sub(*index);
```
dzmitry-lahoda, line 479: some mistake seems . may be should be 

>>> initial_index = 100
>>> final_index = 300
>>> share = 10
>>> total = 100
>>> (final_index-initial_index)*(share/total)
20.0

frame/chaos/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+			})
+		}
+
+		fn claim(to: &Self::AccountId, instance_id: &Self::InstanceId) -> DispatchResult {
+			T::try_mutate_protocol_nft(instance_id, |nft: &mut ChaosStakingNFTOf<T>| {
+				// NOTE(hussein-aitlahcen): user is still able to claim even if his position
+				// 'expired', do we allow that?
+
+				let share = U256::from(nft.stake.saturated_into::<u128>());
+				let total_shares = TotalShares::<T>::get();
```
dzmitry-lahoda, line 459: so if somebody adds share after me, he dillutes my share, also i was first adopter.

dzmitry-lahoda: and goal is to put for as short peruid as possible and take as fast as possible

frame/chaos/src/lib.rs
dzmitry-lahoda: (final_index-initial_index)*(share*(1+(duration_days/max_time)**2)/total) <- make incetive

dzmitry-lahoda: image

in current impl any long term locker is penalized by reduction of hist share by newcomers

dzmitry-lahoda: anywa, current design is unsafe in terms of diluting share of long term holders too much without making short term stakers no so profitable. there is risk of long term stakers to be delluted too much.

dzmitry-lahoda: may be , define StakeStep (equal to 1 day). Than use formula like in WeightToFeePolynomial. So that insteaf of 1*x\**2,  can do term*x\**power for for several values, e.g. 4. Than can fit non decreasing curve to fit parameters from https://math.stackexchange.com/questions/466602/spline-interpolation-that-is-non-decreasing-when-given-non-decreasing-sequence . so we do not have to define ladder, but fit what ever is desired

frame/composable-traits/src/chaos.rs
```diff
@@ -0,0 +1,110 @@
...
+impl<Balance: AtLeast32BitUnsigned + Copy, Indexes> ChaosStakingNFT<Balance, Indexes> {
+	pub fn penalize_early_unstake_amount(
+		&self,
+		now: Timestamp,
+		penalty: Permill,
+	) -> Result<(Balance, Balance), DispatchError> {
+		if self.lock_date.safe_add(&self.lock_duration)? <= now {
+			Ok((self.stake, Balance::zero()))
+		} else {
+			let penalty_amount = penalty.mul_floor(self.stake);
```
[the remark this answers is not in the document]
vimukthi-git: Oh I see that `self.stake` is the unreduced stake amount in this case. But I see that the ordering is hard to enforce like that. May be it's best to pass the stake amount to the function as a parameter.

frame/staking-rewards/src/lib.rs
```diff
@@ -0,0 +1,501 @@
...
+								Ok((*reward_asset, reward))
+							},
+							None => Ok((*reward_asset, Zero::zero())),
+						}
+					})
+					.collect::<Result<Vec<_>, _>>()?;
+
+				let pallet_account = Self::account_id(&nft.asset);
+				for (reward_asset, reward) in rewards {
+					T::Assets::transfer(reward_asset, &pallet_account, to, reward, false)?;
```
vimukthi-git, line 490: How does the protocol ensure that it has enough liquidity in all reward asset types to cover it's reward obligations? Or is that supposed to be managed though external means?

dzmitry-lahoda: i think it would be documented. there is no obligation. rewards are passed by any other protocol (or even person), if there are zero rewars - zero will be given (e.g. protocol did not earn nothing). but yeah, may be check for zero before transfer call?

dzmitry-lahoda: #picasso-chaos describes full protocol afaik

hussein-aitlahcen: > How does the protocol ensure that it has enough liquidity in all reward asset types to cover it's reward obligations? Or is that supposed to be managed though external means?

There is no obligation to reward the user, and the user has to manually claim the reward (meaning that he will know what's the available reward he will be able to claim)

vimukthi-git: > There is no obligation to reward the user

I see, my question was more about say,
A user sees that they could claim some rewards based on the `nft.reward_multiplier` and calls `claim`, but the `pallet_account` does not have the necessary balance. So in that case user has to call again sometime later? Would this also mean that `unstake` calls would fail when there is no balance to pay the claimable reward?

hussein-aitlahcen: I'm not sure I follow, the reward multiplier is not meant to say whether you have some reward to claim or not?

frame/nft/Cargo.toml
```diff
@@ -0,0 +1,53 @@
...
+targets = ["x86_64-unknown-linux-gnu"]
+
+[dependencies.codec]
+default-features = false
+features = ["derive"]
+package = "parity-scale-codec"
+version = "2.0.0"
+
+[dependencies]
+frame-benchmarking = { default-features = false, optional = true, git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.17" }
```
andor0, line 19: where is it used?

frame/staking-rewards/Cargo.toml
```diff
@@ -0,0 +1,61 @@
...
+targets = ["x86_64-unknown-linux-gnu"]
+
+[dependencies.codec]
+default-features = false
+features = ["derive"]
+package = "parity-scale-codec"
+version = "2.0.0"
+
+[dependencies]
+frame-benchmarking = { default-features = false, optional = true, git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.17" }
```
andor0, line 19: where is it used?

frame/staking-rewards/src/lib.rs
vimukthi-git: Based on the slack conversation, it's clear to me the rewards are funded by the protocols calling `StakingReward.transferReward`. Thanks!!

frame/nft/src/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+				value.using_encoded(|v| Self::set_attribute(class, instance, k, v))
+			})
+		}
+
+		fn set_class_attribute(class: &Self::ClassId, key: &[u8], value: &[u8]) -> DispatchResult {
+			Self::ensure_class_exists(class)?;
+			Class::<T>::mutate(class, |entry| {
+				entry.as_mut().map(|nft_class| nft_class.insert(key.into(), value.into()));
+			});
+			Ok(())
```
andor0, line 221: looks like this can rewrite using `try_mutate_exists` as here https://github.com/paritytech/substrate/blob/master/frame/assets/src/functions.rs#L782-L805

hussein-aitlahcen: How would you rewrite this? `try_mutate_exists` does mutate if Ok is returned, in this case we always mutate? I'm not sure I understand the example :thinking:

hussein-aitlahcen: Ah, you mean removing the `ensure_class_exists` and merge it in a `try_mutate` :+1:

frame/chaos/src/lib.rs
hussein-aitlahcen: Updated with reward multiplier applied to the staked amount

frame/chaos/src/lib.rs
hussein-aitlahcen: The epoch mechanism will make this issue less relevant

frame/composable-traits/src/staking_rewards.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+/// Interface for protocol staking.
+pub trait Staking {
+	type AccountId;
+	type AssetId;
+	type Balance;
+	type InstanceId;
+
+	/// Stake an amount of protocol asset. A new NFT representing the user position will be
```
vimukthi-git, line 104: I wonder if this is sufficient documentation to explain the mechanism at work here (our slack chat helped a lot for me :) ). May be a link to further documentation might help? 
eg: https://app.clickup.com/20465559/docs/kghwq-63021 or a readme in the new staking-rewards pallet?

frame/composable-traits/src/staking_rewards.rs
```diff
@@ -0,0 +1,169 @@
...
+		let penalized_amount = self.stake.safe_sub(&penalty_amount)?;
+		Ok((penalized_amount, penalty_amount))
+	}
+
+	pub fn shares(&self) -> u128 {
+		self.reward_multiplier.mul_floor(self.stake.saturated_into::<u128>())
+	}
+
+	pub fn state(&self, now: Timestamp, epoch_start: Timestamp) -> PositionState {
+		if self.lock_date.saturating_add(self.lock_duration) < now {
```
vimukthi-git, line 71: Could be worth writing tests for these functions

dzmitry-lahoda (approved):
i am ok mostly, so need to start work on it to handle rest

hussein-aitlahcen:
@dzmitry-lahoda I've added the feature we discussed about.

frame/composable-traits/src/staking_rewards.rs
```diff
@@ -0,0 +1,183 @@
...
+	pub tagger: AccountId,
+	/// The beneficiary of the rewards distributed after tagging the position.
+	pub beneficiary: AccountId,
+	/// The rewards collected so far, rewards collected after this point will be distributed to the
+	/// `beneficiary`.
+	pub collected_rewards: CollectedRewards,
+}
+
+#[derive(Debug, PartialEq, Eq, Clone, Encode, Decode, TypeInfo)]
+pub struct StakingNFT<AccountId, AssetId, Balance, CollectedRewards> {
```
KaiserKarel, line 53: The naming here stutters a bit, we have a type named `staking_rewards::StakingTag`, which I think would nicer as `staking_rewards::Tag`, locally we can then rename at import if it conflicts.

KaiserKarel (dismissed):

frame/composable-traits/src/staking_rewards.rs
```diff
@@ -0,0 +1,183 @@
...
+	pub lock_date: Timestamp,
+	/// The duration for which this NFT stake was locked.
+	pub lock_duration: DurationSeconds,
+	/// The collected rewards counters at which this NFT was minted, used to compute the rewards.
+	pub collected_rewards: CollectedRewards,
+	/// The reward multiplier.
+	pub reward_multiplier: Perbill,
+	/// A liquidator tagged the staking position.
+	pub tag: Option<StakingTag<AccountId, CollectedRewards>>,
+}
```
dzmitry-lahoda, line 68: should penalty be encoded into fNFT on time of creation? it will ease creation of collateral from fNFT as price will known upfront and staking contract (in fNFT form) will not change its behaviour after it was created.

hussein-aitlahcen: good idea, done!

frame/composable-traits/src/staking_rewards.rs
hussein-aitlahcen: Tagging feature has been removed

benluelo (changes requested):

frame/staking-rewards/README.md
```diff
@@ -0,0 +1,132 @@
+# Staking Rewards
+
+This pallet allow us to reward users for staking certain assets.
```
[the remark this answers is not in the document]
hussein-aitlahcen: The readme is fully outdated, I need to upgrade it to latest changes

PoisonPhang (dismissed):

frame/composable-support/src/abstractions/block_fold.rs
```diff
@@ -0,0 +1,113 @@
...
+			),
+			d @ BlockFold::Done { .. } => return d,
+		};
+		let next_fold_state = match strategy {
+			FoldStrategy::Chunk { number_of_elements } => {
+				let r = iterator.try_fold(
+					(0_u32, current_state),
+					|(processed, state), (key, value)| {
+						let next_state = f(state, key.clone(), value);
+						if processed == number_of_elements.saturating_sub(1) {
```
hussein-aitlahcen, line 83: Need to explain that

frame/composable-support/src/abstractions/block_fold.rs
```diff
@@ -0,0 +1,119 @@
+use codec::{Decode, Encode, FullCodec, MaxEncodedLen};
+use frame_support::{
+	pallet_prelude::{OptionQuery, StorageMap, StorageValue},
+	storage::types::QueryKindTrait,
+	traits::{Get, StorageInstance},
+	ReversibleStorageHasher, StorageHasher,
+};
+use scale_info::TypeInfo;
+
+pub trait FoldStorage<S, K, V> {
```
vimukthi-git, line 10: What is the use of this and where is it used? Just curious

IMO Would be nice to have some docs for understanding and tests to verify the code

hussein-aitlahcen: Folding over a storage, block per block, in a chunk fashion.

hussein-aitlahcen: [Used here for instance](https://github.com/ComposableFi/composable/blob/d114b7fac4dea3da0784b09467cc835b2fccfda4/frame/staking-rewards/src/lib.rs#L426)

hussein-aitlahcen: This avoid having to manually track the folding state/iteration key and make you focus on the actual lambda that fold

benluelo (approved):

PoisonPhang (approved):

PR #832, opened 2022-03-28
CU-29ntvgm Remove unused dependencies

andor0:


seunlanlege (dismissed):

seunlanlege (approved):

dzmitry-lahoda:
i bet it is impossible to automate via some check to prevent to happen again?

dzmitry-lahoda (approved):

andor0:
> i bet it is impossible to automate via some check to prevent to happen again?

[this check exists](https://github.com/ComposableFi/composable/blob/main/.github/workflows/check.yml#L241), but it has `continue-on-error: true` option

dzmitry-lahoda:
wow. whi turned this off? ))) for me it was nit sof often)

PR #836, opened 2022-03-29
Add pull request template

cor:
Checklist currently only includes updating the documentation. Suggestions are most welcome

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,5 +1,6 @@
-## ClickUp
-*Please replace this line with a link to the ClickUp issue*
+## Issue
+*Please replace this line with either a link to the ClickUp issue, or a reference to the GitHub issue.*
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,9 @@
+## ClickUp
```
vimukthi-git, line 1: "ClickUp" sounds more like an internal detail for an OSS project. May be the thing we care about is generally any issue tracking reference ?

vimukthi-git (approved):

PoisonPhang (approved):

.github/pull_request_template.md
cor: I've changed the wording, what do you think about it now?

PoisonPhang (approved):

PR #727, opened 2022-03-07
CU-1vkxbq1 - introduce twap for constant product

hussein-aitlahcen:

PR #841, opened 2022-03-31
[CU-23v38ak] Implement TWAP in pablo.

vivekvpandya:
## Issue
https://app.clickup.com/t/23v38ak

## Description
Completely based on Hussein's work https://github.com/ComposableFi/composable/pull/727


ComposableFiB0t:
Task linked: [CU-23v38ak Add generalized (permissioned, optional & parameterized timeframe) TWAP for any pool](https://app.clickup.com/t/23v38ak)

frame/composable-maths/src/dex/price.rs
```diff
@@ -0,0 +1,66 @@
...
+	previous_price_cumulative: Balance,
+	current_timestamp: Timestamp,
+	current_exchange_rate: Rate,
+) -> Result<(Timestamp, Balance), DispatchError>
+where
+	Balance: SafeArithmetic,
+	Timestamp: AtLeast32Bit + SaturatedConversion + Saturating + Ord + Copy,
+	Convert: ConvertTrait<u128, Balance>,
+{
+	ensure!(current_timestamp > previous_timestamp, ArithmeticError::Underflow);
```
KaiserKarel, line 40: `ArithmeticError::Underflow` Seems the wrong error to me, perhaps use the Other vaiant with a more descriptive message?

frame/pablo/src/lib.rs
```diff
@@ -467,6 +499,139 @@ pub mod pallet {
...
+			)
+			.ok_or(ArithmeticError::Overflow)?)
+		}
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_initialize(_n: T::BlockNumber) -> Weight {
+			let current_timestamp = T::Time::now();
+			Pools::<T>::iter()
```
KaiserKarel, line 525: Do we run this logic every single block? That seems prohibitively expensive to me.

frame/pablo/src/mock.rs
```diff
@@ -137,6 +158,9 @@ impl pablo::Config for Test {
 	type LbpMaxInitialWeight = MaxInitialWeight;
 	type LbpMinFinalWeight = MinFinalWeight;
 	type PoolCreationOrigin = EnsureRoot<AccountId>;
+	type Time = Timestamp;
+	type EnabledPoolTWAP = IsInVec<EnabledPoolTWAP>;
```
KaiserKarel, line 162: This means that we'd need to do a runtime upgrade to add more pools correct? Which means that in order to have a pool with a TWAP, we'd:

1. Add the pool through governance.
2. Do a runtime upgrade.
3. Add the pool to the Pools<T>.

That's too many steps, need to be doable with just extrinsics.

frame/pablo/src/lib.rs
```diff
@@ -284,10 +294,22 @@ pub mod pallet {
...
+
+		/// Time provider.
+		type Time: Time;
+
+		/// The interval between TWAP computations.
+		#[pallet::constant]
+		type TWAPInterval: Get<MomentOf<Self>>;
+
+		/// Provider that tell up whether a specific pool TWAP is enabled or not.
+		type EnabledPoolTWAP: Contains<(CurrencyPair<Self::AssetId>, Self::PoolId)>;
```
KaiserKarel, line 306: So this means we'd need to have a third pallet with extrinsics for adding to the twaps?

frame/pablo/src/lib.rs
```diff
@@ -467,6 +499,139 @@ pub mod pallet {
...
+					let pair = match pool {
+						PoolConfiguration::StableSwap(stable_swap_pool_info) =>
+							stable_swap_pool_info.pair,
+						PoolConfiguration::ConstantProduct(constant_product_pool_info) =>
+							constant_product_pool_info.pair,
+						PoolConfiguration::LiquidityBootstrapping(
+							liquidity_bootstrapping_pool_info,
+						) => liquidity_bootstrapping_pool_info.pair,
+					};
+					T::EnabledPoolTWAP::contains(&(pair, *pool_id))
```
KaiserKarel, line 536: This filter still requires iterating over all pools. Why not iterate over `T::EnabledPoolTWAP` to avoid needless storage access?

frame/pablo/src/lib.rs
```diff
@@ -467,6 +499,139 @@ pub mod pallet {
...
+				pool_base_asset_under_management,
+				pool_quote_asset_under_management,
+			)
+			.ok_or(ArithmeticError::Overflow)?)
+		}
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_initialize(_n: T::BlockNumber) -> Weight {
```
KaiserKarel, line 523: Would be good to break some of this into a separate module and test individual componenets.

frame/pablo/src/lib.rs
```diff
@@ -467,6 +499,139 @@ pub mod pallet {
...
+											TWAPState::<T>::insert(
+												pool_id,
+												TimeWeightedAveragePrice {
+													base_twap: average_price_base,
+													quote_twap: average_price_quote,
+												},
+											);
+											Ok(weight + 1)
+										} else {
+											Ok(weight)
```
KaiserKarel, line 621: This weight calculation seems off.

frame/pablo/src/lib.rs
hussein-aitlahcen: Yeah I agree this was originally done with permissioned creation in mind, meaning the number of pool would be very limited. Let's make it such that `EnabledPoolTWAP` is the iterator

hussein-aitlahcen: Yeah I agree this was originally done with permissioned creation in mind, meaning the number of pool would be very limited. Let's make it such that `EnabledPoolTWAP` is the iterator

vivekvpandya: Okay

frame/pablo/src/lib.rs
hussein-aitlahcen: We can avoid doing that if the interval is always the same for all pool. Which is currently the case so we can probably directly have the extrinsic to enable TWAP in pablo, and setup the initial price cumulative. This would allow the hook to execute that only once the TWAPInterval expired rather than iterating + checking

hussein-aitlahcen: We can avoid doing that if the interval is always the same for all pool. Which is currently the case so we can probably directly have the extrinsic to enable TWAP in pablo, and setup the initial price cumulative. This would allow the hook to execute that only once the TWAPInterval expired rather than iterating + checking

hussein-aitlahcen: @vivekvpandya the `EnabledTWAP` doesn't make sense anymore in Pablo as we have all pools under the same pallet. We can remove it and have a local map that tell whether a pool is enabled or not. With an extrinsic to enable + populate first price cumulative

hussein-aitlahcen: @vivekvpandya the `EnabledTWAP` doesn't make sense anymore in Pablo as we have all pools under the same pallet. We can remove it and have a local map that tell whether a pool is enabled or not. With an extrinsic to enable + populate first price cumulative

vivekvpandya: :+1:

frame/composable-maths/src/dex/price.rs
vivekvpandya: @KaiserKarel any suggestions? I don't think any thing from DispatchError fits. Perhaps add `StaleRequrest` to pablo's Error?

frame/composable-maths/src/dex/price.rs
```diff
@@ -0,0 +1,69 @@
...
+pub fn compute_price_average<Convert, Balance, Timestamp>(
+	current_price_cumulative: Balance,
+	previous_price_cumulative: Balance,
+	elapsed: Timestamp,
+) -> Result<Rate, DispatchError>
+where
+	Convert: ConvertTrait<Balance, u128>,
+	Timestamp: AtLeast32Bit + SaturatedConversion,
+{
+    // previous_price_cumulative = pt
```
vimukthi-git, line 60: Could be part of the the function doc comment.

frame/pablo/src/lib.rs
vimukthi-git: :+1: 
Do we have cyclomatic complexity analysis for these in CI?

frame/composable-maths/src/dex/price.rs
```diff
@@ -0,0 +1,69 @@
+use composable_traits::{
```
vimukthi-git, line 1: It would be good to add tests for this module

frame/pablo/src/lib.rs
```diff
@@ -467,6 +546,104 @@ pub mod pallet {
...
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_initialize(_n: T::BlockNumber) -> Weight {
+            let current_timestamp = T::Time::now();
+            // let twap_enabled_pools : Vec<T::PoolId> = PriceCumulativesState::<T>::iter_keys().collect(); 
+            let twap_enabled_pools : Vec<T::PoolId> = TWAPState::<T>::iter_keys().collect(); 
+            twap_enabled_pools.iter().fold(Weight::zero(), |weight, pool_id| {
+                let rate_base = Self::do_get_exchange_rate(*pool_id, false);
+                let rate_quote = Self::do_get_exchange_rate(*pool_id, true);
```
vimukthi-git, line 576: We could possibly avoid calculating the rate again here for `rate_quote`, by taking `1/rate_base`, saves some storage calls

frame/pablo/src/lib.rs
```diff
@@ -467,6 +546,104 @@ pub mod pallet {
...
+                                    quote_price_cumulative,
+                                    previous_cumulative_price.quote_price_cumulative,
+                                )?;
+                                TWAPState::<T>::insert(
+                                    pool_id,
+                                    TimeWeightedAveragePrice {
+                                        timestamp: current_timestamp,
+                                        base_price_cumulative,
+                                        quote_price_cumulative,
+                                        base_twap: average_price_base,
```
dzmitry-lahoda, line 629: average_price_base can be found from base_price_cumulative. why we store it?

vivekvpandya: IMHO it should not matter much, with storage we get some saving on execution and if we want  to save execution time we use small bit of storage.

dzmitry-lahoda: But that price is not average. It is last price value measured. What us use what last price? It may be used one in the start of block, but than it changes and need to be calculated.

vivekvpandya: isn't it should be calculated after every TWAPInternal time?

frame/pablo/src/lib.rs
vivekvpandya: I tried but code gets ungly specially checking case for None, and Conversion from Rate to Balance etc. So I am not including this in current PR. It is an optimization that can be done later if it really impacts performance.

frame/composable-maths/src/dex/price.rs
```diff
@@ -0,0 +1,66 @@
...
+		Balance::zero(),
+		1_u64,
+		current_exchange_rate,
+	)
+	.map(|(_, x)| x)
+}
+
+pub fn compute_next_price_cumulative<Convert, Balance, Timestamp>(
+	previous_timestamp: Timestamp,
+	previous_price_cumulative: Balance,
```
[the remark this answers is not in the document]
vivekvpandya: To me it feels that these functions are not complicated so IMHO it does not matter much.

frame/pablo/src/lib.rs
vivekvpandya: @KaiserKarel  could you please elaborate ? I don't understand what is the issue.

frame/composable-maths/src/dex/price.rs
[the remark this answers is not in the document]
vimukthi-git: It's not really about avoiding complication, just readability and niceties.. up to you

hussein-aitlahcen:
@vivekvpandya let's move the TWAP computation from the hook to after a swap occured

vivekvpandya:
> @vivekvpandya let's move the TWAP computation from the hook to after a swap occured

@vimukthi-git  has raised concern over impact on weight for swap. Are we okay for that?

hussein-aitlahcen:
> > @vivekvpandya let's move the TWAP computation from the hook to after a swap occured
> 
> @vimukthi-git has raised concern over impact on weight for swap. Are we okay for that?

We are okay with that :)

frame/pablo/src/lib.rs
```diff
@@ -420,6 +440,34 @@ pub mod pallet {
 			)?;
 			Ok(())
 		}
+
+		#[pallet::weight(10_000)]
+		#[transactional]
+		pub fn enable_twap(origin: OriginFor<T>, pool_id: T::PoolId) -> DispatchResult {
+			let _who = ensure_signed(origin)?;
```
vimukthi-git, line 447: What is the rationale behind allowing any signed origin to enable this?

frame/composable-maths/src/dex/price.rs
```diff
@@ -0,0 +1,172 @@
...
+		.ok_or(ArithmeticError::Overflow)?;
+	let current_price_cumulative =
+		previous_price_cumulative.safe_add(&Convert::convert(new_price_cumulative))?;
+	Ok((elapsed, current_price_cumulative))
+}
+
+/// Computes price of asset as per following equation
+/// previous_price_cumulative = pt
+/// current_price_cumulative = (previous_price_cumulative + p * dt)
+/// (current_price_cumulative - previous_price_cumulative) / dt = p
```
vimukthi-git, line 63: May be?
```suggestion
/// (current_price_cumulative - previous_price_cumulative) / dt = twap
```

vimukthi-git (approved):
LGTM, one question though..

frame/pablo/src/lib.rs
vivekvpandya: Yes, I think we should only allow Root or Pool creator to enable this.

vimukthi-git:
@vivekvpandya @hussein-aitlahcen The cost of twap update does not look much when isolated to each pool anyway with each swap, so my question is not that relevant :) .

vimukthi-git (approved):

frame/pablo/src/lib.rs
```diff
@@ -284,10 +290,19 @@ pub mod pallet {
...
+
+		/// The interval between TWAP computations.
+		#[pallet::constant]
+		type TWAPInterval: Get<MomentOf<Self>>;
 	}
 
 	#[pallet::pallet]
 	#[pallet::generate_store(pub(super) trait Store)]
+	// TODO(hussein-aitlahcen): MaxEncodedLen not yet implemented for trait AT `Time::Moment`.
+	#[pallet::without_storage_info]
```
vimukthi-git, line 305: What would be the effect of this?

frame/pablo/src/lib.rs
```diff
@@ -420,6 +440,34 @@ pub mod pallet {
 			)?;
 			Ok(())
 		}
+
+		#[pallet::weight(10_000)]
```
vimukthi-git, line 444: We have to benchmark this after merging #827 I think.

vimukthi-git (dismissed):

frame/pablo/src/lib.rs
vivekvpandya: Yes.

frame/pablo/src/lib.rs
vimukthi-git: Can we just use `#[pallet::unbounded]` on the individual storage item that is affected here like it suggests at https://docs.substrate.io/v3/runtime/macros/#palletgenerate_storage_info

frame/pablo/src/lib.rs
```diff
@@ -467,6 +515,83 @@ pub mod pallet {
...
+								current_timestamp,
+								rate_base,
+							)?,
+							compute_next_price_cumulative::<T::Convert, _, _>(
+								previous_cumulative_price.timestamp,
+								previous_cumulative_price.quote_price_cumulative,
+								current_timestamp,
+								rate_quote,
+							)?,
+						);
```
hussein-aitlahcen, line 557: Why doing this computation here if we check that elapsed >= TWAPInterval after? Doesn't this mean we are going to waste this sometimes?

frame/pablo/src/lib.rs
```diff
@@ -467,6 +515,83 @@ pub mod pallet {
...
+				T::Convert::convert(T::Assets::balance(pair.quote, &pool_account));
+			Ok(Rate::checked_from_rational(
+				pool_base_asset_under_management,
+				pool_quote_asset_under_management,
+			)
+			.ok_or(ArithmeticError::Overflow)?)
+		}
+
+		fn update_twap(pool_id: T::PoolId) -> Result<(), DispatchError> {
+			if TWAPState::<T>::contains_key(pool_id) {
```
hussein-aitlahcen, line 537: Please replace `TWAPState::<T>::contains` and `TWAPState::<T>::get(pool_id).map_or(..` by doing a direct `if let Some(..) = TWAPState...get(..)`

frame/pablo/src/lib.rs
hussein-aitlahcen: Alright I think what we are doing here is somehow wrong, the cumulative should always be updated, only the actual price computation should be done when the `interval` passed

vimukthi-git: Good catch, in fact we are not calculating an average price here because according to the formula and applying the current code,
`twap = [(Price_end * elapsed + previous_cumulative) - previous_cumulative ] / elapsed = Price_end`

Therefore every time spot price changes(exchange and possibly addLiqudity, removeLiquidity) the cumulative should be calculated while keeping other fields of `TimeWeightedAveragePrice` struct in tact. We need to write some good tests here which changes prices during the configured interval, which affects the TWAP.

hussein-aitlahcen (approved):

frame/pablo/src/lib.rs
```diff
@@ -467,6 +533,98 @@ pub mod pallet {
 		pub(crate) fn account_id(pool_id: &T::PoolId) -> T::AccountId {
 			T::PalletId::get().into_sub_account(pool_id)
 		}
+		pub(crate) fn do_get_exchange_rate(
+			pool_id: T::PoolId,
+			swap: bool,
```
KaiserKarel, line 538: Can we make the swap parameter an enum, to better reflect the meaning?

KaiserKarel: Reading `do_foo(10, false)` vs `do_foo(10, PriceRatio::Swapped)`

frame/pablo/src/lib.rs
```diff
@@ -545,15 +705,16 @@ pub mod pallet {
...
+				PoolConfiguration::ConstantProduct(constant_product_pool_info) =>
+					Uniswap::<T>::add_liquidity(
+						who,
+						constant_product_pool_info,
+						pool_account,
+						base_amount,
+						quote_amount,
+						min_mint_amount,
+						keep_alive,
+					)?,
```
KaiserKarel, line 717:
```suggestion
				PoolConfiguration::ConstantProduct(info) =>
					Uniswap::<T>::add_liquidity(
						who,
						info,
						pool_account,
						base_amount,
						quote_amount,
						min_mint_amount,
						keep_alive,
					)?,
```

frame/pablo/src/lib.rs
```diff
@@ -507,12 +666,13 @@ pub mod pallet {
...
-					asset_id,
-					amount,
-				),
+				PoolConfiguration::ConstantProduct(constant_product_pool_info) =>
+					Uniswap::<T>::get_exchange_value(
+						&constant_product_pool_info,
+						&pool_account,
+						asset_id,
+						amount,
+					),
```
KaiserKarel, line 675:
```suggestion
				PoolConfiguration::ConstantProduct(info) =>
					Uniswap::<T>::get_exchange_value(
						&info,
						&pool_account,
						asset_id,
						amount,
					),
```

frame/pablo/src/lib.rs
```diff
@@ -467,6 +533,98 @@ pub mod pallet {
...
+								Ok(())
+							} else {
+								Ok(())
+							}
+						},
+					)
+				} else {
+					Ok(())
+				}
+			})
```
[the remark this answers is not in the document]
hussein-aitlahcen: Ah, didn't checked this change, this code is wrong you should update the cumulative always BEFORE computing he TWAP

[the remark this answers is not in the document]
hussein-aitlahcen: @vivekvpandya

frame/pablo/src/uniswap_tests.rs
```diff
@@ -428,3 +431,259 @@ proptest! {
...
+	}
+
+	fn next_block(n: u64) {
+		Timestamp::set_timestamp(MILLISECS_PER_BLOCK * n);
+		System::set_block_number(n);
+		Pablo::on_initialize(n);
+	}
+
+	#[test]
+	fn twap_1() {
```
KaiserKarel, line 461: Please describe what the test does. If I see in the output:

```
test::twap1 failed
```

vs

```
test::twap_with_two_assets
```
```

```
test::

frame/pablo/src/uniswap_tests.rs
```diff
@@ -428,3 +431,259 @@ proptest! {
...
+			sp_std::if_std! {
+				println!("timestamp {:?}", twap.timestamp);
+				println!("base_twap {:?}", twap.base_twap);
+				println!("base_price_cumulative {:?}", twap.base_price_cumulative);
+			}
+			assert!(twap.base_twap > min_base_price);
+			assert!(twap.quote_twap > min_quote_price);
+			assert!(twap.base_twap < max_base_price);
+			assert!(twap.quote_twap < max_quote_price);
+		});
```
KaiserKarel, line 687: This is quite massive: can we add some helper functions?

frame/pablo/src/lib.rs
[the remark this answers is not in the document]
vivekvpandya: There was no issue in code. However I have refactored a bit as per @KaiserKarel 's suggestion and hope it is fine now.

frame/pablo/src/uniswap_tests.rs
```diff
@@ -428,3 +431,166 @@ proptest! {
...
+					Origin::signed(BOB),
+					pool_id,
+					CurrencyPair::new(BTC, USDT),
+					usdt_value,
+					0,
+					false
+				));
+			};
+			run_to_block_and_swap(5);
+			update_min_max_price();
```
vimukthi-git, line 579: Shouldn't we test that the cumulative prices are updating correctly based on the swaps after these steps?

frame/pablo/src/lib.rs
```diff
@@ -467,6 +541,139 @@ pub mod pallet {
...
+					previous_price_cumulative.timestamp,
+					previous_price_cumulative.quote_price_cumulative,
+					current_timestamp,
+					rate_quote,
+				)?,
+			);
+			Ok((base_price_cumulative, quote_price_cumulative))
+		}
+
+		fn get_average_price(
```
vimukthi-git, line 590: IMO this function belongs in a dedicate type/module for  TWAP

vimukthi-git:
```suggestion
		fn get_twap_price(
```

frame/composable-maths/src/dex/price.rs
```diff
@@ -0,0 +1,170 @@
...
+		.checked_mul_int(1_u128)
+		.ok_or(ArithmeticError::Overflow)?;
+	let current_price_cumulative =
+		previous_price_cumulative.safe_add(&Convert::convert(new_price_cumulative))?;
+	Ok((elapsed, current_price_cumulative))
+}
+
+/// Computes price of asset as per following equation
+/// average_price = (current_price_cumulative - previous_price_cumulative) / elapsed
+pub fn compute_price_average<Convert, Balance, Timestamp>(
```
vimukthi-git, line 62:
```suggestion
pub fn compute_twap<Convert, Balance, Timestamp>(
```

frame/pablo/src/lib.rs
```diff
@@ -467,6 +541,139 @@ pub mod pallet {
...
+							if elapsed >= T::TWAPInterval::get() {
+								let (average_price_base, average_price_quote) =
+									Self::get_average_price(
+										base_price_cumulative,
+										previous_twap_state.base_price_cumulative,
+										quote_price_cumulative,
+										previous_twap_state.quote_price_cumulative,
+										elapsed,
+									)?;
+								*prev_twap_state = Some(TimeWeightedAveragePrice {
```
vimukthi-git, line 659: I wonder if this is the best way to keep twap updated as this is not accurate overtime as per the sampling rate/interval that it is supposed to follow. i.e without swap happening regularly the twap won't be calculated and updated in the storage on the expected sampling rate. Say the interval is 5 mins and the following sequence of updates
1st Min: cumulative calculated
2nd Min: Swap happens -> No twap update
6th Min: swap happens -> Twap updated -> **Not on time**
8th Min: Swap happens -> No Twap update
....
25th Min: Swap happens -> Twap updated -> **Not on time**

So there is a high probability of twap being outdated. Better way might be to build an extrinsic to get the TWAP at a given time by sampling the cumulative, which is always going to be up to date.

frame/pablo/src/types.rs
```diff
@@ -0,0 +1,19 @@
...
+use composable_traits::defi::Rate;
+use scale_info::TypeInfo;
+use sp_runtime::RuntimeDebug;
+
+#[derive(Encode, Decode, TypeInfo, Clone, Default, PartialEq, RuntimeDebug)]
+pub struct TimeWeightedAveragePrice<Timestamp, Balance> {
+	pub timestamp: Timestamp,
+	pub base_price_cumulative: Balance,
+	pub quote_price_cumulative: Balance,
+	pub base_twap: Rate,
```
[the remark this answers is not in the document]
vivekvpandya: Balance can't have floating point values like 1.22. 
I had thought to make `base_price_cumulative` and other as `Rate` but , `Rate` is `FixedU128` under the hood and may be a small type for storing something like cumulative value. So I guess that is why @hussein-aitlahcen  has kept it has `Balance` -> `u128`.

frame/pablo/src/lib.rs
vivekvpandya: Practically interval will be at-least as large as 1 hour. And given that real life pools will have large liquidity so even if TWAP is not updated as soon as interval passes , it will be still meaningful value.
@hussein-aitlahcen @KaiserKarel  any thoughts ?

hussein-aitlahcen: I think this should be fine, especially because we are going to enable it on pools that are "trusted", hopefully with some liquidity/swap happening within the interval. Otherwis we need to reverse back to my initial implementation using block hooks, i.e. updating the cumulative/TWAP in the hook when the interval passed while also updating the cumulative anytime when liquidity is moving.

vimukthi-git (approved):
LGTM

hussein-aitlahcen (approved):

hussein-aitlahcen (changes requested):
Let's implement https://github.com/ComposableFi/composable/pull/841#discussion_r846003456 to avoid any trouble :)

dzmitry-lahoda (approved):

hussein-aitlahcen (approved):

PR #848, opened 2022-04-04
CU-225tx7x - Create Vamm Functionality

Cardosaum:
## Issue
https://app.clickup.com/t/225tx7x

## Description
Adds a function to create a Virtual Automated Market Maker.


ComposableFiB0t:
Task linked: [CU-225tx7x Feature 1 (Create Vamm Functionality)](https://app.clickup.com/t/225tx7x)

frame/vamm/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+//!
+//! ### Extrinsics
+//!
+//! TODO
+//!
+//! not applicable
+//!
+//! ### Implemented Functions
+//!
+//! * [`create`](<Pallet as Vamm>::create)
```
0xangelo, line 44: Did this work? 👀

frame/vamm/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+	//                                             Config Trait
+	// ----------------------------------------------------------------------------------------------------
+
+	// Configure the pallet by specifying the parameters and types on which it depends.
+	#[pallet::config]
+	pub trait Config: frame_system::Config {
+		/// Event type emitted by this pallet. Depends on the runtime's definition of an event.
+		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+
+		/// The `VammId` used by the pallet. Corresponds to the Ids used by the Vamm pallet.
```
0xangelo, line 112: This *is* the Vamm pallet, no?

frame/vamm/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+			+ FullCodec
+			+ MaxEncodedLen
+			+ MaybeSerializeDeserialize
+			+ One
+			+ Parameter
+			+ PartialEq
+			+ TypeInfo
+			+ Zero;
+
+		/// Timestamp to be used for twap calculations and market deprecation.
```
0xangelo, line 127: I've been thinking about using the terms 'stopping' or 'freezing' a market. Thoughts?

frame/vamm/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+//!
+//! ## Overview
+//!
+//! ![](http://www.plantuml.com/plantuml/svg/NSqz2W91343XtbFe0TpqLWk2zyXcuyv09XdoWzTNB2oi7b_rraZqh26dIrUIshbSpYrpnWt0yRKSFLjj5Unacgova0susvWMk0a_Ej0Fm46D7PwEWu64qRiUrsOL_roce7xFA-l-wHi0)
+//!
+//! TODO
+//!
+//! ### Terminology
+//!
+//! * **VAAM**: Acronym for Virtual Automated Market Maker. (TODO: expand definition)
```
0xangelo, line 11: Small typo: **VAMM**

frame/vamm/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+	// ----------------------------------------------------------------------------------------------------
+	//                                            Runtime Events
+	// ----------------------------------------------------------------------------------------------------
+
+	// Pallets use events to inform users when important changes are made.
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
+		/// Emitted after a successful call to the [`create`](Pallet::create) function.
+		Created { vamm_id: VammIdOf<T>, state: VammStateOf<T> },
```
0xangelo, line 192: Might be good to check with others how much information is reasonable to put in an event (maybe Substrate's stack exchange?)

frame/vamm/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+	// Errors inform users that something went wrong.
+	#[pallet::error]
+	pub enum Error<T> {
+		/// This error is thrown when the caller tries to set `base_asset_reserves` to zero.
+		BaseAssetReserveIsZero,
+		/// This error is thrown when the caller tries to set `quote_asset_reserves` to zero.
+		QuoteAssetReserveIsZero,
+		/// This error is thrown when the caller tries to set `peg_multiplier` to zero.
+		PegMultiplierIsZero,
+	}
```
0xangelo, line 208: May be worthwhile to remove the repetitive 'prefixes', e.g., 'Caller tried to set `base_asset_reserves` to zero.

frame/vamm/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+
+		/// Creates a new virtual market.
+		///
+		/// # Overview
+		/// In order for the caller to create new markets, it has to request it
+		/// to the VAMM, which is responsible to keep track of all active
+		/// markets. The VAMM creates the market, inserts it in storage,
+		/// deposits a [`Created`](Event::<T>::Created) event on the
+		/// blockchain and returns the new market ID to the caller.
+		///
```
0xangelo, line 256: Some things to think about: 

Isn't there a difference between a market (like a perpetuals market in the Clearing House, e.g., BTC-PERP) and the market making mechanism used by it? E.g., a Limit Order Book where market makers quote their own prices, or Perp-v2's vAMM where they set range orders in a Uniswap-v3 model, or Cap's where they delegate the quoting of prices completely to the vAMM. What I'm getting at is that the vAMM is not the market itself, IMO, just a part of it (albeit a crucial one). It's as if the vAMM was the accounting procedure that its owner (e.g., the ClearingHouse) uses to keep track of supply and demand and quote prices accordingly, without actually storing and transferring tokens. In that case, it might be better to explicitly say that we're creating and managing virtual **automated** market **makers**.

Another point to consider is that the pallet is not **the** VAMM, but rather protocol for creating and managing several of them. In that respect, it is similar to the `pools` or `vaults` pallets. Perhaps we might even rename it to the `vamms` pallet.

frame/vamm/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+
+		#[transactional]
+		fn create(
+			base_asset_reserves: Self::Balance,
+			quote_asset_reserves: Self::Balance,
+			peg_multiplier: Self::Balance,
+		) -> Result<Self::VammId, DispatchError> {
+			// TODO: (Matheus)
+			// How to ensure that the caller has the right privileges?
+			// (eg. How to ensure the caller is the Clearing House, and not anyone else?)
```
0xangelo, line 292: For future consideration: it might be sufficient to just keep track of who the owner of the vAMM is (defined at the time of creation). That way anyone can create a vAMM but only its owner can modify it. Not sure how to implement it though; might need to verify the caller's signature of something of that nature. Although, this is not an extrinsic, so only Composable's code can access it. Nevertheless, better to ask for help on this matter.

frame/vamm/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+			ensure!(!base_asset_reserves.is_zero(), Error::<T>::BaseAssetReserveIsZero);
+			ensure!(!quote_asset_reserves.is_zero(), Error::<T>::QuoteAssetReserveIsZero);
+			ensure!(!peg_multiplier.is_zero(), Error::<T>::PegMultiplierIsZero);
```
0xangelo, line 295: Good call verifying the inputs before the `try_mutate` call; looks cleaner this way. I think I'll do that too

0xangelo (commented):
Good stuff overall!

frame/composable-traits/src/vamm.rs
```diff
@@ -0,0 +1,79 @@
...
+/// Represents the direction a of a position.
+#[derive(Encode, Decode, MaxEncodedLen, TypeInfo)]
+pub enum SwapDirection {
+	Add,
+	Remove,
+}
+
+/// Data relating to the state of a virtual market.
+#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Clone, Copy, PartialEq, Debug)]
+pub struct VammState<Balance, Timestamp> {
```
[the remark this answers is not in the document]
Cardosaum: I don't foresee it being used by other implementations as of now...

Moved it to `Pallet Types` section :+1:

frame/vamm/src/lib.rs
Cardosaum: Unfortunately not..

Not quite sure how I'm supposed to write it in order to `cargo doc` pick the right reference :cry:

frame/vamm/src/lib.rs
Cardosaum: Yep, might need to write more clearly. The comment seemed a bit off.

Cardosaum: What about the following one?

```
/// The Ids used by the pallet to index each virtual automated market maker created.
```

cc/ @0xangelo

frame/vamm/src/lib.rs
Cardosaum: Don't have a strong position in favor/against any of those.
Which one do you think is better, and why?

frame/vamm/src/lib.rs
Cardosaum: Thank you!

frame/vamm/src/lib.rs
Cardosaum: I'll have a look at it, but the `VammState` should be pretty small, only 3 `u128` and a `Option<i64>` as of now.

If size is a concern, we could remove the `state` tho.

frame/vamm/src/lib.rs
Cardosaum: Alright :+1: 

Made it more succinct.

frame/composable-traits/src/vamm.rs
```diff
@@ -0,0 +1,79 @@
...
+		+ CheckedAdd
+		+ Clone
+		+ Debug
+		+ FullCodec
+		+ MaxEncodedLen
+		+ One
+		+ Parameter
+		+ PartialEq
+		+ TypeInfo
+		+ Zero;
```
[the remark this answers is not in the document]
Cardosaum: Okay :+1: 

Removed them and it didn't break anything, so I think it's a plus? :sweat_smile: 

Thanks for the input.

frame/composable-traits/src/vamm.rs
```diff
@@ -0,0 +1,79 @@
...
+	/// - `quote_asset_reserves`: The amount of quote asset
+	/// - `peg_multiplier`: The constant multiplier responsible to balance quote and base asset
+	///
+	/// ## Returns
+	/// The new virtual market id, if successful.
+	fn create(
+		base_asset_reserves: Self::Balance,
+		quote_asset_reserves: Self::Balance,
+		peg_multiplier: Self::Balance,
+	) -> Result<Self::VammId, DispatchError>;
```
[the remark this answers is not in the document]
Cardosaum: I see your point, and totally agree with it, but I have a question though:
Do you expect the `VammParams` to hold the set of all possible arguments across all functions in the Vamm pallet?

For example, if you go to the [diagram](https://www.plantuml.com/plantuml/svg/NSuxhW8n38NX_gxYiW36FTL8IEp3uvWaakD4VeZjCwAWe3ewN_Nl7cYHG_ukNKcoS9hyjuZbDm33rrPxj3pIsMXgiCP6Sy1zZZKT3-571_eUOAV61a_3mK32Q5thjSrqXjAndku_lYr1_IJihENSDm00), we have arguments `base_asset_amount`, `quote_asset_amount` and `peg_multiplier` for the `create` function. For the `get_price`, however, we have `vamm` and `asset`. Should all of these 5 arguments be included in the `VammParams` struct?

frame/vamm/src/lib.rs
kevin-composable: You can't link directly to the trait implementations

kevin-composable: you can either link to the traits documentation, or not link to anything at all for the trait implementation

kevin-composable: for some reason this is not supported with rustdoc

frame/vamm/src/lib.rs
kevin-composable: Can markets be "continued" or "unfrozen" after they are stoped/frozen?

frame/vamm/src/lib.rs
kevin-composable: You should try to have all validity checks occur before any state mutations

frame/vamm/src/tests.rs
```diff
@@ -0,0 +1,181 @@
...
+		ExtBuilder::default().build().execute_with(|| {
+			assert_eq!(Vamm::vamm_count(), 0u128);
+
+			let vamm_created_ok = Vamm::create(base_asset_reserves, quote_asset_reserves, peg_multiplier);
+			assert_ok!(vamm_created_ok);
+
+			assert_eq!(Vamm::vamm_count(), 1u128);
+		});
+	}
+}
```
kevin-composable, line 181: You can have all proptests under the same proptest! block. I try to include all proptests related to the same set of functionality (creation, depositing, ...) all under the same proptest! block. Also sometimes you want to run some proptests more or less than others, so they can be under a seperate proptest! block that sets a smaller/larger #![proptest_config(ProptestConfig::with_cases(RUN_CASES))] value

frame/vamm/src/tests.rs
```diff
@@ -0,0 +1,181 @@
...
+		ExtBuilder::default().build().execute_with(|| {
+			assert_ok!(Vamm::create(base_asset_reserves, quote_asset_reserves, peg_multiplier));
+		});
+	}
+}
+
+proptest! {
+	#![proptest_config(ProptestConfig::with_cases(RUN_CASES))]
+	#[test]
+	fn create_vamm_zero_base_asset_reserves_error(base_asset_reserves in ZERO_RESERVE..=ZERO_RESERVE, quote_asset_reserves in MINIMUM_RESERVE..=MAXIMUM_RESERVE, peg_multiplier in MINIMUM_RESERVE..=MAXIMUM_RESERVE) {
```
kevin-composable, line 91: you can remove base_asset_reserves in ZERO_RESERVE..=ZERO_RESERVE and just hardcode ZERO_RESERVE like so:

`assert_noop!(Vamm::create(ZERO_RESERVE, quote_asset_reserves, peg_multiplier), Error::<MockRuntime>::BaseAssetReserveIsZero);`

but that is up to you. The same applies to the below functions that check for the same thing with quote_asset_reserves and peg_multiplier

frame/vamm/src/tests.rs
```diff
@@ -0,0 +1,181 @@
...
+
+			assert!(VammMap::<MockRuntime>::contains_key(0u128));
+		});
+	}
+}
+
+proptest! {
+	#![proptest_config(ProptestConfig::with_cases(RUN_CASES))]
+	#[test]
+	fn create_vamm_updates_storage_counter(base_asset_reserves in MINIMUM_RESERVE..=MAXIMUM_RESERVE, quote_asset_reserves in MINIMUM_RESERVE..=MAXIMUM_RESERVE, peg_multiplier in MINIMUM_RESERVE..=MAXIMUM_RESERVE) {
```
kevin-composable, line 171: I think this test `create_vamm_update_counter_succeeds` does what this test does, but better

kevin-composable (approved):

kevin-composable:
Everything looked great! I added a couple comments, but they are just small _potential_ changes here and there.

frame/vamm/src/tests.rs
Cardosaum: Oh, I thought I could only have one function per block. (might be confusing it with propcompose?)

Thank you for the input, there was no need to have different proptest! blocks since all of them was using the same number of test cases.

frame/vamm/src/tests.rs
Cardosaum: Oh, yeah this way is definitely better, I changed to it :+1:

frame/vamm/src/tests.rs
Cardosaum: I think I added `create_vamm_updates_storage_counter` before `create_vamm_update_counter_succeeds`, but forgot to get rid of the old one. 
Thanks for the input! :)

frame/composable-traits/src/vamm.rs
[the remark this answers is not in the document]
0xangelo: I see. I think `VammParams` will be better used in the `create` function only. I don't expect the arguments of the other functions to change much.

frame/vamm/src/lib.rs
0xangelo: Looks great 👌

frame/vamm/src/lib.rs
0xangelo: I'd imagine that they are permanently stopped/frozen. Might have to discuss this with the rest of the team

frame/vamm/src/lib.rs
0xangelo: I just realized that the pallets I mentioned use singular names (`pool` and `vault`) 😅

frame/vamm/src/lib.rs
0xangelo: Yeah, but it seems that the `try_mutate` function will not mutate the corresponding storage item if any error is thrown. At least, in my implementation I checked through tests that the storage is left intact

frame/vamm/src/lib.rs
0xangelo: TBH, I think we could use 'closing a market' as in TradFi

frame/vamm/src/lib.rs
Cardosaum: Just reverted the change from `vamms` to `vamm` :sweat_smile:

frame/vamm/src/lib.rs
Cardosaum: I found a workaround for it :) 

Take a look into the code [here](https://github.com/ComposableFi/composable/blob/07cf93c5d24dcd3e3d969939c6bc0a72f2f397c1/frame/vamm/src/lib.rs#L44-L47).

cc/ @kevin-composable , @0xangelo

frame/vamm/src/lib.rs
```diff
@@ -1,71 +1,64 @@
...
+//! manage virtual automated market makers, abstracting away complexity. It's
+//! important to note that currently just one type of constant function market
+//! maker is supported, namely the `x * y = k`.
 //!
-//! TODO
+//! Below is a diagram showing how the trait and runtime storage looks like and
+//! interact with each other:
 //!
-//! ### Terminology
+//! ![](https://www.plantuml.com/plantuml/svg/hLD1pzCm3BtdLvXT0EdR9y49LQNI92uS9Wv3N2kJcYvQisQssnxG-EqqxIZjEWaXVJbKYUtt4Z-_lj8ZUGAFBCO4j2Si2JO1gufqsphM1gijUh-1dmRwvSvA_0DjN_Hj2AD0k_CUqbGPdMRPhJ2kNs2PKEdDTnJAKOGqHrytPqsWUVV-mnDScbeVPmALkMygTQ5on6FqxOoveC1a8tcBtYSGN_EvU8BkIES4lZfFekZ375AIve6TlOSCru_7NTpUOxJ3i83C2wHFvd_xkwDUzbGu9gkkxXzuw66V_XnNV56MboBqTKiFs_xleQnOjNv8hCYJr7FaTVWMg1YlXasxs-_Xe3LZIPkPRJo6qLqosKiWJM-LUmmBayKrtWmViobwdNQsGXb93efAFP4eDrPN9F6XSr6OXBWbMHT5WVvT9HVM_BIEBvszpdlaq-2vHUfBY8DD9sDdu2IVVe9YqNZyNeqnhwF29hOEswBkpjli9cO27JibCvwUdzcLHyrc8YLn2F8R)
```
0xangelo, line 19: Is the source for this diagram available somewhere?

frame/vamm/src/lib.rs
```diff
@@ -1,71 +1,64 @@
...
 //! ### Example
 //!
-//! TODO
-//!
 //! ## Related Modules
 //!
-//! TODO
-//!
-//! (clearing house)
+//! - [`Clearing House Pallet`](../clearing_house/index.html)
```
0xangelo, line 61: Neat, didn't realize one could do that

frame/vamm/src/lib.rs
```diff
@@ -313,32 +308,30 @@ pub mod pallet {
...
-			ensure!(!quote_asset_reserves.is_zero(), Error::<T>::QuoteAssetReserveIsZero);
-			ensure!(!peg_multiplier.is_zero(), Error::<T>::PegMultiplierIsZero);
+			ensure!(!params.base_asset_reserves.is_zero(), Error::<T>::BaseAssetReserveIsZero);
+			ensure!(!params.quote_asset_reserves.is_zero(), Error::<T>::QuoteAssetReserveIsZero);
+			ensure!(!params.peg_multiplier.is_zero(), Error::<T>::PegMultiplierIsZero);
 
-			VammCounter::<T>::try_mutate(|id| {
-				let old_id = *id;
+			VammCounter::<T>::try_mutate(|next_id| {
+				let id = *next_id;
```
0xangelo, line 321: Hmm, cool. No need to call `.clone()`!

0xangelo: Ah, it's because it has the `Copy` trait

Cardosaum: Yep haha

frame/vamm/src/lib.rs
Cardosaum: Alright, in a recent commit I proceeded with the change: 'Deprecation' -> 'Closing'

PR #858, opened 2022-04-07
Fix full repay borrow

andor0:
## Issue
https://app.clickup.com/t/25drfkk

### DIFF
--- a/frame/lending/src/tests.rs
+++ b/frame/lending/src/tests.rs
@@ -607,8 +607,7 @@ fn test_repay_partial_amount() {
 		);
 		dbg!(alice_limit);
 
-		(1_000..2_000).for_each(process_block);
-		// (2_000..3_000).for_each(process_block);
+		(2_000..3_000).for_each(process_block);
 
 		dbg!(BORROW::units(1) / 10_000);
 		// 100000000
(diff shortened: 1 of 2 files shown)


ComposableFiB0t:
Task linked: [CU-2201k46 Design QA](https://app.clickup.com/t/2201k46)

benluelo:
This has already been implemented: https://github.com/ComposableFi/composable/blob/ben/lending-tests-wip/frame/lending/src/lib.rs#L2029-L2036

frame/lending/src/tests.rs
```diff
@@ -673,7 +679,8 @@ fn test_repay_partial_amount() {
 		dbg!(alice_balance);
 		dbg!(Tokens::balance(COLLATERAL::ID, &ALICE));
 		dbg!(Tokens::balance(COLLATERAL::ID, &BOB));
-		assert!(alice_balance > Tokens::balance(COLLATERAL::ID, &ALICE));
+		// TODO: fix bug with partial repay
```
dzmitry-lahoda, line 682: no sure, the test is that repay all at once == repay several times in part within same block, given accrued interest is not equal to zero

benluelo:
Closed as #867 has the same fix.

PR #862, opened 2022-04-08
mdbook  Addition - Composable Overview/Vision

PoisonPhang:
## Issue
N/A?

## Description
Addition to the mdbook written by @S123Z123
An overview and vision of the Composable parachain

## Checklist

- ~I have updated the cargo docs to reflect changes made by this PR _(if applicable)_~


cor (approved):

dzmitry-lahoda (commented):
may be make image to be lfs? usually that is assets folders

PoisonPhang:
> may be make image to be lfs? usually that is assets folders

I believe @cor planned on converting the project to lfs in a later PR as the work to do so would be quite extensive. However, having assets folders relative to where the assets are used would generally look nice. @dzmitry-lahoda

cor:
Git LFS will be configured later. For now, I disagree with the idea of moving assets to a separate folder for the following reasons:

1. Links to assets are relative, not absolute. determining the relative path of an asset with relation to the page you are writing is a lot harder when you need to walk the directory tree up all the way to the assets folder.
2. Moving pages with their assets to a different location is a lot easier when the assets are stored close to the page, since you can move them without editing the links in the page. With a dedicated assets directory, you would have to edit the links in the page when a page is moved one level deeper, by adding `../` to every link
3. When assets are close to the page, the URLs are more descriptive (and SEO friendly). For example: `docs.composable.finance/pallets/mosaic/schema.png` vs `docs.composable.finance/assets/schema.png`
4. Asset naming collisions are more likely with an assets directory. For example having both `docs.composable.finance/pallets/oracle/schema.png` and `docs.composable.finance/pallets/mosaic/schema.png` is possible when you don't have a shared assets/ directory.

The one exception where an assets directory would make sense is if we had assets that are used on multiple pages (such as a logo). When this happens, we should reconsider adding one.

KaiserKarel (approved):

PR #869, opened 2022-04-11
CU-24ak8ep - Fix merge incompatibilities in Vamm trait

0xangelo:
## Issue
[ClickUp](https://app.clickup.com/t/24ak8ep)

## Description
This PR proposes a way to reconcile the Vamm implementations from last sprint

It also updates the `push` stage git hooks to run Clearing House tests whenever `frame/composable-traits/src/vamm.rs` is modified, since it is a dependency.

### DIFF
--- a/frame/composable-traits/src/vamm.rs
+++ b/frame/composable-traits/src/vamm.rs
@@ -1,7 +1,11 @@
 //! # Virtual Automated Market Maker
 //!
 //! Common traits and data structures for vamm implementation.
+use codec::{Decode, Encode, FullCodec, MaxEncodedLen};
 use frame_support::pallet_prelude::DispatchError;
+use scale_info::TypeInfo;
+use sp_runtime::FixedPointNumber;
+use sp_std::fmt::Debug;
 
 /// Exposes functionality for creation and management of virtual automated market makers.
 ///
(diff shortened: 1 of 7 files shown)


ComposableFiB0t:
Task linked: [CU-24ak8ep Fix merge incompatibilities in Vamm trait](https://app.clickup.com/t/24ak8ep)

frame/composable-traits/src/vamm.rs
```diff
@@ -11,18 +15,27 @@ use frame_support::pallet_prelude::DispatchError;
...
 	type Balance;
-
+	/// Signed fixed point number implementation
+	type Decimal: FixedPointNumber;
+	/// Configuration for creating and initializing a new vAMM instance. May be used in extrinsic
+	/// signatures
+	type VammConfig: FullCodec + MaxEncodedLen + TypeInfo + Debug + Clone + PartialEq;
 	/// The identifier type for each virtual automated market maker.
-	type VammId;
+	type VammId: FullCodec + MaxEncodedLen + TypeInfo;
```
0xangelo, line 24: @Cardosaum lemme know if you disagree with this solution. What motivates this is:
* the associated `VammConfig` type provides a little more generality for its implementation
* the Clearing House doesn't care about the exact attributes of `VammConfig`; it is created and provided by the extrinsic caller via the `MarketConfig` struct
* the Clearing House **does** care about the trait bounds of `VammConfig` here, since it will be indirectly part of an extrinsic signature

This also allows other implementers of the `Vamm` trait to have their own `VammConfig`s. If for any reason some code requires the trait to use the `VammConfig` defined in this file, it can do so by specifying the bound
```rust
type Vamm: Vamm<VammConfig = VammConfig<BalanceType>>;
```

frame/composable-traits/src/vamm.rs
```diff
@@ -11,18 +15,27 @@ use frame_support::pallet_prelude::DispatchError;
...
 
 	/// Create a new virtual automated market maker.
 	///
 	/// ## Returns
 	/// The identifier of the newly created vamm.
-	fn create(config: VammConfig<Self::Balance>) -> Result<Self::VammId, DispatchError>;
+	fn create(config: Self::VammConfig) -> Result<Self::VammId, DispatchError>;
```
0xangelo, line 30: @Cardosaum May I suggest changing this to take in a *reference* to a `VammConfig` instead? I see no particular reason why `create` should take ownership of `config`. It also causes the Clearing House to have to call `vamm_config.clone()` when passing this value to the function in question, which might have an unnecessary resource cost.

If don't have any objections, I can include this change in this PR already.

frame/composable-traits/src/vamm.rs
0xangelo: Yeah, sounds good. If for any reason that changes in the future, we can always remove the bound

PR #871, opened 2022-04-12
refactor(VammConfig): Small changes

Cardosaum:
Response to changes proposed in PR #869

### DIFF
--- a/frame/clearing-house/src/mock/vamm.rs
+++ b/frame/clearing-house/src/mock/vamm.rs
@@ -7,7 +7,7 @@ pub mod pallet {
 	// ----------------------------------------------------------------------------------------------------
 
 	use codec::{Decode, Encode, FullCodec, MaxEncodedLen};
-	use composable_traits::vamm::Vamm;
+	use composable_traits::{defi::DeFiComposableConfig, vamm::Vamm};
 	use frame_support::pallet_prelude::*;
 	use scale_info::TypeInfo;
 	use sp_runtime::{traits::Zero, FixedPointNumber};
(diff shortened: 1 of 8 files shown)


frame/clearing-house/src/mock/vamm.rs
```diff
@@ -98,7 +98,7 @@ pub mod pallet {
 		type VammId = T::VammId;
 
 		#[allow(unused_variables)]
-		fn create(info: Self::VammConfig) -> Result<Self::VammId, DispatchError> {
+		fn create(config: &Self::VammConfig) -> Result<Self::VammId, DispatchError> {
```
0xangelo, line 101: Good catch of the name inconsistency

frame/composable-traits/src/vamm.rs
```diff
@@ -15,23 +15,26 @@ use sp_std::fmt::Debug;
...
 	/// The balance type for an account.
 	type Balance;
+
 	/// Signed fixed point number implementation
 	type Decimal: FixedPointNumber;
+
 	/// Configuration for creating and initializing a new vAMM instance. May be used in extrinsic
 	/// signatures
 	type VammConfig: FullCodec + MaxEncodedLen + TypeInfo + Debug + Clone + PartialEq;
+
```
0xangelo, line 25: Is there a guideline for whether or not to separate trait associated types with a blank line? It seems some pallets do it while others don't. I'm okay with either to be honest

0xangelo (approved):
LGTM. Thanks for the improvements

frame/composable-traits/src/vamm.rs
Cardosaum: I'm not aware of such a guideline for it, but I have a personal preference in favor of black lines between types, since otherwise the code and docs seems a bit clunky.
(No prob to remove the blank lines if needed tho)

0xangelo: No problems, let's keep it this way. Ready to merge then

PR #872, opened 2022-04-12
[Cu-23n7bc2] BondedFinanceTestsFix and Minor Optimizations

kollegian:
Greetings, a minor PR to change the params of bondedFinance Tests and to add an optimized method to mint assets into the wallet. Even though I tried to enable parallelization unfortunately it still fails.

### DIFF
--- a/integration-tests/runtime-tests/test/tests/bondedFinance/txBondedFinanceTests.ts
+++ b/integration-tests/runtime-tests/test/tests/bondedFinance/txBondedFinanceTests.ts
@@ -9,6 +9,7 @@ import {
   txBondedFinanceCancelFailureTest,
   txBondedFinanceCancelSuccessTest, txBondedFinanceCancelSudoSuccessTest
 } from "@composabletests/tests/bondedFinance/testHandlers/cancelTests";
+import {mintAssetsToWallet} from "@composable/utils/mintingHelper";
 
 
 /**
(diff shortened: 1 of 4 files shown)


integration-tests/runtime-tests/test/tests/bondedFinance/txBondedFinanceTests.ts
```diff
@@ -90,7 +96,7 @@ describe('tx.bondedFinance Tests', function() {
    * Runs all tx FAILURE tests for the bondedFinance pallet.
    */
   describe('tx.bondedFinance.offer Failure Tests', function () {
-    if (!testConfiguration.enabledTests.offer_bond__failure.enabled)
+    if (testConfiguration.enabledTests.offer_bond__failure.enabled)
```
Dom-Roth, line 99: I think this should be an inverted if. Since in the case `enabled == false` we want ignore this test block.
If we want to disable this, rather change the bool in the test_configuration.json to false.

Dom-Roth (changes requested):

integration-tests/runtime-tests/test/tests/bondedFinance/txBondedFinanceTests.ts
kollegian: Yes, I missed this. Implemented now.

Dom-Roth (dismissed):
LGTM!

Dom-Roth:
As a next task, we should aim to make the failure tests work correctly again. Then we just have to wait for the Pablo refactoring and all tests will work again and it can be part of CI.

ComposableFiB0t:
Task linked: [CU-23n7bc2 Validation issues](https://app.clickup.com/t/23n7bc2)

kollegian:
> 
Currently they all passing with assertions to check for an error. 
BondedFinance

Dom-Roth (dismissed):

andor0 (approved):

Dom-Roth (approved):

PR #885, opened 2022-04-13
CU-2arav8f Add a running of runtime-test in CI

andor0:
## Issue
https://app.clickup.com/t/2arav8f


scripts/wait_polkadot_launch.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+while true; do
+    sleep 1
+    stdout=$(docker-compose logs --tail=1 | grep "POLKADOT LAUNCH COMPLETE")
```
dzmitry-lahoda, line 5: are there not logs after that message? just not to miss one. can we show some value while waiting? or will see hang?

andor0: this is the last message in logs. other messages are about configuration parameters of relay chain and collators, i.e. duplicate the [config](https://github.com/ComposableFi/composable/blob/andor0/add-runtime-tests-in-ci/scripts/polkadot-launch/composable_and_basilisk.json).

dzmitry-lahoda: if message will never appear, what i will see?

hussein-aitlahcen: Probably worth printing the logs until we get the LAUNCH complete?

andor0: > if message will never appear, what i will see?

a message about timeout

andor0: > Probably worth printing the logs until we get the LAUNCH complete?

done
[example](https://github.com/ComposableFi/composable/runs/6055268569?check_suite_focus=true)

.github/workflows/runtime-tests.yml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Prepare a test environment
+      run: |
+        cd integration-tests/runtime-tests
+        docker-compose -f docker-compose-ci.yml up -d
+
+    - name: Wait when the node will be started
+      timeout-minutes: 5
```
dzmitry-lahoda, line 25: oh

dzmitry-lahoda (approved):

.github/workflows/runtime-tests.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    - name: Run tests
+      run: |
+        cd integration-tests/runtime-tests
+        docker run --rm -v$(pwd):/build -e "ENDPOINT=composable:9988" --workdir /build --network runtime-tests_default --entrypoint=/build/DockerRun.sh node:17
+
+    - name: Cleanup
+      run: |
+        cd integration-tests/runtime-tests
+        docker-compose -f docker-compose-ci.yml down --remove-orphans
```
KaiserKarel, line 38: Can we add this to check.yaml, as we want these to pass for each PR

dzmitry-lahoda: can we make it next time? or optional? 
i think these run long and likely will give false positives.

andor0: > Can we add this to check.yaml, as we want these to pass for each PR

this check depends on composable-sandbox docker image
we should build a new composable-sandbox docker image for every PR if we want to do this. so this will increase check time by about 1h15m

ps: I will add a depend between the tests and composable-sandnox in CI in one of next PRs.

andor0: > can we make it next time? or optional? i think these run long and likely will give false positives.

as a creation of files with weights? using a tag?

hussein-aitlahcen:
Is this PR still relevant?

KaiserKarel:
@andor0 @Dom-Roth Can we rebase this, just run this as a separate workflow on main?

ComposableFiB0t:
Task linked: [CU-2arav8f Add a running of 'integration-tests/runtime-tests' in CI](https://app.clickup.com/t/2arav8f)

hussein-aitlahcen:
Any chance we make the CI pass?

KaiserKarel:
@Dom-Roth still fails :(

PR #886, opened 2022-04-13
Escher feature/vamm get price

Cardosaum:
## Issue
https://app.clickup.com/t/24agzx9

## Description

Adds the `get_price` function for the VAMM Pallet.

## Checklist

- [ ] I have updated the cargo docs to reflect changes made by this PR _(if applicable)_
- [ ] I have updated the `book/` to reflect changes made by this PR _(if applicable)_

PR #893, opened 2022-04-14
CU-244p28y - Open position functionality

0xangelo:
## Issue
https://app.clickup.com/t/244p28y

## Description
Also includes:
- refactorings and style changes
- `Vamm` trait interface changes developed in in #886

### DIFF
--- a/frame/clearing-house/src/lib.rs
+++ b/frame/clearing-house/src/lib.rs
@@ -898,7 +898,7 @@ pub mod pallet {
 
 	// Helper functions - low-level functionality
 	impl<T: Config> Pallet<T> {
-		pub fn get_or_create_position<'a>(
+		fn get_or_create_position<'a>(
 			positions: &'a mut BoundedVec<Position<T>, T::MaxPositions>,
 			market_id: &T::MarketId,
 			market: &Market<T>,
--- a/frame/clearing-house/src/lib.rs
+++ b/frame/clearing-house/src/lib.rs
@@ -246,39 +246,43 @@ pub mod pallet {
 
 	/// Stores the user's position in a particular market
 	#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Debug)]
-	pub struct Position<Decimal, MarketId> {
+	#[scale_info(skip_type_params(T))]
+	#[codec(mel_bound())]
+	pub struct Position<T: Config> {
 		/// The Id of the virtual market
-		pub market_id: MarketId,
+		pub market_id: T::MarketId,
 		/// Virtual base asset amount. Positive implies long position and negative, short.
-		pub base_asset_amount: Decimal,
+		pub base_asset_amount: T::Decimal,
 		/// Virtual quote asset notional amount (margin * leverage * direction) used to open the
 		/// position
-		pub quote_asset_notional_amount: Decimal,
+		pub quote_asset_notional_amount: T::Decimal,
 		/// Last cumulative funding rate used to update this position. The market's latest
 		/// cumulative funding rate minus this gives the funding rate this position must pay. This
 		/// rate multiplied by this position's size (base asset amount * amm price) gives the total
 		/// funding owed, which is deducted from the trader account's margin. This debt is
 		/// accounted for in margin ratio calculations, which may lead to liquidation.
-		pub last_cum_funding: Decimal,
+		pub last_cum_funding: T::Decimal,
 	}
 
 	/// Data relating to a perpetual contracts market
 	#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Debug)]
-	pub struct Market<AssetId, Decimal, VammId> {
+	#[scale_info(skip_type_params(T))]
+	#[codec(mel_bound())]
+	pub struct Market<T: Config> {
 		/// The Id of the vAMM used for price discovery in the virtual market
-		pub vamm_id: VammId,
+		pub vamm_id: T::VammId,
 		/// The Id of the underlying asset (base-quote pair). A price feed from one or more oracles
 		/// must be available for this symbol
-		pub asset_id: AssetId,
+		pub asset_id: AssetIdOf<T>,
 		/// Minimum margin ratio for opening a new position
-		pub margin_ratio_initial: Decimal,
+		pub margin_ratio_initial: T::Decimal,
 		/// Margin ratio below which liquidations can occur
-		pub margin_ratio_maintenance: Decimal,
+		pub margin_ratio_maintenance: T::Decimal,
 		/// Minimum amount of quote asset to exchange when opening a position. Also serves to round
 		/// a trade if it results in closing an existing position
-		pub minimum_trade_size: Decimal,
+		pub minimum_trade_size: T::Decimal,
 		/// The latest cumulative funding rate of this market. Must be updated periodically.
-		pub cum_funding_rate: Decimal,
+		pub cum_funding_rate: T::Decimal,
 		/// The timestamp for the latest funding rate update.
 		pub funding_rate_ts: DurationSeconds,
 		/// The time span between each funding rate update.
@@ -319,15 +323,12 @@ pub mod pallet {
 
 	type AssetIdOf<T> = <T as DeFiComposableConfig>::MayBeAssetId;
 	type BalanceOf<T> = <T as DeFiComposableConfig>::Balance;
-	type MarketIdOf<T> = <T as Config>::MarketId;
 	type DecimalOf<T> = <T as Config>::Decimal;
 	type VammConfigOf<T> = <T as Config>::VammConfig;
 	type VammIdOf<T> = <T as Config>::VammId;
 	type SwapConfigOf<T> = SwapConfig<VammIdOf<T>, BalanceOf<T>>;
 	type SwapSimulationConfigOf<T> = SwapSimulationConfig<VammIdOf<T>, BalanceOf<T>>;
-	type PositionOf<T> = Position<DecimalOf<T>, MarketIdOf<T>>;
 	type MarketConfigOf<T> = MarketConfig<AssetIdOf<T>, DecimalOf<T>, VammConfigOf<T>>;
-	type MarketOf<T> = Market<AssetIdOf<T>, DecimalOf<T>, VammIdOf<T>>;
 
 	// ----------------------------------------------------------------------------------------------------
 	//                                           Runtime Storage
@@ -352,7 +353,7 @@ pub mod pallet {
 		_,
 		Blake2_128Concat,
 		T::AccountId,
-		BoundedVec<PositionOf<T>, T::MaxPositions>,
+		BoundedVec<Position<T>, T::MaxPositions>,
 		ValueQuery,
 	>;
 
@@ -369,7 +370,7 @@ pub mod pallet {
 	/// Maps [MarketId](Config::MarketId) to the corresponding virtual [Market] specs
 	#[pallet::storage]
 	#[pallet::getter(fn get_market)]
-	pub type Markets<T: Config> = StorageMap<_, Blake2_128Concat, T::MarketId, MarketOf<T>>;
+	pub type Markets<T: Config> = StorageMap<_, Blake2_128Concat, T::MarketId, Market<T>>;
 
 	// ----------------------------------------------------------------------------------------------------
 	//                                         Genesis Configuration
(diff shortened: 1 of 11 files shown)


ComposableFiB0t:
Task linked: [CU-244p28y Feature 4 (Open Position Functionality)](https://app.clickup.com/t/244p28y)

frame/clearing-house/src/mock/vamm.rs
```diff
@@ -106,13 +130,51 @@ pub mod pallet {
...
 				Ok(twap)
 			} else {
 				Err(Error::<T>::FailedToCalculateTwap.into())
 			}
 		}
+
+		fn swap(config: &Self::SwapConfig) -> Result<Self::Integer, DispatchError> {
+			let Self::SwapConfig { vamm_id, asset, input_amount, direction, output_amount_limit } =
+				config;
+
```
0xangelo, line 148: should remove this

frame/clearing-house/src/mock/vamm.rs
```diff
@@ -106,13 +130,51 @@ pub mod pallet {
...
+				Some(integer) => Ok(integer),
+				None => Err(Error::<T>::FailedToExecuteSwap.into()),
+			}
+		}
+
+		fn swap_simulation(
+			config: &Self::SwapSimulationConfig,
+		) -> Result<Self::Integer, DispatchError> {
+			let Self::SwapSimulationConfig { vamm_id, asset, input_amount, direction } = config;
+
```
0xangelo, line 159: should remove this

frame/clearing-house/src/tests.rs
```diff
@@ -80,23 +84,71 @@ fn valid_market_config() -> MarketConfig {
 		margin_ratio_initial: FixedI128::from_float(0.1),
 		// liquidate when above 50x leverage
 		margin_ratio_maintenance: FixedI128::from_float(0.02),
+		// 'One cent' of the quote asset
+		minimum_trade_size: FixedI128::checked_from_rational(1, 100).unwrap(),
```
0xangelo, line 88: should just use `from_float` here

frame/clearing-house/src/tests.rs
```diff
@@ -80,23 +84,71 @@ fn valid_market_config() -> MarketConfig {
...
+fn valid_base_asset_amount_limit() -> Balance {
+	FixedI128::checked_from_integer(10).unwrap().into_inner().unsigned_abs()
+}
+
 // ----------------------------------------------------------------------------------------------------
 //                                           Initializers
 // ----------------------------------------------------------------------------------------------------
 
+trait MarginInitializer {
+	fn set_margin(self, account_id: &AccountId, asset_id: AssetId, amount: Balance) -> Self;
```
0xangelo, line 107: should rename to `add_margin`

frame/composable-traits/src/clearing_house.rs
```diff
@@ -38,13 +45,40 @@ pub trait ClearingHouse {
 	/// ## Returns
 	/// The new market's id, if successful
 	fn create_market(config: &Self::MarketConfig) -> Result<Self::MarketId, DispatchError>;
+
+	/// Open a position in a market
+	///
+	/// This may result in the following outcomes:
+	/// * Creation of a whole new position in the market, if one didn't already exist
+	/// * Increase in the size of an existing position, if the trade's direction matches the existion position's one
```
Cardosaum, line 53: "Increase _in_ the size...", does this particle need to be present or would be more natural to just omit it?
(I'm not sure about it, but seems a bit off.)

Cardosaum, line 53: "matches the _existion_ position's one", wouldn't it be '_existing_'?

frame/composable-traits/src/clearing_house.rs
```diff
@@ -38,13 +45,40 @@ pub trait ClearingHouse {
...
+	/// * Decrease in the size of an existing position, if the trade's direction is counter to the existion position's one and its magnitude is smaller than the existing postion's size
```
Cardosaum, line 54: Same comment about the ' _in_ ' particle and ' _existion_ ' applies here I think.
(it applies for the last two items in the list too)

Cardosaum (approved):
Looks good overall :rocket:

frame/clearing-house/src/math.rs
```diff
@@ -0,0 +1,75 @@
...
+) -> Result<T::Decimal, ArithmeticError> {
+	// sign(a) sign(b) | CheckedAdd
+	// ----------------------------
+	//      -1      -1 | Underflow
+	//      -1       1 | Ok
+	//       1      -1 | Ok
+	//       1       1 | Overflow
+	a.checked_add(b).ok_or_else(|| match a.is_positive() {
+		true => Overflow,
+		false => Underflow,
```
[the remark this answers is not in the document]
0xangelo: Just basic refactoring that I forgot to do. I imported `<variant>` directly to avoid specifying the whole enum path but forgot to apply it to all functions

frame/clearing-house/src/math.rs
```diff
@@ -0,0 +1,75 @@
...
+	})
+}
+
+pub fn decimal_checked_mul<T: Config>(
+	a: &T::Decimal,
+	b: &T::Decimal,
+) -> Result<T::Decimal, ArithmeticError> {
+	a.checked_mul(b).ok_or_else(|| match a.is_negative() ^ b.is_negative() {
+		true => ArithmeticError::Underflow,
+		false => ArithmeticError::Overflow,
```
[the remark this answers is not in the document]
0xangelo: That's correct. In that case, the XOR above would yield `false` and the error would be `Overflow`

frame/clearing-house/src/math.rs
```diff
@@ -0,0 +1,75 @@
...
+	a: &T::Decimal,
+	b: &T::Decimal,
+) -> Result<T::Decimal, ArithmeticError> {
+	a.checked_div(b).ok_or_else(|| {
+		if b.is_zero() {
+			ArithmeticError::DivisionByZero
+		} else {
+			match a.is_negative() ^ b.is_negative() {
+				true => ArithmeticError::Underflow,
+				false => ArithmeticError::Overflow,
```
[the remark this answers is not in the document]
0xangelo: Yes, I'm using an [XOR](https://www.tutorialspoint.com/rust/rust_bitwise_operators.htm) for that purpose

frame/composable-traits/src/clearing_house.rs
0xangelo: Not sure TBH. The way I thought about it was "A possible outcome is an increase _in/of_ the size of an existing position"

0xangelo: Good catch, that's a typo

frame/clearing-house/src/math.rs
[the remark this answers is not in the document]
Cardosaum: Oh, yeah you're right.
I confused OR and XOR :see_no_evil:

frame/clearing-house/src/lib.rs
```diff
@@ -460,6 +550,69 @@ pub mod pallet {
...
+		///   existion position's one
+		/// * Decrease in the size of an existing position, if the trade's direction is counter to
+		///   the existion position's one and its magnitude is smaller than the existing postion's
+		///   size
+		/// * Closing of the existing position, if the trade's direction is counter to the existion
+		///   position's one and its magnitude is approximately the existing postion's size
+		/// * Reversing of the existing position, if the trade's direction is counter to the
+		///   existion position's one and its magnitude is greater than the existing postion's size
+		///
+		/// ![](http://www.plantuml.com/plantuml/svg/FOuzgiD030RxTugN0zZgKna2kOUyLhm2hRJeXrm_9aMgZszWOBP8zAmXVpVM9dLGkVptp1bt0CVtUdBssYl8cscIvjfimCF6jC1TwCdGVWSeMYU7b-CWQ4BehEVIhOBWO3ml7c2JTBaCJZPTfw1-2pRIuzeF)
```
Cardosaum, line 581: For what I understood, green arrows are _reads_ and red ones are _writes_, right?
In that case, why `Asset Transfer` has only black arrows?

frame/clearing-house/src/lib.rs
```diff
@@ -598,5 +901,84 @@ pub mod pallet {
...
+							quote_asset_notional_amount: Zero::zero(),
+							last_cum_funding: market.cum_funding_rate,
+						})
+						.map_err(|_| Error::<T>::MaxPositionsExceeded)?;
+					let index = positions.len() - 1;
+					let position = positions
+						.get_mut(index)
+						.expect("Will always succeed if the above push does");
+					(position, index)
+				},
```
Cardosaum, line 946: Even tough we know that all calls to `get_mut` here should succeed, I still think it's a good idea to not panic in case of any errors.
I'd recommend changing `.expect` to `.ok_or` variations, just in case.
(curious to know what you think tho :)

Cardosaum (approved):
What a great PR! 

Learnt a lot reading this, thanks :)

I made some comments in the code, but overall it's in really good quality!
Well done :rocket:

frame/clearing-house/src/lib.rs
```diff
@@ -460,6 +547,69 @@ pub mod pallet {
...
+		/// * Creation of a whole new position in the market, if one didn't already exist
+		/// * Increase in the size of an existing position, if the trade's direction matches the
+		///   existion position's one
+		/// * Decrease in the size of an existing position, if the trade's direction is counter to
+		///   the existion position's one and its magnitude is smaller than the existing postion's
+		///   size
+		/// * Closing of the existing position, if the trade's direction is counter to the existion
+		///   position's one and its magnitude is approximately the existing postion's size
+		/// * Reversing of the existing position, if the trade's direction is counter to the
+		///   existion position's one and its magnitude is greater than the existing postion's size
```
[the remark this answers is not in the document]
0xangelo: Good catch

frame/clearing-house/src/math.rs
```diff
@@ -0,0 +1,147 @@
...
+
+	fn try_mul_(&mut self, other: &Self) -> Result<(), ArithmeticError> {
+		*self = self.try_mul(other)?;
+		Ok(())
+	}
+
+	fn try_div_(&mut self, other: &Self) -> Result<(), ArithmeticError> {
+		*self = self.try_div(other)?;
+		Ok(())
+	}
```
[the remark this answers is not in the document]
0xangelo: Good find! I'll follow that

frame/clearing-house/src/tests.rs
```diff
@@ -237,6 +299,66 @@ prop_compose! {
...
+
+prop_compose! {
+	// Assumes entry_value is positive
+	fn any_short_pnl(entry_value: FixedI128)(
+		exit_value_inner in 1..FixedI128::from(1_000_000_000).into_inner()
+	) -> (FixedI128, FixedI128) {
+		(entry_value, entry_value - FixedI128::from_inner(exit_value_inner))
+	}
+}
+
```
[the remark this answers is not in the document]
0xangelo: I see what you mean. I like the idea but I don't think `prop_assume` is the best option; from its docs:
```markdown

proptest::sugar

macro_rules! prop_assume

---

Rejects the test input if assumptions are not met.

Used directly within a function defined with proptest! or in any function
returning Result<_, TestCaseError>.

This is invoked as prop_assume!(condition, format, args...). condition
is evaluated; if it is false, Err(TestCaseError::Reject) is returned. The
message includes the point of invocation and the format message. format
and args may be omitted to simply use the condition itself as the
message.
```
Instead, I'll try restricting the input to the proptest to be an unsigned integer, and convert it into a decimal inside. That has the benefit of failing at compile time. Plus, I was already converting the input from an integer outside of the function anyway

frame/clearing-house/src/lib.rs
0xangelo: I didn't know how to color then because the accounts in the diagram are just abstractions, not the actual storage items in the balances or tokens pallets. As such, I'd rather leave the coloring for the storage read/writes that are explicit

frame/clearing-house/src/lib.rs
0xangelo: I'm not sure. Let's raise this in next office hours to get the opinion of more people

0xangelo:
Thanks a lot for the feedback @Cardosaum!

PR #903, opened 2022-04-20
Allow single node route on dex-router.

vivekvpandya:
## Checklist

### DIFF
--- a/frame/dex-router/src/lib.rs
+++ b/frame/dex-router/src/lib.rs
@@ -94,8 +94,6 @@ pub mod pallet {
 		NoRouteFound,
 		/// Unexpected node found.
 		UnexpectedNodeFound,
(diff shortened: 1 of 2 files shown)


Dom-Roth (approved):
LGTM!

vimukthi-git (approved):
LGTM

frame/dex-router/src/lib.rs
```diff
@@ -94,8 +94,6 @@ pub mod pallet {
 		NoRouteFound,
 		/// Unexpected node found.
 		UnexpectedNodeFound,
```
dzmitry-lahoda, line 96: not sure I get what error means? what node is and why as user I care about nodes? i care about pools, is that rout from A to B not found? but there is NoRouteFound

PR #906, opened 2022-04-21
DexRouter Improvements:

vivekvpandya:
- Allow single pool route.
- Allow add/remove_liquidity to single pool route.
- Fix update_route/delete_route code to only have single route for (A,B) and (B,A) pair.
- Allow swap in both directions.
- Added dummy weights.
- Updated tests accordingly.
- Adde dex-router pallet to dali runtime.

## Description
This helps frontend engineers to make progress on integration.

## Checklist


frame/dex-router/src/lib.rs
```diff
@@ -121,10 +138,124 @@ pub mod pallet {
...
+				T::AccountId,
+				T::AssetId,
+				T::PoolId,
+				T::Balance,
+				T::MaxHopsInRoute,
+			>>::update_route(&who, asset_pair, route)?;
+			Ok(())
+		}
+
+		#[pallet::weight(T::WeightInfo::exchange())]
```
dzmitry-lahoda, line 159: pallet call docs used to generate scale API and visible in polkadot js. may be add docs in next pr to help qa/integrators

frame/dex-router/src/lib.rs
```diff
@@ -242,43 +402,117 @@ pub mod pallet {
...
 		) -> Result<T::Balance, DispatchError> {
-			let route = Self::get_route(asset_pair).ok_or(Error::<T>::NoRouteFound)?;
+			let (route, reverse) = Self::get_route(asset_pair).ok_or(Error::<T>::NoRouteFound)?;
 			let mut dy_t = amount;
 			let mut dx_t = T::Balance::zero();
-			for pool_id in route.iter().rev() {
-				let currency_pair = T::Pablo::currency_pair(*pool_id)?;
+			let mut forward_iter;
+			let mut backward_iter;
+			let route_iter: &mut dyn Iterator<Item = &T::PoolId> = if !reverse {
```
dzmitry-lahoda, line 437: i believe you may do it withoyt dyn

frame/dex-router/src/lib.rs
```diff
@@ -242,43 +402,117 @@ pub mod pallet {
...
+		fn add_liquidity(
+			who: &T::AccountId,
+			asset_pair: CurrencyPair<T::AssetId>,
+			base_amount: T::Balance,
+			quote_amount: T::Balance,
+			min_mint_amount: T::Balance,
+			keep_alive: bool,
+		) -> Result<(), DispatchError> {
+			let (route, _reverse) = Self::get_route(asset_pair).ok_or(Error::<T>::NoRouteFound)?;
+			if route.len() != 1 {
```
dzmitry-lahoda, line 488: why not get pair directly here?

frame/dex-router/src/lib.rs
```diff
@@ -88,12 +101,16 @@ pub mod pallet {
 
 	#[pallet::error]
 	pub enum Error<T> {
-		/// Number of hops in route exceeded maximum limit.
+		// /// Number of hops in route exceeded maximum limit.
```
dzmitry-lahoda, line 104: comment

runtime/dali/src/lib.rs
```diff
@@ -1051,6 +1051,23 @@ impl pablo::Config for Runtime {
 	type WeightInfo = weights::pablo::WeightInfo<Runtime>;
 }
 
+parameter_types! {
+  #[derive(codec::Encode, codec::Decode, codec::MaxEncodedLen, TypeInfo)]
```
vimukthi-git, line 1055: Why is this needed?

frame/liquidity-bootstrapping/src/lib.rs
```diff
@@ -262,6 +262,7 @@ pub mod pallet {
 		InvalidSaleState,
 		InvalidAmount,
 		CannotRespectMinimumRequested,
+		LpTokenNotRequiedForLbp,
```
vimukthi-git, line 265:
```suggestion
		LpTokenNotRequiredForLbp,
```

frame/dex-router/src/lib.rs
```diff
@@ -216,22 +355,43 @@ pub mod pallet {
 			Ok(())
 		}
 
-		fn get_route(asset_pair: CurrencyPair<T::AssetId>) -> Option<Vec<T::PoolId>> {
-			DexRoutes::<T>::get(asset_pair.base, asset_pair.quote)
-				.map(|DexRoute::Direct(route)| route.into_inner())
+		/// Returns pair of route and bool to indicate if route should be used in reverse direction
```
vimukthi-git, line 358: Consider defining an enum for `Direction` and returning `Direction.Reverse` for reverse routes.

frame/dex-router/src/lib.rs
vivekvpandya: I tried but it does not work. Is it prohibited to use `dyn` in our codebase? if yes could you please explain the details.

frame/dex-router/src/lib.rs
```diff
@@ -242,43 +402,117 @@ pub mod pallet {
...
+		#[transactional]
+		fn add_liquidity(
+			who: &T::AccountId,
+			asset_pair: CurrencyPair<T::AssetId>,
+			base_amount: T::Balance,
+			quote_amount: T::Balance,
+			min_mint_amount: T::Balance,
+			keep_alive: bool,
+		) -> Result<(), DispatchError> {
+			let (route, _reverse) = Self::get_route(asset_pair).ok_or(Error::<T>::NoRouteFound)?;
```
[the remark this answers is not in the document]
vivekvpandya: I don't see any benefit.

frame/dex-router/src/lib.rs
vivekvpandya: not helpful, in current code I can check boolean in if directly which is much simpler compare to enum.

frame/dex-router/src/lib.rs
vivekvpandya: if route gets updated and still remains signle node route it will work fine, otherwise it will be an UnsupportedOperation error.

frame/composable-traits/src/dex.rs
```diff
@@ -258,26 +260,47 @@ pub trait DexRouter<AccountId, AssetId, PoolId, Balance, MaxHops> {
...
+		keep_alive: bool,
+	) -> Result<(), DispatchError>;
+	/// Remove liquidity from underlying dex pool. Works only for route with single pool.
+	fn remove_liquidity(
+		who: &AccountId,
+		asset_pair: CurrencyPair<AssetId>,
+		lp_amount: Balance,
+		min_base_amount: Balance,
+		min_quote_amount: Balance,
+	) -> Result<(), DispatchError>;
```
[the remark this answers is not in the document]
hussein-aitlahcen: Does not serve any purpose? What if another prtocol has a dependency to Amm and I want to use either Pablo or Dex router? Isn't that a purpose? This duplicated code make it way more complicated to maintain than having the same interface implemented by multiple pallet :). Could we try? For the update_route it's out of Amm scope so it's fine to have a Router interface as currently defined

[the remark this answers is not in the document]
vivekvpandya: Okay, so I tried changing it but it requires some more modification like in Amm we don't have min_receive like parameters on buy/sell but in current implementation of DexRouter I have added it. So I think first lets complete this PR and then delete dex pallets except pablo(I think we can delete them now) and refactor DexRouter to have Amm trait and it will also require some small modification to Pablo's AMM .
@hussein-aitlahcen agree?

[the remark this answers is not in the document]
hussein-aitlahcen: It's actually straight forward, the pallet must implements min_receive anyway for user to specify slippage. We currently default to 0 i believe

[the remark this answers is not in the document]
vivekvpandya: Yes but better to remove other dex first so that I don't have to update their code base, and thus keep PR size in limit so reviews are faster and less error prone.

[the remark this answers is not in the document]
hussein-aitlahcen: Ok

hussein-aitlahcen (dismissed):

frame/dex-router/src/lib.rs
```diff
@@ -159,23 +305,26 @@ pub mod pallet {
...
-			let existing_route = DexRoutes::<T>::get(k1, k2);
-			DexRoutes::<T>::insert(k1, k2, DexRoute::Direct(route.clone()));
+			Self::validate_route(asset_pair, &route)?;
+			let base_asset = asset_pair.base;
+			let quote_asset = asset_pair.quote;
+			let existing_route = Self::get_route(asset_pair);
+			// remove any route for base_asset/quote_asset or symmetric quote_asset/base_asset pair.
+			DexRoutes::<T>::remove(base_asset, quote_asset);
+			DexRoutes::<T>::remove(quote_asset, base_asset);
+			DexRoutes::<T>::insert(base_asset, quote_asset, DexRoute::Direct(route.clone()));
```
hussein-aitlahcen, line 315: I don't really understand why we try both way, can't we index by the currency pair and sort it before inserting/removing/indexing on DexRoutes? Also, unless I am missing something, you are not inserting the inverse pair here so in the end it won't work if the user is querying the swapped pair?

frame/dex-router/src/lib.rs
```diff
@@ -217,69 +371,207 @@ pub mod pallet {
...
+				Err(Error::<T>::NoRouteFound.into())
+			}
+		}
+
+		fn lp_token(pool_id: Self::PoolId) -> Result<Self::AssetId, DispatchError> {
+			let (route, _reverse) = Self::get_route(pool_id).ok_or(Error::<T>::NoRouteFound)?;
+			if route.len() != 1 {
+				return Err(Error::<T>::UnsupportedOperation.into())
+			}
+			let pablo_pool_id = route.get(0).ok_or(Error::<T>::UnsupportedOperation)?;
```
hussein-aitlahcen, line 413: I think you can simplify with slice matching
```rust
match routes[..] {
   [route] => T::Pablo:: ...,
   _ => Err(UnsuportedOperation)
}
```

hussein-aitlahcen (changes requested):

frame/dex-router/src/lib.rs
vivekvpandya: if we store sorted pair then to make any operation on pool of route, we have to compare if pool's pair and operation's pair are same or pair needed to be swapped() to make correct operation. 
Instead of above I use same pair as of pool as key in storage and in get_route() method check if request pair is same or reversed and based on this while executing operation on pool use appropriate pair. 
Please check 
```
  2         /// Returns pair of route and bool to indicate if route should be used in reverse direction                                                                                                                                      
  1         /// with assets swapped.                                                                                                                                                                                                         
378         fn get_route(asset_pair: CurrencyPair<T::AssetId>) -> Option<(Vec<T::PoolId>, bool)> {  
```
And also check tests works for both way operations. 
I have also simplified confusing code (on which you have requested changes).

frame/dex-router/src/lib.rs
```diff
@@ -544,18 +553,21 @@ pub mod pallet {
 			keep_alive: bool,
 		) -> Result<(), DispatchError> {
 			let (route, _reverse) = Self::get_route(pool_id).ok_or(Error::<T>::NoRouteFound)?;
-			if route.len() != 1 {
-				return Err(Error::<T>::UnsupportedOperation.into())
+			// if route.len() != 1 {
+			// 	return Err(Error::<T>::UnsupportedOperation.into())
+			// }
+			// let pool_id = route.get(0).ok_or(Error::<T>::UnsupportedOperation)?;
```
hussein-aitlahcen, line 559: No longer required

frame/dex-router/src/lib.rs
```diff
@@ -420,11 +425,15 @@ pub mod pallet {
 			amount: Self::Balance,
 		) -> Result<Self::Balance, DispatchError> {
 			let (route, _reverse) = Self::get_route(pool_id).ok_or(Error::<T>::NoRouteFound)?;
-			if route.len() != 1 {
-				return Err(Error::<T>::UnsupportedOperation.into())
+			// if route.len() != 1 {
+			// 	return Err(Error::<T>::UnsupportedOperation.into())
+			// }
+			// let pablo_pool_id = route.get(0).ok_or(Error::<T>::UnsupportedOperation)?;
+			// T::Pablo::get_exchange_value(*pablo_pool_id, asset_id, amount)
```
hussein-aitlahcen, line 432: Same

frame/dex-router/src/lib.rs
```diff
@@ -407,11 +409,14 @@ pub mod pallet {
 
 		fn lp_token(pool_id: Self::PoolId) -> Result<Self::AssetId, DispatchError> {
 			let (route, _reverse) = Self::get_route(pool_id).ok_or(Error::<T>::NoRouteFound)?;
-			if route.len() != 1 {
-				return Err(Error::<T>::UnsupportedOperation.into())
+			// if route.len() != 1 {
+			// 	return Err(Error::<T>::UnsupportedOperation.into())
+			// }
+			// let pablo_pool_id = route.get(0).ok_or(Error::<T>::UnsupportedOperation)?;
```
hussein-aitlahcen, line 415: Same

frame/dex-router/src/lib.rs
hussein-aitlahcen: Alright :+1:

hussein-aitlahcen (approved):

KaiserKarel (approved):

PR #914, opened 2022-04-23
Fix ED related issue for DEX for creating new tokens.

vivekvpandya:
## Issue
In integration test while creating new token , TokenError::BelowMinimum is hit.

## Checklist


hussein-aitlahcen (dismissed):

dzmitry-lahoda:
this is #xcmp security issue, so please create issue to track fix for that

dzmitry-lahoda (changes requested):
please track issue as it is security issue for #xcmp

dzmitry-lahoda:
it is clear why MAX was not option, but why instead of 0 not put 1_000_000? seems reasonable. at least that will not break #xcmp tested ported from #acala

hussein-aitlahcen:
> it is clear why MAX was not option, but why instead of 0 not put 1_000_000? seems reasonable. at least that will not break #xcmp tested ported from #acala

I think it is fair to hardcode a value for now but a ticket tracking that will be useful. We probably want to have custom ED per asset

hussein-aitlahcen (dismissed):

dzmitry-lahoda (approved):

andor0 (approved):

PR #926, opened 2022-04-28
Update/democracy

ysfkel:
## Issue
*Fork latest version and replace our existing fork (diff is too big to upgrade)
Have both our multi-currency fork and native one on the runtimes to avoid being stuck by our fork*

## Checklist


dzmitry-lahoda (dismissed):

frame/democracy/README.md
```diff
@@ -1,40 +1,135 @@
...
+- [`Call`](https://docs.rs/pallet-democracy/latest/pallet_democracy/enum.Call.html)
 
-#### Changes
+## Overview
 
-- Proposals are identified by their hash + AssetId.
-- Voting is performed using the proposals AssetId to identify what token to lock.
-- During dispatching, GovernanceRegistry is used to lookup the associated Origin.
-- Different Currency traits are used, such as MutateHold.
-- Orml traits MultiCurrency, MultiLockableCurrency and MultiReservableCurrency are used to support multiple assets.
```
dzmitry-lahoda, line 12: please document WHY fork is different from original and please ad hash number into docks you forked from

frame/assets-registry/Cargo.toml
```diff
@@ -3,7 +3,7 @@ authors = ["Composable Developers"]
 edition = "2021"
 homepage = "https://composable.finance"
 name = "pallet-assets-registry"
-version = "0.1.0"
+version = "0.0.1"
```
KaiserKarel, line 6: why are we updating this in this PR?

frame/assets/src/tests/traits.rs
```diff
@@ -320,7 +320,7 @@ mod multicurrency {
 
 		#[test]
 		fn test_minimum_balance_implementation(
-			account in accounts(),
+			_account in accounts(),
```
KaiserKarel, line 323: same here, why is this changing?

frame/democracy/Cargo.toml
```diff
@@ -27,9 +27,9 @@ serde = { version = "1.0.136", optional = true, features = ["derive"] }
 sp-io = { default-features = false, git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.18" }
 sp-runtime = { default-features = false, git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.18" }
 sp-std = { default-features = false, git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.18" }
+orml-tokens = { git = "https://github.com/open-web3-stack/open-runtime-module-library", rev = "2b5d4ce1d08fb54c0007c2055653892d2c93a92e", default-features = false }
```
KaiserKarel, line 30: I think orml-tokens only needs to be a dev-dependency right? Or are we using it in the benches? Then we can still put it behind the benchmarking flag

frame/assets/src/tests/traits.rs
ysfkel: It is not used and the linter is complaining and suggesting the _ for unused variables. I guess it should be removed if it is not intended to be used?

frame/assets-registry/Cargo.toml
ysfkel: Updated now , will push , I though it was the fix to some dependency issue

frame/democracy/Cargo.toml
ysfkel: There was a build error with orml-tokens under dev-dependencies, that fixed it, moved it to dev-dependencies , its fixed now

frame/democracy/README.md
ysfkel: @dzmitry-lahoda  does the comment on line 12 satisfy why it is forked ? "Orml traits MultiCurrency, MultiLockableCurrency and MultiReservableCurrency are used to support multiple assets." or what additions do you suggest ? 

Please explain this "hash number into docks you forked from"

frame/democracy/Cargo.toml
ysfkel: it is used in the test runtime

frame/assets/src/tests/traits.rs
ysfkel: I have removed it altogether , it is unused

frame/democracy/README.md
ysfkel: Pushed new commit with democracy commit hash and url

dzmitry-lahoda (approved):

frame/democracy/Cargo.toml
```diff
@@ -4,37 +4,37 @@ description = "FRAME pallet for democracy with multi currency support"
 edition = "2018"
 homepage = "https://composable.finance"
 license = "Apache-2.0"
-name = "pallet-democracy"
+name = "pallet-multicurrency-democracy"
```
KaiserKarel, line 7:
```suggestion
name = "pallet-multitenant-democracy"
```

KaiserKarel: Or pallet-multi-democracy

frame/democracy/src/benchmarking.rs
```diff
@@ -29,21 +29,22 @@ use frame_system::{Pallet as System, RawOrigin};
...
 fn assert_last_event<T: Config>(generic_event: <T as Config>::Event) {
 	frame_system::Pallet::<T>::assert_last_event(generic_event.into());
 }
 
 fn funded_account<T: Config>(name: &'static str, index: u32) -> T::AccountId {
 	let caller: T::AccountId = account(name, index, SEED);
-	//T::Currency::make_free_balance_be(&caller, BalanceOf::<T>::max_value());
+	// T::Currency::make_free_balance_be(&caller, BalanceOf::<T>::max_value());
 	caller
 }
```
KaiserKarel, line 49: dus the `account` function add funding?

frame/democracy/src/conviction.rs
```diff
@@ -1,6 +1,6 @@
 // This file is part of Substrate.
```
KaiserKarel, line 1:
```suggestion

```

frame/democracy/src/lib.rs
```diff
@@ -1232,13 +1193,26 @@ pub mod pallet {
 			let now = <frame_system::Pallet<T>>::block_number();
 			let (voting, enactment) = (T::VotingPeriod::get(), T::EnactmentPeriod::get());
 			let additional = if who == provider { Zero::zero() } else { enactment };
-			ensure!(now >= since + voting + additional, Error::<T>::TooEarly);
+			ensure!(
+				now >= since.saturating_add(voting).saturating_add(additional),
+				Error::<T>::TooEarly
+			);
```
KaiserKarel, line 1199: Not needed to add in this PR, but chrono has methods on time structs so that you can do:

```
ensure!( since.is_before(now), Error::<T>::TooEarly )
```

@PoisonPhang you reckon we should add an Extension trait to composable_support to implement this for `Moment`?

frame/democracy/src/benchmarking.rs
ysfkel: not clear about this ?

frame/democracy/Cargo.toml
ysfkel: fixed this, moved to dev-dependency and updated feautures

ysfkel:
Worked on this ticket to add proptests and rebased main. requires reviews

ysfkel:
Rebased main. requires review

frame/assets/src/lib.rs
```diff
@@ -99,7 +99,6 @@ pub mod pallet {
 	use frame_system::{ensure_root, ensure_signed, pallet_prelude::OriginFor};
 	use orml_traits::GetByKey;
 	use sp_runtime::DispatchError;
-
```
KaiserKarel, line 102: why remove this line?

frame/democracy/src/lib.rs
```diff
@@ -528,20 +502,11 @@ pub mod pallet {
...
 	>;
 
-	/// Accounts for which there are locks in action which may be removed at some point in the
-	/// future. The value is the block number at which the lock expires and may be removed.
-	///
-	/// TWOX-NOTE: OK ― `AccountId` is a secure hash.
-	#[pallet::storage]
-	#[pallet::getter(fn locks)]
-	pub type Locks<T: Config> = StorageMap<_, Twox64Concat, T::AccountId, T::BlockNumber>;
-
```
KaiserKarel, line 538: Why did we remove the Locks? Howcome those are not necessary when going multicurrency?

frame/democracy/src/lib.rs
```diff
@@ -528,20 +502,11 @@ pub mod pallet {
...
 	/// True if the last referendum tabled was submitted externally. False if it was a public
 	/// proposal.
 	// TODO: There should be any number of tabling origins, not just public and "external"
 	// (council). https://github.com/paritytech/substrate/issues/5322
 	#[pallet::storage]
-	// Usage of ValueQuery was audited by parity and allowed for now.
```
KaiserKarel, line 544: Please leave a comment describing why we use ValueQuery

frame/democracy/src/lib.rs
```diff
@@ -1866,31 +1843,34 @@ impl<T: Config> Pallet<T> {
...
+				let err_amount = T::NativeCurrency::unreserve(&provider, deposit);
+				debug_assert!(err_amount.is_zero());
+				Self::deposit_event(Event::<T>::PreimageUsed {
+					proposal_hash: proposal_id.hash,
+					provider,
+					deposit,
+				});
+
+				let res = proposal
+					.dispatch(frame_system::RawOrigin::Root.into())
```
KaiserKarel, line 1855: We need to use the governance registry for this, we cannot simply dispatch using Root for any AssetId. Please have a look at how the previous fork of the democracy pallet handles this. Alternatively, we can get rid of the governance registry, and instead convert the AssetId to an AccountId, and use that to dispatch.

KaiserKarel: that would allow us to deprecate the governance-registry

frame/democracy/src/tests/cancellation.rs
```diff
@@ -18,79 +18,94 @@
...
 use super::*;
+use frame_support::traits::{fungible::Mutate as FungibleMutet, fungibles::Mutate};
+
+proptest! {
+	#![proptest_config(ProptestConfig::with_cases(1000))]
+
+	#[test]
+	fn cancel_referendum_should_work(
+	asset_id in valid_asset_id(),
+	balance1 in valid_amounts_without_overflow_1()) {
```
KaiserKarel, line 29: why call this `balance1`

frame/democracy/src/tests/fast_tracking.rs
```diff
@@ -19,33 +19,56 @@
...
+					tally: Tally { ayes: 0, nays: 0, turnout: 0 },
+				})
+			);
+		});
+	}
+
+	#[test]
+	fn fast_track_referendum_fails_when_no_simple_majority(
+		asset_id in valid_asset_id(),
+		balance1 in valid_amounts_without_overflow_1()) {
```
KaiserKarel, line 61: why call this balance1?

KaiserKarel (changes requested):
I'm still not seeing tests which actually verify that our multi-tenancy is correct. Furthermore, the way we dispatch the proposal right now is incorrect. See my comments

frame/democracy/src/benchmarking.rs
ysfkel: ill just remove it , it was part of the original tests that came from the substrate repo, they commented it out that way

ysfkel: I mean part of the previous democracy , ill update and push

ysfkel: I just pushed an update , removed funded_account and called account directly

frame/democracy/src/lib.rs
ysfkel: That should be from the merge, the latest substrate democracy pallet does not use it

frame/democracy/src/lib.rs
ysfkel: The latest substrate democracy pallet changed to ValueQuery, should we add more comments ?

frame/democracy/src/lib.rs
ysfkel: Okay , I'm lookt at that and make necessary updates to this

frame/democracy/src/tests/cancellation.rs
ysfkel: ill update this and push

frame/democracy/src/benchmarking.rs
ysfkel: l'll update this and

frame/democracy/src/tests/fast_tracking.rs
ysfkel: l'll update this and push

frame/democracy/src/lib.rs
ysfkel: This has been resolved

frame/democracy/src/tests/fast_tracking.rs
ysfkel: this has been resolved

frame/democracy/src/tests/cancellation.rs
ysfkel: this has been resolved

frame/assets/src/lib.rs
ysfkel: Thats from IDE onsave, ill update

frame/democracy/src/lib.rs
KaiserKarel: Yeah we internally find ValueQuery dangerous, so at least add a comment that it is used because substrate has it.

frame/democracy/src/lib.rs
KaiserKarel: Why not start out this PR from the latest code then, which means we have far fewer diffs?

frame/assets/src/lib.rs
ysfkel: this has been resolved

frame/democracy/src/lib.rs
ysfkel: you mean to get the latest substrate democracy pallet ?

frame/democracy/src/lib.rs
ysfkel: ill update

benluelo:
Will re-add the benchmarks label once the PR is passing all tests

hussein-aitlahcen:
There are commits which are basically unresolved conflicts from a rebase, i.e. accepted and staged conflicts...

ysfkel:
> I'm not entirely sure why we have some commits named `conflicts` and `rebase` if you rebased, as fixing the conflicts is done WHILE rebasing and does not involve pushing anything AFTER having rebased 🤔

There were some issues from the IDE in displaying the rebase, the markings were broken and had to fix those manually

ysfkel:
> I'm still not seeing tests which actually verify that our multi-tenancy is correct. Furthermore, the way we dispatch the proposal right now is incorrect. See my comments

  - The vote function does not accept asset_id, so test cases checking for correct asset_id are not possible.
  - The second test you suggested earlier which checks for reducing the balance of the asset can b achieved, however 
       looking at the code for  [try_vote here](https://github.com/paritytech/substrate/blob/158cdfd1a43a122f8cfbf70473fcd54a3b418f3d/frame/democracy/src/tests.rs#L274 ) and the corresponding test written by the substrate team, it looks like the first check in the try_vote function which appears to compare balances of the same asset will never fail, in their case, it compares the balance of the native asset to the same, as it can be seen in the aye function [here](https://github.com/paritytech/substrate/blob/158cdfd1a43a122f8cfbf70473fcd54a3b418f3d/frame/democracy/src/tests.rs#L274 ), it returns the native asset balance for the specified user, which is the same user calling executing vote. and try_vote compares this same native asset balance with itself, so it is comparing `fee_balance(userA) <= free_balance(userA) `
which is always true,  so im not sure what this comparison is trying to achieve. perhaps you can provide more insight on this .

ysfkel:
> > I'm still not seeing tests which actually verify that our multi-tenancy is correct. Furthermore, the way we dispatch the proposal right now is incorrect. See my comments
> 
> * The voting functionality is tested by the voting.rs tests, thats was not converted to proptest 
> * The vote function does not accept asset_id, so test cases checking for correct asset_id are not possible.
> * The second test you suggested earlier which checks for reducing the balance of the asset can b achieved, however
>   looking at the code for  [try_vote here]
>   (https://github.com/paritytech/substrate/blob/158cdfd1a43a122f8cfbf70473fcd54a3b418f3d/frame/democracy/src/lib.rs#L1366 ) and the corresponding test written by the substrate team, it looks like the first check in the try_vote function which appears to compare balances of the same asset will never fail, in their case, it compares the balance of the native asset to the same, as it can be seen in the aye function [here](https://github.com/paritytech/substrate/blob/158cdfd1a43a122f8cfbf70473fcd54a3b418f3d/frame/democracy/src/tests.rs#L274), it returns the native asset balance for the specified user, which is the same user calling executing vote. and try_vote compares this same native asset balance with itself, so it is comparing `fee_balance(userA) <= free_balance(userA) `
>   which is always true,  so im not sure what this comparison is trying to achieve. perhaps you can provide more insight on this .  I think an explanation of the purpose of  vote configuration `AccountVote<BalanceOf<T>>` will clear this up, it seens like the sender of the

ysfkel:
> There were some issues from the IDE in displaying the rebase, the markings were broken and had to fix those manually

There were some issues from the IDE in displaying the rebase conflicts properly, the markings were broken and had to fix those manually because the IDE didn't pick them up. I am paying attention to such the IDE errors now

KaiserKarel:
> > I'm still not seeing tests which actually verify that our multi-tenancy is correct. Furthermore, the way we dispatch the proposal right now is incorrect. See my comments
> 
> * The voting functionality is tested by the voting.rs tests . [this is one test](https://github.com/ComposableFi/composable/blob/31a24b3647fb8abcc0a7766c44284a9095616b3c/frame/democracy/src/tests/voting.rs#L60) that tests voting on a referendum
> * Also added [this test](https://github.com/ComposableFi/composable/blob/31a24b3647fb8abcc0a7766c44284a9095616b3c/frame/democracy/src/tests/public_proposals.rs#L164) to ensure propose, vote and remove
> * The vote function does not accept asset_id, so test cases checking for correct asset_id are not possible.
> * Let me know if you have any further tests in mind

1. Looking at that test, it uses DEFAULT_ASSET, and only tests the happy path, meaning that we have no checks to see if indeed using a different asset to vote works/not works.

2. "The vote function does not accept asset_id, so test cases checking for correct asset_id are not possible.". They are, by minting a different asset into the account to check that it fails. :) Proptest is great for this, instead of hardcoding

3. I'd like to see a test that verifies that we dispatch with the origin obtained from governance registry.

ysfkel:
> added comments to indicate valuequery is used in the commit at the ti…

ysfkel:
> There are commits which are basically unresolved conflicts from a rebase, i.e. accepted and staged conflicts...

There were some issues (unexpected) from the IDE in displaying the rebase conflicts properly, the markings were broken,  fix those manually because the IDE didn't pick them.  they are resolved now

PR #936, opened 2022-05-01
#CU-2vxc2n7  assets improvements for making non 12 decimals transfers and democracy ceremomy

dzmitry-lahoda:
we need to transfer USDT from statemine parachain, for that we need
- ability of root/democracy to vote adding remote asset
- ability to define ED of remote asset (see readmess of currency factory and assets registry for relevant info)
- and non 12 decimals

this PR does allows for that and improve other things
- proper assets registry
- updates docs and all tests
- some tests refactoring
- all should work as before


frame/assets-registry/src/lib.rs
```diff
@@ -135,53 +99,16 @@ pub mod pallet {
 
 	/// Mapping (local asset, foreign asset) to a candidate status.
```
vivekvpandya, line 100: May be this comment needs update.

Makefile.toml
```diff
@@ -58,15 +58,16 @@ cargo +stable test --package local-integration-tests --features local-integratio
 workspace = false
 
 [tasks.run-local-integration-tests-debug]
+description = "fastest way to build and debug runtime in simulator"
 script = '''
-RUST_BACKTRACE=full RUST_LOG=trace,parity-db=warn,trie=warn,runtime=trace,substrate-relay=trace,bridge=trace,xcmp=trace cargo +stable test  --package local-integration-tests --features local-integration-tests --features picasso  --no-default-features -- --nocapture --test-threads=1
+RUST_BACKTRACE=full RUST_LOG=trace,parity-db=warn,trie=warn,runtime=trace,substrate-relay=trace,bridge=trace,xcmp=trace cargo +stable ltest  --package local-integration-tests --features local-integration-tests --features picasso  --no-default-features -- --nocapture --test-threads=1
```
andor0, line 63: `ltest` ?

frame/assets-registry/src/lib.rs
```diff
@@ -190,298 +117,110 @@ pub mod pallet {
...
-			let _ = ensure_signed(origin.clone())?;
-			Self::ensure_admins_only(origin)?;
-			ensure!(
-				<LocalToForeign<T>>::contains_key(local_asset_id),
-				Error::<T>::LocalAssetIdNotFound
-			);
-
-			<ForeignAssetMetadata<T>>::insert(local_asset_id, metadata);
-			Self::deposit_event(Event::AssetMetadataUpdated(local_asset_id));
+			// note: does not validates if assets exists, not clear what is expected in this case
```
andor0, line 169: can add
```
ensure!(
  <LocalToForeign<T>>::contains_key(asset_id),
  Error::<T>::LocalAssetIdNotFound
);
```

frame/assets-registry/src/lib.rs
```diff
@@ -61,71 +67,29 @@ pub mod pallet {
...
-			+ PartialEq
-			// we wrap non serde type, so until written custom serde, cannot handle that
-			// + MaybeSerializeDeserialize
 			+ Debug
 			+ Clone
 			+ Default
 			+ TypeInfo;
 
 		/// The origin which may set local and foreign admins.
 		type UpdateAdminOrigin: EnsureOrigin<Self::Origin>;
```
andor0, line 77: `UpdateAssetOrigin` ?

frame/assets-registry/src/lib.rs
dzmitry-lahoda: second ensure, really it may be ok to have different ids for same remote.
first, we add assets which are special id from well known, but well known assets are no need to be in list. will attach link to clickup to solve well knowns as usual. than can do check number 1. also as adding is pemissioned, its not a security  to add like that.

frame/assets-registry/src/lib.rs
dzmitry-lahoda: ok, will do it

frame/assets-registry/src/lib.rs
dzmitry-lahoda: was merged from main, will update

frame/assets-registry/src/lib.rs
dzmitry-lahoda: as of not method is ok, can fix first clause after https://app.clickup.com/t/1wtyb38 . second clause is ok as used to handle well knowns.

andor0 (approved):

frame/composable-traits/src/bounded.rs
```diff
@@ -0,0 +1,86 @@
...
+// u32 from 1 to 32 (decimals exponent)
+
+// cannot make const generic because `the type must not depend on the parameter`
+// there is no solution in crates which works with substrate numbers in substrate context
+trait Bound<
+	Inner: PartialOrd + Encode + Decode + MaxEncodedLen + TypeInfo + MaybeSerializeDeserialize + Sized,
+>
+{
+	const MINIMAL: Inner;
+	const MAXIMAL: Inner;
```
benluelo, line 20: MAXIMUM, MINIMUM?

Maybe even just MIN/ MAX would be fine

andor0 (approved):

benluelo (approved):

PR #942, opened 2022-05-03
Update shell.nix

andor0:
## Issue
https://app.clickup.com/t/2a36jxv

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -2,32 +2,26 @@ let
   mozillaOverlay =
     import (builtins.fetchGit {
       url = "https://github.com/mozilla/nixpkgs-mozilla.git";
-      rev = "57c8084c7ef41366993909c20491e359bbb90f54";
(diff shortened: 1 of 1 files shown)


dzmitry-lahoda:
is nix packaged version works right now? is it possible to improve version upgrade in all places at once or only string replace guide is best possible?

andor0:
> is nix packaged version works right now? is it possible to improve version upgrade in all places at once or only string replace guide is best possible?

yes, we use this version now.
I think, can move settings of rust version to `nix/rust.nix` and use it in `*.nix` files.I try to do it in one of next PRs.

dzmitry-lahoda (approved):

hussein-aitlahcen:
Ideally we should remove the `shell.nix` and add a proper shell in the `flake.nix`

hussein-aitlahcen (approved):

andor0:
> Ideally we should remove the `shell.nix` and add a proper shell in the `flake.nix`

I will do it in one of next PRs.

PR #950, opened 2022-05-04
CU-296cyym - Xcvm/cosmwasm via fork of contract/sandbox

hussein-aitlahcen:


ComposableFiB0t:
Task linked: [CU-296cyym XCVM + COSMWASM Cross-chain demo prototype](https://app.clickup.com/t/296cyym)

frame/contracts/Cargo.toml
```diff
@@ -0,0 +1,103 @@
+[package]
+authors = ["Parity Technologies <admin@parity.io>", "Composable Developers"]
+description = "FRAME pallet for WASM contracts"
+edition = "2021"
+homepage = "https://substrate.io"
+license = "Apache-2.0"
+name = "pallet-contracts"
+readme = "README.md"
+repository = "https://github.com/paritytech/substrate/"
```
KaiserKarel, line 9:
```suggestion
repository = "https://github.com/composablefi/composable/"
```

frame/contracts/src/benchmarking/sandbox.rs
```diff
@@ -0,0 +1,55 @@
...
+	/// Creates an instance from the supplied module and supplies as much memory
+	/// to the instance as the module declares as imported.
+	fn from(module: &WasmModule<T>) -> Self {
+		let mut env_builder = EnvironmentDefinitionBuilder::new();
+		let memory = module.add_memory(&mut env_builder);
+		let instance = Instance::new(&module.code, &env_builder, &mut ())
+			.expect("Failed to create benchmarking Sandbox instance");
+		Self { instance, _memory: memory }
+	}
+}
```
KaiserKarel, line 55: Is there a reason for us to implement from instead of a fallible constructor? To me, `From` is for similar types; such as u64 -> u32 etc, here I think a `new -> Result<Self>` is more appropriate.

frame/contracts/src/migration.rs
```diff
@@ -0,0 +1,275 @@
...
+		StorageVersion::new(6).put::<Pallet<T>>();
+	}
+
+	if version < 7 {
+		weight = weight.saturating_add(v7::migrate::<T>());
+		StorageVersion::new(7).put::<Pallet<T>>();
+	}
+
+	weight
+}
```
KaiserKarel, line 57: I don't think we actually need this logic?

KaiserKarel (approved):

frame/contracts/src/migration.rs
hussein-aitlahcen: This is legacy code that will be wiped once we can correctly execute cosmwasm contracts. I didn't drop all ink related code yet

frame/contracts/src/benchmarking/sandbox.rs
hussein-aitlahcen: This is legacy code that will probably need some refactoring anyway, I didn't touch bench/tests yet

PR #847, opened 2022-04-04
Ben/storage nonce

benluelo:
## Issue
No clickup ticket, see this thread: https://composablefinance.slack.com/archives/C02MSA16DCP/p1647439713464129

## Description
Implements a wrapper over incrementing a storage nonce.

## Checklist

- [ ] ~~I have updated the cargo docs to reflect changes made by this PR~~ _(internal change, not applicable)_
- [ ] ~~I have updated the `book/` to reflect changes made by this PR~~ _(internal change, not applicable)_

## Notes
I refactored to use `try_increment()` where I was able to do so easily, however there are several instances that require more refactoring/ knowledge about the specific usecase. I've annotated these instances with `// REVIEW:` comments, if anyone is able to advise me on those that would be great!

PR #954, opened 2022-05-04
Ben/storage counter

benluelo:
## Description
Adds an extension trait(s) for `StorageValue`s that are used as a `Counter`.

This builds off of the `Increment` trait added in the previous PR that added `Nonce` (#847), adding `Decrement`. `Counter` implements both, while `Nonce` only implements `Increment`.

## Question

https://github.com/ComposableFi/composable/blob/b096052bff49fda7d918971be6ca00a57a53dbb5/frame/lending/src/lib.rs#L1121-L1187

Is it more performant to have the majority of the function inside the `try_mutate` closure? It feels unnecessary if the extrinsic is transactional.

## Checklist

- [ ] ~~I have updated the cargo docs to reflect changes made by this PR _(if applicable)_~~
- [ ] ~~I have updated the `book/` to reflect changes made by this PR _(if applicable)_~~


KaiserKarel (dismissed):

frame/privilege/src/lib.rs
```diff
@@ -148,7 +161,70 @@ pub mod pallet {
...
+
+	// pub trait IntoDispatchError<TError, T>
+	// where
+	// 	Self: Into<TError>,
+	// 	TError: Into<DispatchError>,
+	// {
+	// 	fn into_dispatch_error(self) -> DispatchError;
+	// }
+
+	// impl<TError, T> IntoDispatchError<TError, T> for T
```
dzmitry-lahoda, line 192: why it is commented?

frame/privilege/src/lib.rs
```diff
@@ -220,12 +296,11 @@ pub mod pallet {
 			group: PrivilegedGroupOf<Self>,
 			privilege: Privilege,
 		) -> Result<Self::GroupId, DispatchError> {
-			ensure!(GroupCount::<T>::get() < T::MaxGroup::get(), Error::<T>::TooManyGroup);
 			GroupId::<T>::try_mutate(|previous_group_id| {
 				let group_id = previous_group_id.next();
 				*previous_group_id = group_id;
 
-				GroupCount::<T>::mutate(|x| *x += 1);
+				GroupCount::<T>::increment()?;
```
dzmitry-lahoda, line 303: good

frame/privilege/src/lib.rs
benluelo: tried to do something that didn't work, forgot to remove it. removed now :+1:

frame/privilege/src/lib.rs
```diff
@@ -148,7 +161,70 @@ pub mod pallet {
...
 	#[allow(clippy::disallowed_types)]
-	pub type GroupCount<T: Config> = StorageValue<_, u32, ValueQuery>;
+	pub type GroupCount<T: Config> = StorageValue<
+		_,
+		u32,
+		ValueQuery,
+		Counter<ZeroInit, IncrementToMax<T::MaxGroup, TooManyGroup, Error<T>>, SafeDecrement>,
+	>;
+
+	error_to_pallet_error!(
```
dzmitry-lahoda, line 171: overall ok, so I dislike macro. copy paste TooManyGroup 3 times around (macro 2, Error<T> 1), it kind of flat outs error flat into mod. 

better map_err on each counter in case of overflow.


```
error_mod!(
#[pallet::error]
	pub enum Error<T> {
           Foo,
           Bar,
         }

mod error {
   struct Foo(());
   struct Bar(());
}

```

frame/composable-support/src/abstractions/utils/mod.rs
```diff
@@ -0,0 +1,37 @@
...
+/// ```
+///
+/// Note that this assumes that the pallet's `Error` and `Config` types are in scope and not
+/// renamed.
+#[macro_export]
+macro_rules! error_to_pallet_error {
+	($($name:ident -> $to:ident;)+) => {
+		$(
+			#[derive(::core::fmt::Debug, ::core::default::Default, ::core::cmp::PartialEq)]
+			pub struct $name;
```
dzmitry-lahoda, line 22: pub struct $name(()) ?

and put all errors into mod.

not this pr. i am kind of dislike this macro)

benluelo: It's needed for the IncrementToMax incrementor, since enum variants aren't standalone types.

> and put all errors into mod.

What do you mean here?

frame/privilege/src/lib.rs
benluelo: The idea was to remove the need for `map_err`, since it's just boilerplate. I can try to think of a different approach, but I really think this is the best bet.

Perhaps having the macro just take in one ident?

```rust
error_to_pallet_error!(
    TooManyGroup,
)
```

It would force the name of the IncrementToMax error but I think that's ok, since it's all internal anyways.

dzmitry-lahoda: ok oke if `error_mod` will generate for each Error type, and type with private constructor (i.e. you cannot create instance of type).

dzmitry-lahoda: so I am not sure about having macro which needs to map errors manually additionally.

frame/composable-support/src/abstractions/utils/mod.rs
dzmitry-lahoda: see error disucussion

frame/privilege/src/lib.rs
benluelo: How about this?

```rust
error! {
    #[pallet::error]
    #[derive(PartialEqNoBound)]
    pub enum Error<T> {
        /// hello
        #(error)
        ZeroInitIncrementToMaxValueTooLarge,
        OneInitIncrementToMaxValueTooLarge,
        DefaultInitIncrementToMaxValueTooLarge,
    }
}
```

Expanding to:
```rust
pub mod macros {
    #[pallet::error]
    pub enum Error<T> {
        /// hello
        ZeroInitIncrementToMaxValueTooLarge,
        OneInitIncrementToMaxValueTooLarge,
        DefaultInitIncrementToMaxValueTooLarge,
    }

    pub struct ZeroInitIncrementToMaxValueTooLarge;

    impl<T: Config> From<ZeroInitIncrementToMaxValueTooLarge> for Error<T> {
        fn from(_: ZeroInitIncrementToMaxValueTooLarge) -> Error<T> {
            Error::<T>::ZeroInitIncrementToMaxValueTooLarge
        }
    }
}
```

The `#(error)` is arbitrary, all that's needed is some sort of marker. (`#[error]` can't be used due to parsing ambiguities with doc comments. It could be done, but I don't really want to write a TT muncher.)

Dom-Roth (approved):

KaiserKarel (approved):

PR #959, opened 2022-05-05
Add 'lending_currentInterestRate' RPC method

andor0:
## Issue
https://app.clickup.com/t/29gp23n

## Description
I added `lending_currentInterestRate` RPC method. We can add it to RPC of node after we will add pallet-lending to `Picasso` and `Composable` runtimes.

### DIFF
--- a/frame/lending/rpc/src/lib.rs
+++ b/frame/lending/rpc/src/lib.rs
@@ -1,0 +1,66 @@
+use codec::Codec;
+use composable_support::rpc_helpers::SafeRpcWrapper;
+use composable_traits::defi::Rate;
+use core::{fmt::Display, str::FromStr};
+use jsonrpc_core::{Error as RpcError, ErrorCode, Result as RpcResult};
+use jsonrpc_derive::rpc;
+use lending_runtime_api::LendingRuntimeApi;
+use sp_api::ProvideRuntimeApi;
+use sp_blockchain::HeaderBackend;
+use sp_runtime::{generic::BlockId, traits::Block as BlockT};
+use sp_std::sync::Arc;
+
+#[rpc]
+pub trait LendingApi<BlockHash, MarketId>
+where
+	MarketId: FromStr + Display,
+{
+	#[rpc(name = "lending_currentInterestRate")]
+	fn current_interest_rate(
+		&self,
+		market_id: SafeRpcWrapper<MarketId>,
+		at: Option<BlockHash>,
+	) -> RpcResult<SafeRpcWrapper<Rate>>;
+}
+
+pub struct Lending<C, Block> {
+	client: Arc<C>,
+	_marker: sp_std::marker::PhantomData<Block>,
+}
+
+impl<C, M> Lending<C, M> {
+	pub fn new(client: Arc<C>) -> Self {
+		Self { client, _marker: Default::default() }
+	}
+}
+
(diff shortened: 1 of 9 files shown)


frame/lending/src/lib.rs
```diff
@@ -1773,4 +1775,27 @@ pub mod pallet {
...
+		let total_borrowed_from_market_excluding_interest =
+			Pallet::<T>::total_borrowed_from_market_excluding_interest(&market_id)?;
+		let total_available_to_be_borrowed =
+			Pallet::<T>::total_available_to_be_borrowed(&market_id)?;
+		let utilization_ratio = Pallet::<T>::calculate_utilization_ratio(
+			total_available_to_be_borrowed,
+			total_borrowed_from_market_excluding_interest,
+		)?;
+
+		Markets::<T>::try_mutate(market_id, |market_config| {
```
hussein-aitlahcen, line 1791: Why do we try mutate? If get borrow rate is mut then the storage will be mutated? Perhaps just try_get and mutate after, unless I missee something 🤔

andor0: Yes, you are right, we can use `try_get` here. Fixed.

frame/lending/src/lib.rs
```diff
@@ -1773,4 +1775,24 @@ pub mod pallet {
 		pub(crate) accrued_increment: T::Balance,
 		pub(crate) new_borrow_index: FixedU128,
 	}
+
+	/// Retrieve the current interest rate for the given `market_id`.
+	pub fn current_interest_rate<T: Config>(market_id: MarketId) -> Result<Rate, DispatchError> {
```
vimukthi-git, line 1780: It would be best to write some unit test(s) for this function,

runtime/common/Cargo.toml
```diff
@@ -36,6 +36,7 @@ collective = { package = "pallet-collective", git = "https://github.com/parityte
 treasury = { package = "pallet-treasury", git = "https://github.com/paritytech/substrate", default-features = false , branch = "polkadot-v0.9.18" }
 
 composable-traits = { path = "../../frame/composable-traits", default-features = false }
+pallet-lending = { path = "../../frame/lending", default-features = false }
```
vimukthi-git, line 39: Seems unnecessary to import?

runtime/common/src/lib.rs
```diff
@@ -26,6 +26,7 @@ pub use types::*;
 mod types {
 	use codec::{Decode, Encode, MaxEncodedLen};
 	use core::fmt::Debug;
+	pub use pallet_lending::MarketId;
```
vimukthi-git, line 29: Seems unnecessary to import?

frame/lending/rpc/src/lib.rs
```diff
@@ -0,0 +1,66 @@
+use codec::Codec;
+use composable_support::rpc_helpers::SafeRpcWrapper;
+use composable_traits::defi::Rate;
+use core::{fmt::Display, str::FromStr};
+use jsonrpc_core::{Error as RpcError, ErrorCode, Result as RpcResult};
+use jsonrpc_derive::rpc;
+use lending_runtime_api::LendingRuntimeApi;
```
vimukthi-git, line 7: It seems this runtime API has not been declared?

runtime/common/src/lib.rs
andor0: Removed it.

PR #963, opened 2022-05-06
CU-2953d54 

vivekvpandya:
In handle_payout() use T::Currency::repatriate_reserved() to transfer slashed amount to Treasury account.
Update tests accordingly and new test reported by Halborn in audit report.

## Issue
https://app.clickup.com/t/2953d54

## Description
Minor cleanup for unused variables and imports.

## Checklist

### DIFF
--- a/frame/oracle/src/lib.rs
+++ b/frame/oracle/src/lib.rs
@@ -49,7 +49,7 @@ pub mod pallet {
 		dispatch::{DispatchResult, DispatchResultWithPostInfo},
 		pallet_prelude::*,
 		traits::{
-			Currency, EnsureOrigin,
+			BalanceStatus, Currency, EnsureOrigin,
 			ExistenceRequirement::{AllowDeath, KeepAlive},
 			ReservableCurrency,
 		},
@@ -153,6 +153,8 @@ pub mod pallet {
 		type MaxAnswerBound: Get<u32>;
 		/// Upper bound for total assets available for the oracle
 		type MaxAssetsCount: Get<u32>;
+		/// Slashed stakes are transfered to treasury.
+		type TreasuryAccount: Get<Self::AccountId>;
 
 		#[pallet::constant]
 		type MaxHistory: Get<u32>;
@@ -559,6 +561,7 @@ pub mod pallet {
 			let block = frame_system::Pallet::<T>::block_number();
 			let unlock_block = block + T::StakeLock::get();
 			let stake = Self::oracle_stake(signer.clone()).ok_or(Error::<T>::NoStake)?;
+			ensure!(stake > BalanceOf::<T>::zero(), Error::<T>::NoStake);
 			let withdrawal = Withdraw { stake, unlock_block };
 			OracleStake::<T>::remove(&signer);
 			DeclaredWithdraws::<T>::insert(signer.clone(), withdrawal);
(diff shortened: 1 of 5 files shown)


frame/oracle/src/lib.rs
```diff
@@ -692,11 +695,25 @@ pub mod pallet {
 				let min_accuracy = asset_info.threshold;
 				if accuracy < min_accuracy {
 					let slash_amount = asset_info.slash;
-					let try_slash = T::Currency::can_slash(&answer.who, slash_amount);
-					if !try_slash {
-						log::warn!("Failed to slash {:?}", answer.who);
+					let amount_staked = Self::oracle_stake(answer.who.clone())
```
vimukthi-git, line 698: :+1: 
```suggestion
					let new_amount_staked = Self::oracle_stake(answer.who.clone())
```

frame/oracle/src/lib.rs
```diff
@@ -692,11 +695,25 @@ pub mod pallet {
...
+						.unwrap_or_else(|| 0_u32.into())
+						.saturating_sub(slash_amount);
+					OracleStake::<T>::insert(&answer.who, amount_staked);
```
vimukthi-git, line 701: Do we need to insert if the `amount_staked` is zero?

frame/oracle/src/tests.rs
```diff
@@ -955,13 +962,114 @@ fn test_payout_slash() {
...
 		assert_eq!(Balances::free_balance(account_3), 0);
-		assert_eq!(Balances::free_balance(account_4), 90);
+		assert_eq!(Oracle::oracle_stake(account_4), Some(0));
+		assert_eq!(Balances::free_balance(treasury_account), 102);
+		assert_eq!(Balances::free_balance(account_4), 100);
+	});
+}
+
+#[test]
+fn halborn_test_escape_slashing() {
```
vimukthi-git, line 978:
```suggestion
fn halborn_test_bypass_slashing() {
```
?

vimukthi-git (approved):

frame/oracle/src/lib.rs
vivekvpandya: zero value is handled properly in function like remove_stake(), however it may be good to have None instead of zero, but I think both ways are fine and thus not doing any change for this in current PR.

vimukthi-git: Can't we just remove the oracle when there is no stake?

vivekvpandya: we can but don't want to do it in this PR. As I am not much confident about it the code.

frame/oracle/src/lib.rs
```diff
@@ -692,11 +696,25 @@ pub mod pallet {
...
+					);
+					match result {
+						Ok(remaning_val) =>
+							if remaning_val > BalanceOf::<T>::zero() {
+								log::warn!("Only slashed {:?}", slash_amount - remaning_val);
+							},
+						Err(e) => {
+							log::warn!("Failed to slash {:?} due to {:?}", answer.who, e);
+						},
 					}
```
hussein-aitlahcen, line 717: Fine to log this but I believe it's impossible as you are checking that the staked amount >= asset slash right?

hussein-aitlahcen (approved):
LGTM

frame/oracle/src/lib.rs
vivekvpandya: Yes I think this is correct, that change is done in different PR.

vivekvpandya: But it is fine just for sake of completion.

vimukthi-git (approved):
LGTM

PR #969, opened 2022-05-08
Fix benchmarks of pallet-lending

andor0:
## Issue
https://app.clickup.com/t/2azjn0b

## Description
Also I added benchmarks for `liquidate` extrinsic, also I updated tests.


frame/lending/src/lib.rs
```diff
@@ -419,6 +422,8 @@ pub mod pallet {
 		CannotRepayMoreThanTotalDebt,
 
 		BorrowRentDoesNotExist,
+
+		MaxLiquidationBatchSizeExceed,
```
benluelo, line 426:
```suggestion
		MaxLiquidationBatchSizeExceeded,
```

benluelo (commented):
LGTM, just a small nit on spelling

frame/lending/src/lib.rs
andor0: Fixed.

benluelo (approved):

frame/lending/src/lib.rs
```diff
@@ -749,14 +754,18 @@ pub mod pallet {
 		/// liquidation.
 		/// - `origin` : Sender of this extrinsic.
 		/// - `market_id` : Market index from which `borrower` has taken borrow.
-		#[pallet::weight(<T as Config>::WeightInfo::liquidate(borrowers.len() as Weight))]
+		#[pallet::weight(<T as Config>::WeightInfo::liquidate(borrowers.len() as u32))]
```
dzmitry-lahoda, line 757: not sure. may be we should use usize?

dzmitry-lahoda (approved):

frame/lending/src/lib.rs
andor0: a auto-created file with weights (`runtime/dali/src/weights/lending.rs`) will have `fn liquidate(b: u32, ) -> Weight { ... }` independ on what we will use `u32` or `usize`. i.e. we will get a broken build with `usize`.

PR #971, opened 2022-05-09
CU-296cyym - xcvm/cosmwasm

hussein-aitlahcen:
We can instantiate/execute/query a contract from the pallet. Contracts are able to call each other but not instantiate a new contract yet.

- Introduce a `query` extrinsics at the pallet level to call the `Query` export
of an instantiated cosmwasm contract.
- Introduce a sub-messages handler `run_k` that does the depth first handling of
contracts sub-messages on `instantiate` and `execute`, not yet semantically correct w.r.t to transaction behavior defined in https://github.com/CosmWasm/cosmwasm/blob/main/SEMANTICS.md
- Refactor the runtime by moving the loaded module inside it, which change the
dependency direction of executor->runtime to runtime->executor. this make it
possible to call `allocate` and `deallocate` from the runtime
instance (executing module).

Next:

- Handling `reply` on sub-messages handling to notify the origin contract whenever it's sub-message has been handled and implement the associated semantically correct transactional behavior.
- Handling `Instantiate|Migrate|UpdateAdmin|ClearAdmin` sub-messages (probably not worth for the demo).
- Handling crypto related host functions.
- Pass XCVM instructions through the `Custom` sub-message and notify the calling contract of result (require above `reply` handling):
```rust
pub enum CosmosMsg<T = Empty>
where
    T: Clone + fmt::Debug + PartialEq + JsonSchema,
{
    Bank(BankMsg),
    /// This can be defined by each blockchain as a custom extension
    Custom(T),
    Staking(StakingMsg),
    Distribution(DistributionMsg),
    Stargate {
        type_url: String,
        value: Binary,
    },
    Ibc(IbcMsg),
    Wasm(WasmMsg),
} 
```


ComposableFiB0t:
Task linked: [CU-296cyym XCVM + COSMWASM Cross-chain demo prototype](https://app.clickup.com/t/296cyym)

frame/contracts/src/lib.rs
```diff
@@ -385,6 +392,34 @@ pub mod pallet {
...
+				data,
+				None,
+			);
+			if let Ok(retval) = &output.result {
+				if retval.did_revert() {
+					output.result = Err(<Error<T>>::ContractReverted.into());
+				}
+			}
+			output.gas_meter.into_dispatch_result(output.result, T::WeightInfo::call())
+		}
```
KaiserKarel, line 421: I think query can be an rpc method; or does internal_query allow for state changes?

frame/contracts/src/wasm/mod.rs
```diff
@@ -229,68 +226,122 @@ where
...
+							runtime.ext().address().clone(),
+						)),
+					},
+				};
+
+				let info = MessageInfo {
+					sender: Addr::unchecked(T::AccountIdToFromString::convert(
+						runtime.ext().caller().clone(),
+					)),
+					funds: vec![],
```
KaiserKarel, line 266: Should we pass the funds as an argument from the function here?

frame/contracts/src/wasm/mod.rs
```diff
@@ -229,68 +226,122 @@ where
...
+				// this should probably be setup by the caller as well.
+
+				let env = Env {
+					block: BlockInfo {
+						height: 0,
+						time: Timestamp(
+							to_string(&0u128)
+								.map_err(|_| DispatchError::Other("marshall failed"))?,
+						),
+						chain_id: Default::default(),
```
KaiserKarel, line 252: let's already hardcode this to our ID?

KaiserKarel (approved):

frame/contracts/src/wasm/mod.rs
hussein-aitlahcen: The funds are only 1 token in the contracts pallet while it's a vec in cosmwasm, we can pass `vec![("PICA", transferred_amount)]` already :+1: .
But I think when rewritting, we'll make it multi-currency and correctly populate the list of coins with all the transferred coins.

frame/contracts/src/wasm/mod.rs
hussein-aitlahcen: :+1:

frame/contracts/src/lib.rs
hussein-aitlahcen: It just a copy/paste of execute right now

PR #973, opened 2022-05-09
Mchuk/tests/offchain worker

mikolaichuk:
## Issue
https://app.clickup.com/t/2vqed0h

## Description
Off-chain worker test. Checks if we make a liquidation request for a risky position and do not make it for the health one.
The test done in the terms of events and transactions checking.


frame/lending/src/mocks_offchain/mod.rs
```diff
@@ -0,0 +1,535 @@
...
+	);
+	Executive::initialize_block(&header);
+	Timestamp::set_timestamp(MILLISECS_PER_BLOCK * block_number);
+	System::set_block_number(block_number);
+	Executive::apply_extrinsic(extrinsic).unwrap().unwrap();
+}
+
+/// Processes the specified amount alls [`next_block()`] and then calls
+/// [`Lending::finalize`](OnFinalize::on_finalize).
+pub fn process_and_progress_blocks(blocks_to_process: usize) {
```
benluelo, line 487: You can use `composable_tests_helpers::test::block::process_and_progress_blocks` instead of redefining the function (I just merged this earlier today: https://github.com/ComposableFi/composable/blob/a4af564fe79f0ff485b607635add70ba6da25d12/frame/composable-tests-helpers/src/test/block.rs)

frame/lending/src/mocks_offchain/wrapper.rs
```diff
@@ -0,0 +1,150 @@
...
+	Debug,
+	Hash,
+	Serialize,
+	Deserialize,
+	PartialOrd,
+	Ord,
+	MaxEncodedLen,
+	TypeInfo,
+)]
+
```
benluelo, line 33: style, remove blank line here

frame/lending/src/mocks_offchain/wrapper.rs
```diff
@@ -0,0 +1,150 @@
...
+impl CryptoType for UintAuthorityIdWrapper {
+	type Pair = Dummy;
+}
+
+impl AsRef<[u8]> for UintAuthorityIdWrapper {
+	fn as_ref(&self) -> &[u8] {
+		unsafe {
+			let byte: *const _ = &(self.0 .0 as _);
+			std::slice::from_raw_parts(byte, std::mem::size_of::<u64>())
+		}
```
benluelo, line 66: you don't need to use unsafe here: https://doc.rust-lang.org/beta/std/primitive.u64.html#method.to_ne_bytes

frame/lending/src/tests_offchain.rs
```diff
@@ -0,0 +1,159 @@
...
+	}
+}
+
+fn assert_extrinsic_event<T: crate::Config>(
+	result: DispatchResultWithPostInfo,
+	event: <T as crate::Config>::Event,
+) {
+	assert_ok!(result);
+	assert_last_event::<T>(event);
+}
```
benluelo, line 159: these can be imported from `crate::tests` instead of being redefined

frame/lending/src/mocks_offchain/wrapper.rs
```diff
@@ -0,0 +1,150 @@
...
+	}
+}
+
+thread_local! {
+	static ALL_KEYS: RefCell<Vec<UintAuthorityIdWrapper>> = RefCell::new(vec![]);
+}
```
benluelo, line 72: Is the global state necessary? I'm not sure what the purpose is

benluelo: If it is necessary, perhaps add a comment explaining why?

benluelo (changes requested):

frame/lending/src/mocks_offchain/wrapper.rs
```diff
@@ -0,0 +1,150 @@
+use serde::{Deserialize, Serialize};
```
dzmitry-lahoda, line 1: may be rename wrapper to more specific wrapper.

frame/lending/src/tests_offchain.rs
```diff
@@ -0,0 +1,159 @@
...
+		System::assert_has_event(Event::Liquidations(event));
+	});
+}
+
+// HELPERS
+fn default_under_collateralized_warn_percent() -> Percent {
+	Percent::from_float(0.10)
+}
+
+fn create_market<const NORMALIZED_PRICE: u128>(
```
dzmitry-lahoda, line 106: it feels like `create_market<const NORMALIZED_PRICE> already exists in lending. why copy paste? can we parametrize types to avoid copy paste?

frame/lending/src/tests_offchain.rs
```diff
@@ -0,0 +1,159 @@
...
+		set_price(BTC::ID, NORMALIZED::units(38_000));
+		// Header for the fake block to execute off-chain worker
+		let header =
+			Header::new(2, H256::default(), H256::default(), [69u8; 32].into(), Digest::default());
+		// Execute off-chain worker
+		Executive::offchain_worker(&header);
+		// Check if corresponded liquidation transaction has been placed in the pool
+		let tx = pool_state.write().transactions.pop().unwrap();
+		assert!(pool_state.read().transactions.is_empty());
+		let tx = Extrinsic::decode(&mut &*tx).unwrap();
```
dzmitry-lahoda, line 85: not need in this pr, but as part of liquidation offchain testing, ensure that other positions not liquidated

dzmitry-lahoda: you may read on chain state to check it afaik. as off chain worker kind of can read, but not write, on chain state

frame/lending/src/mocks_offchain/wrapper.rs
dzmitry-lahoda: feels like this can be made state of struct?

mikolaichuk: This is supposed to be a wrapper for https://docs.rs/sp-runtime/latest/sp_runtime/testing/struct.UintAuthorityId.html. I need it since UintAuthorityId does not implement frame_system::offchain::AppCrypto. I can  bring it to a more appropriate wrapper template form (for the moment it more looks like reimplementation). However, it will stay the same under the hood. Alternatively, I can completely reimplement this structure. What would be better solution?

frame/lending/src/tests_offchain.rs
dzmitry-lahoda: if you can check events, check that unhealthy positions liquidated and health retained

frame/lending/src/mocks_offchain/authority_id_wrapper.rs
```diff
@@ -85,40 +87,44 @@ impl sp_application_crypto::RuntimeAppPublic for UintAuthorityIdWrapper {
 	type Signature = TestSignature;
 
 	fn all() -> Vec<Self> {
-		ALL_KEYS.with(|l| l.borrow().clone())
+		let authority_id_vec = UintAuthorityId::all();
+		let mut self_vec = Vec::new();
+		for i in 0..authority_id_vec.len() {
+			self_vec.push(Self(authority_id_vec.get(i).unwrap().clone()));
+		}
```
benluelo, line 94:
```suggestion
		self_vec = authority_id_vec.into_iter().map(Self).collect()
```

benluelo (changes requested):

frame/lending/src/tests_offchain.rs
mikolaichuk: >there are no events for healthy positions?
Liquidations pallet has only one event: PositionWasSentToLiquidation

https://github.com/ComposableFi/composable/blob/main/frame/liquidations/src/lib.rs#L116 

My point is that we do not achieve liquidation in our conditions. To liquidate something we need some activity from Liquidations side.

frame/lending/src/mocks_offchain/authority_id_wrapper.rs
mikolaichuk: Added as a separate commit

frame/lending/src/mocks_offchain/mod.rs
mikolaichuk: Done for tests.rs and tests_offchain.rs

frame/lending/src/mocks_offchain/wrapper.rs
mikolaichuk: Fixed

frame/lending/src/tests_offchain.rs
mikolaichuk: Done, fn create_market was generalized. 
Please be aware that now it does not emit events.

frame/lending/src/mocks_offchain/wrapper.rs
mikolaichuk: Implemented as a normal wrapper. If this approach is not suitable, please let me know

frame/lending/src/mocks_offchain/wrapper.rs
mikolaichuk: Done

frame/lending/src/tests_offchain.rs
mikolaichuk: Done

frame/lending/src/tests_offchain.rs
benluelo: It's possible to make it emit events - just be generic over the runtime instead of the pallet type. It will require a bit of refactoring, I can message you on slack if you want to peer program it

benluelo: the basic idea is this: ```rust
crate::Pallet::<Runtime>::create_market(manager, validated.value())
```

**Review by dzmitry-lahoda — DISMISSED** [review]

File: frame/lending/src/tests_offchain.rs (hunk 4 above)
Comment (mikolaichuk) [reply frame/lending/src/tests_offchain.rs]: @benluelo Done

**Review by dzmitry-lahoda — DISMISSED** [review]

File: frame/lending/src/mocks_offchain/mod.rs [hunk 10]
```diff
@@ -0,0 +1,512 @@
...
+			negative: false,
+		};
+		smallvec![one]
+	}
+}
+
+pub struct XcmFake;
+impl Into<Result<cumulus_pallet_xcm::Origin, XcmFake>> for XcmFake {
+	fn into(self) -> Result<cumulus_pallet_xcm::Origin, XcmFake> {
+		todo!("please test via local-integration-tests")
```
benluelo, line 317: if these methods will never be implemented, `unimplemented!` would be more semantic here

frame/lending/src/tests.rs
```diff
@@ -1,50 +1,55 @@
...
 };
+use std::ops::{Div, Mul};
 
 const DEFAULT_MARKET_VAULT_RESERVE: Perquintill = Perquintill::from_percent(10);
 const DEFAULT_MARKET_VAULT_STRATEGY_SHARE: Perquintill = Perquintill::from_percent(90);
 const DEFAULT_COLLATERAL_FACTOR: u128 = 2;
 
+type SystemAccountId<R> = <R as frame_system::Config>::AccountId;
+type SystemOrigin<R> = <R as frame_system::Config>::Origin;
+type SystemEvent<R> = <R as frame_system::Config>::Event;
```
benluelo, line 51: unofficial naming convention for these kinds of type definitions is `TypeNameOf<T>`, so `SystemAccountIdOf<T>`

frame/lending/src/tests.rs
```diff
@@ -1482,18 +1481,33 @@ fn create_simple_vault(
 /// # Panics
 ///
 /// Panics on any errors. Only for use in testing.
-fn create_market<const NORMALIZED_PRICE: u128>(
+pub fn create_market<const NORMALIZED_PRICE: u128, R>(
```
benluelo, line 1484: Use `T` for the Config type

frame/lending/src/tests.rs
```diff
@@ -1506,26 +1520,25 @@ fn create_market<const NORMALIZED_PRICE: u128>(
...
+	if let crate::Event::<R>::MarketCreated { market_id, vault_id, .. } = pallet_event {
+		(market_id, vault_id)
 	} else {
 		panic!(
-			"System::events() did not contain the market creation event. Found {:#?}",
+			"There is no market creation event in System::events(). Found: {:#?}",
 			system_events
-		)
+		);
 	}
```
benluelo, line 1542: I think assert_extrinsic_event could be used here, I wrote this test before I wrote that function and never changed the test to use it

frame/lending/src/tests.rs
```diff
@@ -1578,24 +1591,38 @@ fn create_simple_market() -> (MarketIndex, VaultId) {
 ///
 /// Panics on any errors and checks that the last event was `CollateralDeposited` with the correct/
 /// expected values.
-fn mint_and_deposit_collateral<T: crate::Config>(
-	account: &sp_core::sr25519::Public,
+pub fn mint_and_deposit_collateral<R>(
```
benluelo, line 1594: Again, use `T` for the config type

benluelo: (Or something more verbose like `Runtime`)

benluelo (commented):
LGTM, just a few style nits

frame/lending/src/tests.rs
mikolaichuk: @benluelo As far as I understand, here we use the latest system event to extract market_id and vault_id, since dispatchable function do not return such information. assert_extrincis_event would not help us in this case.

frame/lending/src/mocks_offchain/mod.rs
mikolaichuk: Done for tests.rs and tests_offchain.rs

frame/lending/src/tests.rs
mikolaichuk: Done

frame/lending/src/tests.rs
mikolaichuk: Done

frame/lending/src/tests.rs
mikolaichuk: Done

frame/lending/src/tests.rs
benluelo: Yep, totally forgot we returned those values. Disregard haha

benluelo (approved):

dzmitry-lahoda (approved):

PR #974, opened 2022-05-09
parachain-staking

r2st:
## Issue
https://app.clickup.com/20465559/docs/kghwq-74067/kghwq-29547

## Description
Included the pallet parachain-staking, the build is successful and test cases passed. Will implement other features described in clickup

## Checklist

- [ ] Need to check if it is working with existing code in runtime (yet to add it to runtime)


seunlanlege:
Can't really access the clickup issue, but i would strongly prefer that we offload the maintenance burden of this pallet to the purestake team and just pull it in as a git dependency, seeing as they got a web3 grant to develop it. If we need any changes made to it, then we should pr upstream to them:

https://github.com/PureStake/moonbeam/blob/master/pallets/parachain-staking/README.md

PR #975, opened 2022-05-09
parachain-collators-staking

r2st:
## Issue
https://app.clickup.com/20465559/docs/kghwq-74067/kghwq-29547

## Description
Included the pallet parachain-staking as  patch, the build is successful. Will implement other features described in clickup

## Checklist

- [ ] Need to check if it is working with existing code in runtime (yet to add it to runtime)

### DIFF
--- a/frame/parachain-collators-staking/Cargo.toml
+++ b/frame/parachain-collators-staking/Cargo.toml
@@ -1,0 +1,37 @@
+[package]
+authors = ["Composable Developers"]
+edition = "2021"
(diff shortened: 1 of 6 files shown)


seunlanlege:
Hmm so we're not using the parachain-staking pallet from moonbeam?

r2st:
> Hmm so we're not using the parachain-staking pallet from moonbeam?

we are using it but as a patch as you suggested.

Martin-Kiss (approved):

PR #981, opened 2022-05-11
Remove dependency on our polkadot and substrate forks

KaiserKarel:
I think we no longer need our own polkadot and substrate forks, and thus we can deprecate our own forked repo.


r2st (approved):
LGTM

andor0 (approved):

PR #986, opened 2022-05-11
NFT Tests (CU-25enyrn)

benluelo:
## Description
Adds some tests to pallet-nft. More tests to come.

### Changes
- Refactored some common test utilities into `composable-tests-helpers`
- `NFT` -> `Nft` in type names (in progress)
- Fixed bug in `Transfer` implementation

### Status
- [x] `nonfungibles::Inspect` success tests
- [x] `nonfungibles::Inspect` failure tests
- [x] `nonfungibles::Create` success tests
- [x] `nonfungibles::Create` failure tests
- [x] `nonfungibles::Transfer` success tests
- [x] `nonfungibles::Transfer` failure tests
- [x] `nonfungibles::Mutate` success tests (partially complete)
- [x] `nonfungibles::Mutate` failure tests (partially complete)
- [x] `FinancialNftProvider` success tests
- [ ] ~~`FinancialNftProvider` failure tests~~ there doesn't seem to be any failure cases here (other than hitting numeric bounds on storage.)

## Checklist

- ~~[ ] I have updated the cargo docs to reflect changes made by this PR _(if applicable)_~~
- ~~[ ] I have updated the `book/` to reflect changes made by this PR _(if applicable)_~~


ComposableFiB0t:
Task linked: [CU-25enyr6 fNFT pallet implementation based on nonfungibles](https://app.clickup.com/t/25enyr6)

ComposableFiB0t:
Task linked: [CU-25enyrn Tests](https://app.clickup.com/t/25enyrn)

frame/nft/src/test/impls/transfer.rs
```diff
@@ -0,0 +1,302 @@
...
+		mint_nft_and_assert,
+		mock::{new_test_ext, Event, Test},
+		ALICE, BOB, CHARLIE,
+	},
+	AccountIdOf, Instance, NftInstanceId, OwnerInstances, Pallet,
+};
+
+#[test]
+fn simple() {
+	//! Tests a simple transfer between 2 accounts, with only 1 total NFT existing.
```
dzmitry-lahoda, line 27: wow, that works? nice. so i would pefer on test because it is possible)

frame/nft/src/test/mock.rs
```diff
@@ -0,0 +1,81 @@
...
+		NodeBlock = Block,
+		UncheckedExtrinsic = UncheckedExtrinsic,
+	{
+		System: frame_system::{Pallet, Call, Config, Storage, Event<T>},
+		Timestamp: pallet_timestamp::{Pallet, Call, Storage},
+		Nft: crate::{Pallet, Storage , Event<T>},
+	}
+);
+
+impl crate::Config for Test {
```
dzmitry-lahoda, line 31: i prefer Runtime word instead of Test. code is more polymorphic than with benchmarks and real runtimes.

frame/nft/src/test/impls/transfer.rs
benluelo: I just thought it looked a bit strange like this:

```rust
/// Tests a simple transfer between 2 accounts, with only 1 total NFT existing.
#[test]
fn simple() {
```

with the test attribute between the comment and the test (also weird if you put the attribute above the comment, just can't win :laughing:)

I'm not super opinionated either way though, I can use the regular outer comment :+1:

frame/nft/src/test/impls/mutate.rs
```diff
@@ -0,0 +1,235 @@
...
+			assert_noop!(
+				Pallet::<MockRuntime>::mint_into(&NftClass::STAKING, &NEW_NFT_ID, &ALICE),
+				DispatchError::from(crate::Error::<MockRuntime>::InstanceAlreadyExists)
+			);
+		})
+	}
+}
+
+/// Tests the implementation of [`Mutate::burn_from`].
+mod burn_from {
```
dzmitry-lahoda, line 73: afaik polkadot renamed its burn_from to burn, can rename later either as separate task

frame/nft/src/lib.rs
```diff
@@ -224,7 +226,9 @@ pub mod pallet {
 							Some(instances) => {
```
dzmitry-lahoda, line 226: would it be possible to use ValueQuery with empty map to instead of map wrapped in option and simplify handling?

frame/nft/src/lib.rs
```diff
@@ -224,7 +226,9 @@ pub mod pallet {
...
 								instances.remove(&(*class, *instance));
 							},
-							None => {},
+							None => {
+								debug_assert!(false, "unreachable")
```
dzmitry-lahoda, line 230: i wold prefer if let with such empty map. or panic!.

frame/nft/src/lib.rs
benluelo: Was wondering the same myself - I'll look into it tomorrow

frame/nft/src/test/impls/mutate.rs
benluelo: Sounds like a problem for whoever does the upgrade 😁

frame/nft/src/lib.rs
benluelo: Will review this when reviewing ValueQuery

frame/nft/src/test/helpers.rs
```diff
@@ -0,0 +1,172 @@
...
+		(ALICE, BTreeMap::new()),
+		"owner should be ALICE, with no attributes"
+	);
+
+	NEW_NFT_ID
+}
+
+/// Mints many NFTs into the specified account and checks that they were created properly,
+/// returning the ids of the newly created NFTs.
+pub(crate) fn mint_many_nfts_and_assert<const AMOUNT: usize>(
```
dzmitry-lahoda, line 93: amount  is not part of type, neither we can have loop unroll here. it seems like overengineeirng

dzmitry-lahoda: passing parameters would be more expected imho

benluelo: I just wanted to be able to do array destructuring when crating multiple nfts is all. I can change it to parameter if it's an issue

dzmitry-lahoda (approved):

frame/nft/src/test/helpers.rs
dzmitry-lahoda: blame to me. you preserved array len.

hussein-aitlahcen (approved):

PR #1002, opened 2022-05-12
Escher feature/vamm move price

Cardosaum:
## Issue
CU-24ah01w

## Description
- Adds `move_price` functionality to the Vamm


ComposableFiB0t:
Task linked: [CU-24ah01w Feature 5 (Move Price Functionality)](https://app.clickup.com/t/24ah01w)

frame/vamm/src/tests/create_vamm.rs
```diff
@@ -0,0 +1,159 @@
...
+};
+use composable_traits::vamm::{Vamm as VammTrait, VammConfig};
+use frame_support::{assert_noop, assert_ok};
+use proptest::prelude::*;
+
+proptest! {
+	#![proptest_config(ProptestConfig::with_cases(RUN_CASES))]
+	#[test]
+	#[allow(clippy::disallowed_methods)]
+	fn create_vamm(
```
kevin-composable, line 14: I would change name to create_vamm_correctly_updates_runtime_storage

frame/vamm/src/tests/get_price.rs
```diff
@@ -0,0 +1,83 @@
...
+			} else {
+				assert_ok!(
+					TestPallet::get_price(0, AssetType::Base),
+					Decimal::from_inner(quote_peg.unwrap().checked_div(base_asset_reserves).unwrap()))
+			}
+		})
+	}
+}
+
+proptest! {
```
kevin-composable, line 48: both of these proptest tests can be in the same `proptest`! block

frame/vamm/src/tests/swap_asset.rs
```diff
@@ -0,0 +1,477 @@
...
+		ExtBuilder {
+			vamm_count: 1,
+			vamms: vec![(0, vamm_state)]
+		}.build().execute_with(|| {
+			// For event emission
+			run_to_block(1);
+
+			let swap = TestPallet::swap(&swap_config);
+
+			if let Ok(swap) = swap {
```
kevin-composable, line 126: shouldn't you call `ensure_ok!(swap)` first and then call `system::assert_last_event(...)`

frame/vamm/src/tests/swap_asset.rs
```diff
@@ -0,0 +1,477 @@
...
+			let invariant_after = TestPallet::compute_invariant(
+				vamm_after_swap.unwrap().base_asset_reserves,
+				vamm_after_swap.unwrap().quote_asset_reserves,
+			).unwrap();
+
+			let invariant_epsilon = invariant_before / U256::exp10(8);
+			let invariant_delta = invariant_before.max(invariant_after)
+				- invariant_before.min(invariant_after);
+
+			assert_eq!(invariant_delta / vamm_state.invariant, U256::zero());
```
kevin-composable, line 388: does this default/round to zero if invariant_delta < vamm_state.invariant? trying something like `assert_eq!(1u128 / 2u128, 0u128)` is true even though it shouldn't necessarily be

kevin-composable (approved):
Again, I am very pleased by the work you and Angelo are doing; keep it up! I just had a couple small comments

frame/vamm/src/tests/create_vamm.rs
Cardosaum: Thanks for the input.
I changed it :+1:

frame/vamm/src/tests/get_price.rs
Cardosaum: Tests moved to the same block :+1:

frame/vamm/src/tests/swap_asset.rs
Cardosaum: You're correct, but then we stumble up in the error we're facing with some specific swaps (the error I'm yet to tackle next week), so this was the way I found to test that events were being emitted correct despite some specific swaps failing. 

I should've added a TODO comment about changing it later tho, sorry about it.

frame/vamm/src/tests/swap_asset.rs
Cardosaum: Good catch!

Actually I forgot to remove those `assert_eq!` as they are doing essentially nothing since we have `assert!(invariant_delta <= invariant_epsilon)` in the next line (for the 3 instances of this test: both assets, only base and only quote).

I removed those unnecessary asserts, thanks for the input! :)

frame/vamm/src/tests/get_price.rs
```diff
@@ -0,0 +1,83 @@
...
+					TestPallet::get_price(0, AssetType::Base),
+					ArithmeticError::Overflow);
+			} else if quote_peg.unwrap().checked_div(base_asset_reserves).is_none() {
+				assert_noop!(
+					TestPallet::get_price(0, AssetType::Base),
+					ArithmeticError::DivisionByZero);
+			} else {
+				assert_ok!(
+					TestPallet::get_price(0, AssetType::Base),
+					Decimal::from_inner(quote_peg.unwrap().checked_div(base_asset_reserves).unwrap()))
```
0xangelo, line 42: I think you should convert `quote_peg` and `base_asset_reserves` to `Decimal` **before** the division, no?
Can you check that doing the above vs. what you're doing now yields the same results? Try with `quote_peg` being half of `base_asset_reserves` and check if the price is `0.5`

Cardosaum:
> Again, I am very pleased by the work you and Angelo are doing; keep it up! I just had a couple small comments

Thank you so much for the valuable feedback!
It's really important for me and I'm happy to receive them as they help me to learn more and ship better quality code :slightly_smiling_face:

frame/vamm/src/tests/swap_asset.rs
```diff
@@ -0,0 +1,477 @@
...
+		],
+	}
+	.build()
+	.execute_with(|| {
+		let swap = TestPallet::swap(&swap_config);
+		let vamm_after_swap = VammMap::<MockRuntime>::get(0);
+		let vamm_expected = VammMap::<MockRuntime>::get(1);
+
+		assert_eq!(swap, Ok(16_666_666_666_667));
+		assert_eq!(vamm_after_swap, vamm_expected);
```
0xangelo, line 192: Maybe nit-picky, but did you need to push the expected state into storage to compare it later? I think you can just use it inside the closure (I use configs created outside the closure for `execute_with` in some of my tests).

Also remember that you may use `assert_ok!(swap, 16_666_666_666_667);`

frame/vamm/src/tests/swap_asset.rs
```diff
@@ -0,0 +1,477 @@
...
+
+		assert_eq!(swap, Ok(39_215_686_275));
+		assert_eq!(vamm_after_swap, vamm_expected);
+	})
+}
+
+proptest! {
+	#![proptest_config(ProptestConfig::with_cases(1))]
+	#[test]
+	fn multiple_swaps_dont_diverge_from_original_invariant_both_base_and_quote(
```
0xangelo, line 360: Nice

frame/vamm/src/tests/swap_base_asset.rs
```diff
@@ -0,0 +1,115 @@
...
+		mut swap_config in get_swap_config(
+			TestSwapConfig {
+				direction: Some(Direction::Remove),
+				vamm_id: Some(0),
+				asset: Some(AssetType::Base),
+				..Default::default()}),
+		input_amount in balance_range_upper_half(),
+		base_asset_reserves in balance_range_lower_half(),
+	) {
+		prop_assume!(input_amount > base_asset_reserves);
```
0xangelo, line 27: I think you can get rid of this by generating `input_amount` and `base_asset_reserves` jointly in a `prop_compose`. Should make shrinking inputs better in case errors are found

0xangelo (commented):
Looking good so far. Just have a few comments.

frame/vamm/src/lib.rs
```diff
@@ -272,8 +287,23 @@ pub mod pallet {
 	#[pallet::event]
 	#[pallet::generate_deposit(pub(super) fn deposit_event)]
 	pub enum Event<T: Config> {
-		/// Emitted after a successful call to the [`create`](Pallet::create) function.
+		/// Emitted after a successful call to [`create`](Pallet::create) function.
```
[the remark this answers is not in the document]
Cardosaum: Alright, thanks for the input! 
Changed it and the other instances (`swap` and `move_price`)

frame/vamm/src/lib.rs
```diff
@@ -481,16 +532,319 @@ pub mod pallet {
 			todo!()
 		}
 
-		#[allow(unused_variables)]
+		/// Performs the swap of the desired asset agains the vamm.
```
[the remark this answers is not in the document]
Cardosaum: good catch!

frame/vamm/src/lib.rs
```diff
@@ -299,6 +331,17 @@ pub mod pallet {
 		/// Tried to add some amount of asset to Vamm but it would exceeds the
 		/// supported maximum value.
 		TradeExtrapolatesMaximumSupportedAmount,
+		/// Tried to perform operation agains a closed Vamm.
```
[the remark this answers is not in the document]
Cardosaum: good catch here too, thanks! :)

frame/vamm/src/lib.rs
```diff
@@ -481,16 +532,319 @@ pub mod pallet {
...
+				Self::compute_invariant(config.base_asset_reserves, config.quote_asset_reserves)?;
+
+			vamm_state.base_asset_reserves = config.base_asset_reserves;
+			vamm_state.quote_asset_reserves = config.quote_asset_reserves;
+			vamm_state.invariant = invariant;
+
+			// Update runtime storage.
+			VammMap::<T>::insert(&config.vamm_id, vamm_state);
+
+			// Deposit swap event into blockchain.
```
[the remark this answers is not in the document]
Cardosaum: Oh.. good catch.

I need to be more careful when doing copy+paste :P

frame/vamm/src/tests/swap_base_asset.rs
Cardosaum: actually, in the way `range_upper` and `range_lower` were made it would be technically impossible to create values that invalidate this `prop_assume`, so this assert-like statement there is just to be pedantic about the requirement we're expecting for this proptest.

Do you think I should refactor it tho?

frame/vamm/src/tests/get_price.rs
Cardosaum: Interesting enough both this version and the one converting to `Decimal` before handle the same results (all tests passes for both of them...)

I'll need to investigate if further next week than. 

Thank you for the comment on this!
I'll create a new task for it on clickup.

frame/vamm/src/tests/swap_asset.rs
Cardosaum: Yeah, technically I didn't need to insert it into storage. I'll change it, thanks for the input! :)

Regarding the `assert_ok`, I sometimes forget it exists :see_no_evil: haha

frame/vamm/src/tests/swap_base_asset.rs
0xangelo: Ah, I did not check the definition of those functions. Well, if by definition they satisfy the property in line 27, then I think removing the `prop_assume!` is warranted.

I would still prefer to generate then jointly, though, to avoid errors in the future if you change the implementation of either of two functions. Your call really

0xangelo (approved):
LGTM. Congrats!

frame/vamm/src/tests/get_price.rs
Cardosaum: I really want to thank you for raising attention to this section. 

Yeah, the values should be converted in `Decimal` before the computation, but the problem of both `Decimal` and `u128` passing all tests was due to the fact that the range of valid input values for the variables was too permissive, so the fuzz tests didn't really catch the issue with the `Decimal`s. 

I fixed this error and also create individual tests for each specific case. (something in the lines of [this comment here](https://github.com/ComposableFi/composable/pull/1013#discussion_r874066000))

Thanks again! :)

PR #1014, opened 2022-05-16
[cu-25ek05u] Pablo Fee & Staking Rewards Distribution Docs

vimukthi-git:
## Issue

## Description

## Checklist


ComposableFiB0t:
Task linked: [CU-25ek05u Fee approach - xPBLO distribution of PBLO minted based on fees in treasury](https://app.clickup.com/t/25ek05u)

benluelo:
Why use adoc instead of markdown?

vimukthi-git:
Main advantage is it's easier to write with embedded plantuml. There are other benefits like better formatting. Anyway I have my script that converts adoc to md as you may see.

rfcs/0004-pablo-distribution.adoc
```diff
@@ -0,0 +1,443 @@
...
+    duration_presets: NO_END,
+    reward_assets: [BASE, QUOTE]),
+    early_unstake_penaly: 0
+);
+stop
+----
+
+The interface for the configurations does not currently exist on staking_rewards pallet, hence it has to be implemented.
+
+===== LP fNFT Creation When Receiving LP Tokens for Adding Liquidity
```
vimukthi-git, line 233: Capture what happens when removing liquidity

rfcs/0004-pablo-distribution.adoc
```diff
@@ -0,0 +1,443 @@
...
+        usecase "Set PBLO Reward Rate" as UC9
+        usecase "Set PBLO Reward\nAllocation Percentage" as UC12
+        usecase "Configure Total\nPBLO Supply" as UC14
+    }
+    usecase "Stake PICA" as UC1 <<Staking>>
+    usecase "Earn PICA Reward" as UC3 <<Staking>>
+    usecase "Earn Other\nToken(s) Reward" as UC10 <<Staking>>
+    usecase "Set PICA Reward Rate" as UC8
+    usecase "Set PICA Reward\nAllocation Percentage" as UC11
+    usecase "Configure Total\nPICA Supply" as UC15
```
vimukthi-git, line 76: Remove as irrelevant

rfcs/0004-pablo-distribution.adoc
```diff
@@ -0,0 +1,443 @@
...
+    package Pablo {
+        usecase "Stake PBLO" as UC2 <<Staking>>
+        usecase "Earn PBLO Reward" as UC4 <<Staking>>
+        usecase "Stake LP Token" as UC5 <<Staking>>
+        usecase "Earn Pablo LP\nTrading Fees" as UC6
+        usecase "Earn Pablo Protocol\nTrading Fees" as UC13 <<Staking>>
+        usecase "Set Pool Protocol\nFee Percentage" as UC7
+        usecase "Set PBLO Reward Rate" as UC9
+        usecase "Set PBLO Reward\nAllocation Percentage" as UC12
+        usecase "Configure Total\nPBLO Supply" as UC14
```
vimukthi-git, line 69: remove as irrelevant

rfcs/0004-pablo-distribution.adoc
```diff
@@ -0,0 +1,443 @@
...
+from=transaction_origin,
+amount=x,
+keep_alive=false
+);
+stop
+----
+
+==== PBLO Staker Trading Fee Distribution
+
+This is the reward a `PBLO` staker receives from the trading fees of Pablo pools. It is equal to the protocol fee charged on Pablo pools. This can be accomplished by calling the already exisiting `StakingReward.transfer_reward` interface as follows. According to product there is also a need to convert whatever the fee asset in to PBLO to create an demand/additional value for PBLO.
```
vimukthi-git, line 273:
```suggestion
This is the reward a `PBLO` staker receives from the trading fees of Pablo pools. It is equal to the protocol fee charged on Pablo pools. This can be accomplished by calling the already exisiting `StakingReward.transfer_reward` interface as follows. According to product there is also a need to convert whatever the fee asset in to PBLO to create a demand/additional value for PBLO.
```

dzmitry-lahoda:
- can we have image assets to be lfs? or our story to have monorepo will end fast
- not sure, can it be a more clear separation of biz docs vs tech docs?

dzmitry-lahoda:
most of us can edit md, and all docs inside repo are md and all docs i have read on whole polkadot ecosystem too. so may we retain only md file in final pr?

rfcs/0004-pablo-distribution.adoc
```diff
@@ -0,0 +1,443 @@
...
+
+[Pool] <<exists>>
+
+node "Governance" {
+    [Configuration] --> [RewardsConfig]
+    [Configuration] --> [FeeConfig]
+}
+----
+
+TODO: Suggest rename of staking-reward pallet to pallet-earn or pallet-rewards.
```
dzmitry-lahoda, line 172: may be we can rename later when have more rewards mechanics, not now

rfcs/0004-pablo-distribution.adoc
```diff
@@ -0,0 +1,443 @@
...
+
+TODO: What to do for part of protocol fees that should be transferred to treasury eventually?
```
dzmitry-lahoda, line 174: we can decide later as separate task i guess

rfcs/0004-pablo-distribution.adoc.md
```diff
@@ -0,0 +1,400 @@
+Design Proposal: Pablo Fees & Staking Rewards Distribution
+==========================================================
+
+Table of Contents
+
+-   [1. Abstract](#_abstract)
```
vimukthi-git, line 6: Fix links here

vimukthi-git:
@dzmitry-lahoda 
> can we have image assets to be lfs? or our story to have monorepo will end fast ...

What do you mean by `lfs` ? You mean the size of the images being high? I don't know if that is a going to be a problem. But I'm also ok with moving the design docs to another repo if it's ok with @cor

vimukthi-git:
@dzmitry-lahoda 

>most of us can edit md, and all docs inside repo are md and all docs i have read on whole polkadot ecosystem too. so may we retain only md file in final pr?

I'm OK with that, I use mostly md too. Normally I use adoc to create the initial document structure and then keep editing the md. However there would be some effort there in committing the plantuml source with that approach.

rfcs/0004-pablo-distribution.adoc
vimukthi-git: The reason to suggest this is that with this design we would already be using the staking-rewards pallet to calculate and disburse LP fees. So it makes sense to have a more generic name like `pallet-earn`. @hussein-aitlahcen any ideas?

cor:
> @dzmitry-lahoda
> 
> > can we have image assets to be lfs? or our story to have monorepo will end fast ...
> 
> What do you mean by `lfs` ? You mean the size of the images being high? I don't know if that is a going to be a problem. But I'm also ok with moving the design docs to another repo if it's ok with @cor
> 
> > not sure, can it be a more clear separation of biz docs vs tech docs?
> 
> IMHO we need more integration of biz side and tech docs to reduce the communication gap in translating requirements to implementation :)

@vimukthi-git LFS is out of scope for this PR. Migrating to LFS may be nice in the future once we hit storage issues, but for now we're not doing it yet as it requires rebasing the entire repo. Please deploy them like this and merge this PR :)

frame/composable-traits/src/staking_rewards.rs
```diff
@@ -204,6 +205,26 @@ pub trait StakingReward {
...
+	///
+	/// Arguments
+	///
+	/// * `asset` - the protocol asset to reward.
+	/// * `reward_asset` - the reward asset to transfer.
+	/// * `from` - the account to transfer the reward from.
+	/// * `schedule` - the schedule that decides the rewarding rate.
+	/// * `keep_alive` - whether to keep alive or not the `from` account while transferring the
+	///   reward.
+	fn transfer_vested_reward(
```
vimukthi-git, line 221: Remove

dzmitry-lahoda:
@cor 
> @vimukthi-git LFS is out of scope for this PR. Migrating to LFS may be nice in the future once we hit storage issues, but for now we're not doing it yet as it requires rebasing the entire repo. Please deploy them like this and merge this PR :)

not sure what you mean? LFS does not requires any rebases. 

also later will be too late. each binary added not to lfs bloats history, so you have to purge history, which we will not do. so it makes monorepo soon to end...

dzmitry-lahoda:
> IMHO we need more integration of biz side and tech docs to reduce the communication gap in translating requirements to implementation :)

hm, i think wisdom is different. do not mix is right.

dzmitry-lahoda:
> I'm ok with that, I use mostly md too. Normally I use adoc to create the initial document structure and then keep editing the md. However there would be some effort there in committing the plantuml source with that approach.

as reference to file in md or as ```plantuml

rfcs/0004-pablo-distribution.adoc
dzmitry-lahoda: let look if we really has 2 composable pallets, actually whole composable is `earn` runtime. so i would not be so broad.

rfcs/0004-pablo-distribution.adoc
```diff
@@ -0,0 +1,480 @@
+= Design Proposal: Pablo Fees & Staking Rewards Distribution
+:math:
+:stem:
+:toc:
+:toclevels: 4
+:sectnums:
```
dzmitry-lahoda, line 6: in MD there is pluging to insert toc, but yeah, not sure how automated it is. other, may be github md supports toc too

rfcs/0004-pablo-distribution.md
```diff
@@ -0,0 +1,447 @@
...
+fees) mechanism while considering various options and capturing the
+discussions about the subject.
+
+`TODO summarise the mechanism`
+
+## 2. Background
+
+### 2.1. PBLO Token Initial Distribution
+
+According to the [tokenomics design](http://link) the Pablo is supposed
```
dzmitry-lahoda, line 70: is it broken?

vimukthi-git: There is a design though I have not received the link yet, hence it's kind of a to do

rfcs/0004-pablo-distribution.adoc
vimukthi-git: I'm pretty sure I have tried most md plugins in the past, very hard to get a nice numbered/linked/multilevel toc with md.

rfcs/0004-pablo-distribution.adoc
```diff
@@ -0,0 +1,480 @@
...
+
+"But if required this effect can be negated by increasing the trading fee by a " delta a_n " while at the same time subtracting it from the total fees paid out already to liquidity providers as follows,"
+
+"New trading fee " = x_(adjusted) = l / (sum l_m) * (sum a_n + delta a_n)
+
+"For " r_(th) " liquidity provider,"
+
+delta a = (sum_(r+1) l_m) / (sum l_r) * sum a_n
+
+= x_(adjusted) = l_(r) / (sum_(r+1) l_m + sum l_r) * (sum a_n + (sum_(r+1) l_m) / (sum l_r) * sum a_n )
```
dzmitry-lahoda, line 242: not sure from where `sum l_r` appears in denumerator.

also can you recheck how `x_(adjusted) ` is simplified into `l_r / (sum l_r) * sum a_n`.

it is so similar to `x = l / (sum l_m) * sum a_n` what is difference of `m` and `r`?

rfcs/0004-pablo-distribution.adoc
```diff
@@ -0,0 +1,480 @@
...
+
+"With this adjusted value all later additions to LP shares have been negated when receiving fees for earlier LPs."
+
+
+++++
+
+[#_pblo_staker_trading_fee_distribution]
+==== PBLO Staker Trading Fee Distribution
+
+This is the reward a `PBLO` staker receives from the trading fees of Pablo pools. It is equal to the protocol fee charged on Pablo pools. This can be accomplished by calling the already existing `StakingReward.transfer_reward` interface as follows. According to product there is also a need to convert whatever the fee asset in to PBLO to create a demand/additional value for PBLO.
```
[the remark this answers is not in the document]
dzmitry-lahoda: i do not think conversion needed. you can make pablo price in LP + lp = total. each each user possition is sum of his pablo with multiplier + lp. if you want more pablo used - raise pablo price. and give away nfts to user to specfici pools kind of. so config is price of palbo vs lp. so you incease rewards share for in pool as one dimensions, and change ration of pablo vs pool token as other dimensions.

rfcs/0004-pablo-distribution.adoc
```diff
@@ -0,0 +1,480 @@
...
+
+x _n = P_(current) . s_n / (sum s_n)
+
+------------------------------------------------
+
+"When adding a new staker n+1, existing stakers(n) reward would be,"
+
+x_n = P_(current) . s_n / (sum s_(n+1))
+
+"As this is less than what is expected above, an adjustment " delta P " to total reward pool can be made to allow realtime reward calculations,"
```
dzmitry-lahoda, line 324: i doubt it will work. so you may try. may be i do not know.

rfcs/0004-pablo-distribution.adoc
```diff
@@ -0,0 +1,480 @@
...
+----
+include::../frame/composable-traits/src/staking_rewards.rs[lines=103..122]
+----
+This data structure can be modified to support the RP approach by adding/adjusting/removing some fields.
+
+[source,rust]
+----
+pub struct StakingNFT<AccountId, AssetId, Balance, Epoch, Rewards> {
+    /// Reward Pool ID from which pool to allocate rewards for this
+    pub reward_pool_id: ID,
```
dzmitry-lahoda, line 373: so resonable to add such kind id to staked positions

rfcs/0004-pablo-distribution.adoc
```diff
@@ -0,0 +1,480 @@
...
+    pub stake: Balance,
+
+    /// New: Pool share received for this NFT
+    pub share: Balance,
+
+    /// New: reduced rewards for the fNFT (d_n)
+    pub reduction: Balance,
+
+    /// The reward epoch at which this NFT will start yielding rewards. This is no longer needed.
+    /// pub reward_epoch_start: Epoch,
```
dzmitry-lahoda, line 388: can handle, nice

rfcs/0004-pablo-distribution.adoc
```diff
@@ -0,0 +1,480 @@
...
+
+    /// List of reward asset/pending rewards. This is no longer needed as pending rewards are pooled
+    /// pub pending_rewards: Rewards,
```
dzmitry-lahoda, line 391: not sure about it.  can make nft with stake only so when it is rewarded all goes to stake. so it is locked compunding variant.

rfcs/0004-pablo-distribution.adoc
```diff
@@ -0,0 +1,480 @@
...
+[appendix]
+== Fee Distribution Q&A
+
+Based on the current setup following questions arise when deciding on the distribution of these fees to relevant liquidity providers, owners and stakers.
+
+. A Protocol Fee for all pools in Pablo (or even protocol pallets other than Pablo)?
++
+Does it make sense to define a protocol fee percentage on top of the pool owner fees of the pools so that the protocol fee can be used as the pot out of which the stakers are rewarded? Initially the Protocol Fee = Pool Owner Fee as the pools are owned by Composable. Assumption here is that the stakers would indeed still get a reward out of third party created pool fees.
++
+*Comment:* While having a protocol funding mechanism is valuable, initially the protocol fees should zero or minimal.
```
dzmitry-lahoda, line 470: PICA is Coposable fee i guess. composable should motivate to use more pica. creating pools also motivatrs for PICA

rfcs/0004-pablo-distribution.adoc
```diff
@@ -0,0 +1,480 @@
...
+. How does the system reward PICA stakers? Wouldn't the Pablo protocol needs some parameter to define how much of its swap fee or protocol fee as referred to above would go to PICA holders? Or do we assume that PICA stakers do not get a reward out of the Pablo pool fees?
+.. If Pablo does reward PICA stakers, the system might need a common interface that directs those funds out of Pablo.
+.. If Pablo does reward PICA stakers, the system might need to have a treasury parameter that defines the percentage that goes out to PICA holders that can be adjusted overtime.
```
dzmitry-lahoda, line 473: each pool can define ratio of lp to pica to pablo. and split revenue here. at same time it defines ration, it may define 3 protocol ids and store array in pablo. when palo rewards, it loops over pools.

rfcs/0004-pablo-distribution.adoc
```diff
@@ -0,0 +1,480 @@
...
+----
+
+Now in order to allow redeeming the above fNFT, following data structures needs to be tracked,
+
+
+
+[#_rewardaccumulationhook]
+==== [Needs modification] RewardAccumulationHook
+
+Following algorithm is to added as part of the existing https://github.com/ComposableFi/composable/blob/main/frame/staking-rewards/src/lib.rs#L363[block hook] in staking rewards pallet. As it is only accumulating new rewards for an upcoming epoch, the code is proposed to be run inside a new state `State:AccumulatingRewards`.
```
dzmitry-lahoda, line 414: i think pablo can have its own loop calling stakin-rewards, no need to modify staking-rewards. so there can be addtional pallet for configuration of lp/pablo/pica prices relation.

rfcs/0004-pablo-distribution.adoc
```diff
@@ -0,0 +1,480 @@
...
+----
+start
+->State:AccumulatingRewards;
+:rewardConfigIndex = read(rewardConfigCount) - 1;
+repeat :rewardConfig = read(rewardConfigs[rewardConfigIndex--]);
+    :reward_allocation = Assets::balance_of(reward_allocation_acc,
+            rewardConfig.asset);
+    :epochReward = reward_allocation * rewardConfig.reward_rate;
+    :stakerReward = epochReward;
+    :perAssetRewardAllocationIndex = read(perAssetRewardAllocationCount) - 1;
```
dzmitry-lahoda, line 426: can add additional step into reward hook, which can be initialized with amount and rate. so that on each ecoch reward is put into staked fnft. that is so called linear vesting so. i think we should have linear vesting which can vest into  account of rewards. so may be need to replace reward EP by accounts for rewards. and compos with linear vesting.

rfcs/0004-pablo-distribution.adoc
vimukthi-git: I understand your concern, but you may think of this document as source code of the actual document like latex source code of a document. This when converted, creates the actual document, that is a markdown file. Some others can decide to edit the markdown file generated later when the doc is more stable, which is totally fine. I've been following this workflow for some years now.

rfcs/0004-pablo-distribution.adoc
vimukthi-git: > not sure from where sum l_r appears in denumerator.

Just a trick to get the math to work by dividing the full sum to the part before `0 to r th` element and `r+1 to m th` element. `m` is the total number of liquidity additions(LPs).

>also can you recheck how x_(adjusted) is simplified into l_r / (sum l_r) * sum a_n.

Yes, it is correct. 

`sum_(r+1) l_m + sum l_r` parts inside and outside the brackets cancel each other.

> it is so similar to x = l / (sum l_m) * sum a_n what is difference of m and r?

`r` is one of the earlier LPs, `m` is the total number of LPs at the end.

vimukthi-git: Note that trading fees optimization is optional though. It's there because I was trying to understand the problems with calculating trading fees for new LPs without taking into account the dilution to earlier LPs

rfcs/0004-pablo-distribution.adoc
vimukthi-git: For the cases we are considering above (5 points), this works as the math suggests here. But there may be cases which I have not considered in relation to compounding, extending etc., that I may need your help in resolving.

rfcs/0004-pablo-distribution.adoc
vimukthi-git: Yeah, this was necessary for making the concept of a reward pool inside the staking rewards pallet

rfcs/0004-pablo-distribution.adoc
vimukthi-git: Note that this part is outdated.

rfcs/0004-pablo-distribution.adoc
vimukthi-git: Updated things now

rfcs/0004-pablo-distribution.md
```diff
@@ -0,0 +1,447 @@
...
+        to PICA holders that can be adjusted overtime.
+
+        **Comment:** PICA stakers would not be rewarded from the Pablo
+        fees. PICA stakers are rewarded in newly minted PICA(or PBLO
+        later), Mechanism to transfer the PICA tokens for stakers does
+        not exist, need to be built.
+
+3.  Does it make sense to define a Pool Owner Fee(Protocol Fee as
+    referred to above) for LBPs that goes out to Pablo holders reward
+    pool?
```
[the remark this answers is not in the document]
vimukthi-git: Just Q&A

rfcs/0004-pablo-distribution.adoc
vimukthi-git: Compounding is out of scope now as explained on slack :)  so resolving

rfcs/0004-pablo-distribution.adoc
```diff
@@ -0,0 +1,626 @@
...
+stop
+----
+
+==== Split Position
+
+[plantuml,images/split-position,png]
+----
+start
+->assetId\npositionId\nratio;
+:rewards_pool = readStorage(assetId);
```
vimukthi-git, line 516:
```suggestion
```

rfcs/0004-pablo-distribution.adoc
```diff
@@ -0,0 +1,626 @@
...
+==== Update Reward Pool
+
+[plantuml,images/update-reward-pool,png]
+----
+start
+->assetId\nreward_update\n;
+:rewards_pool = readStorage(assetId);
+while (reward = reward_update.rewards.has_next) is (yes)
+    :assert max_reward not exceeded;
+    :reward.total_rewards += reward_update[reward.asset_id].amount;
```
vimukthi-git, line 553: Before updating the values, need to calculate the epoch rewards using previous values and add to the reward pool

rfcs/0004-pablo-distribution.adoc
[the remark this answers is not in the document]
dzmitry-lahoda: i more of next
1. you have 1000 USDC + 1000 USDT pool.
2. you say that 20% of fees will go to PBLO stakers and 80% to LP stakers. 
3. it is same as saying that PBLT cost 1/4 LP. 
4. so people are more to stake LP and bring assets.
5. but if say PBLO 80% and 20% LP, than 4x
6. so that people would chace pablo.
7. In code of Pablo pallet, can send rewards to 2 staking pools, using 1/4 ratio.

KaiserKarel (dismissed):

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+:toc:
+:toclevels: 4
+:sectnums:
+:sectnumlevels: 4
+
+== Abstract
+
+This document proposes the Pablo distribution(token and pool trading fees) mechanism while considering various options and capturing the discussions about the subject.
+
+`TODO summarise the mechanism`
```
dzmitry-lahoda, line 15: it seems summary is what we need first hand. because need to have clear team goals

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+== Background
+
+=== PBLO Token Initial Distribution
+
+According to the http://link[tokenomics design] the Pablo is supposed to be distributed as follows,
+
+* Ecosystem fund - x
+* Farming rewards - y
+* LBP Auction - z
+* ...
```
dzmitry-lahoda, line 26: not sure what ... means. how we can act on it?

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+
+TODO fill above with details from the original sheet.
```
dzmitry-lahoda, line 28: please refer original shit here. i was not shared with it.

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+
+The farming rewards are incentives for liquidity provider(LP)s who stake their LP tokens for Pablo pools. How much reward is allocated as incentive for each pool is to be decided by governance.
+
+=== Pool Fees
+
+Composable intends to distribute some percentage of the swap(transaction) fees captured by the pools in Pablo dex pallet as rewards to users who stake their `PBLO` tokens using the staking-rewards pallet interface. The idea is to incentivize the continuous owning of the staked `PICA` and `PBLO` to earn these yields which increases the value of the ecosystem overall by increasing the desirability of the staked assets.
```
dzmitry-lahoda, line 34: No need to do, but nice when `staking-rewards` is reference to that pallet

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+
+At the time of writing Pablo has the following fee parameters other than for liquidity bootstrapping pools(LBP) which do not charge fees,
+
+. LP Fee - A percentage of the trading fee that is distributed to liquidity providers based on the number of liquidity provider(LP) tokens they minted at the time of providing the liquidity.
+. Pool Owner Fee - A percentage of the trading fee that is distributed to the pool owner.
+
+==== LP Fee Distribution
+
+This is yet to be implemented in Pablo, hence the idea is that it can be addressed in the context of this proposal.
```
dzmitry-lahoda, line 43: it can be done in summary. and document remaining parts written as if it is already done

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+
+== Use Cases
+
+Following is a summary of use cases omitting the UI specific use cases for brevity.
```
dzmitry-lahoda, line 47: you may omit ommitting notes for bravity too

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+.. Same pool can receive multiple types of tokens as rewards.
+. The system MUST support accumulating the LP share of Pablo trading fees.
+. Pablo trading fees(LP fee part) MUST be disbursed according to LP token share of each LP. Fees are accumulated towards increasing liquidity in a pool while allowing LPs to redeem the fee share with their LP tokens at a preferred time.
+
+=== PBLO Stakers
+
+. System MUST allow staking of PBLO. This must be implemented through the fNFT mechanism with multiple time period unlocks being possible for users.
+. The system MUST accumulate the rewards share for PBLO holders who stake PBLO token, out of the PBLO supply allocated for them.
+. The system MUST support accumulating the (stakers) reward part of the Pablo trading fees.
+. The system must support rewards being distributed on granular basis - e.g every 6 or 12 hours.
```
dzmitry-lahoda, line 112: on granularity of day less than day, 
may be

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+. The users MUST be able to claim the rewards once distributed.
+. The system SHOULD support rewards in the form of fNFTs.
+
+=== PICA Stakers
+
+. System MUST allow staking of PICA. This must be implemented through the fNFT mechanism with multiple time period unlocks being possible for users.
+. The system MUST accumulate the rewards share for PICA holders who stake PICA token, out of the PICA supply allocated for them.
+. The system MUST support accumulating any token rewards other than PICA for PICA stakers.
+. The system must support rewards being distributed on granular basis - e.g every 6 or 12 hours.
+. The users MUST be able to claim the rewards once distributed.
```
dzmitry-lahoda, line 122: to claim distributed rewards eventually.

may be.

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+=== System Overview
+
+[plantuml,images/pablo-distribution-verview,png]
+----
+skinparam component {
+  backgroundColor<<exists>> LightGreen
+  backgroundColor<<remove>> Red
+}
+
+node "Pallet-Staking-Rewards\n(Pallet-Rewards?)" {
```
dzmitry-lahoda, line 152: so we should decide here. so fNFT is not inside staking-rewards pallet. as we agreed it is outside.

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+    StakingReward - [Reward Logic]
+    Staking - [fNFT Storage/Logic]
+    Node "BatchProcess" {
+        [RewardDisbursementHook]
+        [RewardsConfig] --> [RewardAccumulationHook]
+    }
+}
+
+[fNFT Storage/Logic] <<exists>>
+[RewardDisbursementHook] <<remove>>
```
dzmitry-lahoda, line 162: no sure what Remove means. May be remove Remove?

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+[Pool] <<exists>>
+
+
+node "Governance" {
+    [Configuration] --> [FeeConfig]
+    [Treasury]--> StakingReward: Transfer LP, PBLO\nor other token rewards
+}
+----
+
+TODO: What to do for part of protocol fees that should be transferred to treasury eventually as treasury does not stake it's PBLO?
```
dzmitry-lahoda, line 180: can remove it. as RFC does not decides on it. 

may be it should not stake. afaik treasury is used in emergencies of some kinds or rewarding. 

so kind of remove TODO, to RFC would be final.

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+        [RewardDisbursementHook]
+        [RewardsConfig] --> [RewardAccumulationHook]
+    }
+}
+
+[fNFT Storage/Logic] <<exists>>
+[RewardDisbursementHook] <<remove>>
+
+node "Pallet-Pablo" {
+    [Pool] --> StakingReward : transfer\ntrading fees\nfor PBLO stakers
```
dzmitry-lahoda, line 165: afaik I see this is image. can be make images LFS or remove these? are we agreed pushing images in monorepo? if yes - i am ok.

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+    [Configuration] --> [FeeConfig]
+    [Treasury]--> StakingReward: Transfer LP, PBLO\nor other token rewards
+}
+----
+
+TODO: What to do for part of protocol fees that should be transferred to treasury eventually as treasury does not stake it's PBLO?
+
+=== Pallet-Pablo
+
+In order to 1. support LP staking 2. LP trading fee distribution and 3. PBLO staking reward using trading fees, following changes are proposed for https://github.com/ComposableFi/composable/tree/main/frame/pablo[Pallet-Pablo].
```
dzmitry-lahoda, line 184: [Pallet-Pablo] link has no readme, usually it is references to be clear what is it. at least one line summary

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+
+[#_feeconfig]
+==== FeeConfig
+
+Each pool in Pablo defines a fee percentage to be charged for each trade.Except for LBPs other pools also define an owner fee that is a percentage out of the main trading fee. The `FeeConfig` is a new abstraction over all fees that could be charged on a pool to allow for extension. At this time a 100% of the owner fee should be defined as a new field `protocol_fee`.
```
dzmitry-lahoda, line 189: LBPs comment is detail which may be moved from RFC to pablo pallet readme to make less text in RFC and more targeted.

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+
+[plantuml,images/pablo-fee-config,png]
+----
```
dzmitry-lahoda, line 192: not sure. why this is class, while later same type of information is rust Struct? can we handle this consistently?

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+    +owner_fee_rate: Permill,
+    +protocol_fee_rate: Permill,
+}
+----
+
+*Existing code must be modified to use this data structure*.
+
+Given this,
+----
+fee = // calculation depends on the pool type: based on the fee_rate
```
dzmitry-lahoda, line 204: please make this python

```

owner_fee_rate = 20/100
protocol_fee_rate = 100/100

fee = 100
owner_fee = fee * owner_fee_rate * (1 - protocol_fee_rate)
protocol_fee = owner_fee * protocol_fee_rate

print(protocol_fee)
print(owner_fee)
```
I reject PR until it is fixed on why both fees are 0

dzmitry-lahoda (changes requested):
please fix error with 0 fees

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
vimukthi-git: This is actually the opposite of how normal technical documentation process work, you write the abstract at the end. 
Order roughly is as follows,
1. Background
2. Requirements
3. Method
4. Implementation/Milestones etc
5. Abstract

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+
+This document proposes the Pablo distribution(token and pool trading fees) mechanism while considering various options and capturing the discussions about the subject.
+
+`TODO summarise the mechanism`
+
+== Background
+
+=== PBLO Token Initial Distribution
+
+According to the http://link[tokenomics design] the Pablo is supposed to be distributed as follows,
```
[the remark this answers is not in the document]
vimukthi-git: The broken link is intentional as we don't have a public shareable doc.

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
vimukthi-git: This will completed in time for now it's a TODO overall.

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
vimukthi-git: It's not shareable at the moment

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
vimukthi-git: The design proposals are written before the implementation. So it is written in proposal form to be implemented

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
vimukthi-git: I'm not sure I follow what you mean? Could you clarify or give an example?

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
dzmitry-lahoda:
```
protocol_fee = fee * owner_fee_rate * (1 - protocol_fee_rate) * protocol_fee_rate


y = C (1 - x) * x

y = C x - C x ^ 2

optimal,

C = 2 C x  

x = 1/ 2
```

maximal protocol amount will be in case of its fee is 0.5

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
vimukthi-git: I consider this a NIT, so won't address for now

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
vimukthi-git: The requirements are written as the action a party can take, the implementation details such as when they could be claimed are upto the method.

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
vimukthi-git: Good point let me make a change

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
vimukthi-git: This is exactly what we are doing to these hook in the pallet, which has already been removed

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
vimukthi-git: I consider this discussion settled as multiple parties gave me the go ahead to merge with images.

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
vimukthi-git: Pablo readme is not of concern to this doc.

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
vimukthi-git: It's just laying out things that are relevant to fee calculations for this design.
Readme can be updated at a later point.

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
vimukthi-git: Good point I can switch this to rust

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
vivekvpandya: so owner_fee goes to PBLO holder, if its is zero then nothing goes to them.
AFAIU protocol_fee is value transferred to liquidity providers. So having protocol_fee_rate zero makes sense only perhaps for LBP. 
So Fee= protocol_fee - (protocol_fee * owner_fee_percent)

vimukthi-git: `owner_fee_rate` is intended to capture the total fee that goes out to owner plus the protocol token (PBLO) holders,

as default for this can be found already implemented at https://github.com/ComposableFi/composable/blob/main/frame/composable-traits/src/dex.rs#L175

vimukthi-git: The idea is that PBLO owners would only get a cut of the fee of the owner of the pool, if they don't charge, we don't charge. But by default the pool owner fee is zero and PBLO holders get the full owner fee.

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
dzmitry-lahoda: so expectancy if there is reward then it is always possible to claim immediately? so automtic reward stops if there is no ability to claim?

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
dzmitry-lahoda: Not sure, just mark in code what you whant to remove and no need to have that in RFC.

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
dzmitry-lahoda: remove this text

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
dzmitry-lahoda: please summarize in PR, not in comment

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
dzmitry-lahoda: may you tag here who can approve that?

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+
+total liquidity stem:[= sum l_m]
+
+fees and liquidity returned for an LP amount stem:[l = l / (sum l_m) * (sum l_m + sum a_n)]
+
+stem:[= l + l / (sum l_m) * sum a_n]
+
+trading fees received stem:[= x = l / (sum l_m) * sum a_n]
+
+"When pool size stem:[sum l_m] increases the amount of trading fees received stem:[x] reduces for a particular LP position.
```
dzmitry-lahoda, line 230: "

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+
+For large pool sizes of stem:[sum l_m] (steady state) this effect is negligible, hence it's a good enough strategy to distribute fees.
```
dzmitry-lahoda, line 232: not sure what is comment about. this works for anu pool sizes and by design honest (if we keep in mind using dillution protection)

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+
+But if required this effect can be negated by increasing the trading fee by a stem:[delta a_n] while at the same time subtracting it from the total fees paid out already to liquidity providers. Refer <<_trading_fee_inflation_to_avoid_dilution_of_lps>>.
```
dzmitry-lahoda, line 234: both total fees and payouts should be kind if increaced. no sure what substracting means here

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+owner_fee = fee * owner_fee_rate * (1 - protocol_fee_rate);
+protocol_fee = owner_fee * protocol_fee_rate;
+----
+For all pools launched at the Picasso launch following values would be set for these configs
+----
+owner_fee_rate = 20%
+protocol_fee_rate = 100% // all owner fees goes to composable to be distributed as rewards
+----
+
+==== LP Trading Fee Distribution
```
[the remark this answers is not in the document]
dzmitry-lahoda: for me this is misleading chapter

rfcs/0005-pablo-distribution.md
```diff
@@ -0,0 +1,667 @@
...
+Pablo pools. It is equal to the protocol fee charged on Pablo pools.
+This can be accomplished by calling the already existing
+`StakingReward.transfer_reward` interface as follows. According to
+product there is also a need to convert whatever the fee asset in to
+PBLO to create a demand/additional value for PBLO.
+
+<img src="0005-pablo-distribution-assets/images/images/pablo-fNFT-pblo-staking-fee-distro.png" width="510" height="346" alt="pablo fNFT pblo staking fee distro" />
+
+Will it need a change in
+[this](https://github.com/ComposableFi/composable/blob/main/frame/composable-traits/src/staking_rewards.rs#L96)
```
dzmitry-lahoda, line 314: you can use relative linkin in markdown. and link is broken now

rfcs/0005-pablo-distribution.md
```diff
@@ -0,0 +1,667 @@
...
+#### 5.2.3. PBLO Staker Trading Fee Distribution
+
+This is the reward a `PBLO` staker receives from the trading fees of
+Pablo pools. It is equal to the protocol fee charged on Pablo pools.
+This can be accomplished by calling the already existing
+`StakingReward.transfer_reward` interface as follows. According to
+product there is also a need to convert whatever the fee asset in to
+PBLO to create a demand/additional value for PBLO.
+
+<img src="0005-pablo-distribution-assets/images/images/pablo-fNFT-pblo-staking-fee-distro.png" width="510" height="346" alt="pablo fNFT pblo staking fee distro" />
```
dzmitry-lahoda, line 311: swapping anything to pablo should not have fee? so fee could have assymetry gearing toward pablo. so we need assymetry in pools to allow one sided fees. right?

rfcs/0005-pablo-distribution.md
```diff
@@ -0,0 +1,667 @@
...
+
+This section covers how the staking rewards are distributed using the
+[staking rewards
+pallet](https://github.com/ComposableFi/composable/tree/main/frame/staking-rewards).
+
+#### 5.3.1. Analysis of Reward Calculations
+
+In order to create the necessary reward pool as well as the rewarding
+rate for stakers the following model can be used. It tries to address
+the following constraints (TODO move to requirements)
```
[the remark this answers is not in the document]
dzmitry-lahoda: document you shared as examples was full of todos?

rfcs/0005-pablo-distribution.md
```diff
@@ -0,0 +1,667 @@
...
+
+1.  Allow <span id="rate">specification of the reward rate for a
```
dzmitry-lahoda, line 329: dynamically changing rate [postponed]

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
vimukthi-git: Why should it be removed? It's providing context that UI specific use cases are not covered by this doc.

rfcs/0005-pablo-distribution.md
vimukthi-git: these would be addressed with a later PR after staking rewards is done, no point in going into details now

rfcs/0005-pablo-distribution.md
vimukthi-git: It is not, it will be implemented with the first iteration, refer section 5.3.7

rfcs/0005-pablo-distribution.md
```diff
@@ -0,0 +1,667 @@
...
+    possible for users to just re-stake their earned assets.
+
+To analyze the requirement fully, let’s define the following terms for a
+given staking reward pool,
+
+Pre-defined reward rate (say per second) <img src="0005-pablo-distribution-assets/images/stem-6fb32a8803a6d58cd54908033a2556f9.png" width="23" height="6" alt="stem 6fb32a8803a6d58cd54908033a2556f9" /></span>
+
+Pre-defined reward calculation epoch in seconds <img src="0005-pablo-distribution-assets/images/stem-6184b58307a1dc90934a6a7051a42ceb.png" width="22" height="8" alt="stem 6184b58307a1dc90934a6a7051a42ceb" /></span>
+
+Reward per calculation epoch <img src="0005-pablo-distribution-assets/images/stem-b219ff7e7a0df744c99c2e11229a1ded.png" width="33" height="8" alt="stem b219ff7e7a0df744c99c2e11229a1ded" /></span>
```
dzmitry-lahoda, line 358: looks like r.t which is supposed to be r*t? like multiplication?

rfcs/0005-pablo-distribution.md
```diff
@@ -0,0 +1,667 @@
...
+
+Previous total reward pool before the current epoch <img src="0005-pablo-distribution-assets/images/stem-53fadade13e71b863963af9a23b28b71.png" width="25" height="8" alt="stem 53fadade13e71b863963af9a23b28b71" /></span>
+
+Assuming there is a per epoch calculation which adds to the pool, the
```
dzmitry-lahoda, line 362: there is no need for epoches. rate is per second.

rfcs/0005-pablo-distribution.md
```diff
@@ -0,0 +1,667 @@
...
+Existing <img src="0005-pablo-distribution-assets/images/stem-39e8c7852cdbd74b28d331353778e128.png" width="21" height="9" alt="stem 39e8c7852cdbd74b28d331353778e128" /></span>
+staker reward,
+
+<img src="0005-pablo-distribution-assets/images/stem-e1359ae7d0fae29ebf9e42efcaa5536e.png" width="111" height="18" alt="stem e1359ae7d0fae29ebf9e42efcaa5536e" /></span>
+
+##### 5.3.1.1. When adding a new staker <span class="image"><img src="0005-pablo-distribution-assets/images/stem-0b46f732c83c0e66067b0e50c2156089.png" width="29" height="8" alt="stem 0b46f732c83c0e66067b0e50c2156089" /></span>, existing stakers(<span class="image"><img src="0005-pablo-distribution-assets/images/stem-55a049b8f161ae7cfeb0197d75aff967.png" width="9" height="6" alt="stem 55a049b8f161ae7cfeb0197d75aff967" /></span>) reward would be,
+
+<img src="0005-pablo-distribution-assets/images/stem-569c4bf984a23f18046277fd561e89a3.png" width="126" height="20" alt="stem 569c4bf984a23f18046277fd561e89a3" /></span>
+
+As this is less than what is expected above, an adjustment " delta P "
```
dzmitry-lahoda, line 384: not sure why not to render as $delta_P$?

rfcs/0005-pablo-distribution.md
vimukthi-git: There has to be a batch process that calculates the reward for the pool so epoch is how often that batch job runs

rfcs/0005-pablo-distribution.md
```diff
@@ -0,0 +1,667 @@
...
+<img src="0005-pablo-distribution-assets/images/stem-828ec270409cb6ff5cfc583587d0eae9.png" width="142" height="18" alt="stem 828ec270409cb6ff5cfc583587d0eae9" /></span>
+← (1)
+
+In general,
+
+<img src="0005-pablo-distribution-assets/images/stem-0f2f030a4f8a3c172e968af2768a3ec8.png" width="349" height="11" alt="stem 0f2f030a4f8a3c172e968af2768a3ec8" /></span>
+
+##### 5.3.1.2. When removing a staker from the pool the above addition step has to be reverted
+
+This can be done by subtracting the <img src="0005-pablo-distribution-assets/images/stem-7c4ec4f9c189cb8f3edb39740e43c33f.png" width="16" height="10" alt="stem 7c4ec4f9c189cb8f3edb39740e43c33f" /></span>
```
dzmitry-lahoda, line 409: > This can be done by subtracting the stem 7c4ec4f9c189cb8f3edb39740e43c33f of the staker from the total pool and reducing the pool shares accordingly.

please clarify.

when we add new stake, we add delta_p to total pool, and remove d_n from stake of each user (or new users)? but then when we remove share, we should remove d_n from pool. so we addd delta_p and remove d_n. what is relation of delta_p to d_n?

rfcs/0005-pablo-distribution.md
```diff
@@ -0,0 +1,667 @@
...
+
+<img src="0005-pablo-distribution-assets/images/stem-38d917fea7c6a7a47eb1aa77edd4da97.png" width="169" height="20" alt="stem 38d917fea7c6a7a47eb1aa77edd4da97" /></span>
+
+<img src="0005-pablo-distribution-assets/images/stem-5fcfbc0bc69ee8b8f356ce2bbfb42002.png" width="190" height="22" alt="stem 5fcfbc0bc69ee8b8f356ce2bbfb42002" /></span>
+
+<img src="0005-pablo-distribution-assets/images/stem-e1359ae7d0fae29ebf9e42efcaa5536e.png" width="111" height="18" alt="stem e1359ae7d0fae29ebf9e42efcaa5536e" /></span>
+
+**Therefore, the existing staker receives the same reward as before**
+
+To compensate for this new adjustment, a reduction <img src="0005-pablo-distribution-assets/images/stem-7c4ec4f9c189cb8f3edb39740e43c33f.png" width="16" height="10" alt="stem 7c4ec4f9c189cb8f3edb39740e43c33f" /></span>
```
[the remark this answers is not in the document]
vimukthi-git: Not going to explain/relitigate again as I have explained this in multiple occasions before

rfcs/0005-pablo-distribution.md
```diff
@@ -0,0 +1,667 @@
...
+are tracked for each staker, they can always claim the new reward share
+from <img src="0005-pablo-distribution-assets/images/stem-32efe856de4078991a47242cc1d89349.png" width="43" height="11" alt="stem 32efe856de4078991a47242cc1d89349" /></span>
+later.
+
+##### 5.3.1.4. When extending an existing position
+
+Extension of an existing staker position can be treated in the same way
+as adding a new staker as the following relationship holds with the new
+stake <img src="0005-pablo-distribution-assets/images/stem-9849cee8ec3e29bf6d2ea80a64d995dd.png" width="23" height="10" alt="stem 9849cee8ec3e29bf6d2ea80a64d995dd" /></span>
+and the corresponding inflation <img src="0005-pablo-distribution-assets/images/stem-e7d319c4dcb739d8e91edd37454e20e8.png" width="25" height="10" alt="stem e7d319c4dcb739d8e91edd37454e20e8" /></span>,
```
dzmitry-lahoda, line 427: we do not mint. so we do not inflate

rfcs/0005-pablo-distribution.md
```diff
@@ -0,0 +1,667 @@
...
+
+new staker to add <img src="0005-pablo-distribution-assets/images/stem-1b75a50a55357d9a7a8d3ecbb06df470.png" width="180" height="20" alt="stem 1b75a50a55357d9a7a8d3ecbb06df470" /></span>
+← (2)
+
+Now with (1) + (2),
+
+<img src="0005-pablo-distribution-assets/images/stem-6a000622e842e98de57502915826da7b.png" width="245" height="20" alt="stem 6a000622e842e98de57502915826da7b" /></span>
+
+Therefore, same computation as before with <img src="0005-pablo-distribution-assets/images/stem-9849cee8ec3e29bf6d2ea80a64d995dd.png" width="23" height="10" alt="stem 9849cee8ec3e29bf6d2ea80a64d995dd" /></span>
```
dzmitry-lahoda, line 436: no sure that i see this formula works well . imagine i added 100 PBLO for 4 years. than in 2 years he adds 1000000000 PBLO. but its leverage will be like for 4 years. really he locked for 2 years.

rfcs/0005-pablo-distribution.md
```diff
@@ -0,0 +1,667 @@
...
+number of shares added to the staker position works as expected.
+
+##### 5.3.1.5. When splitting an existing position
+
+As the total reward pool is not affected the splitting is just creating
+a new position using some ratio. If the ratio is <img src="0005-pablo-distribution-assets/images/stem-603de94498e154610e3066ec63603017.png" width="25" height="9" alt="stem 603de94498e154610e3066ec63603017" /></span>
```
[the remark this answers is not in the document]
dzmitry-lahoda: if k= 0.1 than xn1 will be more than xn original. are you proposing to add loop hole to protocol?

rfcs/0005-pablo-distribution.md
```diff
@@ -0,0 +1,667 @@
...
+        /// The reward epoch at which this NFT will start yielding rewards.
+        pub reward_epoch_start: Epoch,
+        /// List of reward asset/pending rewards.
+        pub pending_rewards: Rewards,
+        /// The date at which this NFT was minted.
+        pub lock_date: Timestamp,
+        /// The duration for which this NFT stake was locked.
+        pub lock_duration: DurationSeconds,
+        /// The penalty applied if a staker unstake before the end date.
+        pub early_unstake_penalty: Penalty<AccountId>,
```
dzmitry-lahoda, line 476: no sure, why Penalty depends on AccountId?

rfcs/0005-pablo-distribution.md
```diff
@@ -0,0 +1,667 @@
...
+        /// The reward multiplier.
+        pub reward_multiplier: Perbill,
+    }
+
+This data structure can be modified to support the RP approach by
+adding/adjusting/removing some fields.
+
+    /// This can possibly be renamed as `RewardPoolPosition`
+    pub struct StakingNFT<RewardPoolId, AccountId, AssetId, Balance> {
```
dzmitry-lahoda, line 485: hm, other struct was called StakingNFT too....

rfcs/0005-pablo-distribution.md
```diff
@@ -0,0 +1,667 @@
...
+        pub reward_pool_id: RewardPoolId,
+
+        /// The staked asset.
+        pub asset: AssetId,
+        /// The original stake this NFT was minted for.
+        /// Used for calculating the pool share.
+        pub stake: Balance,
+
+        /// [Adding] Pool share received for this NFT
+        pub share: Balance,
```
dzmitry-lahoda, line 496: not sure, for share we either will just mult each time or use assets. why would we store it? we always can store later if that is optimal. but if we store, we need to track minted assets and this share to be 1 to 1? what you was going to avoid?

rfcs/0005-pablo-distribution.md
```diff
@@ -0,0 +1,667 @@
...
+
+        /// [Adding] reduced rewards by asset for the fNFT (d_n)
+        pub reductions: BtreeMap<AssetId, Balance>,
+
+        /// [Removing] The reward epoch at which this NFT will start yielding rewards. This is no longer needed.
+        /// pub reward_epoch_start: Epoch,
+
+        /// [Removing] List of reward asset/pending rewards. This is no longer needed as pending rewards are pooled
```
dzmitry-lahoda, line 504: already removed. not sure why merge something which is not here

rfcs/0005-pablo-distribution.md
```diff
@@ -0,0 +1,667 @@
...
+        /// Total rewards including inflation for adjusting for new stakers joining the pool. All stakers
+        /// in a pool are eligible to receive a part of this value based on their share of the pool.
+        pub total_rewards: Balance,
+
+        /// A book keeping field to track the actual total reward without the reward inflation caused
+        /// by new stakers joining the pool.
+        pub total_inflation: Balance,
+
+        /// Upper bound on the `actual_total_rewards`.
+        pub max_rewards: Balance,
```
dzmitry-lahoda, line 537: i do not agree with this implementaion. max_rewards can be defined in terms of rate and time of rate active.

rfcs/0005-pablo-distribution.md
```diff
@@ -0,0 +1,667 @@
...
+
+        /// The rewarding rate that increases the pool `total_reward` (and `actual_total_reward`)
```
dzmitry-lahoda, line 539: > `total_reward` (and `actual_total_reward`)

are these should end with s?

rfcs/0005-pablo-distribution.md
vimukthi-git: Yup, we can later modify based on any new naming, as long as the meaning is obvious this is ok

rfcs/0005-pablo-distribution.md
vimukthi-git: Not sure what you mean, can you provide an example?

rfcs/0005-pablo-distribution.md
vimukthi-git: You changed the existing code according to new design we agreed, this document is the design that proposes changes to previous version, so we keep it as it is

rfcs/0005-pablo-distribution.md
```diff
@@ -0,0 +1,667 @@
...
+        pub asset: AssetId,
+        /// The original stake this NFT was minted for.
+        /// Used for calculating the pool share.
+        pub stake: Balance,
+
+        /// [Adding] Pool share received for this NFT
+        pub share: Balance,
+
+        /// [Adding] reduced rewards by asset for the fNFT (d_n)
+        pub reductions: BtreeMap<AssetId, Balance>,
```
dzmitry-lahoda, line 499: just clarify, when `reductions` field updated? what operation/extrinsic/trait is called to updated this field?

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -0,0 +1,632 @@
...
+
+stem:[x_n = (P_(current) + delta P) . s_n / (sum s_(n+1))]
+
+stem:[x_n = P_(current) . (1 + s_(n+1) / (sum s_n)) . s_n / (sum s_(n+1))]
+
+stem:[x _n = P_(current) . s_n / (sum s_n)]
+
+**Therefore, the existing staker receives the same reward as before **
+
+To compensate for this new adjustment, a reduction stem:[d_n] of reward for each staker needs to be tracked,
```
dzmitry-lahoda, line 320: is d_n == delta_p? :)

rfcs/0005-pablo-distribution.md
dzmitry-lahoda: so why we call this inflation?

dzmitry-lahoda (approved):
i think that formula for fees is broken and not usable (while code done is) and extend position will not work.

rfcs/0004-pablo-distribution.adoc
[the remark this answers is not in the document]
vimukthi-git: These are configurable params per pool so not relevant to this design

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
vimukthi-git: This is correct

rfcs/0005-pablo-distribution.md
vimukthi-git: reductions are updated when `stake` is called, check the algorithms section

rfcs/0005-pablo-distribution.md
vimukthi-git: Just math notation

rfcs/0005-pablo-distribution.md
vimukthi-git: I will clarify, we need to avoid boosting for extend position operations.

rfcs/0005-pablo-distribution.md
vimukthi-git: corrected

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
vimukthi-git: corrected

KaiserKarel (approved):

andor0 (approved):

PR #1015, opened 2022-05-17
Add test for Halborn audit issue HAL-03 

vivekvpandya:
IMPROPER LENDING MARKET CONFIGURATION LEADS TO MULTIPLE VULNERABILITIES

## Issue
https://app.clickup.com/t/2953czv

## Description
Just adding a simple test showing the issue. 

## Checklist

### DIFF
--- a/frame/lending/src/tests.rs
+++ b/frame/lending/src/tests.rs
@@ -512,7 +512,7 @@ fn old_price() {
 		const BORROW_AMOUNT: u128 = 30;
 		const SECOND_PRICE: u128 = 3;
 
-		let (market, vault) = create_market::<FIRST_PRICE, Runtime>(
+		let (market, vault) = create_market::<Runtime, FIRST_PRICE>(
 			USDT::instance(),
 			BTC::instance(),
 			*ALICE,
@@ -1129,12 +1129,12 @@ fn test_repay_total_debt() {
 #[test]
 fn liquidation() {
 	new_test_ext().execute_with(|| {
-		let (market_id, vault) = create_market::<50_000, Runtime>(
+		let (market_id, vault) = create_market::<Runtime, 50_000>(
 			USDT::instance(),
 			BTC::instance(),
 			*ALICE,
 			DEFAULT_MARKET_VAULT_RESERVE,
-			MoreThanOneFixedU128::saturating_from_rational(2, 1),
+			MoreThanOneFixedU128::saturating_from_rational(2_u128, 1_u128),
 		);
 
 		let collateral = BTC::units(100);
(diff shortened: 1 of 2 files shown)


frame/lending/src/tests.rs
```diff
@@ -1181,6 +1181,62 @@ fn zero_amount_collateral_deposit() {
 	})
 }
 
+#[test]
+fn halborn_issue_market_owner_profits_by_lquidate() {
```
dzmitry-lahoda, line 1185: `lquidate` typo

dzmitry-lahoda, line 1185: may you add harbor_0003_... id here?

frame/lending/src/tests.rs
```diff
@@ -1181,6 +1181,62 @@ fn zero_amount_collateral_deposit() {
...
+		let collateral_factor = MoreThanOneFixedU128::saturating_from_rational(2000_u128, 99_u128);
+		let updatable = UpdateInput {
+			collateral_factor,
+			under_collateralized_warn_percent: Percent::from_float(1.1),
+			liquidators: vec![],
+			interest_rate_model: InterestRateModel::default(),
+		};
+		let updatable = updatable.try_into_validated::<UpdateInputVaild>().unwrap();
+		assert_ok!(Lending::update_market(Origin::signed(*ALICE), market_id, updatable));
+		// ALICE loan must be liquidated now.
```
vimukthi-git, line 1235: Just to understand, Why would ALICE `update_market`/collateral factor to get their own loan liquidated?

vivekvpandya: Sorry for confusion, ALICE is creator of the market so she can update the configs. This is just to show problematic case.

vivekvpandya: Updated test to avoid confusion.

vimukthi-git (approved):
Seems like a critical issue to fix no?

vivekvpandya:
> Seems like a critical issue to fix no?

So its bit complicated issue and most likely we need some sort of governance mechanism , so perhaps will not be done in current sprint.

vimukthi-git:
@vivekvpandya Then may be we should create a task already and prioritize no? Or is there any other process to handle this?

hussein-aitlahcen:
Perhaps worth thinking about removing the ability to change the collateral factor for open markets and only allow the users to do so via closing + creating a new one?

vivekvpandya:
This is ticket for https://app.clickup.com/t/2953d0u . Can we close this PR as it just intends to add test?

dzmitry-lahoda (approved):

dzmitry-lahoda (approved):

vimukthi-git (approved):

vimukthi-git:
Approved on the understanding that another team/dev is handling the fix for this.

PR #1016, opened 2022-05-17
Add test for dex-router which shows a cycle in the route.

vivekvpandya:
## Issue
https://app.clickup.com/t/2953czv

## Description
Just adding test to show issue.

## Checklist

### DIFF
--- a/frame/composable-traits/src/defi.rs
+++ b/frame/composable-traits/src/defi.rs
@@ -96,6 +96,22 @@ impl<AssetId: PartialEq> PartialEq for CurrencyPair<AssetId> {
 
 impl<AssetId: PartialEq> Eq for CurrencyPair<AssetId> {}
 
+impl<AssetId: Ord> Ord for CurrencyPair<AssetId> {
+	fn cmp(&self, other: &Self) -> core::cmp::Ordering {
+		let base_ordering = self.base.cmp(&other.base);
+		if base_ordering.is_eq() {
+			return self.quote.cmp(&other.quote)
+		}
+		base_ordering
+	}
+}
+
+impl<AssetId: Ord> PartialOrd for CurrencyPair<AssetId> {
+	fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
+		Some(self.cmp(other))
+	}
+}
+
 impl<AssetId: Copy + PartialEq> CurrencyPair<AssetId> {
 	pub fn swap(&self) -> Self {
 		Self { base: self.quote, quote: self.base }
--- a/frame/composable-traits/src/defi.rs
+++ b/frame/composable-traits/src/defi.rs
@@ -164,6 +164,31 @@ impl<AssetId: PartialEq> AsRef<[AssetId]> for CurrencyPair<AssetId> {
 	}
 }
 
+#[cfg(test)]
+mod test_currency_pair {
+	use super::*;
+	#[test]
+	fn test_cmp_for_currency_pair() {
+		let pair1: CurrencyPair<u8> = CurrencyPair::new(0_u8, 1_u8);
+		let pair2: CurrencyPair<u8> = CurrencyPair::new(1_u8, 2_u8);
+		let pair3: CurrencyPair<u8> = CurrencyPair::new(0_u8, 1_u8);
+		let pair4: CurrencyPair<u8> = CurrencyPair::new(1_u8, 0_u8);
+		assert_eq!(pair1.cmp(&pair2), core::cmp::Ordering::Less);
+		assert_eq!(pair1.cmp(&pair4), core::cmp::Ordering::Less);
+		assert_eq!(pair1.cmp(&pair3), core::cmp::Ordering::Equal);
+		assert_eq!(pair2.cmp(&pair1), core::cmp::Ordering::Greater);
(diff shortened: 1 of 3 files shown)


frame/dex-router/src/tests.rs
```diff
@@ -148,6 +161,40 @@ fn get_route_tests() {
 	});
 }
 
+#[test]
+fn halborn_route_with_cycle() {
```
dzmitry-lahoda, line 165: may you add harbor number id here?

frame/dex-router/src/tests.rs
```diff
@@ -148,6 +161,40 @@ fn get_route_tests() {
...
 }
 
+#[test]
+fn halborn_hal11_route_with_cycle() {
+	new_test_ext().execute_with(|| {
+		let currency_pair = CurrencyPair { base: USDT, quote: USDC };
+		assert_eq!(DexRouter::get_route(currency_pair), None);
+
+		let dex_route =
+			vec![create_usdt_usdc_pool(), create_usdc_usdt_pool(), create_usdt_usdc_pool()];
```
vimukthi-git, line 171: So this problem only happens if someone updates a cyclic route? Shouldn't we validate at the time of update then?

vimukthi-git (approved):

frame/dex-router/src/tests.rs
vivekvpandya: so if want validation for this case then we can improve, or we just trust that users won't do such routes...

vimukthi-git: I see, Is it hard to validate?

vivekvpandya: no, I think we can use https://docs.substrate.io/rustdocs/latest/sp_std/collections/btree_set/struct.BTreeSet.html to detect duplicates. 
By tracking pool_id in set we can eliminate second case in the test. And by tracking (base,quote) pair in set we can eliminate both cases.

vimukthi-git: I see, seems the second case would be eliminated with tracking base/quote in a set anyway.

vivekvpandya: Yes that seem good way.

vimukthi-git (dismissed):
LGTM :+1:

frame/dex-router/src/lib.rs
```diff
@@ -118,6 +118,8 @@ pub mod pallet {
 		CanNotRespectMinAmountRequested,
 		/// Unsupported operation.
 		UnsupportedOperation,
+		/// Route with possible loop is not allowed.
+		LoopSuspectedInRouteWhileValidation,
```
vimukthi-git, line 122: Nit: 
```suggestion
		LoopSuspectedInRouteUpdate,
```

frame/dex-router/src/lib.rs
```diff
@@ -285,6 +288,11 @@ pub mod pallet {
 							sp_std::if_std! {
 								println!("pool_id {:?} pair.quote {:?} pair.base {:?} val {:?}", *iter, pair.quote, pair.base, val);
 							}
+							if pair_set.contains(&pair) {
+								return Err(Error::<T>::LoopSuspectedInRouteWhileValidation.into())
+							} else {
+								pair_set.insert(pair);
```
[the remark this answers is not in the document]
vimukthi-git: I see that `insert` returns false for existing element, perhaps we can use that https://doc.rust-lang.org/nightly/src/alloc/collections/btree/set.rs.html#788-790

frame/composable-traits/src/defi.rs
```diff
@@ -96,6 +96,22 @@ impl<AssetId: PartialEq> PartialEq for CurrencyPair<AssetId> {
 
 impl<AssetId: PartialEq> Eq for CurrencyPair<AssetId> {}
 
+impl<AssetId: Ord> Ord for CurrencyPair<AssetId> {
+	fn cmp(&self, other: &Self) -> core::cmp::Ordering {
```
vimukthi-git, line 100: Needs a test

frame/composable-traits/src/defi.rs
```diff
@@ -96,6 +96,22 @@ impl<AssetId: PartialEq> PartialEq for CurrencyPair<AssetId> {
...
+		let base_ordering = self.base.cmp(&other.base);
+		if base_ordering.is_eq() {
```
vimukthi-git, line 102: Seems this function would not be avoiding duplicates as it does not return "equal" for two pairs that are the swapped version of each other.  Because dex router supports reverse route iteration, we should consider swapped versions as equal no?

vivekvpandya: So currently we don't store pairs as part of route, we just store pool_id and based on pair passed in exchange()/sell()/buy() operation we use all nodes in normal direction or reverse direction. So it is not possible to have route with say A-B and then same pool in reverse direction (B-A) at some point in route. So we don't need to worry about swapped version here.

To clarify more if we have a route BTC-ETH, ETH-USDT then I can exchange with two currency pair, either USDT-BTC (in that case it uses normal direction while performing any extrinsic on the route and if we do BTC-USDT then all operations on pool will be executed in reverse direction.

vimukthi-git (approved):
LGTM

hussein-aitlahcen (approved):

PR #1021, opened 2022-05-18
Refactor node to allow for different RPCs in different runtimes

benluelo:
## Issue

https://app.clickup.com/t/1y23x2d

## Description

Very much a work in progress; likely to change significantly. I wouldn't bother reviewing until this PR is out of draft.

## Checklist


dzmitry-lahoda (dismissed):

node/src/rpc.rs
```diff
@@ -5,60 +5,101 @@
...
+		deps.clone(),
+	);
 
-	io.extend_with(PabloApi::to_delegate(Pablo::new(client)));
+	<FullClient<RuntimeApi, Executor> as ProvideRuntimeApi<OpaqueBlock>>::Api::extend_with_lending_api(
+		&mut io,
+		deps.clone(),
+	);
+
+	// io.extend_with(PabloApi::to_delegate(Pablo::new(client)));
```
dzmitry-lahoda, line 102: pablo will be releases soon, earlier than lending, so may be return it back?

node/src/service.rs
```diff
@@ -262,12 +281,17 @@ async fn start_node_impl<RuntimeApi, Executor>(
 	id: ParaId,
 ) -> sc_service::error::Result<TaskManager>
 where
+	// Block: BlockT,
```
dzmitry-lahoda, line 284: remove commented?

node/src/service.rs
```diff
@@ -354,10 +378,13 @@ where
 				deny_unsafe,
 			};
 
+			// let _: () = client;
 			Ok(rpc::create(deps))
 		})
 	};
 
+	// let _: () = task_manager;
```
dzmitry-lahoda, line 386: comments

dzmitry-lahoda (commented):
i see. i like this solution, similar to what simulator does. please clean up

node/src/rpc.rs
benluelo: I think keeping all the RPC apis implemented this way is ideal - keeps it consistent and allows us to remove/add rpcs as we need to

benluelo: see latest commits for boilerplate impl

seunlanlege (dismissed):

node/src/client.rs
```diff
@@ -21,11 +21,25 @@ use std::sync::Arc;
...
+pub(crate) type ComposableClient = FullClient<composable_runtime::RuntimeApi, ComposableExecutor>;
 #[cfg(feature = "dali")]
-type DaliClient = FullClient<dali_runtime::RuntimeApi, DaliExecutor>;
+pub(crate) type DaliClient = FullClient<dali_runtime::RuntimeApi, DaliExecutor>;
+
+/// A client instance of Picasso.
+#[derive(Clone)]
+pub enum Client {
+	/// Picasso client type
+	Picasso(Arc<PicassoClient>),
```
KaiserKarel, line 34: So Picasso becomes our base?

KaiserKarel: Perhaps it is more logical to have composable be that base, as it should have the least amount of features. Then `picasso` is additive, and `dali` is basically `full`.

runtime/dali/src/lib.rs
```diff
@@ -1163,6 +1164,17 @@ mod benches {
 }
 
 impl_runtime_apis! {
+	impl lending_runtime_api::LendingRuntimeApi<Block, MarketId> for Runtime {
+		fn current_interest_rate(_market_id: MarketId) -> SafeRpcWrapper<Rate> {
+			SafeRpcWrapper(
+				// TODO: Actually implement this
+				Rate::zero()
+				// lending::BorrowIndex::<Runtime>::get(market_id)
+				// 	.unwrap_or_else(Rate::zero)
```
KaiserKarel, line 1173: Let's return Rate::max() here, just for security reasons

KaiserKarel, line 1173: Let's return Rate::max() here, just for security reasons

KaiserKarel (dismissed):

KaiserKarel (approved):

seunlanlege (approved):

PR #1026, opened 2022-05-19
[cu-2f7bbge] Pablo FeeConfig Refactoring

vimukthi-git:
## Issue
https://app.clickup.com/t/2f7bbge

## Description
- Refactoring fee handling to be based on a single config for all pools
- Disbursements of fees are in a single place for all pools now to allow for future changes

refer: https://github.com/ComposableFi/composable/blob/1021dad42dc6515960e53d9b9b6107bfadecffa4/rfcs/0004-pablo-distribution.md#521-feeconfig

## Checklist


ComposableFiB0t:
Task linked: [CU-2f7bbge 5.2.1 Implement FeeConfig on pallet-pablo across all 3 types of pools.](https://app.clickup.com/t/2f7bbge)

frame/pablo/src/lib.rs
```diff
@@ -650,6 +629,19 @@ pub mod pallet {
 			}
 			Ok(())
 		}
+
+		#[transactional]
+		fn disburse_fees(
```
vimukthi-git, line 634: All fee disbursements for all pools are here now

frame/composable-traits/src/dex.rs
```diff
@@ -101,6 +105,88 @@ pub trait Amm {
...
+			protocol_fee,
+			asset_id,
+		}
+	}
+}
+
+impl Mul<Permill> for FeeConfig {
+	type Output = Self;
+
+	fn mul(self, rhs: Permill) -> Self::Output {
```
vimukthi-git, line 181: Needs a test

frame/composable-traits/src/dex.rs
```diff
@@ -101,6 +105,88 @@ pub trait Amm {
...
+)]
+pub struct Fee<AssetId, Balance> {
+	// total fee
+	pub fee: Balance,
+	/// Amount of the fee pool charges for the exchange, this goes to liquidity providers.
+	pub lp_fee: Balance,
+	/// Amount of the fee that goes out to the owner of the pool
+	pub owner_fee: Balance,
+	/// Amount of the protocol fees(for PBLO holders) out of owner_fees.
+	pub protocol_fee: Balance,
```
hussein-aitlahcen, line 120: If I understand correctly, owner fee will  be 20% and protocol fee will be 100% here?

hussein-aitlahcen (approved):

frame/composable-traits/src/dex.rs
vimukthi-git: Yes!

vivekvpandya: So just to confirm, in above setting 20% of lp_fee will be gone to owner_fee and 100% of that goes to PBLO holders right?

frame/composable-traits/src/dex.rs
```diff
@@ -11,6 +14,7 @@ use sp_runtime::{
 	ArithmeticError, DispatchError, Permill,
 };
 use sp_std::vec::Vec;
+use std::ops::Mul;
```
vivekvpandya, line 17: better use `sp_std` ?

frame/composable-traits/src/dex.rs
vivekvpandya: IIUC this is required due to add_liquidity() from StableSwap ?
I think if we just change fee_rate then that is enough right? I mean for purpose of StableSwap that is enough.
In that case better not have this mul() implemenation rater just fee_rate locally in add_liquidity()

frame/composable-traits/src/dex.rs
vimukthi-git: Stupid IDE haha

frame/composable-traits/src/dex.rs
vimukthi-git: You are correct. To clarify and just to define terms, we can call the total fee = trading fee.
Then out of trading fee 80% is `lp_fee`, 20% is separated as `owner_fee`. At the beginning 100% of the `owner_fee` is allocated as `protocol_fee` which goes to PBLO holders.

frame/composable-traits/src/dex.rs
vimukthi-git: > IIUC this is required due to add_liquidity() from StableSwap ?

yeah.

> I think if we just change fee_rate then that is enough right?

Yeah, there is a mistake in the code here. But I wonder why there are no tests that fail for this stuff. I will add. About the `Mul` implementation, it's just to make the code cleaner. The idea is to allow increasing or decreasing the fee rate this way.

frame/pablo/src/lib.rs
```diff
@@ -650,6 +629,19 @@ pub mod pallet {
...
+			who: &T::AccountId,
+			owner: &T::AccountId,
+			fees: &Fee<T::AssetId, T::Balance>,
+		) -> Result<(), DispatchError> {
+			if !fees.owner_fee.is_zero() {
+				T::Assets::transfer(fees.asset_id, who, owner, fees.owner_fee, false)?;
+			}
+			// TODO other fees handling
```
vivekvpandya, line 642: Do we need this TODO?

frame/pablo/src/lib.rs
```diff
@@ -717,15 +709,22 @@ pub mod pallet {
...
+							who,
+							info.clone(),
+							pool_account.clone(),
+							base_amount,
+							quote_amount,
+							min_mint_amount,
+							keep_alive,
+						)?;
+					// imbalance fees
+					Self::disburse_fees(&pool_account, &info.owner, &base_fee)?;
```
vivekvpandya, line 724: Suggestion: using `base_owner_fee` name or similar makes it easy to read that only owner fee is required to transfer, and other portion of the fee is already deposited into pool.

vivekvpandya (approved):

frame/pablo/src/lib.rs
vimukthi-git: Yes, those are the next tasks

frame/pablo/src/lib.rs
vimukthi-git: May be you can clarify what you mean here? no fees are transferred before this step as of now.

vivekvpandya: Oh I see. Makes sense now.

hussein-aitlahcen (approved):

vivekvpandya (approved):

PR #1038, opened 2022-05-21
CU-2bbgg1k - Configurable constant product

hussein-aitlahcen:
## Issue
[Clickup issue](https://app.clickup.com/t/2bbgg1k)

## Description
Allow users to configure constant product pool weights. Prior to this PR, constant product pool were hardcoded with 50/50 weights. This PR allow arbitrary weights for base/quote assets. With the following property:
$$\sum_{i=1}^n{w_i}=1$$

As we have $n=2$, the implementation is expecting $w_1$ with $w_1 != 1$ to be provided and compute the according $w_2=1-w_1$.

Note: just wanted to try mathjax

## Checklist

### DIFF
--- a/frame/composable-traits/src/dex.rs
+++ b/frame/composable-traits/src/dex.rs
@@ -233,6 +233,10 @@ pub struct ConstantProductPoolInfo<AccountId, AssetId> {
 	pub lp_token: AssetId,
 	/// Amount of the fee pool charges for the exchange
 	pub fee_config: FeeConfig,
+	/// The weight of the base asset. Must hold `1 = base_weight + quote_weight`
+	pub base_weight: Permill,
+	/// The weight of the quote asset. Must hold `1 = base_weight + quote_weight`
+	pub quote_weight: Permill,
 }
 
 #[derive(Copy, Clone, PartialEq, Eq)]
--- a/frame/dex-router/src/benchmarking.rs
+++ b/frame/dex-router/src/benchmarking.rs
@@ -72,6 +72,7 @@ where
 		owner: owner.clone(),
 		pair: CurrencyPair::new(pica, ksm),
 		fee_config: FeeConfig::zero(),
+		base_weight: Permill::from_percent(50),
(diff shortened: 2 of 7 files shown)


ComposableFiB0t:
Task linked: [CU-2bbgg1k Create Pool Module - constant weights addition](https://app.clickup.com/t/2bbgg1k)

vivekvpandya (approved):

KaiserKarel:
@hussein-aitlahcen needs rebase

dzmitry-lahoda (approved):

frame/composable-traits/src/dex.rs
```diff
@@ -152,6 +152,10 @@ pub struct ConstantProductPoolInfo<AccountId, AssetId> {
 	pub fee: Permill,
 	/// Amount of the fee goes to owner of the pool
 	pub owner_fee: Permill,
+	/// The weight of the base asset. Must hold `1 = base_weight + quote_weight`
+	pub base_weight: Permill,
```
vimukthi-git, line 156: Wouldn't it be more forward compatible if we define this as an array of weights? Given that we want to support multi asset eventually? Obviously we can reject the case with more than 2 for now

vimukthi-git (approved):
LGTM, one minor question though

frame/composable-traits/src/dex.rs
hussein-aitlahcen: If we do so we should do it for add/remove liquidity as well and integrate a vec everywhere

vimukthi-git (approved):
LGTM

frame/composable-traits/src/dex.rs
KaiserKarel: Note that the refactor will make a storage migration more difficult; I'd already refactor this to a Vec<Permill>, and check that len == 2; that way we can remove the check to allow multiweighted pools

KaiserKarel (approved):

PR #1042, opened 2022-05-23
fix(Runtime Integration Tests): Fix Mosaic test occasionally failing

Dom-Roth:
## Issue
The last Mosaic test was failing in about 1 of 10 tries.

## Description
The error was a `TxStillLocked` message, the easy fix here is to retry the transaction in this case.
Might not be the best way of solving this, though this is currently blocking CI implementation of the Integration Tests, so we wanted a fix asap.

This should fix it, though since the issue was appearing at random, and there was no way to reliably reproduce it, I ran it 3 times and it didn't fail. So I hope this band-aid keeps it together.


kollegian (approved):

Juliansv (approved):
LGTM

PR #1049, opened 2022-05-23
Update bash script to check for changes

haroldsphinx:
- Create self manage github actions to check for changes in files and folders
- Create bash script to check and output the file that changed
- Export the script as github actions and created a workflow for it


KaiserKarel:
Just going to close this. Spoke offline and this is not a good idea.

PR #1051, opened 2022-05-23
extend fnft staking with amount

dzmitry-lahoda:
## Issue
https://app.clickup.com/t/2xw8zx1

## Description
extends staking with amount, see README.ME, added tests and ticket to understand requirements

## Checklist

### DIFF
--- a/frame/composable-traits/src/staking_rewards.rs
+++ b/frame/composable-traits/src/staking_rewards.rs
@@ -189,15 +189,6 @@ pub trait Staking {
 	/// * `to` the account to transfer the final claimed rewards to.
 	fn unstake(instance_id: &Self::InstanceId, to: &Self::AccountId) -> DispatchResult;
 
-	/// Unstake an actual staked position, represented by a NFT.
-	///
-	/// Arguments
-	///
-	/// * `instance_id` the ID uniquely identifiying the NFT from which we will compute the
-	///   available rewards.
-	/// * `to` the account to transfer the final claimed rewards to.
-	fn unstake(instance_id: &Self::InstanceId, to: &Self::AccountId) -> DispatchResult;
-
 	/// Claim the current rewards.
 	///
 	/// Arguments
--- a/frame/composable-traits/src/staking_rewards.rs
+++ b/frame/composable-traits/src/staking_rewards.rs
@@ -104,7 +104,7 @@ pub struct StakingConfig<AccountId, DurationPresets, RewardAssets> {
 pub struct StakingNFT<AccountId, AssetId, Balance, Epoch, Rewards> {
 	/// The staked asset.
 	pub asset: AssetId,
-	/// The original stake this NFT was minted for.
+	/// The original stake this NFT was minted for or updated NFT with increased stake amount.
 	pub stake: Balance,
 	/// The reward epoch at which this NFT will start yielding rewards.
 	pub reward_epoch_start: Epoch,
--- a/frame/composable-traits/src/staking_rewards.rs
+++ b/frame/composable-traits/src/staking_rewards.rs
@@ -189,6 +189,15 @@ pub trait Staking {
 	/// * `to` the account to transfer the final claimed rewards to.
 	fn unstake(instance_id: &Self::InstanceId, to: &Self::AccountId) -> DispatchResult;
 
+	/// Unstake an actual staked position, represented by a NFT.
+	///
+	/// Arguments
+	///
+	/// * `instance_id` the ID uniquely identifiying the NFT from which we will compute the
+	///   available rewards.
+	/// * `to` the account to transfer the final claimed rewards to.
+	fn unstake(instance_id: &Self::InstanceId, to: &Self::AccountId) -> DispatchResult;
+
 	/// Claim the current rewards.
 	///
 	/// Arguments
@@ -197,7 +206,6 @@ pub trait Staking {
 	/// * `instance_id` the ID uniquely identifiying the NFT from which we will compute the
 	///   available rewards.
 	/// * `to` the account to transfer the rewards to.
-	/// * `strategy` the strategy used to claim the rewards.
 	fn claim(instance_id: &Self::InstanceId, to: &Self::AccountId) -> DispatchResult;
 }
 
--- a/frame/composable-traits/src/staking_rewards.rs
+++ b/frame/composable-traits/src/staking_rewards.rs
@@ -110,7 +110,7 @@ pub struct StakingNFT<AccountId, AssetId, Balance, Epoch, Rewards> {
 	pub reward_epoch_start: Epoch,
 	/// List of reward asset/pending rewards.
 	pub pending_rewards: Rewards,
-	/// The date at which this NFT was minted.
+	/// The date at which this NFT was minted or to wich lock was extended too.
 	pub lock_date: Timestamp,
 	/// The duration for which this NFT stake was locked.
 	pub lock_duration: DurationSeconds,
--- a/frame/staking-rewards/src/lib.rs
+++ b/frame/staking-rewards/src/lib.rs
@@ -75,8 +75,7 @@ pub mod pallet {
 		ArithmeticError, Perbill, SaturatedConversion,
 	};
 	use sp_std::collections::btree_map::BTreeMap;
-
-//use crate::math::honest_stake_increase;
+	use crate::math::honest_locked_stake_increase;
 
 	pub(crate) type EpochId = u128;
 	pub(crate) type AccountIdOf<T> = <T as frame_system::Config>::AccountId;
@@ -137,7 +136,8 @@ pub mod pallet {
 		EpochNotFound,
 		PalletIsBusy,
 		ImpossibleState,
-		OnlyOwnerOfPositionCanDoThis
+		OnlyOwnerOfPositionCanDoThis,
+		CannotIncreaseStakedAmountBecauseOfLimitedArithmetic,
 	}
 
 	#[pallet::config]
@@ -151,7 +151,7 @@ pub mod pallet {
 		/// The ID that uniquely identify an asset.
 		type AssetId: AssetId + Ord;
 
-		type Balance: Balance + TryFrom<u128>;
+		type Balance: Balance + TryFrom<u128> + Into<u128>;
 
 		/// The underlying currency system.
 		type Assets: FungiblesInspect<
@@ -479,15 +479,17 @@ pub mod pallet {
 													)
 													.map_err(|_| ArithmeticError::Overflow)?;												
 												if let Some(amount) = PendingAmountExtensions::<T>::take(&nft_id) {
-													// nft.lock_date 
-													// += 
-													// honest_stake_increase(
-													// 	nft.early_unstake_penalty.value, 
-													// 	nft.stake, 
-													// 	amount, nft.lock_duration,
-													// 	 now - nft.lock_date)
-													// 	 .try_into();
-													nft.stake = nft.stake.safe_add(&amount)?;
+													let time_lock = honest_locked_stake_increase(
+														nft.early_unstake_penalty.value, 
+														nft.stake.into(), 
+														amount.into(), 
+														nft.lock_duration.into(),
+														 (now - nft.lock_date).into())
+														 .ok_or(Error::<T>::CannotIncreaseStakedAmountBecauseOfLimitedArithmetic)?;
+
+													nft.lock_date = nft.lock_date.safe_add(&time_lock)?;
+													
+														 nft.stake = nft.stake.safe_add(&amount)?;
 												}
 											}
 										},
--- a/frame/staking-rewards/src/lib.rs
+++ b/frame/staking-rewards/src/lib.rs
@@ -361,7 +361,7 @@ pub mod pallet {
 		}
 
 		/// Splits fNFT position into several chunks with various amounts, but with same exposure.
-		/// fNFT splitted earns reward in current epoch proportial to split.
+		/// fNFT splitted earns reward in current epoch proportional to split.
 		/// Can split only at  `State::WaitingForEpochEnd` state.
 		///
 		/// `origin` - owner of fNFT
@@ -427,6 +427,8 @@ pub mod pallet {
 											// TODO: https://app.clickup.com/t/2xw5fca
 										},
 										PositionState::LockedRewarding => {
+											// TODO: return here increased share if one of assets is
+											// same as staked
 											let shares = nft.shares();
 											for (reward_asset, pending_reward) in
 												nft.pending_rewards.clone().into_iter()
@@ -443,6 +445,8 @@ pub mod pallet {
 													reward,
 													total_shares,
 												)?;
+												// TODO: if adding asset which is staked, increase
+												// total
 												nft.pending_rewards
 													.try_insert(
 														reward_asset,
--- a/frame/staking-rewards/src/lib.rs
+++ b/frame/staking-rewards/src/lib.rs
@@ -49,7 +49,7 @@ pub mod pallet {
 		math::safe::{safe_multiply_by_rational, SafeAdd, SafeSub},
 	};
 	use composable_traits::{
-		financial_nft::{FinancialNftProtocol, NftClass, NftVersion},
+		financial_nft::{FinancialNftProtocol, NftClass, NftVersion, FinancialNftProvider},
 		staking_rewards::{
 			Penalty, PenaltyOutcome, PositionState, Staking, StakingConfig, StakingNFT,
 			StakingReward,
@@ -76,7 +76,7 @@ pub mod pallet {
 	};
 	use sp_std::collections::btree_map::BTreeMap;
 
-use crate::math::honest_stake_increase;
+//use crate::math::honest_stake_increase;
 
 	pub(crate) type EpochId = u128;
 	pub(crate) type AccountIdOf<T> = <T as frame_system::Config>::AccountId;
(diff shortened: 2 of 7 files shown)


frame/staking-rewards/src/math.rs
```diff
@@ -0,0 +1,67 @@
...
+type Balance = u128;
+
+// keep in sync with python math
+pub fn honest_locked_stake_increase(
+	after_penalized: Perbill,
+	original_amount: Balance,
+	new_amount: Balance,
+	duration: DurationSeconds,
+	passed: DurationSeconds,
+) -> Option<DurationSeconds> {
```
dzmitry-lahoda, line 15: example of working with precision

frame/staking-rewards/src/lib.rs
```diff
@@ -425,6 +447,7 @@ pub mod pallet {
 										PositionState::Pending => {},
 										PositionState::Expired => {
 											// TODO: https://app.clickup.com/t/2xw5fca
+											log::into!("Expired: {:?}", &nft);
```
saruman9, line 450:
```suggestion
											log::info!("Expired: {:?}", &nft);
```

frame/staking-rewards/src/lib.rs
```diff
@@ -453,6 +481,20 @@ pub mod pallet {
...
+														nft.early_unstake_penalty.value,
+														nft.stake.into(),
+														amount.into(),
+														nft.lock_duration.into(),
+														 (now - nft.lock_date).into())
+														 .ok_or(Error::<T>::CannotIncreaseStakedAmountBecauseOfLimitedArithmetic)?;
+													nft.lock_date =
+														nft.lock_date.safe_add(&time_lock)?;
+													nft.stake = nft.stake.safe_add(&amount)?;
+												}
```
hussein-aitlahcen, line 497: I think you need to do that on the registering phase as you are increasing the share of the user while rewarding the other users, i.e. you dillute the others

frame/staking-rewards/src/lib.rs
```diff
@@ -370,21 +380,32 @@ pub mod pallet {
...
 			origin: OriginFor<T>,
 			instance_id: InstanceIdOf<T>,
 			balance: T::Balance,
 		) -> DispatchResult {
-			Err(DispatchError::Other("no implemented. TODO: insert update for next fold"))
+			let owner = ensure_signed(origin)?;
+			let position = T::get_protocol_nft::<StakingNFTOf<T>>(&instance_id)?;
+			T::ensure_protocol_nft_owner::<StakingNFTOf<T>>(&owner, &instance_id)?;
+			let protocol_account = Self::account_id(&position.asset);
+			T::Assets::transfer(position.asset, &owner, &protocol_account, balance, false)?;
```
hussein-aitlahcen, line 402: Good to add a `keep_alive` parameter to the extrinsic :)

frame/staking-rewards/src/lib.rs
hussein-aitlahcen: Also not sure why we do a complex computation, we should just refresh the lock, i.e. `lock_date = date(call(extend_stake))` ?

dzmitry-lahoda: > I think you need to do that on the registering phase as you are increasing the share of the user while rewarding the other users, i.e. you dillute the others

huh, exactly what I am fixing. i do not increase total stake. I will increase pending shared stake merged on registering. right now it does not dillutes for sure. but problem is that it does not increase. if fixing increase.

dzmitry-lahoda: > Also not sure why we do a complex computation, we should just refresh the lock, i.e. lock_date = date(call(extend_stake)) ?

if you extend statke without using time of lock as value you incentivize creating new nft because old one captures some value lost during extensions. 

i.e. setting date of nft the date is extended will not be used by bots. as it will loose value.

frame/staking-rewards/src/lib.rs
dzmitry-lahoda: ok.

frame/staking-rewards/src/lib.rs
dzmitry-lahoda: so if formula not complex, than it is same as not having extension at all technically as people better create more accounts and fnfts than so the value is rolled.

ComposableFiB0t:
Task linked: [CU-2xw8zx1 Extend the NFT staked amount (without burning)](https://app.clickup.com/t/2xw8zx1)

ComposableFiB0t:
Task linked: [CU-27qz1eh Extend position by refreshing the time lock](https://app.clickup.com/t/27qz1eh)

dzmitry-lahoda:
r+

vivekvpandya (approved):

frame/composable-support/src/abstractions/block_fold.rs
```diff
@@ -133,3 +133,39 @@ impl<S, K> BlockFold<S, K> {
...
+
+#[cfg(all(test, feature = "std"))]
+mod tests {
+	use frame_support::{Identity, StorageMap, StoragePrefixedMap};
+	use sp_io::TestExternalities;
+
+	#[test]
+	fn proves_partial_drain_possible() {
+		/// based on tests from frame_support, but there is no such test to show off partial drain
+		/// and docs does not tell that drain happens iff you iterate element, not just by calling
```
andor0, line 145:
```suggestion
		/// and docs do not tell that drain happens if you iterate element, not just by calling
```

frame/staking-rewards/src/lib.rs
```diff
@@ -95,11 +99,26 @@ pub mod pallet {
...
 	pub enum State {
-		WaitingForEpochEnd,
-		Rewarding,
-		Registering,
+		/// epoch runs, cannot modify active positions, only append modification queues
+		Running,
+		/// stakes positins rewarded`
+		/// Processed by `fold` as updates are done in place
+		Distributing,
+		// TODO: decide what is better first- duration or amount
```
andor0, line 114:
```suggestion
		// TODO: decide what is better the first duration or
```

frame/staking-rewards/src/lib.rs
```diff
@@ -95,11 +99,26 @@ pub mod pallet {
...
+		/// amount is added regardless of position state
+		PendingAmounts,
+		/// time extended  regardless of position state
```
andor0, line 117:
```suggestion
		/// time extended regardless of position state
```

frame/staking-rewards/src/lib.rs
```diff
@@ -95,11 +99,26 @@ pub mod pallet {
...
+		PendingDurations,
+		/// working with pending operations, registerin new stakers and asset total increases
```
andor0, line 119:
```suggestion
		/// working with pending operations, registering new stakers and asset total increases
```

frame/staking-rewards/src/lib.rs
```diff
@@ -356,7 +392,13 @@ pub mod pallet {
...
 			let nft_owner = T::get_protocol_nft_owner::<StakingNFTOf<T>>(&instance_id)?;
 			let to = if nft_owner == who { to } else { nft_owner };
-			<Self as Staking>::claim(&instance_id, &to)?;
+			let (asset, compound) = <Self as Staking>::claim(&instance_id, &to)?;
+			if compound > T::Balance::zero() {
+				RunningTotalShares::<T>::try_mutate(asset, |total_shares| -> DispatchResult {
+					*total_shares = total_shares.safe_sub(&compound.into())?;
+					Ok(())
+				})?;
+			}
```
andor0, line 401: looks like, there should be:
* a error if `compound == T::Balance::zero() `
* a event before `Ok(())`

frame/staking-rewards/src/lib.rs
```diff
@@ -378,13 +420,28 @@ pub mod pallet {
 		}
 
 		/// Extends fNFT position stake. Applied only to next epoch.
+		///
+		/// Is better then unstake and restake more in case of penalty.
+		/// But not incetivises to making many many small stakes one on other to game over system.
```
andor0, line 425:
```suggestion
		/// But not incentivises to making many small stakes one on other to game over system.
```

frame/composable-traits/src/staking_rewards.rs
```diff
@@ -120,13 +126,15 @@ pub struct StakingNFT<AccountId, AssetId, Balance, Epoch, Rewards> {
 	pub reward_multiplier: Perbill,
 }
 
+/// implemented by instances which know their share of something biggers
```
vimukthi-git, line 129:
```suggestion
/// implemented by instances which know their share of something bigger
```

frame/staking-rewards/src/lib.rs
```diff
@@ -95,11 +99,26 @@ pub mod pallet {
 	#[allow(dead_code)]
 	pub(crate) type PenaltyOf<T> = Penalty<AccountIdOf<T>>;
 
+	// NOTE:
+	// some state processed by batched brain, as these just merged into main collections
+	// some states are processed by fold, because updates happens in place
+	// fold can be made drain if to use A/B storages and drain one and fill other depending on epoch
```
vimukthi-git, line 105: Any reference to what A/B storages means?

frame/staking-rewards/src/lib.rs
```diff
@@ -228,20 +252,21 @@ pub mod pallet {
 	>;
 
 	#[pallet::type_value]
-	pub fn SharesOnEmpty<T: Config>() -> u128 {
-		u128::zero()
+	pub fn SharesOnEmpty<T: Config>() -> ShareAmount {
+		ShareAmount::zero()
 	}
 
+	//// active running total shares
```
vimukthi-git, line 259: Would be nice to have better docs on why this storage exists

frame/staking-rewards/src/lib.rs
```diff
@@ -228,20 +252,21 @@ pub mod pallet {
...
-		(AssetIdOf<T>, AssetIdOf<T>),
-		u128,
-		ValueQuery,
-		SharesOnEmpty<T>,
-	>;
+	#[pallet::getter(fn running_total_shares)]
+	pub type RunningTotalShares<T: Config> =
+		StorageMap<_, Twox64Concat, AssetIdOf<T>, ShareAmount, ValueQuery, SharesOnEmpty<T>>;
+
+	/// pending total shares to be applied on next run
```
vimukthi-git, line 265: Would be nice to have better docs on why this storage exists

frame/staking-rewards/src/lib.rs
```diff
@@ -262,6 +287,18 @@ pub mod pallet {
...
+	/// amount of assets which will be added to position on next epoch
+	#[pallet::storage]
+	#[pallet::getter(fn pending_amount_extensions)]
+	pub type PendingAmountExtensions<T: Config> =
+		StorageMap<_, Twox64Concat, InstanceIdOf<T>, T::Balance, OptionQuery>;
+
+	/// time to extend position with
+	#[pallet::storage]
+	#[pallet::getter(fn pending_duration_extensions)]
+	pub type PendingDurationExtensions<T: Config> =
```
vimukthi-git, line 299: Why do we need to track the pending duration extensions? Is it because the reward multiplier is affected by this?

frame/staking-rewards/src/lib.rs
```diff
@@ -378,13 +423,29 @@ pub mod pallet {
 		}
 
 		/// Extends fNFT position stake. Applied only to next epoch.
+		///
+		/// Is better then unstake and restake more in case of penalty.
+		/// But not incentivises to making many small stakes one on other to game over system.
```
vimukthi-git, line 428:
```suggestion
		/// But this may incentivize making many small stakes one on top each other to game the system.
```
?

frame/staking-rewards/src/lib.rs
```diff
@@ -378,13 +423,29 @@ pub mod pallet {
...
-			_origin: OriginFor<T>,
-			_instance_id: InstanceIdOf<T>,
-			_balance: T::Balance,
+			origin: OriginFor<T>,
+			instance_id: InstanceIdOf<T>,
+			balance: T::Balance,
+			keep_alive: bool,
 		) -> DispatchResult {
-			Err(DispatchError::Other("no implemented. TODO: insert update for next fold"))
+			// TODO: consider attack adding near zero stake each block, preventing stream to finish
```
vimukthi-git, line 437: Shouldn't we set a minimum stake amount to call this then?

frame/staking-rewards/src/lib.rs
```diff
@@ -378,13 +423,29 @@ pub mod pallet {
...
+			let owner = ensure_signed(origin)?;
+			T::ensure_protocol_nft_owner::<StakingNFTOf<T>>(&owner, &instance_id)?;
+			let position = T::get_protocol_nft::<StakingNFTOf<T>>(&instance_id)?;
+			let protocol_account = Self::account_id(&position.asset);
+			T::Assets::transfer(position.asset, &owner, &protocol_account, balance, keep_alive)?;
```
vimukthi-git, line 442: I wonder if it's a good idea to transfer to the protocol account(unless a sub account) as protocol account also stores rewards, there are cases where staked asset is same as rewarded asset

frame/staking-rewards/src/lib.rs
```diff
@@ -457,59 +539,159 @@ pub mod pallet {
...
-						CurrentState::<T>::set(State::WaitingForEpochEnd);
+				State::PendingAmounts => {
+					for (nft_id, amount) in PendingAmountExtensions::<T>::drain()
+						.take(T::ElementToProcessPerBlock::get() as usize)
+					{
+						any = true;
+						let result = T::try_mutate_protocol_nft(
+							&nft_id,
+							|nft: &mut StakingNFTOf<T>| -> DispatchResult {
+								let old_shares = nft.shares();
```
vimukthi-git, line 575: Note: With the new proposed design this code block would not be necessary as new shares are accounted for by the pool by having reward inflation for each rewarded asset without having to wait for the epoch to finish.

frame/staking-rewards/src/lib.rs
```diff
@@ -457,59 +539,159 @@ pub mod pallet {
...
+					}
+				},
+				State::PendingDurations => {
+					for (nft_id, new_lock) in PendingDurationExtensions::<T>::drain()
+						.take(T::ElementToProcessPerBlock::get() as usize)
+					{
+						any = true;
+						let result = T::try_mutate_protocol_nft(
+							&nft_id,
+							|nft: &mut StakingNFTOf<T>| -> DispatchResult {
```
vimukthi-git, line 606: Note: With the new proposed design this code block would not be necessary as new shares are accounted for by the pool by having reward inflation for each rewarded asset without having to wait for the epoch to finish.

frame/staking-rewards/src/lib.rs
dzmitry-lahoda: total stake is depend on this. you should not modify total stake until reward happend.

frame/staking-rewards/src/lib.rs
dzmitry-lahoda: total stake is depend on this. you should not modify total stake until reward happend.

frame/staking-rewards/src/lib.rs
dzmitry-lahoda: total stake is depend on this. you should not modify total stake until reward happend.

will add docs in next PRs

frame/staking-rewards/src/lib.rs
```diff
@@ -354,9 +391,17 @@ pub mod pallet {
 		) -> DispatchResultWithPostInfo {
 			let who = ensure_signed(origin)?;
 			// Only the owner is able to select an arbitrary `to` account.
+			let (asset, compound) = <Self as Staking>::claim(&instance_id, &to)?;
 			let nft_owner = T::get_protocol_nft_owner::<StakingNFTOf<T>>(&instance_id)?;
-			let to = if nft_owner == who { to } else { nft_owner };
-			<Self as Staking>::claim(&instance_id, &to)?;
+			ensure!(nft_owner == who, Error::<T>::OnlyPositionOwnerCanClaim,);
```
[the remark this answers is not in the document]
dzmitry-lahoda: yes, seems can.

frame/staking-rewards/src/lib.rs
dzmitry-lahoda: there is separate task for that, just added additional TODO.

frame/staking-rewards/src/lib.rs
dzmitry-lahoda: i will improve this next prs. there will be protocol and asset, and locked compounding vs not.

frame/staking-rewards/src/lib.rs
dzmitry-lahoda: it is like having 2 bottles. one is empty and one is full. one bottle is added with some a little, but than at some point other bottled fully drained into bottle which was added a few. and they switch position. will document that in PR. this is hack for blockhain not having real concurent queues.

dzmitry-lahoda: like A/B deploy. you have A or B.

dzmitry-lahoda: and switch

frame/staking-rewards/src/lib.rs
dzmitry-lahoda: i am not sure i can act on this in this PR. would it be ok to make changes in next PRs?

frame/staking-rewards/src/lib.rs
dzmitry-lahoda: there is task to handle that in clickup, just articulated concrere place

frame/staking-rewards/src/lib.rs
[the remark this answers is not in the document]
vimukthi-git: So shall we change?

PR #1065, opened 2022-05-25
Mck/add with transaction

mikolaichuk:
## Issue
https://app.clickup.com/t/2yk43dm
https://app.clickup.com/t/2z4bgvf

## Description
To avoid whole liquidation request failing due to one position's fail, 
liquidation request logic moved to a separate function liquidate_position(). 
Now we can continue the loop even if one request was failed. 
Each separate liquidation request is wrapped in storage transaction. 
Test_liquidate_multiple() was refactored. 
Checking of max liquidation batch size exceeding was moved to a separate test.
Storage cleanup after liquidation added. Corresponded tests updated.


frame/lending/src/lib.rs
```diff
@@ -775,8 +776,18 @@ pub mod pallet {
...
 			);
-			Self::liquidate_internal(&sender, &market_id, borrowers.clone())?;
-			Self::deposit_event(Event::LiquidationInitiated { market_id, borrowers });
+			// Unwrapped because of upper check.
+			// Perhaps borrowers ids should be in the bounded vector.
+			let borrowers =
+				BoundedVec::<_, T::MaxLiquidationBatchSize>::try_from(borrowers).unwrap();
+			let subjected_borrowers = Self::liquidate_internal(&sender, &market_id, borrowers);
+			// if at least one borrower was affected then liquidation been initiated
+			if subjected_borrowers.len() != 0 {
```
dzmitry-lahoda, line 785: You may use BiBoundedVec and avoid this check.

dzmitry-lahoda:
just converted to draft if you do not mind?

dzmitry-lahoda (approved):

frame/lending/src/lib.rs
mikolaichuk: BoundedVec used here because of liquidate_internal() function. It uses BoundedVec to avoid explicit MaxLiquidationBatchSize checking. Using BiBoundedVec in the liquidate_interanl() looks restrictive since potential user of this function may want to pass empty vector of borrowers there (for some reason). Perhaps we can use BoundedVec in dispatchable liquidate() function too.

frame/lending/src/lib.rs
```diff
@@ -776,8 +777,18 @@ pub mod pallet {
 				borrowers.len() <= T::MaxLiquidationBatchSize::get() as usize,
 				Error::<T>::MaxLiquidationBatchSizeExceeded
 			);
-			Self::liquidate_internal(&sender, &market_id, borrowers.clone())?;
-			Self::deposit_event(Event::LiquidationInitiated { market_id, borrowers });
+			// Unwrapped because of upper check.
+			// Perhaps borrowers ids should be in the bounded vector.
+			let borrowers = BoundedVec::<_, T::MaxLiquidationBatchSize>::try_from(borrowers)
```
dzmitry-lahoda, line 782: ok_or_err ?

dzmitry-lahoda: in theory you could use oneline with BiBoundedVec instead of 3 line checks.

also you do not check lower len > 0 (so at least one elemement in vector).

frame/lending/src/lib.rs
```diff
@@ -1034,8 +1102,14 @@ pub mod pallet {
 
 	// private helper functions
 	impl<T: Config> Pallet<T> {
-		fn get_market(market_id: &MarketIndex) -> Result<MarketConfigOf<T>, DispatchError> {
-			Markets::<T>::get(market_id).ok_or_else(|| Error::<T>::MarketDoesNotExist.into())
+		/// Returns pair of market's id and market (as 'MarketConfing') via market's id
+		/// - `market_id` : Market index as a key in 'Markets' storage
+		fn get_market(
+			market_id: &MarketIndex,
+		) -> Result<(&MarketIndex, MarketConfigOf<T>), DispatchError> {
```
dzmitry-lahoda, line 1109: not sure why return index?

frame/lending/src/lib.rs
```diff
@@ -846,41 +857,98 @@ pub mod pallet {
...
+		/// - `liquidator` : Liquidator's account id.
+		/// - `market_pair` : Index and configuration of the market from which tokens were borrowed.
+		/// - `account` : Borrower's account id whose debt are going to be liquidated.
+		fn liquidate_position(
+			liquidator: &<Self as DeFiEngine>::AccountId,
+			market_pair: &(&<Self as Lending>::MarketId, MarketConfigOf<T>),
+			borrow_asset: <T as DeFiComposableConfig>::MayBeAssetId,
+			account: &<Self as DeFiEngine>::AccountId,
+		) -> Result<(), DispatchError> {
+			let (market_id, market) = market_pair;
```
dzmitry-lahoda, line 872: not sure, usually there is no such rust patter with pair passed and then split. can do differently?

frame/lending/src/lib.rs
```diff
@@ -846,41 +857,98 @@ pub mod pallet {
...
+					borrow_asset,
+					collateral_to_liquidate,
+					unit_price,
+				);
+				T::Liquidation::liquidate(
+					&source_target_account,
+					sell,
+					market.liquidators.clone(),
+				)?;
+				// TODO: @mikolaichuk: Justify why is it fine to move borrow rent
```
dzmitry-lahoda, line 892: as for dicussion is happend to slack, what we decided? i assume it is safe

frame/lending/src/lib.rs
```diff
@@ -846,41 +857,98 @@ pub mod pallet {
...
+			} else {
+				return Err(DispatchError::Other(
+					"Tried liquidate position which is not supposed to be liquidated",
+				))
+			}
+			Ok(())
+		}
+
+		/// Liquidates position for each borrower in the vector within mentioned market.
+		/// Returns a vector of borrowers' account ids whose positions were requested to
```
dzmitry-lahoda, line 912: were send to liquidation succeffully, i think.
because requestred via lending
and liquidaitation does not know about positions.
so may be refrase

frame/lending/src/lib.rs
```diff
@@ -846,41 +857,98 @@ pub mod pallet {
...
-							deposit,
-							false,
-						)?;
+				// Wrap liquidate position request in a storage transaction.
+				// So, in the case of any error state's changes will not be committed
+				let storage_transaction_succeeded = with_transaction(|| {
+					let liquidation_response_result =
+						Self::liquidate_position(liquidator, &market_pair, borrow_asset, account);
+					if let Err(error) = liquidation_response_result {
+						log::error!("Creation of liquidation request for position {:?} {:?} was failed: {:?}",
```
dzmitry-lahoda, line 933: may be return error out of tx and log warning, not error. 
error usually handled layer +1 out of scope and errors will allow spam

frame/lending/src/tests.rs
```diff
@@ -51,6 +51,16 @@ type SystemAccountIdOf<T> = <T as frame_system::Config>::AccountId;
 type SystemOriginOf<T> = <T as frame_system::Config>::Origin;
 type SystemEventOf<T> = <T as frame_system::Config>::Event;
 
+// Bounds for configuration generic type, used in create market helpers.
+pub trait ConfigBound:
```
dzmitry-lahoda, line 55: nice

frame/lending/src/lib.rs
mikolaichuk: Agree

frame/lending/src/lib.rs
mikolaichuk: BoundedVec used here because of liquidate_internal() function. It uses BoundedVec to avoid explicit MaxLiquidationBatchSize checking. Using BiBoundedVec in the liquidate_interanl() looks restrictive since potential user of this function may want to pass empty vector of borrowers there (for some reason). Perhaps we can use BoundedVec in dispatchable liquidate() function too.

mikolaichuk: Regarding unwrap_or_default(), the same situation. Would not it be too restrictive emit en error because of empty borrowers vector?

frame/lending/src/lib.rs
```diff
@@ -846,41 +857,98 @@ pub mod pallet {
...
+		/// - `market_pair` : Index and configuration of the market from which tokens were borrowed.
+		/// - `account` : Borrower's account id whose debt are going to be liquidated.
+		fn liquidate_position(
+			liquidator: &<Self as DeFiEngine>::AccountId,
+			market_pair: &(&<Self as Lending>::MarketId, MarketConfigOf<T>),
+			borrow_asset: <T as DeFiComposableConfig>::MayBeAssetId,
+			account: &<Self as DeFiEngine>::AccountId,
+		) -> Result<(), DispatchError> {
+			let (market_id, market) = market_pair;
+			if Self::should_liquidate(market_id, account)? {
```
andor0, line 873: I would use `ensure!` here:
```
ensure!(
  Self::should_liquidate(market_id, account)?,
  DispatchError::Other("Tried liquidate position which is not supposed to be liquidated")
);

let collateral_to_liquidate = Self::collateral_of_account(market_id, account)?;
let source_target_account = Self::account_id(market_id);

// ...
```

mikolaichuk: Done, thanks!

frame/lending/src/lib.rs
mikolaichuk: Anyway we return Result here. Log level was changed.

frame/lending/src/lib.rs
mikolaichuk: Donce

frame/lending/src/lib.rs
mikolaichuk: We can use structure with two public fields.... I need this pair because of: https://github.com/ComposableFi/composable/pull/1065#discussion_r884573202

dzmitry-lahoda (approved):

andor0 (approved):

PR #1093, opened 2022-06-01
Pablo rpc

vivekvpandya:
## Issue
CU-2dvd2an

## Description
Adding RPC to get base/quote asset estimate for redeeming lp tokens.

## Checklist

### DIFF
--- a/frame/composable-traits/src/dex.rs
+++ b/frame/composable-traits/src/dex.rs
@@ -381,6 +381,15 @@ pub struct PriceAggregate<PoolId, AssetId, Balance> {
 	pub spot_price: Balance, // prices based on any other stat such as TWAP goes here..
 }
 
+/// RedeemableAssets for given amount of lp tokens.
+#[derive(RuntimeDebug, Encode, Decode, Default, Clone, PartialEq, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct RedeemableAssets<Balance> {
+	pub lp_tokens: Balance,
+	pub base_assets: Balance,
+	pub quote_assets: Balance,
+}
+
 #[cfg(test)]
 mod tests {
 	use crate::dex::{Fee, FeeConfig};
--- a/frame/composable-traits/src/dex.rs
+++ b/frame/composable-traits/src/dex.rs
@@ -36,7 +36,9 @@ pub trait Amm {
 	fn redeemable_assets_for_given_lp_tokens(
 		pool_id: Self::PoolId,
 		lp_amount: Self::Balance,
-	) -> Result<(Self::Balance, Self::Balance), DispatchError>;
+	) -> Result<RedeemableAssets<Self::AssetId, Self::Balance>, DispatchError>
+	where
+		Self::AssetId: sp_std::cmp::Ord;
 
 	/// Returns the amount of LP tokens that would be recieved by adding the given amounts of base
 	/// and quote.
--- a/frame/composable-traits/src/dex.rs
+++ b/frame/composable-traits/src/dex.rs
@@ -13,7 +13,7 @@ use sp_runtime::{
 	traits::{CheckedMul, CheckedSub},
 	ArithmeticError, DispatchError, Permill,
 };
-use sp_std::{ops::Mul, vec::Vec, collections::btree_map::BTreeMap};
+use sp_std::{collections::btree_map::BTreeMap, ops::Mul, vec::Vec};
 
 /// Trait for automated market maker.
 pub trait Amm {
@@ -384,10 +384,11 @@ pub struct PriceAggregate<PoolId, AssetId, Balance> {
 /// RedeemableAssets for given amount of lp tokens.
 #[derive(RuntimeDebug, Encode, Decode, Default, Clone, PartialEq, TypeInfo)]
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
-pub struct RedeemableAssets<AssetId, Balance> 
-    where AssetId: Ord
+pub struct RedeemableAssets<AssetId, Balance>
+where
+	AssetId: Ord,
 {
-    pub assets: BTreeMap<AssetId, Balance>
+	pub assets: BTreeMap<AssetId, Balance>,
 }
 
 #[cfg(test)]
--- a/frame/composable-traits/src/dex.rs
+++ b/frame/composable-traits/src/dex.rs
@@ -13,7 +13,7 @@ use sp_runtime::{
 	traits::{CheckedMul, CheckedSub},
 	ArithmeticError, DispatchError, Permill,
 };
-use sp_std::{ops::Mul, vec::Vec};
+use sp_std::{ops::Mul, vec::Vec, collections::btree_map::BTreeMap};
 
 /// Trait for automated market maker.
 pub trait Amm {
@@ -384,10 +384,10 @@ pub struct PriceAggregate<PoolId, AssetId, Balance> {
 /// RedeemableAssets for given amount of lp tokens.
 #[derive(RuntimeDebug, Encode, Decode, Default, Clone, PartialEq, TypeInfo)]
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
-pub struct RedeemableAssets<Balance> {
-	pub lp_tokens: Balance,
-	pub base_assets: Balance,
-	pub quote_assets: Balance,
+pub struct RedeemableAssets<AssetId, Balance> 
+    where AssetId: Ord
+{
+    pub assets: BTreeMap<AssetId, Balance>
 }
 
 #[cfg(test)]
--- a/frame/composable-traits/src/dex.rs
+++ b/frame/composable-traits/src/dex.rs
@@ -32,6 +32,12 @@ pub trait Amm {
 
 	fn lp_token(pool_id: Self::PoolId) -> Result<Self::AssetId, DispatchError>;
 
+	/// Returns the amount of base & quote asset redeemable for given amount of lp token.
+	fn redeemable_assets_for_given_lp_tokens(
+		pool_id: Self::PoolId,
+		lp_amount: Self::Balance,
+	) -> Result<(Self::Balance, Self::Balance), DispatchError>;
+
 	/// Returns the amount of LP tokens that would be recieved by adding the given amounts of base
 	/// and quote.
 	fn amount_of_lp_token_for_added_liquidity(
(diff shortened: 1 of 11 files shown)


ComposableFiB0t:
Task linked: [CU-2dvd2an RPC of remove liquidity](https://app.clickup.com/t/2dvd2an)

frame/pablo/rpc/src/lib.rs
```diff
@@ -118,4 +126,25 @@ where
 			)))
 		})
 	}
+
+	fn redeemable_assets_for_given_lp_tokens(
+		&self,
+		pool_id: SafeRpcWrapper<PoolId>,
+		lp_amount: SafeRpcWrapper<Balance>,
+		at: Option<<Block as BlockT>::Hash>,
+	) -> RpcResult<(SafeRpcWrapper<Balance>, SafeRpcWrapper<Balance>)> {
```
vimukthi-git, line 135: I think we should avoid creating interfaces that assume fixed number of assets in a pool while also ignoring assetIds, like 2 in this case. We can create a data structure to return here that has the assetId+balance list.

frame/composable-traits/src/dex.rs
```diff
@@ -375,6 +381,15 @@ pub struct PriceAggregate<PoolId, AssetId, Balance> {
 	pub spot_price: Balance, // prices based on any other stat such as TWAP goes here..
 }
 
+/// RedeemableAssets for given amount of lp tokens.
+#[derive(RuntimeDebug, Encode, Decode, Default, Clone, PartialEq, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct RedeemableAssets<Balance> {
+	pub lp_tokens: Balance,
+	pub base_assets: Balance,
+	pub quote_assets: Balance,
```
dzmitry-lahoda, line 390: _amount may be?  at least what is in extrinsic typocaly used is good

frame/dex-router/src/lib.rs
```diff
@@ -438,6 +438,17 @@ pub mod pallet {
...
 		}
 
+		fn redeemable_assets_for_given_lp_tokens(
+			pool_id: Self::PoolId,
+			lp_amount: Self::Balance,
+		) -> Result<(Self::Balance, Self::Balance), DispatchError> {
+			let (route, _reverse) = Self::get_route(pool_id).ok_or(Error::<T>::NoRouteFound)?;
+			match route[..] {
+				[pool_id] => T::Pablo::redeemable_assets_for_given_lp_tokens(pool_id, lp_amount),
+				_ => Err(Error::<T>::UnsupportedOperation.into()),
```
dzmitry-lahoda, line 448: specific error could be awesome. not general one.

frame/pablo/rpc/src/lib.rs
```diff
@@ -118,4 +126,25 @@ where
...
+
+		let at = BlockId::hash(at.unwrap_or_else(|| self.client.info().best_hash));
+
+		// calling ../../runtime-api
+		let runtime_api_result = api.redeemable_assets_for_given_lp_tokens(&at, pool_id, lp_amount);
+		runtime_api_result.map_err(|e| {
+			RpcError::Call(CallError::Custom(ErrorObject::owned(
+				9876,
+				"Something wrong",
+				Some(format!("{:?}", e)),
```
dzmitry-lahoda, line 146: if other rpc do string errors i am ok. just confirm?

frame/pablo/runtime-api/src/lib.rs
```diff
@@ -29,5 +29,10 @@ sp_api::decl_runtime_apis! {
 			base_asset_amount: SafeRpcWrapper<Balance>,
 			quote_asset_amount: SafeRpcWrapper<Balance>,
 		) -> SafeRpcWrapper<Balance>;
+
```
dzmitry-lahoda, line 32: oh, see amounts are around, can we be consistent in namings?

frame/pablo/src/lib.rs
```diff
@@ -322,6 +322,43 @@ pub mod pallet {
...
+						pool_quote_aum,
+						Self::Convert::convert(lp_issued),
+					)?);
+					Ok((base_amount, quote_amount))
+				},
+
+				// for LBP just return (0,0) as it does not have LP Token
+				PoolConfiguration::LiquidityBootstrapping(_) =>
+					Ok((Self::Balance::zero(), Self::Balance::zero())),
+			}
```
dzmitry-lahoda, line 360: in other places when lbp is different, error returned afaik

dzmitry-lahoda (approved):

frame/dex-router/src/lib.rs
vivekvpandya: On dex router certain operations are only applicable to route with a single pool. And this new function is one of them thus given error is fine here.

frame/composable-traits/src/dex.rs
```diff
@@ -375,6 +381,15 @@ pub struct PriceAggregate<PoolId, AssetId, Balance> {
 	pub spot_price: Balance, // prices based on any other stat such as TWAP goes here..
 }
 
+/// RedeemableAssets for given amount of lp tokens.
+#[derive(RuntimeDebug, Encode, Decode, Default, Clone, PartialEq, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct RedeemableAssets<Balance> {
```
vimukthi-git, line 387:
```suggestion
pub struct RedeemableAssets<AssetId, Balance> {
```

frame/composable-traits/src/dex.rs
```diff
@@ -375,6 +381,15 @@ pub struct PriceAggregate<PoolId, AssetId, Balance> {
...
+	pub lp_tokens: Balance,
```
vimukthi-git, line 388: My original suggestion was something like this, so that we can expand the number of assets and their balances returned without breaking clients.
```suggestion
	pub BTreeMap<AssetId, Balance> assets;
```
Can remove all the other fields here as well as knowing the base/quote asset is not important when redeeming

vimukthi-git (commented):
Just minor mods

frame/pablo/src/lib.rs
vivekvpandya: Yes that's true, but for LBP's remove_liquidity() lp_amount, min_base_amount, max_base_amount does not matter. So returning Error prevents call to LBP's remove_liquidity() function hence I just returned (0,0). Although I realized that I need to move check for CannotRespectMinimumRequested into individual match arms for StableSwap and ConstantProduct.

frame/pablo/src/lib.rs
```diff
@@ -322,6 +322,43 @@ pub mod pallet {
 		type TWAPInterval: Get<MomentOf<Self>>;
 
 		type WeightInfo: WeightInfo;
+
+		fn redeemable_assets_for_given_lp_tokens(
+			pool: PoolConfigurationOf<Self>,
+			pool_account: Self::AccountId,
+			lp_amount: Self::Balance,
+		) -> Result<(Self::Balance, Self::Balance), DispatchError> {
```
vimukthi-git, line 330: What's preventing us from using the same `RedeemableAssets` data structure here as well?

vimukthi-git: We need a test for this function

vivekvpandya: > What's preventing us from using the same `RedeemableAssets` data structure here as well?

What is need for that? Given RedeemableAssets will have map now. Why to iterate over map?

vimukthi-git: Not sure I understand, which place exactly you are referring to where iteration occurs?

vivekvpandya: I am using this new function in remove_liquidity(), there I need to query map to get base/quote asset balance.

vimukthi-git: Since other pallets integrate with Pablo using this trait, this would save us from some refactorings in the future across multiple pallets when there are more than two redeemable assets in the pool. Also having the assetId instead of just balances to avoid mistakes.

vimukthi-git: Hmm I also notice that this particular function follows a different pattern than what we follow for other trait impls where the `Amm` trait impl is used by the `Config` trait instead of the other way around. I think we may change that around if there is no blocker for that :)

frame/dex-router/src/lib.rs
dzmitry-lahoda: ::OnlyApplicableToRouteWithASinglePool

frame/pablo/src/lib.rs
dzmitry-lahoda: it is not (0,0). people will burn PICA for doing nothin. may be error them for doing stupid work?

frame/pablo/src/lib.rs
dzmitry-lahoda: yep. i fixed all that in lending to be like that.

frame/dex-router/src/lib.rs
vivekvpandya: I prefer separate PR as it changes DexRouter Errors and possible that break some integration test.

frame/pablo/src/lib.rs
vivekvpandya: burn PICA in RPC, or in remove_liquidity() ? I don't think it does any unnecessary work in LBP as it just return (0,0).

frame/pablo/rpc/src/lib.rs
benluelo: This is something that I have to review still, every rpcs errors are like this right now 👍

benluelo (dismissed):

frame/pablo/rpc/src/lib.rs
```diff
@@ -118,4 +126,25 @@ where
 			)))
 		})
 	}
+
+	fn redeemable_assets_for_given_lp_tokens(
+		&self,
+		pool_id: SafeRpcWrapper<PoolId>,
+		lp_amount: SafeRpcWrapper<Balance>,
+		at: Option<<Block as BlockT>::Hash>,
+	) -> RpcResult<RedeemableAssets<SafeRpcWrapper<AssetId>, SafeRpcWrapper<Balance>>> {
```
saadjhk, line 135: @vivekvpandya should it provide balance for both assets in pair, i see no asset id paramter here either to distinguish which asset will be released

vimukthi-git (approved):
LGTM

vimukthi-git (approved):

benluelo (approved):

PR #1095, opened 2022-06-01
use TWAP in lending

comfi01:
## Issue
CU-2tk08z9

## Description
lending: allow use twap from oracle

### DIFF
--- a/frame/composable-traits/src/oracle.rs
+++ b/frame/composable-traits/src/oracle.rs
@@ -92,10 +92,12 @@ pub trait Oracle {
 
 	/// Time Weighted Average Price
 	fn get_twap(
-		of: Self::AssetId,
+		asset_id: Self::AssetId,
 		weighting: Vec<Self::Balance>,
+		amount: Self::Balance,
 	) -> Result<Self::Balance, DispatchError>;
 
+	/// Length of price history
 	fn depth_of_history(of: Self::AssetId) -> usize;
 
 	/// How much of `quote` for unit `base` Oracle suggests to take.
--- a/frame/composable-traits/src/oracle.rs
+++ b/frame/composable-traits/src/oracle.rs
@@ -96,6 +96,8 @@ pub trait Oracle {
 		weighting: Vec<Self::Balance>,
 	) -> Result<Self::Balance, DispatchError>;
 
+	fn depth_of_history(of: Self::AssetId) -> usize;
+
 	/// How much of `quote` for unit `base` Oracle suggests to take.
 	/// Up to oracle how it decides ratio.
 	/// If there is no direct trading pair, can estimate via common pair (to which all currencies
--- a/frame/lending/src/lib.rs
+++ b/frame/lending/src/lib.rs
@@ -1117,9 +1117,18 @@ pub mod pallet {
 			asset_id: <T as DeFiComposableConfig>::MayBeAssetId,
 			amount: T::Balance,
 		) -> Result<T::Balance, DispatchError> {
-			<T::Oracle as Oracle>::get_price(asset_id, amount)
-				.map(|p| p.price)
-				.map_err(|_| Error::<T>::AssetPriceNotFound.into())
+			const WINDOW: usize = 3;
+			let prices_length = <T::Oracle as Oracle>::depth_of_history(asset_id);
+			if prices_length == 0 {
+				<T::Oracle as Oracle>::get_price(asset_id, amount)
+					.map(|p| p.price)
+					.map_err(|_| Error::<T>::AssetPriceNotFound.into())
+			} else {
+				let weights_length = if prices_length < WINDOW { prices_length } else { WINDOW };
+				// make flat weights
+				let weights = vec![(100_u128 / (weights_length as u128)).into(); weights_length];
+				<T::Oracle as Oracle>::get_twap(asset_id, weights)
+			}
 		}
 
 		/// Some of these checks remain to provide better errors. See [this clickup task](task) for
--- a/frame/lending/src/lib.rs
+++ b/frame/lending/src/lib.rs
@@ -1113,6 +1113,7 @@ pub mod pallet {
 				.ok_or_else(|| Error::<T>::MarketDoesNotExist.into())
 		}
 
+		/// Get TWAP from oracle. If history of prices is empty then return latest price.
 		fn get_price(
 			asset_id: <T as DeFiComposableConfig>::MayBeAssetId,
 			amount: T::Balance,
@@ -1127,7 +1128,7 @@ pub mod pallet {
 				let weights_length = if prices_length < WINDOW { prices_length } else { WINDOW };
 				// make flat weights
 				let weights = vec![(100_u128 / (weights_length as u128)).into(); weights_length];
-				<T::Oracle as Oracle>::get_twap(asset_id, weights)
+				<T::Oracle as Oracle>::get_twap(asset_id, weights, amount)
 			}
 		}
 
--- a/frame/oracle/src/lib.rs
+++ b/frame/oracle/src/lib.rs
@@ -405,10 +405,20 @@ pub mod pallet {
 		}
 
 		fn get_twap(
-			of: Self::AssetId,
+			asset_id: Self::AssetId,
 			weighting: Vec<Self::Balance>,
+			amount: Self::Balance,
 		) -> Result<Self::Balance, DispatchError> {
-			Self::get_twap(of, weighting)
+			let price = Self::get_twap(asset_id, weighting)?;
+			let unit = 10_u128
+				.checked_pow(Self::LocalAssets::decimals(asset_id)?)
+				.ok_or(DispatchError::Arithmetic(ArithmeticError::Overflow))?;
+			let price = multiply_by_rational(price.into(), amount.into(), unit)
+				.map_err(|_| DispatchError::Arithmetic(ArithmeticError::Overflow))?;
+			let price = price
+				.try_into()
+				.map_err(|_| DispatchError::Arithmetic(ArithmeticError::Overflow))?;
+			Ok(price)
 		}
 
 		fn depth_of_history(asset_id: T::AssetId) -> usize {
--- a/frame/oracle/src/lib.rs
+++ b/frame/oracle/src/lib.rs
@@ -411,6 +411,10 @@ pub mod pallet {
 			Self::get_twap(of, weighting)
 		}
 
+		fn depth_of_history(asset_id: T::AssetId) -> usize {
+			Self::price_history(asset_id).len()
+		}
+
 		fn get_ratio(
 			pair: composable_traits::defi::CurrencyPair<Self::AssetId>,
 		) -> Result<sp_runtime::FixedU128, DispatchError> {


frame/lending/src/lib.rs
```diff
@@ -1113,13 +1113,23 @@ pub mod pallet {
...
+			const WINDOW: usize = 3;
+			let prices_length = <T::Oracle as Oracle>::depth_of_history(asset_id);
+			if prices_length == 0 {
+				<T::Oracle as Oracle>::get_price(asset_id, amount)
+					.map(|p| p.price)
+					.map_err(|_| Error::<T>::AssetPriceNotFound.into())
+			} else {
+				let weights_length = if prices_length < WINDOW { prices_length } else { WINDOW };
+				// make flat weights
+				let weights = vec![(100_u128 / (weights_length as u128)).into(); weights_length];
```
dzmitry-lahoda, line 1130: i would prefer usage of Percent struct

frame/lending/src/lib.rs
```diff
@@ -1113,13 +1113,23 @@ pub mod pallet {
...
 		fn get_price(
 			asset_id: <T as DeFiComposableConfig>::MayBeAssetId,
 			amount: T::Balance,
 		) -> Result<T::Balance, DispatchError> {
-			<T::Oracle as Oracle>::get_price(asset_id, amount)
-				.map(|p| p.price)
-				.map_err(|_| Error::<T>::AssetPriceNotFound.into())
+			const WINDOW: usize = 3;
+			let prices_length = <T::Oracle as Oracle>::depth_of_history(asset_id);
+			if prices_length == 0 {
```
dzmitry-lahoda, line 1123: may be move it to <T::Oracle as Oracle>::get_price_weghted(asset_id, amount, weights)

comfi01: do you mean move all logic to `Oracle` side and remove `depth_of_history`? I would do it, but what should be result of the new method `Result<Option<Price>>` or `Result<Price>` with fallback to `get_price` inside?

frame/oracle/src/lib.rs
```diff
@@ -393,22 +393,33 @@ pub mod pallet {
...
+		fn get_price_weighted(
+			asset_id: Self::AssetId,
+			amount: Self::Balance,
+		) -> Result<Self::Balance, DispatchError> {
+			const WINDOW: usize = 3;
+			let prices_length = Self::depth_of_history(asset_id);
+			if prices_length == 0 {
+				Self::get_price(asset_id, amount).map(|p| p.price)
+			} else {
+				let weights_length = if prices_length < WINDOW { prices_length } else { WINDOW };
```
benluelo, line 409:
```suggestion
				let weights_length = prices_length.min(WINDOW);
```

frame/oracle/src/lib.rs
```diff
@@ -401,12 +405,12 @@ pub mod pallet {
...
-			const WINDOW: usize = 3;
 			let prices_length = Self::depth_of_history(asset_id);
 			if prices_length == 0 {
 				Self::get_price(asset_id, amount).map(|p| p.price)
 			} else {
-				let weights_length = if prices_length < WINDOW { prices_length } else { WINDOW };
+				let twap_window = Self::TwapWindow::get() as usize;
+				let weights_length = prices_length.min(twap_window);
 				// make flat weights
 				let weight = Percent::from_percent(100) / weights_length;
```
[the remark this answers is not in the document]
benluelo: @comfi01

benluelo (approved):

dzmitry-lahoda (dismissed):
it is better then it was, hope it iss more secure, but please leave comment on top of why it is averages over 3 prices and why lending needs that.

dzmitry-lahoda (approved):

benluelo (dismissed):

frame/oracle/src/lib.rs
```diff
@@ -981,6 +1003,26 @@ pub mod pallet {
 			Ok(weighted_average)
 		}
 
+		fn depth_of_history(asset_id: T::AssetId) -> usize {
```
vimukthi-git, line 1006: Curious, why is this a separate function?

frame/oracle/src/lib.rs
```diff
@@ -944,20 +959,27 @@ pub mod pallet {
 		#[allow(clippy::indexing_slicing)] // to get CI to pass
 		pub fn get_twap(
 			asset_id: T::AssetId,
-			mut price_weights: Vec<T::PriceValue>,
+			price_weights: Vec<Percent>,
```
vimukthi-git, line 962: Why are the weights being passed from outside?

frame/oracle/src/lib.rs
```diff
@@ -386,29 +389,41 @@ pub mod pallet {
...
-			let price = multiply_by_rational(price.into(), amount.into(), unit)
-				.map_err(|_| DispatchError::Arithmetic(ArithmeticError::Overflow))?;
-			let price = price
-				.try_into()
-				.map_err(|_| DispatchError::Arithmetic(ArithmeticError::Overflow))?;
+			let price = Self::quote(asset_id, price, amount)?;
 			Ok(Price { price, block })
 		}
 
+		fn get_price_weighted(
```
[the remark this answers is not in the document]
comfi01: it's helper for getting TWAP with flat distribution of weights.

frame/oracle/src/lib.rs
comfi01: i removed it

frame/oracle/src/lib.rs
vimukthi-git: I don't see the reason why this function should accept weights externally, but it's ok as it already exists.

frame/composable-traits/src/oracle.rs
```diff
@@ -90,10 +93,17 @@ pub trait Oracle {
...
 
+	/// Getting time weighted price for setted amount.
+	fn get_price_weighted(
+		asset_id: Self::AssetId,
+		amount: Self::Balance,
+	) -> Result<Self::Balance, DispatchError>;
+
 	/// Time Weighted Average Price
+	/// May be used to minimize a kurtosis impact on the price.
 	fn get_twap(
```
vimukthi-git, line 104: I think we can just remove this function all together as there are no usages. Also specifying weights for prices externally seems like a bad idea

frame/composable-traits/src/oracle.rs
```diff
@@ -90,10 +92,15 @@ pub trait Oracle {
 		Self::get_price(asset, unit).map(|_| true)
 	}
 
+	fn get_price_weighted(
```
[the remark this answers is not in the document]
vimukthi-git:
```suggestion
	fn get_twap(
```

frame/oracle/src/lib.rs
```diff
@@ -386,29 +389,48 @@ pub mod pallet {
...
 		) -> Result<Self::Balance, DispatchError> {
-			Self::get_twap(of, weighting)
+			let prices_length = Self::price_history(asset_id).len();
+			if prices_length == 0 {
+				Self::get_price(asset_id, amount).map(|p| p.price)
+			} else {
+				let twap_window = Self::TwapWindow::get().into();
+				let weights_length = prices_length.min(twap_window);
+				// make flat weights
+				let weight = Percent::from_rational(1, weights_length);
```
vimukthi-git, line 416: We can just remove the weight calculation as the TWAP for flat weights is just equal to 
`(sum of prices for the window) / window_length` where `window_length = min(prices_length, twap_window_length)`. i.e no need to call `get_twap`

frame/oracle/src/lib.rs
```diff
@@ -380,35 +382,40 @@ pub mod pallet {
 		}
 	}
 
-	impl<T: Config> Oracle for Pallet<T> {
+	impl<T: Config> Oracle for Pallet<T>
+	where
```
vimukthi-git, line 386: This bound can be removed as block numbers can be reliably `unique_saturated_into` u128 where necessary.

vimukthi-git (commented):
Calculation looks good, minor adjustments are needed

frame/oracle/src/lib.rs
```diff
@@ -940,51 +947,23 @@ pub mod pallet {
...
-
-		fn price_values_sum(price_values: &[T::PriceValue]) -> T::PriceValue {
-			price_values
-				.iter()
-				.fold(T::PriceValue::from(0_u128), |acc, b| acc.saturating_add(*b))
+			price: T::PriceValue,
+			amount: T::PriceValue,
+		) -> Result<T::PriceValue, DispatchError>
+		where
+			u128: From<<T as frame_system::Config>::BlockNumber>,
```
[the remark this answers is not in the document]
comfi01: Yup, for use `BlockNumber` as weight in calculations.

frame/oracle/src/lib.rs
```diff
@@ -1116,4 +1095,71 @@ pub mod pallet {
...
+						.unwrap_or_default()
+						.into()
+				})
+				.collect();
+
+			let weights_sum: u128 = weights.iter().sum();
+
+			let prices_sum: u128 = prices
+				.iter()
+				// skip helper for diff calculation
```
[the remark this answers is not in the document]
vimukthi-git: As I understand skipping 1 here is incorrect as the first price in the list need to be multiplied by the first weight in the weight list. To compensate the last price in the list has to be multiplied by `current_block_number - last_price_block_number`

frame/oracle/src/lib.rs
```diff
@@ -940,51 +947,86 @@ pub mod pallet {
...
+						.unwrap_or_default()
+						.unique_saturated_into()
 				})
-				.collect::<Vec<_>>();
-			let current_price = Self::prices(asset_id);
-			let current_weighted_price = last_weight.mul(current_price.price).div(precision);
+				.collect();
 
-			weighted_prices.push(current_weighted_price);
+			let weights_sum: u128 = weights.iter().sum();
```
vimukthi-git, line 980: We can add an `ensure` to assert that `weight_sum = latest_block - earliest block in the window`

vimukthi-git (approved):
LGTM :rocket:

frame/composable-traits/src/oracle.rs
```diff
@@ -92,17 +91,12 @@ pub trait Oracle {
 		Self::get_price(asset, unit).map(|_| true)
 	}
 
-	fn get_price_weighted(
+	/// Getting time weighted price for setted amount.
```
benluelo, line 94:
```suggestion
	/// Get the time weighted price for the `amount` of `asset_id`.
```

frame/oracle/src/lib.rs
```diff
@@ -358,8 +360,8 @@ pub mod pallet {
 		PriceNotFound,
 		/// Stake exceeded
 		ExceedStake,
-		/// Price weight must sum to 100
-		MustSumTo100,
+		/// Order of prices history for twap
```
benluelo, line 363: Not sure if this doc comment makes sense

frame/oracle/src/lib.rs
```diff
@@ -955,56 +947,76 @@ pub mod pallet {
...
+			let mut prices: Vec<_> = historical_prices
 				.iter()
-				.fold(Some(Percent::zero()), |a, b| a.and_then(|a| a.checked_add(b)));
-			ensure!(sum == Some(precision), Error::<T>::MustSumTo100);
-
-			let mut price_weights: Vec<T::PriceValue> =
-				price_weights.into_iter().map(|v| (v.deconstruct() as u128).into()).collect();
+				.rev()
+				.enumerate()
+				.filter_map(|(i, price)| if i < twap_window { Some(price) } else { None })
```
benluelo, line 962: can you just do `.take(twap_window)`?

frame/oracle/src/lib.rs
```diff
@@ -955,56 +947,76 @@ pub mod pallet {
...
+			let current_price =
+				Prices::<T>::try_get(asset_id).map_err(|_| Error::<T>::PriceNotFound)?;
+			prices.push(&current_price);
+
+			let weights: Vec<u128> = prices
+				.windows(2)
+				.map(|w| {
+					w.get(0)
+						.and_then(|a| w.get(1).map(|b| (a, b)))
+						.and_then(|(a, b)| b.block.checked_sub(&a.block))
```
benluelo, line 974: nit: more descriptive variable names would be good, i have no idea what `a` and `b` are here

frame/oracle/src/lib.rs
```diff
@@ -955,56 +947,76 @@ pub mod pallet {
...
-							historical_prices[historical_prices.len() - price_weights.len() + i]
-								.price,
-						)
-						.div(precision)
+				// skip 1 for diff calculation
+				.skip(1)
+				.map(|e| -> u128 { e.price.into() })
+				.zip(weights)
+				.filter_map(|(price, weight)| {
+					if weights_sum == 0_u128 {
```
benluelo, line 999: perhaps use `Zero::is_zero()`

vimukthi-git (approved):

benluelo (approved):

PR #1107, opened 2022-06-02
Cu 2j8gc2a story 1 mvp of rebalance without get apy mike belous

belousm:
## Issue

## Description

## Checklist

### DIFF
--- a/frame/instrumental-strategy-pablo/src/lib.rs
+++ b/frame/instrumental-strategy-pablo/src/lib.rs
@@ -24,9 +24,8 @@ pub mod pallet {
 		dispatch::{DispatchError, DispatchResult},
 		pallet_prelude::*,
 		storage::bounded_btree_set::BoundedBTreeSet,
+		traits::fungibles::{Inspect, Mutate, MutateHold, Transfer},
 		transactional, Blake2_128Concat, PalletId,
-		traits::{fungibles::{Inspect, Mutate, MutateHold, Transfer},
-		},
 	};
 	use sp_runtime::traits::{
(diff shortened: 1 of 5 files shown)


frame/instrumental-strategy-pablo/src/lib.rs
```diff
@@ -210,6 +227,39 @@ pub mod pallet {
...
 		#[transactional]
 		fn do_rebalance(_vault_id: &T::VaultId) -> DispatchResult {
+			let task = T::Vault::available_funds(vault_id, &Self::account_id())?;
+
+			let action = match task {
+				FundsAvailability::Withdrawable(balance) => {
+					let vault_account = T::Vault::account_id(vault_id);
+					Pools::iter_keys().for_each(|pool_id| {
+						if T::Pablo::currency_pair(pool_id).unwrap() == PICA  && 
+						T::Vault::asset_id.unwrap() == PICA {
```
kevin-composable, line 237: What is the purpose behind this call

> T::Vault::asset_id.unwrap() == PICA

kevin-composable: Does Pablo differentiate between EthxPICA and PICAxEth pools?

belousm: @kevin-composable So, here I tried to implement something like finding pool which contains PICA asset. And I also check whether we are trying to add/remove liquidity exactly with PICA. Because as I understand it, we are adding/removing liquidity to pool PICA/some asset. And so if we're trying to add PICA, it'll be base asset, not quote. Hence, we’ll need to pass the asset to the function parameter as the base asset.

belousm: Is my reasoning correct, or can it be implemented in some other way?

PR #1123, opened 2022-06-06
[Merged by Bors] - [cu-27xkb7w] Adding Oracle Reward Model

vimukthi-git:
## Issue
https://app.clickup.com/t/27xkb7w


ComposableFiB0t:
Task linked: [CU-27xkb7w Control Oracle inflation](https://app.clickup.com/t/27xkb7w)

frame/oracle/src/lib.rs
```diff
@@ -163,7 +169,7 @@ pub mod pallet {
 		type MaxPrePrices: Get<u32>;
 
 		/// The weight information of this pallet.
-		type WeightInfo: WeightInfo;
+		type OracleWeightInfo: WeightInfo;
```
dzmitry-lahoda, line 172: Not sure why is renamed neeed? Any weight info is of pallet. If you want to share configs may be look into DeFiComposableConfig example

frame/oracle/src/lib.rs
```diff
@@ -204,6 +210,13 @@ pub mod pallet {
 	pub type AssetsCount<T: Config> =
 		StorageValue<_, u32, ValueQuery, Nonce<ZeroInit, SafeIncrement>>;
 
+	#[pallet::storage]
+	#[pallet::getter(fn reward_rate)]
+	#[allow(clippy::disallowed_types)]
+	/// Annualized reward rate for oracles. This is used to inflate currency supply when issuing the
+	/// reward per each block.
+	pub type RewardRate<T: Config> = StorageValue<_, Perbill, ValueQuery>;
```
dzmitry-lahoda, line 218: likely this is not enough as kt does not tracks previos issue (bias and slope in parlance of curve afaik)

dzmitry-lahoda: so we move reward rates into each pallet?

frame/oracle/src/lib.rs
```diff
@@ -739,6 +757,23 @@ pub mod pallet {
 			}
 		}
 
+		fn possible_reward(standard_reward: BalanceOf<T>) -> BalanceOf<T> {
+			let max_reward_per_block = Self::max_reward_per_block();
+			sp_std::cmp::min(standard_reward, max_reward_per_block)
+		}
+
+		pub fn max_reward_per_block() -> BalanceOf<T> {
+			let total_issuance = <T as Config>::Currency::total_issuance();
```
dzmitry-lahoda, line 766: we have rewards in LBP/vesting/crowdloan. it seems usually it tracks previos reward and ups it more inf next step is more than gap step from previous. so not sure what goal it here, may be design should be different

dzmitry-lahoda: may you leave comment in code? 
as i got it is some safety measure, so it would be clear from code.

frame/oracle/src/lib.rs
vimukthi-git: >1.<T as pallet_oracle:Config>::WeigthInfo on call side
1.1. pallet_timestamp::Config do you use WeigthInfo from that trait? feels like bad smell.
1.2 i seen never such was done in any pallet and there are option to do diffently

I don't use the `WeightInfo` from that trait but there is a conflict since they both have the same name. I don't think there are other pallets extending  `pallet_timestamp::Config` trait to access it's storage.

>3. let slot_duration: u64 = <T as pallet_timestamp::Config>::MinimumPeriod::get. You can add type MinimumPeriod into pallet_oracle::Config and init it in runtime.

I considered this, but then the code would allow changing the pallet_oracle::Config without the actual `MinimumPeriod` of the chain having changed. The way I use this now is the way aura and babe use it refer, https://github.com/paritytech/substrate/blob/f318350b6410ef47e438aae24c0b4779373d0a7b/frame/babe/src/lib.rs#L468

Not sure about 3.1 and 3.2, let me check

frame/oracle/src/lib.rs
```diff
@@ -721,7 +739,7 @@ pub mod pallet {
 						slash_amount,
 					));
 				} else {
-					let reward_amount = asset_info.reward;
+					let reward_amount = Self::possible_reward(asset_info.reward);
```
hussein-aitlahcen, line 742: Everything is looking fine, but I have one question: right now if the reward of an asset exhaust the block reward, the remaining oracles won't get rewared. Would it be better to divide the "penalty" over all the rewards for the current block?

frame/oracle/src/mock.rs
```diff
@@ -131,8 +151,37 @@ where
 pub type AssetId = u128;
 pub type PriceValue = u128;
 parameter_types! {
- pub const TreasuryAccountId : AccountId= sr25519::Public([10u8; 32]);
+	 pub const TreasuryAccountId : AccountId= sr25519::Public([10u8; 32]);
+	pub const OraclePalletId: PalletId = PalletId(*b"plt_orac");
 }
+
+pub struct MockRewardModel();
+impl OracleRewardModel<Balance, Moment> for MockRewardModel {
```
vimukthi-git, line 159: TODO: Add this to each runtime as a common model with tests

frame/oracle/src/lib.rs
```diff
@@ -69,11 +70,17 @@ pub mod pallet {
...
+		},
 		AccountId32, ArithmeticError, FixedPointNumber, FixedU128, KeyTypeId as CryptoKeyTypeId,
-		PerThing, Percent, RuntimeDebug,
+		PerThing, Percent, RuntimeDebug, SaturatedConversion,
+	};
+	use sp_std::{
+		borrow::ToOwned, collections::btree_set::BTreeSet, fmt::Debug, str, vec, vec::Vec,
 	};
-	use sp_std::{borrow::ToOwned, fmt::Debug, str, vec, vec::Vec};
+	use std::ops::Add;
```
vimukthi-git, line 83: remove

KaiserKarel: Can you use the Add suggestion feature? Makes it easier to understand the context.

vimukthi-git: ah note to my self, not to use `std` :D

frame/composable-traits/src/time.rs
```diff
@@ -15,6 +15,7 @@ pub const ONE_HOUR: DurationSeconds = 60 * 60;
 /// current notion of year will take away 1/365 from lenders and give away to borrowers (as does no
 /// accounts to length of year)
 pub const SECONDS_PER_YEAR_NAIVE: DurationSeconds = 365 * 24 * ONE_HOUR;
+pub const MS_PER_YEAR_NAIVE: DurationSeconds = SECONDS_PER_YEAR_NAIVE * 1000;
```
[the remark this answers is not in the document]
vimukthi-git: Thanks let me check that out, I was wondering if there is nicer way to add this actually.

KaiserKarel (approved):

KaiserKarel (approved):

frame/oracle/src/lib.rs
```diff
@@ -311,7 +338,9 @@ pub mod pallet {
 		/// Oracle slashed. \[oracle_address, asset_id, amount\]
 		UserSlashed(T::AccountId, T::AssetId, BalanceOf<T>),
```
vivekvpandya, line 339: May be rename this too? In next PR?

vimukthi-git:
Since this PR may seem overwhelming with the design doc changes, here's a summary of changes

- Control the inflation done to reward Oracles by setting a max controlled by governance via two params: cost per oracle and number of oracles to support
- Distribute the per block rewards fairly among assets by using a weighted asset reward scheme

The design doc is basically justifying these choices with requirements and model

frame/oracle/src/lib.rs
vimukthi-git: Yeah makes sense, will do so later

KaiserKarel (approved):

vivekvpandya (approved):

PR #1127, opened 2022-06-08
Do not allow to increase collateral factor of open market.

vivekvpandya:
## Issue
https://app.clickup.com/t/2953d0u

## Description
For an open lending market increase of collateral factor can trigger loans to be liquidated. So such updates are not allowed.

## Checklist

### DIFF
--- a/frame/lending/src/tests.rs
+++ b/frame/lending/src/tests.rs
@@ -1,4 +1,3 @@
-
 //! (events can be obtained from System pallet as in banchmarking.rs before this commit)
 //! TODO: OCW of liquidations (like in Oracle)
 //! TODO: test on small numbers via proptests - detect edge case what is minimal amounts it starts
--- a/frame/lending/src/tests.rs
+++ b/frame/lending/src/tests.rs
@@ -1,3 +1,4 @@
+
 //! (events can be obtained from System pallet as in banchmarking.rs before this commit)
 //! TODO: OCW of liquidations (like in Oracle)
 //! TODO: test on small numbers via proptests - detect edge case what is minimal amounts it starts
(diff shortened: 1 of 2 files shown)


frame/lending/src/tests.rs
```diff
@@ -300,255 +301,255 @@ fn can_update_market() {
...
-					alice_total_debt_with_interest
-				);
-			},
-			_ => panic!(
-				"Unexpected value for System::events(); found {:#?}",
-				system_events
-			),
-		}
-	});
+            System::set_block_number(1); // ensure block is non-zero
```
benluelo, line 304: Is this all just whitespace change?

benluelo (approved):

mikolaichuk:
@vivekvpandya sorry, have commited changes instead of request

frame/lending/src/tests.rs
vivekvpandya: I think it is done by cargo fmt

vivekvpandya:
> @vivekvpandya sorry, have commited changes instead of request

Could you please explain it? To me updated check seems wrong, or I am missing some context.

mikolaichuk:
> > @vivekvpandya sorry, have commited changes instead of request
> 
> Could you please explain it? To me updated check seems wrong, or I am missing some context.

@vivekvpandya we update 'market.collateral.factor' with 'input.collateral.factor'. Market collateral factor should not be increased, so if `input.collateral.factor > market.collateral.factor` we  should not update 'market.collateral.factor' with new value.

frame/lending/src/tests.rs
```diff
@@ -1589,9 +1586,12 @@ fn market_owner_cannot_retroactively_liquidate() {
...
 		};
 		// ALICE is the creater of the market.
-		assert_ok!(Lending::update_market(Origin::signed(*ALICE), market_id, updatable));
-		// BOB loan must be liquidated now.
-		assert_eq!(Lending::should_liquidate(&market_id, &BOB), Ok(true));
+		assert_noop!(
+			Lending::update_market(Origin::signed(*ALICE), market_id, updatable),
+			Error::<Runtime>::CannotIncreaseCollateralFactorOfOpenMarket
+		);
+		// if above update was succeeded BOB's loan would have to be liquidated.
```
mikolaichuk, line 1593: Should we drop this accert?

mikolaichuk (approved):

frame/lending/src/tests.rs
```diff
@@ -1589,9 +1586,12 @@ fn market_owner_cannot_retroactively_liquidate() {
...
+		assert_eq!(Lending::should_liquidate(&market_id, &BOB), Ok(false));
```
mikolaichuk, line 1594: Should we drop this assert?

mikolaichuk (approved):

frame/lending/src/tests.rs
vivekvpandya: @mikolaichuk ensure macro retruns error if its condition is not satisfied . So my original check was correct and your check is wrong
```rust
					ensure!(
						input.collateral_factor > market.collateral_factor,
						Error::<T>::CannotIncreaseCollateralFactorOfOpenMarket
					);
```
with above check function will only allow updates when input > market's collateral.

vivekvpandya: Also after my change updates which increases collateral factor should not be allowed and hence following assert is correct.
```rust
assert_eq!(Lending::should_liquidate(&market_id, &BOB), Ok(false));
```

dzmitry-lahoda (commented):
was it product requirement? if this is temporal solution before governance, can we have a comment?

mikolaichuk (dismissed):

dzmitry-lahoda:
> Market collateral factor should not be increased, so if input.collateral.factor > market.collateral.factor we should not update 'market.collateral.factor' with new value.

why not? and if has special treatment untlike other parameters which can change curve and any other curve parameter, may be separate extinsic?

mikolaichuk:
> > Market collateral factor should not be increased, so if input.collateral.factor > market.collateral.factor we should not update 'market.collateral.factor' with new value.
> 
> why not? and if has special treatment untlike other parameters which can change curve and any other curve parameter, may be separate extinsic?

I believe  this question is addressed to @vivekvpandya.
I do not now about other parameters, but if market owner can change collateral factor after the market was started, he can initiate liquidation for all positions within the market by changing collateral factor to inadequate value. Is it expected behaviour?

vivekvpandya:
> > > Market collateral factor should not be increased, so if input.collateral.factor > market.collateral.factor we should not update 'market.collateral.factor' with new value.
> > 
> > 
> > why not? and if has special treatment untlike other parameters which can change curve and any other curve parameter, may be separate extinsic?
> 
> I believe this question is addressed to @vivekvpandya. I do not know about other parameters, but if market owner can change collateral factor after the market was started, he can initiate liquidation for all positions within the market by changing collateral factor to inadequate value. Is it expected behaviour?
For now I am told to not allow increase in collateral factor. We can keep clickup ticket open for discussion more better solutions.

dzmitry-lahoda:
>  but if market owner can change collateral factor after the market was started, he can initiate liquidation for all positions within the market by changing collateral factor to inadequate value. Is it expected behaviour?

yes. yes.

also he can change curve, increase IR curve parameters from 2 to 99 and hit liquidation.

vivekvpandya:
> > but if market owner can change collateral factor after the market was started, he can initiate liquidation for all positions within the market by changing collateral factor to inadequate value. Is it expected behaviour?
> 
> yes. yes.
> 
> also he can change curve, increase IR curve parameters from 2 to 99 and hit liquidation.

I would suggest IR should not be up-gradable once market is open. If we agree it can be done in separate PR with test.

mikolaichuk:
> > but if market owner can change collateral factor after the market was started, he can initiate liquidation for all positions within the market by changing collateral factor to inadequate value. Is it expected behaviour?
> 
> yes. yes.
> 
> also he can change curve, increase IR curve parameters from 2 to 99 and hit liquidation.

I would not borrow money on those terms :)

dzmitry-lahoda:
> I would not borrow money on those terms :)


do you have 1M usd? :)

> I would suggest IR should not be up-gradable once market is open. If we agree it can be done in separate PR with test.

I suggest to get approve from PO and write doc updates into reame or mdbook about.

mikolaichuk:
> do you have 1M usd? :)
Not yet :(

benluelo (approved):

mikolaichuk (approved):

mikolaichuk:
bors merge

mikolaichuk:
bors ping

PR #1141, opened 2022-06-13
Pablo saad feat/apollo

saadjhk:
## Issue
[Integrate Apollo into Pablo](https://app.clickup.com/t/2exrdvy)
## Description
Add Apollo integration for Price updates on UI

## Checklist

- [x] Update app wide prices of assets.
- [x] Update LP token prices.


diddt (approved):

KaiserKarel:
Question: does this mean that for a given asset price we first:

1. Check the price provided by the pablo RPC.
2. If no price is returned, fall back to Apollo.
3. If no price available for apollo, omit/fall back to pyth?

If so, I suggest we add an RPC to pablo, which performs 1. and 2. Thoughts @ComposableFi/parachain-finance?

dzmitry-lahoda:
@KaiserKarel if target to simply get price no matter what to have as much liquidity as possible then such trait and rpc with local price discovery -> applo sounds great. so gitbook should be documented too about that. (composable solution may appear  to be like choosing what to check first and some assets quality tiers)

diddt (dismissed):

saadjhk:
@KaiserKarel prices for pablo rpc gives FE spot price, however apollo gives price in USD, they are 2 separate prices

vimukthi-git:
@KaiserKarel @dzmitry-lahoda I had a call with @saadjhk on the requirement. Here it is,
- FE needs to have a list Pablo and/or other types of asset meta-data (ids, symbols, images etc.,) to show in screens
- Currently this list is hard coded, may be later we need to have an RPC for this (no task yet) and cache the result.
- FE needs to query the price of these assets at various stages(based on screen) from Apollo (this PR implements it).

Given this requirement and since we are working on integrating Pablo prices to Apollo through a price server(https://app.clickup.com/t/2f3k45w - which allows some flexibility) , I think it makes sense not to integrate Apollo into Pablo for price fetching. What do you think?

KaiserKarel:
> @KaiserKarel @dzmitry-lahoda I had a call with @saadjhk on the requirement. Here it is,
> 
> * FE needs to have a list Pablo and/or other types of asset meta-data (ids, symbols, images etc.,) to show in screens
> * Currently this list is hard coded, may be later we need to have an RPC for this (no task yet) and cache the result.
> * FE needs to query the price of these assets at various stages(based on screen) from Apollo (this PR implements it).
> 
> Given this requirement and since we are working on integrating Pablo prices to Apollo through a price server(https://app.clickup.com/t/2f3k45w - which allows some flexibility) , I think it makes sense not to integrate Apollo into Pablo for price fetching. What do you think?

I agree. Let's keep them separate, however let's make sure that in the FE, we display the prices that pablo will use in internals. If we use apollo to display BTC price at 10k, and Pablo has a spot price of 8k, then the FE should display 8k.

vimukthi-git:
I see so the goal is to provide a price that is executable in Pablo at the given moment? I think it's good to have that spot price taken from Pablo as you suggest as long as there is a pool for the required pair (eg: PBLO+USD).

vimukthi-git:
> simply have the RPC perform that trade and return the value? 
You mean like the simulation PR we have open right? I don't know exactly how that execution would work though, need to check.

frontend/ui-pablo/app/components/Organisms/swap/SwapForm.tsx
```diff
@@ -28,24 +28,49 @@ import { SwapRoute } from "./SwapRoute";
...
+    let quote = getAssetOnChainId(DEFAULT_NETWORK_ID, swaps.ui.quoteAssetSelected);
+    if (quote) {
+      let quoteId = quote.toString()
+      if (apollo[quoteId]) {
+        return new BigNumber(apollo[quoteId])
+      }
+    }
+    return new BigNumber(0);
+  }, [swaps.ui.quoteAssetSelected, apollo])
+
```
easteregg, line 73: Suggestion for later, or if we have time, maybe now: 

```tsx
function getTokenPriceInUsd(token) { 
  if (token === "none") return new BigNumber(0);
    let quote = getAssetOnChainId(DEFAULT_NETWORK_ID, token)?.toString();
    return apollo[quoteId] ? return new BigNumber(apollo[quoteId] : return new BigNumber(0);
}

saadjhk: can do, already have a hook.

saadjhk:
Created a PR that extends this so closing this one. @easteregg please review https://github.com/ComposableFi/composable/pull/1190

PR #1147, opened 2022-06-14
New runtime upgrade

alexkdefi:
## Issue
https://github.com/ComposableFi/SRE/issues/76
## Description
GA for runtime upgrade 

## Checklist


.github/workflows/new-runtime-upgrade.yml
```diff
@@ -0,0 +1,141 @@
+name: "NodeJs runtime upgrade"
```
KaiserKarel, line 1: Are we upgrading Nodejs?

.github/workflows/new-runtime-upgrade.yml
```diff
@@ -0,0 +1,141 @@
...
+
+      - name: Install
+        working-directory: scripts/github/
+        run: yarn install
+
+      - name: Upgrade runtime
+        working-directory: scripts/github/
+        run: |
+          FILE_NAME=picasso_runtime.wasm
+          yarn upgrade-runtime --root_key=${{ secrets.PICASSO_ROOT_KEY }} --rpc_ws_url=wss://rpc.composablefinance.ninja --path=./${FILE_NAME}
```
KaiserKarel, line 141: FYI, once we launch, we will no longer have access to `root`. These 2 actions would then need to become: submit runtime upgrade to chain for proposal.

KaiserKarel (changes requested):
Please squash your commits, to remove the unnecessary test-run messages.

.github/workflows/new-runtime-upgrade.yml
ch00mba: ok. That would be multisig right ? Which is basically another extrinsic. Correct ? @KaiserKarel

alexkdefi: ok

alexkdefi:
Bors do squash

PR #1154, opened 2022-06-15
Cu 31zpfez feature 3 set pool id for asset mike belous

belousm:
## Issue
https://app.clickup.com/t/31zpfez

## Description
More specifically, calling PabloStrategy::set_pool_id_for_asset(asset_id, pool_id) should store a mapping of asset_id -> pool_id in the pools runtime storage object.

## Checklist

### DIFF
--- a/frame/instrumental-strategy-pablo/src/lib.rs
+++ b/frame/instrumental-strategy-pablo/src/lib.rs
@@ -24,9 +24,8 @@ pub mod pallet {
 		dispatch::{DispatchError, DispatchResult},
 		pallet_prelude::*,
 		storage::bounded_btree_set::BoundedBTreeSet,
+		traits::fungibles::{Inspect, Mutate, MutateHold, Transfer},
 		transactional, Blake2_128Concat, PalletId,
-		traits::{fungibles::{Inspect, Mutate, MutateHold, Transfer},
-		},
 	};
 	use sp_runtime::traits::{
 		AccountIdConversion, AtLeast32BitUnsigned, CheckedAdd, CheckedMul, CheckedSub, One, Zero,
@@ -304,7 +303,7 @@ pub mod pallet {
 						T::Balance::zero(),
 						balance_of_asset,
 					)?;
-				}
+				},
 				FundsAvailability::Equilibrable => {
 					();
 				},
--- a/frame/instrumental-strategy-pablo/src/lib.rs
+++ b/frame/instrumental-strategy-pablo/src/lib.rs
@@ -14,7 +14,6 @@ pub mod pallet {
 	//                                   Imports and Dependencies
 	// -------------------------------------------------------------------------------------------
 	use codec::{Codec, FullCodec};
-	use composable_support::math::safe::SafeArithmetic;
 	use composable_traits::{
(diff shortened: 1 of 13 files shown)


ComposableFiB0t:
Task linked: [CU-31zpfez Feature 3 (set_pool_id_for_asset)](https://app.clickup.com/t/31zpfez)

frame/instrumental-strategy-pablo/src/lib.rs
```diff
@@ -184,6 +246,20 @@ pub mod pallet {
...
 		}
 
+		#[transactional]
+		fn set_pool_id_for_asset(
+			asset_id: T::AssetId,
+			pool_id: T::PoolId,
+		) -> Result<(), DispatchError> {
+			if Pools::<T>::contains_key(asset_id) {
+				Pools::<T>::mutate(asset_id, |_| pool_id);
+				Ok(())
```
kevin-composable, line 256: These `Ok(())`'s can be moved out of the if else block since both conditions result in it

frame/instrumental-strategy-pablo/src/lib.rs
```diff
@@ -184,6 +246,20 @@ pub mod pallet {
 			})
 		}
 
+		#[transactional]
+		fn set_pool_id_for_asset(
+			asset_id: T::AssetId,
+			pool_id: T::PoolId,
+		) -> Result<(), DispatchError> {
+			if Pools::<T>::contains_key(asset_id) {
+				Pools::<T>::mutate(asset_id, |_| pool_id);
```
kevin-composable, line 255: This logic is good, but here is where we need to start the logic of moving assets from one pool to the other.

kevin-composable: At a high level this makes me think that we need to store the strategies state of each asset.

kevin-composable: Something like:
`enum State {
Normal,
Transfering
`
`struct PoolConfig {
pool_id,
state: State
}`

and this can be stored in the `Pools` StorageMap

frame/instrumental-strategy-pablo/src/lib.rs
```diff
@@ -209,7 +285,57 @@ pub mod pallet {
 
 	impl<T: Config> Pallet<T> {
 		#[transactional]
-		fn do_rebalance(_vault_id: &T::VaultId) -> DispatchResult {
+		fn do_rebalance(vault_id: &T::VaultId) -> DispatchResult {
```
kevin-composable, line 288: I would separate the inner match arm's logic into their own functions, so you have something like:

```
match task {
	FundsAvailability::Withdrawable(balance) => {
	        Self::withdraw(...)
	} ...
```

kevin-composable (approved):
I just had a couple small refactoring changes, but the logic is there. Let me know if you have any questions on moving forward with the transferring of funds from one pool_id to another

frame/instrumental-strategy-pablo/src/lib.rs
belousm: Got it! Will refactor

PR #1157, opened 2022-06-16
Dryrun rpc

vivekvpandya:
## Issue
https://app.clickup.com/t/30xdrqf

## Description
RPC method to check if add_liquidity() and remove_liquidity() will success or not.

## Checklist

### DIFF
--- a/frame/composable-traits/src/dex.rs
+++ b/frame/composable-traits/src/dex.rs
@@ -33,7 +33,7 @@ pub trait Amm {
 	fn lp_token(pool_id: Self::PoolId) -> Result<Self::AssetId, DispatchError>;
 
 	/// Returns the amount of base & quote asset redeemable for given amount of lp token.
-	fn redeemable_assets_for_given_lp_tokens(
+	fn redeemable_assets_for_lp_tokens(
 		pool_id: Self::PoolId,
 		lp_amount: Self::Balance,
 		min_base_amount: Self::Balance,
@@ -44,12 +44,24 @@ pub trait Amm {
 
 	/// Returns the amount of LP tokens that would be recieved by adding the given amounts of base
 	/// and quote.
-	fn add_liquidity_dryrun(
+	fn simulate_add_liquidity(
 		who: &Self::AccountId,
 		pool_id: Self::PoolId,
 		base_amount: Self::Balance,
 		quote_amount: Self::Balance,
 	) -> Result<Self::Balance, DispatchError>;
+
+	/// Returns the amount of base/quote assets that would be recieved by removing the given
+	/// amounts of lp tokens.
+	fn simulate_remove_liquidity(
+		who: &Self::AccountId,
+		pool_id: Self::PoolId,
+		lp_amount: Self::Balance,
+		min_base_amount: Self::Balance,
+		min_quote_amount: Self::Balance,
+	) -> Result<RemoveLiquiditySimulationResult<Self::AssetId, Self::Balance>, DispatchError>
+	where
+		Self::AssetId: sp_std::cmp::Ord;
 
 	/// Get pure exchange value for given units of given asset. (Note this does not include fees.)
 	/// `pool_id` the pool containing the `asset_id`.
@@ -121,16 +133,6 @@ pub trait Amm {
 		min_receive: Self::Balance,
 		keep_alive: bool,
 	) -> Result<Self::Balance, DispatchError>;
-
-	fn remove_liquidity_dryrun(
-		who: &Self::AccountId,
-		pool_id: Self::PoolId,
-		lp_amount: Self::Balance,
-		min_base_amount: Self::Balance,
-		min_quote_amount: Self::Balance,
-	) -> Result<RemoveLiquidityDryrunResult<Self::AssetId, Self::Balance>, DispatchError>
-	where
-		Self::AssetId: sp_std::cmp::Ord;
 }
 
 /// Pool Fees
@@ -406,10 +408,10 @@ where
 	pub assets: BTreeMap<AssetId, Balance>,
 }
 
-/// RemoveLiquidityDryrunResult for given amount of lp tokens.
+/// RemoveLiquiditySimulationResult for given amount of lp tokens.
 #[derive(RuntimeDebug, Encode, Decode, Default, Clone, PartialEq, TypeInfo)]
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
-pub struct RemoveLiquidityDryrunResult<AssetId, Balance>
+pub struct RemoveLiquiditySimulationResult<AssetId, Balance>
 where
 	AssetId: Ord,
 {
--- a/frame/composable-traits/src/dex.rs
+++ b/frame/composable-traits/src/dex.rs
@@ -36,8 +36,7 @@ pub trait Amm {
 	fn redeemable_assets_for_lp_tokens(
 		pool_id: Self::PoolId,
 		lp_amount: Self::Balance,
-		min_base_amount: Self::Balance,
-		min_quote_amount: Self::Balance,
+		min_expected_amounts: BTreeMap<Self::AssetId, Self::Balance>,
 	) -> Result<RedeemableAssets<Self::AssetId, Self::Balance>, DispatchError>
 	where
 		Self::AssetId: sp_std::cmp::Ord;
@@ -47,9 +46,10 @@ pub trait Amm {
 	fn simulate_add_liquidity(
 		who: &Self::AccountId,
 		pool_id: Self::PoolId,
-		base_amount: Self::Balance,
-		quote_amount: Self::Balance,
-	) -> Result<Self::Balance, DispatchError>;
+		amounts: BTreeMap<Self::AssetId, Self::Balance>,
+	) -> Result<Self::Balance, DispatchError>
+	where
+		Self::AssetId: sp_std::cmp::Ord;
 
 	/// Returns the amount of base/quote assets that would be recieved by removing the given
 	/// amounts of lp tokens.
@@ -57,8 +57,7 @@ pub trait Amm {
 		who: &Self::AccountId,
 		pool_id: Self::PoolId,
 		lp_amount: Self::Balance,
-		min_base_amount: Self::Balance,
-		min_quote_amount: Self::Balance,
+		min_expected_amounts: BTreeMap<Self::AssetId, Self::Balance>,
 	) -> Result<RemoveLiquiditySimulationResult<Self::AssetId, Self::Balance>, DispatchError>
 	where
 		Self::AssetId: sp_std::cmp::Ord;
--- a/frame/composable-traits/src/dex.rs
+++ b/frame/composable-traits/src/dex.rs
@@ -44,7 +44,8 @@ pub trait Amm {
 
 	/// Returns the amount of LP tokens that would be recieved by adding the given amounts of base
 	/// and quote.
-	fn amount_of_lp_token_for_added_liquidity(
+	fn add_liquidity_dryrun(
+		who: &Self::AccountId,
 		pool_id: Self::PoolId,
 		base_amount: Self::Balance,
 		quote_amount: Self::Balance,
--- a/frame/composable-traits/src/dex.rs
+++ b/frame/composable-traits/src/dex.rs
@@ -36,6 +36,8 @@ pub trait Amm {
 	fn redeemable_assets_for_given_lp_tokens(
 		pool_id: Self::PoolId,
 		lp_amount: Self::Balance,
+		min_base_amount: Self::Balance,
+		min_quote_amount: Self::Balance,
 	) -> Result<RedeemableAssets<Self::AssetId, Self::Balance>, DispatchError>
 	where
 		Self::AssetId: sp_std::cmp::Ord;
@@ -118,6 +120,16 @@ pub trait Amm {
 		min_receive: Self::Balance,
 		keep_alive: bool,
 	) -> Result<Self::Balance, DispatchError>;
+
+	fn remove_liquidity_dryrun(
+		who: &Self::AccountId,
+		pool_id: Self::PoolId,
+		lp_amount: Self::Balance,
+		min_base_amount: Self::Balance,
+		min_quote_amount: Self::Balance,
+	) -> Result<RemoveLiquidityDryrunResult<Self::AssetId, Self::Balance>, DispatchError>
+	where
+		Self::AssetId: sp_std::cmp::Ord;
 }
 
 /// Pool Fees
@@ -387,6 +399,16 @@ pub struct PriceAggregate<PoolId, AssetId, Balance> {
 #[derive(RuntimeDebug, Encode, Decode, Default, Clone, PartialEq, TypeInfo)]
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 pub struct RedeemableAssets<AssetId, Balance>
+where
+	AssetId: Ord,
+{
+	pub assets: BTreeMap<AssetId, Balance>,
+}
+
+/// RemoveLiquidityDryrunResult for given amount of lp tokens.
+#[derive(RuntimeDebug, Encode, Decode, Default, Clone, PartialEq, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct RemoveLiquidityDryrunResult<AssetId, Balance>
 where
 	AssetId: Ord,
 {
--- a/frame/composable-traits/src/dex.rs
+++ b/frame/composable-traits/src/dex.rs
@@ -41,8 +41,8 @@ pub trait Amm {
 	where
 		Self::AssetId: sp_std::cmp::Ord;
 
-	/// Returns the amount of LP tokens that would be recieved by adding the given amounts of base
-	/// and quote.
+	/// Simulate add_liquidity computations, on success returns the amount of LP tokens
+	/// that would be recieved by adding the given amounts of base and quote.
 	fn simulate_add_liquidity(
 		who: &Self::AccountId,
 		pool_id: Self::PoolId,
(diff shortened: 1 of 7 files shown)


dzmitry-lahoda (dismissed):

frame/composable-traits/src/dex.rs
```diff
@@ -36,13 +36,16 @@ pub trait Amm {
 	fn redeemable_assets_for_given_lp_tokens(
 		pool_id: Self::PoolId,
 		lp_amount: Self::Balance,
+		min_base_amount: Self::Balance,
+		min_quote_amount: Self::Balance,
```
KaiserKarel, line 40: If this `/// Returns the amount of base & quote asset redeemable for given amount of lp token.`, why do we need to add min_base and min_quote parameters?

frame/composable-traits/src/dex.rs
```diff
@@ -36,13 +36,16 @@ pub trait Amm {
 	fn redeemable_assets_for_given_lp_tokens(
```
KaiserKarel, line 36: nit: 
```suggestion
	fn redeemable_assets_for_lp_tokens(
```

frame/pablo/rpc/src/lib.rs
```diff
@@ -32,22 +33,26 @@ where
 		PriceAggregate<SafeRpcWrapper<PoolId>, SafeRpcWrapper<AssetId>, SafeRpcWrapper<Balance>>,
 	>;
 
-	#[method(name = "pablo_expectedLpTokensGivenLiquidity")]
-	fn expected_lp_tokens_given_liquidity(
+	#[method(name = "pablo_add_liquidity_dryrun")]
+	fn add_liquidity_dryrun(
```
KaiserKarel, line 37: Aren't RPCs always dry runs?

frame/pablo/rpc/src/lib.rs
```diff
@@ -32,22 +33,26 @@ where
...
 		pool_id: SafeRpcWrapper<PoolId>,
 		base_asset_amount: SafeRpcWrapper<Balance>,
 		quote_asset_amount: SafeRpcWrapper<Balance>,
 		at: Option<BlockHash>,
 	) -> RpcResult<SafeRpcWrapper<Balance>>;
 
-	#[method(name = "pablo_redeemableAssetForGivenLpTokens")]
-	fn redeemable_assets_for_given_lp_tokens(
+	#[method(name = "pablo_remove_liquidity_dryrun")]
+	fn remove_liquidity_dryrun(
```
KaiserKarel, line 47: Aren't RPCs always dry runs?

frame/pablo/rpc/src/lib.rs
```diff
@@ -100,8 +106,9 @@ where
 		})
 	}
 
-	fn expected_lp_tokens_given_liquidity(
+	fn add_liquidity_dryrun(
```
KaiserKarel, line 109: Aren't RPCs always dry runs?

frame/composable-traits/src/dex.rs
```diff
@@ -118,6 +121,16 @@ pub trait Amm {
 		min_receive: Self::Balance,
 		keep_alive: bool,
 	) -> Result<Self::Balance, DispatchError>;
+
+	fn remove_liquidity_dryrun(
+		who: &Self::AccountId,
+		pool_id: Self::PoolId,
+		lp_amount: Self::Balance,
+		min_base_amount: Self::Balance,
```
vimukthi-git, line 129: We need to introduce asset Ids (lists of assets) for these types of calls, names like `base` and `quote` doesn't cut it anymore as we want to move towards multi asset pools.

frame/composable-traits/src/dex.rs
```diff
@@ -36,13 +36,16 @@ pub trait Amm {
 	fn redeemable_assets_for_given_lp_tokens(
 		pool_id: Self::PoolId,
 		lp_amount: Self::Balance,
+		min_base_amount: Self::Balance,
```
vimukthi-git, line 39: Same as the other comment

frame/dex-router/src/lib.rs
```diff
@@ -422,29 +422,54 @@ pub mod pallet {
...
+					pool_id,
+					lp_amount,
+					min_base_amount,
+					min_quote_amount,
+				),
+				_ => Err(Error::<T>::UnsupportedOperation.into()),
+			}
+		}
+
+		fn remove_liquidity_dryrun(
```
vimukthi-git, line 457:
```suggestion
		fn simulate_remove_liquidity(
```

frame/dex-router/src/lib.rs
```diff
@@ -422,29 +422,54 @@ pub mod pallet {
 			}
 		}
 
-		fn amount_of_lp_token_for_added_liquidity(
+		fn add_liquidity_dryrun(
```
vimukthi-git, line 425:
```suggestion
		fn simulate_add_liquidity(
```

frame/composable-traits/src/dex.rs
```diff
@@ -393,6 +406,16 @@ where
 	pub assets: BTreeMap<AssetId, Balance>,
 }
 
+/// RemoveLiquidityDryrunResult for given amount of lp tokens.
+#[derive(RuntimeDebug, Encode, Decode, Default, Clone, PartialEq, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct RemoveLiquidityDryrunResult<AssetId, Balance>
```
vimukthi-git, line 412:
```suggestion
pub struct RemoveLiquiditySimulationResult<AssetId, Balance>
```

frame/pablo/src/lib.rs
```diff
@@ -745,22 +697,33 @@ pub mod pallet {
...
+					ensure!(
+						T::Assets::reducible_balance(pool.pair.base, who, false) >= base_amount,
+						Error::<T>::NotEnoughLiquidity
+					);
+					ensure!(
+						T::Assets::reducible_balance(pool.pair.quote, who, false) >= quote_amount,
+						Error::<T>::NotEnoughLiquidity
+					);
+					Ok(amount_of_lp_token_to_mint)
+				},
```
KaiserKarel, line 726: I'd prefer for this to be methods on pool, so that the body of this function is reduced:

```
impl PoolConfiguration {
    // Returns number of LP tokens for the given pool
    fn lp_for_liquidity(...) -> Balance
}
```

Then the body can be:

```
let pool = Self::get_pool(pool_id)?;
let pool_account = Self::account_id(&pool_id);
ensure!(T::Assets::reducible_balance(pool.pair.base, who, false) >= base_amount,
Error::<T>::NotEnoughLiquidity
);
ensure!(
T::Assets::reducible_balance(pool.pair.quote, who, false) >= quote_amount,
Error::<T>::NotEnoughLiquidity
);
let lp = pool.lp_for_liquidity(base_amount, quote_amount);
```

frame/composable-traits/src/dex.rs
vimukthi-git: Probably best to have a common data structure that captures an `asset_id` and a `balance`

frame/composable-traits/src/dex.rs
vivekvpandya: Because we don't want a check for CannotRespectMinimuAmount in frontend. with those two parameters it is done in RPC.

frame/composable-traits/src/dex.rs
vivekvpandya: Is if fine to do that in separate change? AFAIK these RPCs are priority for FE.

frame/composable-traits/src/dex.rs
KaiserKarel: :+1:

frame/composable-traits/src/dex.rs
vivekvpandya: Also I request to have small PRs which has changes related to the subject of given click-up ticket. This allows reviewers to have fast reviews and make overall progress faster. IMHO diverging to other large subject is not good.

vimukthi-git: Unfortunately, there is no design document, but I also explained to you the same requirement before. We have ongoing discussions on supporting multi-asset stable swap soon. So this is not hypothetical. Since assetIds are expected to be provided through these functions, so for now, we can assert the order of assets provided as base first for now. Later we can switch the implementation easily.

vimukthi-git: >Also I request to have small PRs which has changes related to the subject of given click-up ticket.

I understand, but this function signature is being changed in the current PR, so I see that it's not going to cause a change in PR scope.

KaiserKarel (changes requested):
Would want to see the discussion regarding multi assets pool resolved before merging this.

frame/composable-traits/src/dex.rs
dzmitry-lahoda: @vimukthi-git 
```
We need to introduce asset Ids (lists of assets) for these types of calls, names like base and quote
``` 
this is future requirements, came after current requirement to get things done. for me it feels by process should be separate PR

frame/composable-traits/src/dex.rs
```diff
@@ -33,21 +33,33 @@ pub trait Amm {
...
 		lp_amount: Self::Balance,
+		min_expected_amounts: Vec<Self::Balance>,
 	) -> Result<RedeemableAssets<Self::AssetId, Self::Balance>, DispatchError>
 	where
 		Self::AssetId: sp_std::cmp::Ord;
 
 	/// Returns the amount of LP tokens that would be recieved by adding the given amounts of base
 	/// and quote.
-	fn amount_of_lp_token_for_added_liquidity(
+	fn simulate_add_liquidity(
```
dzmitry-lahoda, line 46: docs should be canged too. so may be in other PR not to block merge. i am very agains adding new work into already decide scope until work doen is useles. it is not useless in this case.

frame/composable-traits/src/dex.rs
vimukthi-git: @dzmitry-lahoda I do not consider this a new requirement nor was it not known before about this concept. You can refer this comment in a previous PR https://github.com/ComposableFi/composable/pull/1093#discussion_r886559668 where this same thing was explained.

vivekvpandya: There is already a task for assessment https://app.clickup.com/t/2gh7cjb . I changed functions to use Vec<> but as we see there is also possibilities of using HashMap. IMHO these kind of discussion should be part of design. Also when we complete 3 pool support in pallet then FE also need to use it. So even for them it is a rework.

vimukthi-git: @KaiserKarel solution makes sense. Only requirement from my end is to prevent us from assuming two assets at the interface level from now until we have a design/tasks to implement multi asset.

frame/composable-traits/src/dex.rs
vivekvpandya: updated.

dzmitry-lahoda (dismissed):

KaiserKarel (approved):

frame/pablo/rpc/src/lib.rs
```diff
@@ -32,22 +33,24 @@ where
...
 		pool_id: SafeRpcWrapper<PoolId>,
-		base_asset_amount: SafeRpcWrapper<Balance>,
-		quote_asset_amount: SafeRpcWrapper<Balance>,
+		amounts: BTreeMap<SafeRpcWrapper<AssetId>, SafeRpcWrapper<Balance>>,
 		at: Option<BlockHash>,
 	) -> RpcResult<SafeRpcWrapper<Balance>>;
 
-	#[method(name = "pablo_redeemableAssetForGivenLpTokens")]
-	fn redeemable_assets_for_given_lp_tokens(
+	#[method(name = "pablo_simulate_remove_liquidity")]
```
benluelo, line 45: these need to be camel case: pablo_simulateRemoveLiquidity (same for add rpc)

frame/pablo/src/lib.rs
```diff
@@ -129,6 +129,44 @@ pub mod pallet {
...
+					T::Convert::convert(quote_amount),
+					pool_base_aum,
+					pool_quote_aum,
+				)?;
+				Ok(T::Convert::convert(amount_of_lp_token_to_mint))
+			},
+			PoolConfiguration::LiquidityBootstrapping(_pool) => Ok(T::Balance::zero()),
+		}
+	}
+
```
benluelo, line 169: perhaps move this to the bottom of the file? feels out of place here

runtime/dali/src/lib.rs
```diff
@@ -1281,44 +1281,50 @@ impl_runtime_apis! {
...
-		pool_id: SafeRpcWrapper<PoolId>,
-		lp_amount: SafeRpcWrapper<Balance>
-	) -> RedeemableAssets<SafeRpcWrapper<CurrencyId>, SafeRpcWrapper<Balance>> {
+		fn simulate_remove_liquidity(
+			who: SafeRpcWrapper<AccountId>,
+			pool_id: SafeRpcWrapper<PoolId>,
+			lp_amount: SafeRpcWrapper<Balance>,
+			min_expected_amounts: BTreeMap<SafeRpcWrapper<CurrencyId>, SafeRpcWrapper<Balance>>,
+		) -> RemoveLiquiditySimulationResult<SafeRpcWrapper<CurrencyId>, SafeRpcWrapper<Balance>> {
+			let min_expected_amounts: BTreeMap<CurrencyId, Balance> = BTreeMap::from_iter(min_expected_amounts.iter().map(|(k, v)| (k.0, v.0)));
```
benluelo, line 1306:
```suggestion
			let min_expected_amounts: BTreeMap<_, _> = min_expected_amounts.iter().map(|(k, v)| (k.0, v.0)).collect();
```

runtime/dali/src/lib.rs
```diff
@@ -1281,44 +1281,50 @@ impl_runtime_apis! {
...
 
-		fn expected_lp_tokens_given_liquidity(
+		fn simulate_add_liquidity(
+			who: SafeRpcWrapper<AccountId>,
 			pool_id: SafeRpcWrapper<PoolId>,
-			base_asset_amount: SafeRpcWrapper<Balance>,
-			quote_asset_amount: SafeRpcWrapper<Balance>,
+			amounts: BTreeMap<SafeRpcWrapper<CurrencyId>, SafeRpcWrapper<Balance>>,
 		) -> SafeRpcWrapper<Balance> {
+			let amounts: BTreeMap<CurrencyId, Balance> = BTreeMap::from_iter(amounts.iter().map(|(k, v)| (k.0,v.0)));
```
benluelo, line 1289:
```suggestion
			let amounts: BTreeMap<_, _> = amounts.iter().map(|(k, v)| (k.0,v.0)).collect();
```

frame/composable-traits/src/dex.rs
```diff
@@ -33,20 +33,34 @@ pub trait Amm {
...
-		quote_amount: Self::Balance,
-	) -> Result<Self::Balance, DispatchError>;
+		amounts: BTreeMap<Self::AssetId, Self::Balance>,
+	) -> Result<Self::Balance, DispatchError>
+	where
+		Self::AssetId: sp_std::cmp::Ord;
+
+	/// Simulate remove_liquidity computations, on success returns the amount of base/quote assets
+	/// that would be recieved by removing the given amounts of lp tokens.
+	fn simulate_remove_liquidity(
```
hussein-aitlahcen, line 56: Why do we add another function for that, RPC are readonly so you can directly use the usual add/remove liquidity function from a RPC to dry run?

vivekvpandya: I was not aware about that. So I added a similar function which just check if balances are enough on all sides.

vivekvpandya: @hussein-aitlahcen so if I call a function which has #transactional from RPC , it will not change any thing in storage ?

vimukthi-git: I also would like to know a definitive answer to this question :) , what happens when an RPC call is made that updates storage? Does it avoid any storage updates if there is no explicit initiation of an extrinsic?

hussein-aitlahcen: It's a readonly operation that is not going to do anthing to the storage. So we are free to declare RPC that call functions called by extrinsics safely.

hussein-aitlahcen: We need to be careful though as a node can be ddos, but if it's a RPC only node we can probably scale it. Computing the liquidity is quite fast

hussein-aitlahcen: By the way, you should check out RPCs:
```
    system_dryRun
    system_dryRunAt
```

vivekvpandya: @saadjhk could you please verify https://polkadot.js.org/docs/substrate/rpc/#dryrunextrinsic-bytes-at-blockhash-applyextrinsicresult if that can satisfy your needs?

KaiserKarel (approved):

PoisonPhang (dismissed):

vimukthi-git (approved):
LGTM, though I see there is a question regarding proper impl of dry runs.

frame/composable-traits/src/dex.rs
saadjhk: @hussein-aitlahcen those RPCs require signed extrinsics, FE will have to prompt users for signatures for each dryrun

KaiserKarel (dismissed):

vivekvpandya:
> 

@benluelo 

> This PR was included in a batch that successfully built, but then failed to merge into main. It will not be retried.
> 
> Additional information:
> 
> ```json
> {"message":"Waiting on code owner review from ComposableFi/parachain-rpc.","documentation_url":"https://docs.github.com/articles/about-protected-branches"}
> ```

KaiserKarel:
@vivekvpandya You might need to reopen this PR, as Github kind of messes up required reviewers with team changes.

PR #1122, opened 2022-06-06
ICS20 Implementation

Wizdave97:
## Issue

Closes https://github.com/ComposableFi/composable/issues/1044

## Description
*Please replace this line with info that is not in the ClickUp ticket, eg: "also bumps dependency foo version"*

## Checklist

- [ ] I have updated the cargo docs to reflect changes made by this PR _(if applicable)_
- [ ] I have updated the `book/` to reflect changes made by this PR _(if applicable)_

PR #1158, opened 2022-06-16
Fix pallet ibc dependencies build errors.

Wizdave97:
## Issue

## Description

## Checklist

### DIFF
--- a/frame/ibc/Cargo.toml
+++ b/frame/ibc/Cargo.toml
@@ -24,7 +24,7 @@ frame-system = { git = "https://github.com/paritytech/substrate", branch = "polk
 pallet-timestamp = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.22", default-features = false, optional = true }
 parachain-info = { git = "https://github.com/paritytech/cumulus", default-features = false, branch = "polkadot-v0.9.22" }
 prost = { version = "0.10", default-features = false, optional = true }
-simple-iavl = { git = "https://github.com/Wizdave97/simple-avl", default-features = false, optional = true }
+simple-iavl = { git = "https://github.com/ComposableFi/simple-avl", rev = "7f68d854e82b3e422565df7100dedd79e8ffa346", default-features = false, optional = true }
 sp-core = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.22", default-features = false }
 sp-io = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.22", default-features = false }
 sp-runtime = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.22", default-features = false }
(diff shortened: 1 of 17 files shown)


hussein-aitlahcen (approved):

frame/ibc/Cargo.toml
```diff
@@ -33,32 +33,36 @@ sp-std = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v
...
+ics23 = { git = "https://github.com/composablefi/ics23", branch = "seun-v0.8.0-alpha", default-features = false, optional = true }
+tendermint = { git = "https://github.com/composableFi/tendermint-rs", branch = "seun-0.23.7", default-features = false, optional = true }
+tendermint-proto = { git = "https://github.com/composableFi/tendermint-rs", branch = "seun-0.23.7", default-features = false }
 
 # local deps
 ibc-primitives = { path = "ibc-primitives", default-features = false }
 ibc-trait = { path = "ibc-trait", default-features = false }
 pallet-ibc-ping = { path = "../ibc-ping", default-features = false }
 
+#temporary dependencies
```
KaiserKarel, line 47: Please describe why they are temporary, so other developers know when they can be removed.

frame/ibc/src/host_functions.rs
```diff
@@ -55,4 +51,60 @@ impl HostFunctionsProvider for HostFunctions {
...
+		use ripemd::Digest;
+		let mut hasher = ripemd::Ripemd160::new();
+		hasher.update(message);
+		let hash = hasher.finalize();
+		let mut res = [0u8; 20];
+		res.copy_from_slice(&hash);
+		res
+	}
+
+	fn ed25519_verify(signature: &[u8; 64], msg: &[u8], pubkey: &[u8]) -> bool {
```
KaiserKarel, line 67: I'd personally change the signature to be `ed25519_verify(signature: &[u8; 64], msg: &[u8], pubkey: &ed25519::Public) -> bool` . That way the caller can evaluate if it is a pubkey mismatch, or invalid pubkey. Now we hide that fact.

frame/ibc/src/impls.rs
```diff
@@ -683,9 +673,14 @@ where
 			ctx.channel_counter().map_err(|_| IbcHandlerError::ChannelInitError)?;
 		let channel_id = ChannelId::new(channel_counter);
 		// Signer does not matter in this case
-		let value = MsgChannelOpenInit { port_id, channel: channel_end, signer: Signer::new("") }
-			.encode_vec()
-			.map_err(|_| IbcHandlerError::ChannelInitError)?;
+		let value = MsgChannelOpenInit {
+			port_id,
+			channel: channel_end,
+			signer: Signer::from_str("pallet_ibc")
```
KaiserKarel, line 679: Can we make the `"pallet_ibc"` a constant to avoid typos etc. I'm seeing `"relayer"` used in the benchmarks and wondering what the difference is.

frame/ibc/src/lib.rs
```diff
@@ -450,7 +451,7 @@ pub mod pallet {
 				counterparty,
 				version: Some(version),
 				delay_period,
-				signer: Signer::new(""),
+				signer: Signer::from_str("pallet_ibc").map_err(|_| Error::<T>::DecodingError)?,
```
KaiserKarel, line 454: Can we make Signer a const to avoid doing the work during runtime; and reducing the error noise?

KaiserKarel (approved):

frame/ibc/Cargo.toml
Wizdave97: They'll be removed from this pallet in this PR https://github.com/ComposableFi/composable/pull/1122
because the runtime interface where they are used would be moved to it's own crate.

frame/ibc/src/host_functions.rs
Wizdave97: The trait definition is specified in `ibc-rs`, will keep this in mind for the next update.

frame/ibc/src/lib.rs
Wizdave97: The `Signer` struct holds a `String`, so  it can't be represented as a constant.

frame/ibc/src/host_functions.rs
```diff
@@ -55,4 +51,60 @@ impl HostFunctionsProvider for HostFunctions {
...
+			return false
+		};
+		sp_io::crypto::ed25519_verify(&signature, msg, &public_key)
+	}
+
+	fn sha2_256(message: &[u8]) -> [u8; 32] {
+		sp_io::hashing::sha2_256(message)
+	}
+
+	fn sha2_512(message: &[u8]) -> [u8; 64] {
```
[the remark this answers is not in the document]
Wizdave97: Don't quite understand what you mean here.

benluelo (approved):

frame/ibc/Cargo.toml
KaiserKarel: Alright, but in the case that you randomly fall sick or whatever, let's add that.

frame/ibc/src/host_functions.rs
KaiserKarel: Ah alright gotcha

frame/ibc/src/lib.rs
KaiserKarel: :( Badly designed APIs lol

frame/ibc/src/host_functions.rs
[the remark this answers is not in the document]
KaiserKarel: So if you have a function that returns a buffer/data, it increases the stack size. I however may have a buffer laying around that I wanted to dispose of. This API:

```
fn sha2_512(message: &[u8], mut [0u8; 64]) -> [u8; 64] {}
```

Is in general more flexible. But it really is a nit.

KaiserKarel (approved):

PR #1163, opened 2022-06-16
Vault pallet - Added function to calculate shares from asset

Matteocolella1995:
## Issue
I needed a function to calculate the amount of `lp_tokens` corresponding to an amount of asset given the current vault conditions. 

## Description
Added a trait function `calculate_lp_tokens_from_asset_amount` and its pallet implementation.

### DIFF
--- a/frame/composable-traits/src/vault.rs
+++ b/frame/composable-traits/src/vault.rs
@@ -113,7 +113,7 @@ pub trait Vault {
 		lp_amount: Self::Balance,
 	) -> Result<Self::Balance, DispatchError>;
 
-	fn calculate_lp_tokens_from_asset_amount(
+	fn amount_of_lp_token_for_added_liquidity(
 		vault_id: &Self::VaultId,
 		asset_amount: Self::Balance,
 	) -> Result<Self::Balance, DispatchError>;
--- a/frame/composable-traits/src/vault.rs
+++ b/frame/composable-traits/src/vault.rs
@@ -112,6 +112,11 @@ pub trait Vault {
 		vault_id: &Self::VaultId,
 		lp_amount: Self::Balance,
 	) -> Result<Self::Balance, DispatchError>;
+
+	fn calculate_lp_tokens_from_asset_amount(
+		vault_id: &Self::VaultId,
+		asset_amount: Self::Balance,
+	) -> Result<Self::Balance, DispatchError>;
 }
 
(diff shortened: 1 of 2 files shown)


kevin-composable (approved):

KaiserKarel (approved):

frame/vault/src/lib.rs
```diff
@@ -849,6 +849,25 @@ pub mod pallet {
...
 
+		fn do_calculate_lp_tokens_from_asset_amount(
+			vault_id: &T::VaultId,
+			vault: &VaultInfo<T>,
+			asset_amount: T::Balance,
+		) -> Result<BalanceOf<T>, DispatchError> {
+			let total_lp_issuance = T::Currency::total_issuance(vault.lp_token_id);
+			let aum = Self::assets_under_management(vault_id)?;
+
+			let a = <T::Convert as Convert<T::Balance, u128>>::convert(asset_amount);
```
dzmitry-lahoda, line 860: so you have functions which receives 3 numbers, converts and converts back?

dzmitry-lahoda: which is same as on lines 81x

dzmitry-lahoda: can we manage some code reuse?

dzmitry-lahoda: in this pr

frame/composable-traits/src/vault.rs
```diff
@@ -112,6 +112,11 @@ pub trait Vault {
 		vault_id: &Self::VaultId,
 		lp_amount: Self::Balance,
 	) -> Result<Self::Balance, DispatchError>;
+
+	fn calculate_lp_tokens_from_asset_amount(
```
dzmitry-lahoda, line 116: in AMM DEX traits (pablo) they call it `amount_of_lp_token_for_added_liquidity`

frame/vault/src/lib.rs
Matteocolella1995: You're right, I have refactored the code with your suggestions, let me know if it's better now or if I should change something else too

frame/composable-traits/src/vault.rs
Matteocolella1995: I have renamed the function in the last commit together with the refactor you suggested above

dzmitry-lahoda (approved):

KaiserKarel (approved):

PR #1168, opened 2022-06-17
Bors 🤖 do squash

vimukthi-git:
Bors 🤖 do squash

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -11,3 +11,4 @@ timeout_sec = 7200
 up_to_date_approvals = true
 update_base_for_deletes = true
 use_codeowners = true
+use_squash_merge = true


vimukthi-git:
bors+

dzmitry-lahoda (dismissed):

andor0 (dismissed):

KaiserKarel:
Moving discussion from slack to here. The question is; do we want to squash PRs?

vimukthi-git:
May be take a vote? We all agree the current state is not good, just need a solution that everyone commits to. Options are
- squash with github default
- rebase on your branch before merging with a neat history (if I understand correctly): a bit hard to enforce I guess though.

benluelo:
We could add checking the history to the review process, although it won't be possible to enforce that with a check (other than checking every commit with the CI pipeline, but I'm not sure if that would be worth it)

benluelo:
I think we should set standards for commit content & messages, whatever those standards are I don't personally care but if we all use the same format then it would make reviewing a lot easier

KaiserKarel:
> I think we should set standards for commit content & messages, whatever those standards are I don't personally care but if we all use the same format then it would make reviewing a lot easier

There are CI checks for message standards

benluelo:
> There are CI checks for message standards

That's sweet, I didn't know that! I think integrating something like that would be super helpful, help us all build better git habits

vimukthi-git:
@hussein-aitlahcen @benluelo @KaiserKarel @dzmitry-lahoda Shall we make this change for now until the proper rules/system can be set up (which just resets things to previous state before integrating bors. I'm ok with what @benluelo suggests for later though)? Longer we wait the more our main branch looks untidy (which for me is the main concern). 

Otherwise I can close this PR also.

PoisonPhang (dismissed):

KaiserKarel (approved):

PoisonPhang (dismissed):

benluelo (approved):

KaiserKarel (approved):

vimukthi-git:
Closing this to reopen as another PR to fix the bors problem

PR #1171, opened 2022-06-19
#CU-2z45zj3 reorganize lending pallet

comfi01:
## Issue

CU-2z45zj3

## Description
This is part of the larger refactoring that I want done for Angular. Rn I think that the create structure is a little cluttered and can be reorganized to follow the structure of our other projects.


frame/lending/src/helpers.rs
```diff
@@ -0,0 +1,616 @@
...
+};
+use sp_runtime::{
+	traits::{One, Saturating, Zero},
+	ArithmeticError, DispatchError, FixedPointNumber, FixedU128, Percent, Perquintill,
+};
+use sp_std::vec::Vec;
+
+// private helper functions
+impl<T: Config> Pallet<T> {
+	pub(crate) fn do_create_market(
```
benluelo, line 41: This is crate public, should it be under that impl block?

benluelo (approved):
Looks great to me, this refactor is long overdue 🚀

comfi01:
bors merge

PR #1173, opened 2022-06-20
Wasm runtime upgrade

alexkdefi:
## Issue
https://github.com/ComposableFi/SRE/issues/76
## Description
GA for runtime upgrade 

## Checklist


haroldsphinx (approved):

KaiserKarel (approved):

.github/workflows/new-runtime-upgrade.yml
```diff
@@ -0,0 +1,141 @@
...
+        options:
+          - dali
+          - composable
+          - picasso
+      release_selection:
+        type: string
+        required: true
+        description: 'Release version. Example: v2.2.0'
+
+jobs:
```
dzmitry-lahoda, line 19: if fills if condirion should define set of vars, and then all blocks run without much change. 
so instead of copy paste 3 times you paramterize.
in future that may be template of set of jobs tasks chosen per runtime upgrade.
so not sure if current copy paste is good now.

alexkdefi: Its a temporary logic.

PR #1174, opened 2022-06-20
added property test to check if vault balanced state holds

mikolaichuk:
## Issue
https://app.clickup.com/t/1y23xd8

## Description
Added property test to check if vault balanced state holds

## Checklist

### DIFF
--- a/frame/lending/src/tests/vault.rs
+++ b/frame/lending/src/tests/vault.rs
@@ -1,4 +1,12 @@
 use super::prelude::*;
+use crate::tests::borrow;
+use composable_traits::vault::{FundsAvailability, StrategicVault, Vault as VaultTrate};
+use frame_support::{
+	traits::{fungible::Mutate, fungibles::Mutate as FungiblesMutate},
+	PalletId,
+};
+use sp_core::sr25519::Public;
+use sp_runtime::{traits::AccountIdConversion, Perquintill};
 
 #[test]
 fn vault_takes_part_of_borrow_so_cannot_withdraw() {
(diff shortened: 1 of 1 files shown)


frame/lending/src/tests/vault.rs
```diff
@@ -67,3 +75,115 @@ fn test_vault_market_can_withdraw() {
 		);
 	});
 }
+
+// Generates well funded accounts
+fn generate_accounts(amount: u128) -> Vec<AccountId> {
+	let mut accounts = Vec::new();
+	let pid = PalletId(*b"markttst");
```
dzmitry-lahoda, line 82: you do not need to derive acounts from pallet id,

dzmitry-lahoda: it will be simpler to bump +1 existing accounts to get next

pub static CHARLIE: Lazy<AccountId> = Lazy::new(|| {
	AccountId::from_raw(hex!("0000000000000000000000000000000000000000000000000000000000000002"))
});

dzmitry-lahoda:
so it does not covers that you cannot borrow when unbalanced and when lender puts back into vault it starts to grow back.

so if fix account creation (no need to derive from pallet id), it is good. other pallets somewhere create many accounts.

dzmitry-lahoda (approved):

frame/lending/src/tests/vault.rs
mikolaichuk: @dzmitry-lahoda to bump existing account we have to bump value in string or in [u8;32] which is kind of tricky. into_sub_account() do it for us. I have changed generate_accounts() implementation to remove "pallet" references

mikolaichuk:
> so it does not covers that you cannot borrow when unbalanced and when lender puts back into vault it starts to grow back.

@dzmitry-lahoda It looks like we can borrow form the market which is related to unbalanced vault.
https://github.com/ComposableFi/composable/blob/863fa6907de2749b99868b8437594d174c769f14/frame/lending/src/tests/vault.rs#L204
PLMK if this behavior is not expected or you meant something different.

>lender puts back into vault it starts to grow back

Do you mean such check? 
https://github.com/ComposableFi/composable/blob/863fa6907de2749b99868b8437594d174c769f14/frame/lending/src/tests/vault.rs#L207

frame/lending/src/tests/vault.rs
```diff
@@ -67,3 +73,144 @@ fn test_vault_market_can_withdraw() {
 		);
 	});
 }
+
+// Struct for accounts generating
+#[derive(Encode, Decode, Default, PartialEq, Debug)]
+struct AccountPrefix([u8; 6]);
+impl TypeId for AccountPrefix {
```
dzmitry-lahoda, line 80: oh, not sure why it is needed, account can be anything not need for prefix, but ok.

dzmitry-lahoda (approved):

frame/lending/src/tests/vault.rs
dzmitry-lahoda: you may have u32 number, serde it and make trailing zeroes. other pallets do that. `trailing` key word.

dzmitry-lahoda:
> Do you mean such check?
summetry, balanced into other side, people start borrow. but ok.

dzmitry-lahoda:
> PLMK if this behavior is not expected or you meant something different.

if marked should give back to vault, then borrower should not be able to borrow on this block or at worse on next block. so likely you found bug. if lending is negative balance it should not give amount out. likely cc @kevin-composable

mikolaichuk:
> if marked should give back to vault, then borrower should not be able to borrow on this block or at worse on next block. so likely >you found bug. if lending is negative balance it should not give amount out. likely cc @kevin-composable

Sounds reasonable to me. If we have to re-balance vault-market pair, we do not know for sure how many assets we need for this operation. So, there is a non-zero probability that we not able to re-balance after borrowing. 

@dzmitry-lahoda @kevin-composable  
Added additional check in can_borrow() function:
https://github.com/ComposableFi/composable/blob/2e573ecde7c4f432a6a7d335528d229da866fed4/frame/lending/src/helpers.rs#L246	

Please approve if you ok with those changes

frame/composable-traits/src/vault.rs
```diff
@@ -181,6 +181,12 @@ pub trait StrategicVault: Vault {
 		from: &Self::AccountId,
 		amount: Self::Balance,
 	) -> Result<(), DispatchError>;
+
+	// Check if valut and strategy are balanced.
```
kevin-composable, line 185: valut -> vault

frame/lending/src/helpers.rs
```diff
@@ -243,6 +243,11 @@ impl<T: Config> Pallet<T> {
 			Error::<T>::MarketIsClosing
 		);
 
+		ensure!(
+			T::Vault::is_vault_balanced(&market.borrow_asset_vault, &market_account)?,
+			Error::<T>::CannotBorrowFromMarketWithUnbalancedVault
+		);
+
```
kevin-composable, line 250: can you merge these two ensures into one function, so we aren't querying available_funds back to back?

kevin-composable: I think a better way of doing this check is creating a function e.g. `can_borrow_from_vault`. This function can contain the logic for querying `available_funds` and return:
1. Error::<T>::MarketIsClosing if `available_funds` returns FundsAvailability::MustLiquidate
2. Error::<T>::CannotBorrowFromMarketWithUnbalancedVault if if `available_funds` returns FundsAvailability::Withdrawable(balance)
3. Ok() otherwise

kevin-composable: I think the logic of `is_vault_balanced` is better contained within the `lending` pallet rather than the `vault` pallet

frame/lending/src/tests/vault.rs
mikolaichuk: Done

frame/composable-traits/src/vault.rs
mikolaichuk: Done

frame/vault/src/lib.rs
```diff
@@ -1049,6 +1049,20 @@ pub mod pallet {
 				Ok(())
 			})
 		}
+		// Check if valut and strategy are balanced.
```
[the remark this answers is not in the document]
mikolaichuk: Done

frame/lending/src/helpers.rs
mikolaichuk: Done

mikolaichuk:
@dzmitry-lahoda @kevin-composable Changed accounts generation function. Now accounts are generating from `u32`.
Moved logic of vault balancing check and market availability to the separate `can_borrow_from_vault()` function

dzmitry-lahoda (approved):

benluelo (approved):

mikolaichuk:
bors merge

PR #1194, opened 2022-06-23
Dali stage

haroldsphinx:
- setup CI for dali-stage

### DIFF
--- a/.github/workflows/devnet-stage-release.yml
+++ b/.github/workflows/devnet-stage-release.yml
@@ -57,42 +57,42 @@ jobs:
 
       - name: Upload Artifacts
         run: |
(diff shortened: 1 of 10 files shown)


.github/workflows/new-runtime-upgrade.yml
```diff
@@ -0,0 +1,141 @@
+name: "WASM runtime upgrade"
```
andor0, line 1: Doesn't this job update dali-stage? Does it update other clusters/networks?

haroldsphinx: No it doesnt, it's a stray commit which I have rebased

seunlanlege (approved):

alexkdefi (approved):
👍

PR #1202, opened 2022-06-24
[CU-313atr5, CU-31md7r5] Create Staking Reward Pool

vimukthi-git:
## Issue
https://app.clickup.com/t/313atr5
https://app.clickup.com/t/31md7r5


ComposableFiB0t:
Task linked: [CU-313atr5 Staking-Rewards Pallet](https://app.clickup.com/t/313atr5)

frame/staking-rewards/src/lib.rs
```diff
@@ -85,11 +132,150 @@ pub mod pallet {
...
+	pub type RewardPools<T: Config> = StorageMap<_, Blake2_128Concat, T::PoolId, RewardPoolOf<T>>;
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Create a new reward pool based on the config.
+		///
+		/// Emits `RewardPoolCreated` event when successful.
+		#[pallet::weight(T::WeightInfo::create_reward_pool())]
+		#[transactional]
+		pub fn create_reward_pool(
```
vimukthi-git, line 190: TODO bench

frame/staking-rewards/src/lib.rs
```diff
@@ -85,11 +132,150 @@ pub mod pallet {
 		#[pallet::constant]
 		type PalletId: Get<PalletId>;
 
+		/// Maximum number of staking duration presets allowed.
+		#[pallet::constant]
+		type MaxStakingDurationPresets: Get<u32>;
+
+		/// Maximum number of reward configurations per pool.
+		#[pallet::constant]
+		type MaxRewardConfigsPerPool: Get<u32> + TypeInfo;
```
vimukthi-git, line 141:
```suggestion
		type MaxRewardConfigsPerPool: Get<u32>;
```

frame/composable-traits/src/staking/mod.rs
```diff
@@ -1,35 +1,138 @@
...
+/// TODO refer to the relevant section in the design doc.
+#[derive(RuntimeDebug, PartialEq, Eq, Clone, Encode, Decode, TypeInfo)]
+pub struct Reward<AssetId, Balance> {
+	/// asset id of the reward
+	pub asset_id: AssetId,
+
+	/// Total rewards including inflation for adjusting for new stakers joining the pool. All
+	/// stakers in a pool are eligible to receive a part of this value based on their share of the
+	/// pool.
+	pub total_rewards: Balance,
```
dzmitry-lahoda, line 27: this is not part of public API. I think that public API types should reveal only end results, not interals. so this likely to go to staking-rewards/src/types.rs

frame/staking-rewards/src/lib.rs
```diff
@@ -85,11 +132,150 @@ pub mod pallet {
...
-	#[pallet::generate_store(pub(super) trait Store)]
+	#[pallet::generate_store(pub (super) trait Store)]
 	#[pallet::without_storage_info]
 	pub struct Pallet<T>(_);
+
+	#[pallet::storage]
+	#[pallet::getter(fn pool_count)]
+	#[allow(clippy::disallowed_types)]
+	pub type RewardPoolCount<T: Config> =
+		StorageValue<_, T::PoolId, ValueQuery, Counter<ZeroInit, SafeIncrement, SafeDecrement>>;
```
vimukthi-git, line 177:
```suggestion
		StorageValue<_, T::PoolId, ValueQuery, Nonce<ZeroInit, SafeIncrement>>;
```

frame/composable-traits/src/staking/mod.rs
```diff
@@ -1,35 +1,138 @@
...
+	/// The staked asset id of the reward pool.
+	pub asset_id: AssetId,
+
+	/// rewards accumulated
+	pub rewards: BoundedBTreeMap<AssetId, Reward<AssetId, Balance>, RewardsLength>,
+
+	/// Total shares distributed among stakers
+	pub total_shares: Balance,
+
+	/// Pool would stop adding rewards to pool at this block number.
```
dzmitry-lahoda, line 76: we have rate,  stop block and max rewards. so if rewards max will be reached before block stop, than effectively max is less than expected. how would you handel that in UI to end users? i proposed solution which will not have such an issue

dzmitry-lahoda: soltion was to have only reward rate and block stop. not max

frame/composable-traits/src/staking/mod.rs
dzmitry-lahoda: Please split into RewardInput to upserts rewards an RewardState to store at rest.

frame/composable-traits/src/staking/mod.rs
```diff
@@ -70,32 +173,27 @@ pub trait Staking {
...
-	/// * `duration` the staking duration (must be one of the predefined presets). Unstaking before
-	///   the end trigger the unstake penalty.
+	/// * `who` the account to transfer the stake from.
+	/// * `amount` the amount to stake. the end trigger the unstake penalty.
 	/// * `keep_alive` whether to keep the `from` account alive or not while transferring the stake.
-	/// * `config_index` - reference to one of  staking configs for `pool_id`
 	fn stake(
 		who: &Self::AccountId,
 		pool_id: &Self::PoolId,
-		config_index: u8,
```
dzmitry-lahoda, line 88: how would user choose possible lock config our of several possible then when he stakes?

frame/composable-traits/src/staking/rewards.rs
```diff
@@ -1,26 +0,0 @@
-use crate::time::{DurationSeconds, Timestamp};
```
dzmitry-lahoda, line 1: i think rewards should be separate config from staking as it can be used separately and diffent domain. SOLID engineering.

frame/composable-traits/src/staking/mod.rs
```diff
@@ -1,35 +1,138 @@
...
-	pub lock: Option<Lock>,
-	/// The reward multiplier.
-	pub reward_multiplier: Perbill,
+	pub stake: Balance,
+
+	/// Pool share received for this NFT
+	pub share: Balance,
+
+	/// Reduced rewards by asset for the position (d_n)
+	pub reductions: BoundedBTreeMap<AssetId, Balance, RewardsLength>,
```
dzmitry-lahoda, line 132: this is not traits, not public API of pallet

frame/composable-traits/src/staking/mod.rs
```diff
@@ -1,35 +1,138 @@
...
 	/// The original stake this NFT was minted for or updated NFT with increased stake amount.
-	pub real_stake: Balance,
-	/// List of reward asset/pending rewards.
-	pub rewards: Rewards,
-	pub lock: Option<Lock>,
-	/// The reward multiplier.
-	pub reward_multiplier: Perbill,
+	pub stake: Balance,
+
+	/// Pool share received for this NFT
```
dzmitry-lahoda, line 128: its position, not NFT

frame/staking-rewards/src/weights.rs
```diff
@@ -1,3 +1,11 @@
...
+use frame_support::dispatch::Weight;
 
-impl WeightInfo for () {}
+pub trait WeightInfo {
+	fn create_reward_pool() -> Weight;
+}
+
+impl WeightInfo for () {
+	fn create_reward_pool() -> Weight {
+		todo!()
```
dzmitry-lahoda, line 9: not sure, may be put 1000 into pallet, than this. likely CI will not pass now as there are no bench.

frame/staking-rewards/src/test/runtime.rs
```diff
@@ -189,7 +189,23 @@ parameter_types! {
...
+	fn get() -> u32 {
+		10
+	}
+}
+
+#[derive(PartialEq, Eq, Copy, Clone, Encode, Decode, MaxEncodedLen, TypeInfo, core::fmt::Debug)]
+pub struct MaxRewardConfigsPerPool;
+impl Get<u32> for MaxRewardConfigsPerPool {
+	fn get() -> u32 {
+		10
```
dzmitry-lahoda, line 204: Use GetU32<10>

frame/staking-rewards/src/test/runtime.rs
```diff
@@ -172,7 +172,7 @@ parameter_types! {
 	pub const NativeAssetId: CurrencyId = PICA::ID;
 }
 
-impl pallet_assets::Config for Runtime {
+impl pallet_assets::Config for Test {
```
dzmitry-lahoda, line 175: i am agains it . it is NOT Test. It is Runtime for Test. Why you name Runtime to be Test? It is not test tests. Right? So it uses simplfied Runtime.

dzmitry-lahoda: I integrate this pallet into benches and into Dali, it is much easier if Runtime is named same in all places. please change back.

frame/composable-traits/src/staking/mod.rs
```diff
@@ -43,12 +146,12 @@ pub trait ProtocolStaking {
 	type AccountId;
 	type AssetId;
 	type Balance;
-	type PoolId;
+	type RewardPoolId;
```
dzmitry-lahoda, line 149: Not sure, I am ok with PoolId in context of Staking. Parent namespace defines what pool id is. Not sure why to mix and produce repetion. Runtim(Protocol(Pallet(Pool(Position))))

frame/composable-traits/src/staking/mod.rs
```diff
@@ -106,16 +204,16 @@ pub trait Staking {
...
 		remove_amount: Self::Balance,
 	) -> DispatchResult;
 
-	/// `ratio` - how much of share to retain in original position.
+	/// `ratio` - how much of share to retain in the original position.
 	fn split(
 		who: &Self::AccountId,
-		instance_id: &Self::PositionId,
+		position: &Self::PositionId,
 		ratio: Permill,
```
dzmitry-lahoda, line 215: in math, it was 1/2, 1/3, 1/4. So it could not define all possible percentages.  When using usually stuff like (1-p) where where 0<=p<=1.  iam ok with current ration so.

frame/staking-rewards/src/lib.rs
```diff
@@ -44,16 +42,53 @@ pub use pallet::*;
...
-	pub enum Event<T: Config> {}
+	pub enum Event<T: Config> {
+		/// Pool with specified id `T::PoolId` was created successfully by `T::AccountId`.
+		RewardPoolCreated {
+			/// Id of newly created pool.
+			pool_id: T::PoolId,
+			/// Owner of the pool.
+			owner: T::AccountId,
+			/// End block
+			end_block: T::BlockNumber,
```
dzmitry-lahoda, line 81: end block of pool? Why we mix rate config with pool config? Pool = Rate + Rewards + Staking.  if to define rewards configuration State separate from Input, then each reward will have rate and end block. end block in Input is one, but is split per reward.

frame/staking-rewards/src/lib.rs
```diff
@@ -44,16 +42,53 @@ pub use pallet::*;
...
+		},
+	}
+
 	#[pallet::error]
-	pub enum Error<T> {}
+	pub enum Error<T> {
+		/// Error when creating reward configs.
+		RewardConfigError,
+		/// Invalid end block number provided for creating a pool.
+		InvalidEndBlock,
```
dzmitry-lahoda, line 90: please be more specific. what is invalid? Likely is less than current block.

frame/staking-rewards/src/lib.rs
```diff
@@ -44,16 +42,53 @@ pub use pallet::*;
...
+			/// End block
+			end_block: T::BlockNumber,
+		},
+	}
+
 	#[pallet::error]
-	pub enum Error<T> {}
+	pub enum Error<T> {
+		/// Error when creating reward configs.
+		RewardConfigError,
```
dzmitry-lahoda, line 88: Not sure, `Error` and `RewardConfig**Error**`, Why repetions? We do not do this

frame/staking-rewards/src/lib.rs
```diff
@@ -65,10 +100,22 @@ pub mod pallet {
 		type Balance: Parameter + Member + BalanceLike + FixedPointOperand;
 
 		/// The reward pool ID type.
-		type PoolId: Parameter + Member + Clone + FullCodec;
+		/// Type representing the unique ID of a pool.
+		type PoolId: FullCodec
```
dzmitry-lahoda, line 104: Here is PoolId and it is clear it staking pool fron mod name, same should be in traits

frame/staking-rewards/src/lib.rs
dzmitry-lahoda, line 141: u8

frame/staking-rewards/src/lib.rs
```diff
@@ -85,11 +132,150 @@ pub mod pallet {
 		#[pallet::constant]
 		type PalletId: Get<PalletId>;
 
+		/// Maximum number of staking duration presets allowed.
+		#[pallet::constant]
+		type MaxStakingDurationPresets: Get<u32>;
```
dzmitry-lahoda, line 137: u8

frame/composable-traits/src/staking/mod.rs
```diff
@@ -1,35 +1,138 @@
...
+	pub reward_rate: Perbill,
+}
+
+/// Categorize the reward pool by it's incentive characteristics and expose
+/// initial configuration parameters.
+/// TODO refer to the relevant section in the design doc.
+#[derive(RuntimeDebug, Encode, Decode, MaxEncodedLen, Clone, PartialEq, Eq, TypeInfo)]
+pub enum RewardPoolConfiguration<AccountId, AssetId, Balance, BlockNumber, DurationPresets> {
+	/// A pool with an adjustable reward rate to be used as incentive.
+	RewardRateBasedIncentive {
```
dzmitry-lahoda, line 103: you mix things here. There will be extrinsic to add reward manually and also set rate to 0. So you traying to build non generic API. 

not sure I see multiplier for share in config

frame/composable-traits/src/staking/mod.rs
```diff
@@ -1,35 +1,138 @@
...
+		// possible lock config for this reward
+		lock: LockConfig<DurationPresets>,
+	},
+}
+
+/// Staking typed fNFT, usually can be mapped to raw fNFT storage type. A position identifier
+/// should exist for each position when stored in the runtime storage.
+/// TODO refer to the relevant section in the design doc.
+#[derive(Debug, PartialEq, Eq, Clone, Encode, Decode, TypeInfo)]
+pub struct Stake<RewardPoolId, AssetId, Balance, RewardsLength: Get<u32>> {
```
dzmitry-lahoda, line 121: u8

frame/staking-rewards/src/lib.rs
```diff
@@ -85,11 +132,150 @@ pub mod pallet {
...
+		/// Emits `RewardPoolCreated` event when successful.
+		#[pallet::weight(T::WeightInfo::create_reward_pool())]
+		#[transactional]
+		pub fn create_reward_pool(
+			origin: OriginFor<T>,
+			pool_config: RewardPoolConfigurationOf<T>,
+		) -> DispatchResult {
+			T::RewardPoolCreationOrigin::ensure_origin(origin)?;
+			let (owner, pool_id, end_block) = match pool_config {
+				RewardRateBasedIncentive {
```
dzmitry-lahoda, line 196: Other incetives this code will be same. So not sure why split a such way

frame/staking-rewards/src/lib.rs
```diff
@@ -85,11 +132,150 @@ pub mod pallet {
...
+					owner,
+					asset_id,
+					initial_reward_config,
+					end_block,
+					lock,
+				} => {
+					ensure!(
+						end_block > frame_system::Pallet::<T>::current_block_number(),
+						Error::<T>::InvalidEndBlock
```
dzmitry-lahoda, line 205: please be more specific in error.
please use Validated from configuration.

dzmitry-lahoda: please use UnixTime trate as dep like in other pallets

frame/staking-rewards/src/lib.rs
```diff
@@ -85,11 +132,150 @@ pub mod pallet {
...
+					);
+					let pool_id = RewardPoolCount::<T>::increment()?;
```
dzmitry-lahoda, line 207: may be split this code into function so can use trait as pools can be created by other protocols.

frame/staking-rewards/src/test/runtime.rs
```diff
@@ -200,5 +216,13 @@ impl pallet_staking_rewards::Config for Runtime {
 	type UnixTime = Timestamp;
```
dzmitry-lahoda, line 216: so you can use UnixTime trait to check block I guess, not direct deps on other pallet.

frame/composable-traits/src/staking/mod.rs
vimukthi-git: I think that is over refactoring,
1. Get things work according to the design
2. Then refactor if things make sense,

frame/composable-traits/src/staking/mod.rs
```diff
@@ -1,35 +1,138 @@
...
+	pub total_rewards: Balance,
+
+	/// A book keeping field to track the actual total reward without the
+	/// reward dilution adjustment caused by new stakers joining the pool.
+	pub total_dilution_adjustment: Balance,
+
+	/// Upper bound on the `total_rewards - total_dilution_adjustment`.
+	pub max_rewards: Balance,
+
+	/// The rewarding rate that increases the pool `total_reward` (and `actual_total_reward`)
```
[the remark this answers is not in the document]
vimukthi-git: That has been removed in favor of tracking `total_dilution_adjustment`
`total_rewards - total_dilution_adjustment` gives the `actual_total_reward`

frame/composable-traits/src/staking/mod.rs
vimukthi-git: That is totally fine those are independent variables. If there's a lot of staking than we expect then the max will be reached earlier

frame/composable-traits/src/staking/mod.rs
```diff
@@ -1,35 +1,138 @@
...
+/// Reward configurations for a given asset type.
+#[derive(RuntimeDebug, PartialEq, Eq, Clone, MaxEncodedLen, Encode, Decode, TypeInfo)]
+pub struct RewardConfig<AssetId, Balance> {
+	/// asset id of the reward
+	pub asset_id: AssetId,
+
+	/// Upper bound on the `total_rewards - total_dilution_adjustment`.
+	pub max_rewards: Balance,
+
+	/// The rewarding rate that increases the pool `total_reward` (and `actual_total_reward`)
```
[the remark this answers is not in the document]
vimukthi-git: Mistake
```suggestion
	/// The rewarding rate that increases the pool `total_reward`
```

frame/composable-traits/src/staking/mod.rs
vimukthi-git: Setting reward rate = 0 is a valid state. Any pool would support manual increase of rewards. Not a problem

frame/composable-traits/src/staking/mod.rs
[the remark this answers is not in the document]
vimukthi-git:
```suggestion
	/// The rewarding rate that increases the pool `total_reward` (and `total_rewards - total_dilution_adjustment`)
```

frame/composable-traits/src/staking/mod.rs
```diff
@@ -1,35 +1,138 @@
...
+#[derive(RuntimeDebug, Encode, Decode, MaxEncodedLen, Clone, PartialEq, Eq, TypeInfo)]
+pub enum RewardPoolConfiguration<AccountId, AssetId, Balance, BlockNumber, DurationPresets> {
+	/// A pool with an adjustable reward rate to be used as incentive.
+	RewardRateBasedIncentive {
+		/// Protocol or the user account that owns this pool
+		owner: AccountId,
+		/// The staked asset id of the reward pool.
+		asset_id: AssetId,
+		/// Pool would stop adding rewards to pool at this block number.
+		end_block: BlockNumber,
```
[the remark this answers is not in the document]
vimukthi-git: Right will rename

frame/composable-traits/src/staking/mod.rs
vimukthi-git:
```suggestion
pub struct Stake<RewardPoolId, AssetId, Balance, RewardsLength: Get<u8>> {
```

frame/composable-traits/src/staking/mod.rs
vimukthi-git:
```suggestion
	/// Pool share received for this Position
```

frame/composable-traits/src/staking/mod.rs
vimukthi-git:
```suggestion
	reductions: BoundedBTreeMap<AssetId, Balance, RewardsLength>,
```

frame/composable-traits/src/staking/mod.rs
vimukthi-git: This is to prevent confusions between Pablo pools and staking pools, it's for human readability

frame/composable-traits/src/staking/mod.rs
vimukthi-git: Makes sense, but the name wasn't clear to me. Let me make a change. Thanks

frame/composable-traits/src/staking/mod.rs
vimukthi-git: Yeah the ratio is not meant to be above 1. Regardless all we care about is that is a way to define the ratio for splitting

frame/staking-rewards/src/lib.rs
vimukthi-git: My conceptual design does not separate between those things for simplicity, let's not over refactor things for now.
But I'm open to making restructures when we have the complete system built.

frame/staking-rewards/src/lib.rs
vimukthi-git: Any suggestions for renaming this to be better? I also don't like this name

frame/staking-rewards/src/lib.rs
vimukthi-git: We don't know how other incentives/logic would look for now. This just leaves us the flexibility to branch the code cleanly later

frame/staking-rewards/src/lib.rs
vimukthi-git: Not really sure about the `UnixTime` trait suggestion, could you link to an example? Here we don't need time

frame/staking-rewards/src/test/runtime.rs
vimukthi-git: I'm just following the convention among substrate developers, which makes sense to me to differentiate between a Test runtime and an actual runtime when searching through code.

frame/staking-rewards/src/lib.rs
vimukthi-git: u32 is required as per the bounds on `BoundedBTreeMap` impls

frame/staking-rewards/src/lib.rs
vimukthi-git: u32 is required as per the bounds on `BoundedBTreeMap` impls

frame/composable-traits/src/staking/mod.rs
vimukthi-git: u32 is required as per the bounds on `BoundedBTreeMap` impls

frame/staking-rewards/src/test/runtime.rs
vimukthi-git: Here we need this struct to implement `TypeInfo`

frame/composable-traits/src/staking/rewards.rs
vimukthi-git: I do not agree as the idea of the design is rewards pool is an inherent part of the staking pool, without rewards there is no staking. But also as explained in another comment, first priority is to get things functionally working according to the design, then we refactor things if it makes sense.

frame/composable-traits/src/staking/mod.rs
dzmitry-lahoda: > /// Pool would stop adding rewards to pool at this block number.
> end_block: BlockNumber,
setting end_block to now also will end rate based rewarding at same time.

so there are 2 variables which can be set to  ensure it is no RewardRateBasedIncetive, not cleary why then it is called such?

frame/composable-traits/src/staking/mod.rs
dzmitry-lahoda: how max will be reached earlier? what if max reached, and end_block not. so we have end_block not reached, but there is no rewards. it is noncece.

frame/staking-rewards/src/lib.rs
dzmitry-lahoda: remove Error suffix as suggested

dzmitry-lahoda: as it is done in most other places

frame/staking-rewards/src/lib.rs
dzmitry-lahoda: what prevents you to rename to right now and add later more checks with other names?

frame/staking-rewards/src/lib.rs
dzmitry-lahoda: i do not agree. it add ambiguility. you can still init without intetive settings setting end block now or setting rate to 0. it is up to UI to make such kind of sloppy enum, not this pallet. if this is needed as some wrapper than can do it off pallet or as separate helper extrinsic.

frame/staking-rewards/src/lib.rs
dzmitry-lahoda: it was team decision to use Validated. if you want to override please ensure you settle that decision.  you validate big configuration. 

example, please validate asset_id exists in extinsic call.

frame/staking-rewards/src/test/runtime.rs
dzmitry-lahoda: please send me reference

frame/composable-traits/src/staking/mod.rs
vimukthi-git: It is natural to set the reward rate to zero because we would no longer want to incentivize a certain pool after launching and running for a while. As I explained above it's part of how the incentive works, it can be turned off when owner/governance decides to do so.

frame/staking-rewards/src/lib.rs
vimukthi-git: https://app.clickup.com/t/2n9rhpd

frame/composable-traits/src/staking/mod.rs
vimukthi-git: It's not non-sense, there are two ways to control the rewards campaign, either it runs out of time or runs out of funds, perfectly normal.

frame/staking-rewards/src/lib.rs
dzmitry-lahoda: I think this enum should not be part of main API, and should be part of layer on top.

frame/composable-traits/src/staking/mod.rs
dzmitry-lahoda: please get approve on adding non public stuff into composable traits

frame/composable-traits/src/staking/mod.rs
dzmitry-lahoda: max rewads = time * rate. having to ways of control does not makes any sense except creating ambiguity. 
one may choose reward amount, then get rate to fit that.

dzmitry-lahoda: please provide concrete example when it will not be the case

dzmitry-lahoda: why to introduce stuff when there is no use case?

frame/composable-traits/src/staking/mod.rs
dzmitry-lahoda: enum Incetives {
  RewardRateBasedIncentive(SharedConfig)
}

enum(struct) == sturct(enum) => you produce something not useful

frame/composable-traits/src/staking/mod.rs
dzmitry-lahoda: but math was different. if was easy to make math and code be same.

frame/composable-traits/src/staking/mod.rs
dzmitry-lahoda: not sure, for me confusing double naming. I do not think it is good naming.

frame/staking-rewards/src/lib.rs
dzmitry-lahoda: in SE simplicity is SOLID. conceptual desing = code too.

frame/staking-rewards/src/lib.rs
dzmitry-lahoda: please get approve not using validate as it was decision of team and leads before.

dzmitry-lahoda: please estimate that task if need too

frame/staking-rewards/src/test/runtime.rs
dzmitry-lahoda: frame_support::traits::ConstU32<16>;

frame/staking-rewards/src/types.rs
```diff
@@ -1,2 +0,0 @@
-//! Internal types used inside pallet, example for storage.
```
dzmitry-lahoda, line 1: please move private types of pallet here

runtime/dali/src/lib.rs
```diff
@@ -867,20 +867,31 @@ impl crowdloan_rewards::Config for Runtime {
...
 
 impl pallet_staking_rewards::Config for Runtime {
 	type Event = Event;
-	type Share = Balance;
 	type Balance = Balance;
-	type PoolId = u16;
+	type RewardPoolId = u16;
 	type PositionId = u128;
-	type MayBeAssetId = CurrencyId;
+	type AssetId = CurrencyId;
```
dzmitry-lahoda, line 886: It is not asset, it is may be asset. please retain the difference or split these 2

frame/composable-traits/src/staking/mod.rs
vimukthi-git: This is not non-public stuff, it's your assumption

frame/staking-rewards/src/test/runtime.rs
dzmitry-lahoda: in Lending we switched from Blocks to Seconds afaik for some reason were Karol was in discussion.

frame/composable-traits/src/staking/mod.rs
vimukthi-git: Let me explain one last time on top of many other times I have explained to you online/offline/chat :sweat_smile: 

- max rewards = time * rate  + any additional transferred rewards
- `rate` is adjusted(governance/gauge etc for incentivizing staking or the opposite) so it's not possible to predict that end_block will end with the max reward possible
- `any additional transferred rewards` increases . Then the pool can satisfy it's maximum reward goal earlier than expected, then there is no need to keep. To control this scenario we need a max reward independent of the end block/time. 
- Also the rewarded assets are independent of each other, one can reach the max reward for it before the others and before the overall reward pool has expired.
- You can see this type of dynamic uniswap clones, most recent one I used is https://app.stellaswap.com/farm

frame/staking-rewards/src/test/runtime.rs
vimukthi-git: I'm actually thinking we could remove this `end_block` field as it creates more trouble than help and probably irrelevant to the overall design as originally intended. cc @KaiserKarel

frame/staking-rewards/src/lib.rs
vimukthi-git: The api should be forward compatible so I would like this to be part of the API

vimukthi-git: cc @KaiserKarel

frame/staking-rewards/src/test/runtime.rs
vimukthi-git: Could you point to where that `ConstU16` implements `TypeInfo` ?

frame/staking-rewards/src/types.rs
vimukthi-git: As I mentioned in other places, let's not over refactor for now, which puts a lot constraints on getting stuff done. Once we have the overall thing working we can do whatever makes sense.

frame/composable-traits/src/staking/mod.rs
KaiserKarel: I think that design is fine. Governance just needs the knobs to actually configure this.

frame/composable-traits/src/staking/mod.rs
KaiserKarel: enum(struct) with the #[non_exhaustive] attribute is fine to communicate possible future changes. It should be zero overhead anyways.

frame/staking-rewards/src/lib.rs
KaiserKarel: Let's go with the enum for now. We can always create anothe, better designed extrinsic and migrate towards that.

KaiserKarel (approved):

frame/staking-rewards/src/test/runtime.rs
vimukthi-git: Fixed the problem in a different way, thanks

frame/staking-rewards/src/test/runtime.rs
vimukthi-git: Just check the frame pallets in substrate repo

frame/staking-rewards/src/lib.rs
vimukthi-git: Done

PoisonPhang (approved):
LGTM!

frame/composable-traits/src/staking/mod.rs
dzmitry-lahoda: i use word public as inter crate and inter pallet usage. modifications provided do make negative unclaimable non existing rewards as part of commont types. while it is just detail of implementation.  do we put internal storage states into composable traits? afaik we do not do this.

dzmitry-lahoda: i suggested tonput these types inside subject pallet as per my comments.

dzmitry-lahoda: so for me it ok if @KaiserKarel will approve this and we will take the direction of put internal pallet details used purely for bookkeeping(these number do not mean anything useful) as part of composable traits.

dzmitry-lahoda: so I think it is weird

frame/composable-traits/src/staking/mod.rs
dzmitry-lahoda: 4. about assets and pool have different lifecycles. i did not say differently. each asset can have its end block.

dzmitry-lahoda: 2. not sure why need to set max to reward explicitly?  why would somebody adding reward into the pool after there is  no need to do so? will be other pallet?  
and i do propose to predict anything.

dzmitry-lahoda: 3. not sure. in those case should not rate to be set to zero? so pool is not rewardinf anymore?

dzmitry-lahoda: i imagine UI. reward rate is nice, end block is far ahead. and pool not rewarding at all. but wait, there is reward rate, why pool not rewarding?

dzmitry-lahoda: it possible, that hook logic for rate could track previous reward and calculated by per asset block. if somebody rewarded our of rate via extrinsic, it can minus block for proper amount.   no need for max to ve stored on chain and  no incosistencies in ui. so max reward = already rewarded + rate * remaining time.

dzmitry-lahoda: so what i say that max reward is derived field and not having it in state will make UI more consistent.

frame/staking-rewards/src/lib.rs
dzmitry-lahoda: hm. i proposed remove enum. enum here is overenineering imho. 

not sure why we handling backward compatibility on this level. never seen such pallets before. so it is also no the substrate design.

dzmitry-lahoda: i do not see a reason why we overcomplicate api and going agains how other extrinsics designed, and that is on top that enum creates ambiguilty.

dzmitry-lahoda: @KaiserKarel if you are ok please approve.

dzmitry-lahoda (approved):

PR #1203, opened 2022-06-24
Add event TWAPUpdated, update tests.

vivekvpandya:
In liquiditybootstraping do not delete pool on remove_liquidity.

## Issue
CU-2gg4ayr

## Description
Emit TWAPUpdated event so that external entities can subscribe to it.

## Checklist

### DIFF
--- a/frame/pablo/src/lib.rs
+++ b/frame/pablo/src/lib.rs
@@ -710,16 +710,12 @@ pub mod pallet {
 				let updated_twap = TWAPState::<T>::try_mutate(
 					pool_id,
 					|prev_twap_state| -> Result<Option<TWAPStateOf<T>>, DispatchError> {
-						let res = update_twap_state::<T>(
+						update_twap_state::<T>(
 							base_price_cumulative,
 							quote_price_cumulative,
 							prev_twap_state,
-						);
-						if res.is_ok() {
-							Ok(prev_twap_state.clone())
-						} else {
-							Ok(None)
-						}
+						)
+						.map_or_else(|_| Ok(None), |_| Ok(prev_twap_state.clone()))
 					},
 				)?;
 				if let Some(updated_twap) = updated_twap {
--- a/frame/pablo/src/lib.rs
+++ b/frame/pablo/src/lib.rs
@@ -1000,6 +1000,7 @@ pub mod pallet {
 							min_quote_amount,
 						)?;
 					Self::update_twap(pool_id)?;
+					Pools::<T>::remove(pool_id);
 					Self::deposit_event(Event::PoolDeleted { pool_id, base_amount, quote_amount });
 				},
 			}
--- a/frame/pablo/src/lib.rs
+++ b/frame/pablo/src/lib.rs
@@ -218,6 +218,15 @@ pub mod pallet {
 			/// Charged fees.
 			fee: Fee<T::AssetId, T::Balance>,
 		},
+		/// TWAP updated.
+		TWAPUpdated {
+			/// Pool id on which exchange done.
+			pool_id: T::PoolId,
+			/// TWAP Timestamp
+			timestamp: MomentOf<T>,
+			/// Map of asset_id -> twap
+			twaps: BTreeMap<T::AssetId, Rate>,
+		},
 	}
 
 	#[pallet::error]
@@ -591,6 +600,18 @@ pub mod pallet {
 				);
 				if result.is_ok() {
 					weight += 1;
+					if let Some(updated_twap) = TWAPState::<T>::get(pool_id) {
+						if let Ok(currency_pair) = Self::currency_pair(pool_id) {
+							Self::deposit_event(Event::<T>::TWAPUpdated {
+								pool_id,
+								timestamp: updated_twap.timestamp,
+								twaps: BTreeMap::from([
+									(currency_pair.base, updated_twap.base_twap),
+									(currency_pair.quote, updated_twap.quote_twap),
+								]),
+							});
+						}
+					}
 				}
 			}
 			weight
@@ -674,7 +695,7 @@ pub mod pallet {
 		}
 
 		fn update_twap(pool_id: T::PoolId) -> Result<(), DispatchError> {
-			// update price cumulatives
+			let currency_pair = Self::currency_pair(pool_id)?; // update price cumulatives
 			let (base_price_cumulative, quote_price_cumulative) =
 				PriceCumulativeState::<T>::try_mutate(
 					pool_id,
@@ -686,16 +707,32 @@ pub mod pallet {
 				quote_price_cumulative != T::Balance::zero()
 			{
 				// update TWAP
-				let _ = TWAPState::<T>::try_mutate(
+				let updated_twap = TWAPState::<T>::try_mutate(
 					pool_id,
-					|prev_twap_state| -> Result<(), DispatchError> {
-						update_twap_state::<T>(
+					|prev_twap_state| -> Result<Option<TWAPStateOf<T>>, DispatchError> {
+						let res = update_twap_state::<T>(
 							base_price_cumulative,
 							quote_price_cumulative,
 							prev_twap_state,
-						)
+						);
+						if res.is_ok() {
+							Ok(prev_twap_state.clone())
+						} else {
+							Ok(None)
+						}
 					},
-				);
+				)?;
+				if let Some(updated_twap) = updated_twap {
+					Self::deposit_event(Event::<T>::TWAPUpdated {
+						pool_id,
+						timestamp: updated_twap.timestamp,
+						twaps: BTreeMap::from([
+							(currency_pair.base, updated_twap.base_twap),
+							(currency_pair.quote, updated_twap.quote_twap),
+						]),
+					});
+				}
+				return Ok(())
 			}
 			Ok(())
 		}
--- a/frame/pablo/src/lib.rs
+++ b/frame/pablo/src/lib.rs
@@ -219,7 +219,7 @@ pub mod pallet {
 			fee: Fee<T::AssetId, T::Balance>,
 		},
 		/// TWAP updated.
-		TWAPUpdated {
+		TwapUpdated {
 			/// Pool id on which exchange done.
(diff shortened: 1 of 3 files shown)


ComposableFiB0t:
Task linked: [CU-2gg4ayr Emit event from pablo for twap updates](https://app.clickup.com/t/2gg4ayr)

frame/pablo/src/lib.rs
```diff
@@ -591,6 +600,18 @@ pub mod pallet {
 				);
 				if result.is_ok() {
 					weight += 1;
+					if let Some(updated_twap) = TWAPState::<T>::get(pool_id) {
```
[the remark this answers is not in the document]
vivekvpandya: Its per pool , do you mean a separate event with Vec<TWAPUpdated> ?

frame/pablo/src/lib.rs
```diff
@@ -686,16 +707,32 @@ pub mod pallet {
...
 							base_price_cumulative,
 							quote_price_cumulative,
 							prev_twap_state,
-						)
+						);
+						if res.is_ok() {
+							Ok(prev_twap_state.clone())
+						} else {
+							Ok(None)
+						}
```
PoisonPhang, line 722: NIT: A `match` statement would be more concise here

frame/pablo/src/lib.rs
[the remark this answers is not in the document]
vivekvpandya: Not sure how this can help? Is it efficient for network layer , for runtime to emit event, or just because it is in on_initialize()? Is there any document from substrate for this kind of aggrigation? Also in update_twap() it is any way for a single pool so there I can't aggregate it. ( We can use event with a Vec of len 1) But I want to be sure that there is a real concern.

frame/pablo/src/liquidity_bootstrapping.rs
```diff
@@ -203,7 +203,6 @@ impl<T: Config> LiquidityBootstrapping<T> {
 		let base_amount = repatriate(pool.pair.base)?;
 		let quote_amount = repatriate(pool.pair.quote)?;
 
-		Pools::<T>::remove(pool_id);
```
dzmitry-lahoda, line 206: what is reason to terain pool  if there is no liquidity?

frame/pablo/src/lib.rs
```diff
@@ -686,16 +707,28 @@ pub mod pallet {
...
 						)
+						.map_or_else(|_| Ok(None), |_| Ok(prev_twap_state.clone()))
 					},
-				);
+				)?;
+				if let Some(updated_twap) = updated_twap {
+					Self::deposit_event(Event::<T>::TWAPUpdated {
+						pool_id,
+						timestamp: updated_twap.timestamp,
+						twaps: BTreeMap::from([
```
dzmitry-lahoda, line 725: you do not need tree here. building tree does hashing and burns resources. you may just use vec.

dzmitry-lahoda: because anyway it was tree and no need to prove that again

frame/pablo/src/lib.rs
```diff
@@ -591,6 +600,18 @@ pub mod pallet {
 				);
 				if result.is_ok() {
 					weight += 1;
+					if let Some(updated_twap) = TWAPState::<T>::get(pool_id) {
+						if let Ok(currency_pair) = Self::currency_pair(pool_id) {
+							Self::deposit_event(Event::<T>::TWAPUpdated {
+								pool_id,
+								timestamp: updated_twap.timestamp,
+								twaps: BTreeMap::from([
+									(currency_pair.base, updated_twap.base_twap),
```
dzmitry-lahoda, line 609: just to double check, that is possible that base_twap current == previous base_twap ?

dzmitry-lahoda: so question is if there was no trades, then should we see events - i think not.

frame/pablo/src/liquidity_bootstrapping.rs
vivekvpandya: we may want some basic details about pool. like in this PR, remove_liquidity() triggers TWAPUpdated event and for that I need pool's asset ids. if I allow pool to be deleted here I need to emit event from this specialized part of remove_liquidity() call. 
IMO we it would be better to have a separate extrinsic to delete pool.

frame/pablo/src/lib.rs
vivekvpandya: To support more than two asset I am advised to use tree with asset_id -> balance mapping instead of using vec where we need to assume which balance for which asset. 
For performance I think we can benchmark and see how much difference is there due to hashing and tree creation.

frame/pablo/src/lib.rs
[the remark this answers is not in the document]
vimukthi-git: I don't have a source for substrate other than general understanding of event based systems and batching, when you have a single event,
1. The network overhead is lower as there are fewer packets to be received and encoding/decoding load is lower
2. Any storage system only needs to store one set of event meta data instead of many, so efficient to store

But there could be advantages to not batching as well
1. Receivers might be able to process the msgs in a parallel/concurrent way
2. Receivers with lower resources can manage smaller events easily

I just saw a recent thread about this here https://substrate.stackexchange.com/questions/3455/emit-events-from-hooks-like-on-initialize, may be we can upvote/add detail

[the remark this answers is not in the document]
vivekvpandya: :sweat_smile: I created that thread in hope to get some insights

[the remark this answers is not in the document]
vimukthi-git: Cool, I didn't know that you are a "llvm ninja" :laughing:

frame/pablo/src/liquidity_bootstrapping.rs
dzmitry-lahoda: i have seen that one extinsic can do 2 events. its ok.

dzmitry-lahoda: may be ok if some ED is retained on pool account

vivekvpandya: I will more Pools::<T>::remove() call to generic remove_liquidity() call where LBP is handled.

frame/pablo/src/lib.rs
vivekvpandya: IIRC there is no hash map available in substrate std.

frame/pablo/src/lib.rs
```diff
@@ -218,6 +218,15 @@ pub mod pallet {
 			/// Charged fees.
 			fee: Fee<T::AssetId, T::Balance>,
 		},
+		/// TWAP updated.
+		TWAPUpdated {
```
benluelo, line 222: NIT: Name should be `TwapUpdated`

frame/pablo/src/lib.rs
benluelo: Correct: https://docs.rs/sp-std/latest/sp_std/collections/index.html

dzmitry-lahoda: insert into btree is O(logN) insert into vec is O(1). why use tree and making log N in hook? this is attach vector.

frame/pablo/src/uniswap_tests.rs
```diff
@@ -779,7 +779,7 @@ mod twap {
 			let unit = 1_000_000_000_000_u128;
 			let initial_btc = 100_u128 * unit;
 			let initial_usdt = 100_u128 * unit;
-			let pool_id =
+			let pool_identifier =
```
vimukthi-git, line 782: why the rename?

frame/pablo/src/lib.rs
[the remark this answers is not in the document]
vimukthi-git: Let's keep an eye on the stackexchange thread and change if there is a authoritative answer for this

vimukthi-git (dismissed):
LGTM

frame/pablo/src/uniswap_tests.rs
vivekvpandya: it was conflicting name for newly added assert at line 849

frame/pablo/src/lib.rs
vimukthi-git: As this is basically an internal operation and the number of TWAP enabled pools are limited, the hook is already O(n) so additional `O(logN)` is still O(n). If there is a way to get rid of O(n) we should definitely do it though like my suggestion on the other thread for event emitting

dzmitry-lahoda:
you may link SE links in code if code needs such comment

frame/pablo/src/lib.rs
dzmitry-lahoda: uniseap had 1000 plus pools, there contracta with 3000 pools.

dzmitry-lahoda: i guess comlosable want the same.

dzmitry-lahoda: i do bot get your reasoning

vimukthi-git: TWAP is only for certain pools we enable it for. So all pools won't have twap enabled

vimukthi-git: But adding a comment about these technical considerations in the code makes sense btw

PoisonPhang (approved):
LGTM

frame/pablo/src/lib.rs
dzmitry-lahoda: until hash is consumed in memory it useles. so now we are doing useless job on chain. instead of doing simple vec. i do not get that logic.

dzmitry-lahoda: what comments? when there is not map user there is not need for map.

vivekvpandya: Map will be used by code which processes event, to identify which is base_amount and which is quote_amount.
Also using map vs vec is yet to be finalized with https://app.clickup.com/t/2gh7cjb
so perhaps after that we may not use map and use vec all the places.

PR #1206, opened 2022-06-24
Removed unused patches

PoisonPhang:
## Issue
N/A

## Description
For a while now, our unused patches have been breaking the functionality of RA. This will correct it for the meantime, but we will need to continue to evaluate which patches we have in use in the future


benluelo (commented):
A comment at the top of the patch explaining why some patches are commented out would be good, otherwise lgtm

dzmitry-lahoda (dismissed):

KaiserKarel (changes requested):
Could we, instead of commenting, plainly remove them?

KaiserKarel (approved):

dzmitry-lahoda (approved):

PR #1211, opened 2022-06-26
CU-225wvg7 - Update funding guard rails

0xangelo:
## Issue
[CU-225wvg7](https://app.clickup.com/t/225wvg7)

## Description
This PR completes the subtasks of the main ClickUp one, linked above, related to additional validity checks. It includes:
- chore(CH): update TWAPs before funding rate update
- chore(CH): add update_funding price spread guards
- test(CH): adapt to new TWAP update rules
- test(CH): expect TWAP updates in update_funding
- test(CH): try liquidate with unrealized funding
- fix(CH): use proper upper bound when clipping oracle
- fix(CH): set TWAP to current price if needed when updating Oracle TWAP
- fix(CH): clip current oracle price to withing 33% of last TWAP
- refactor(CH): group helpers by main functionality

### DIFF
--- a/frame/clearing-house/src/lib.rs
+++ b/frame/clearing-house/src/lib.rs
@@ -2317,12 +2317,12 @@ pub mod pallet {
 			twap: &T::Decimal,
 		) -> Result<T::Decimal, DispatchError> {
 			let oracle_twap_spread = oracle.try_sub(twap)?;
-			let oracle_33pct = oracle.try_div(&T::Decimal::saturating_from_integer(3))?;
-			Ok(if oracle_twap_spread.saturating_abs() > oracle_33pct {
+			let twap_33pct = twap.try_div(&T::Decimal::saturating_from_integer(3))?;
+			Ok(if oracle_twap_spread.saturating_abs() > twap_33pct {
 				if oracle > twap {
-					twap.try_add(&oracle_33pct)?
+					twap.try_add(&twap_33pct)?
 				} else {
-					twap.try_sub(&oracle_33pct)?
+					twap.try_sub(&twap_33pct)?
 				}
 			} else {
 				*oracle
--- a/frame/clearing-house/src/lib.rs
+++ b/frame/clearing-house/src/lib.rs
@@ -2265,7 +2265,7 @@ pub mod pallet {
 				oracle = math::clip(
 					oracle,
 					last_oracle.try_sub(&last_oracle_10bp)?,
-					last_oracle.try_sub(&last_oracle_10bp)?,
+					last_oracle.try_add(&last_oracle_10bp)?,
 				);
 
 				// TODO(0xangelo): consider further guard rails
--- a/frame/clearing-house/src/lib.rs
+++ b/frame/clearing-house/src/lib.rs
@@ -1247,6 +1247,10 @@ pub mod pallet {
 				Error::<T>::UpdatingFundingTooEarly
 			);
 
+			// Update TWAPs *before* funding rate calculations
+			Self::update_oracle_twap(&mut market)?;
+			T::Vamm::update_twap(market.vamm_id, None, None)?;
+
 			Self::do_update_funding(market_id, &mut market, now)?;
 
 			Markets::<T>::insert(market_id, market);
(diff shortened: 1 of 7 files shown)


ComposableFiB0t:
Task linked: [CU-225wvg7 Feature 5 (Update Market Funding Functionality)](https://app.clickup.com/t/225wvg7)

frame/clearing-house/src/lib.rs
```diff
@@ -2241,10 +2285,18 @@ pub mod pallet {
 	// Oracle update helpers
 	impl<T: Config> Pallet<T> {
 		fn update_oracle_twap(market: &mut Market<T>) -> Result<(), DispatchError> {
-			let mut oracle = Self::clip_around_twap(
-				&Self::oracle_price(market.asset_id)?,
-				&market.last_oracle_twap,
-			)?;
+			let mut oracle = Self::oracle_price(market.asset_id)?;
+			ensure!(oracle.is_positive(), Error::<T>::InvalidOracleReading);
```
Cardosaum, line 2289: In line 508 the definition of `InvalidOracleReading` is: *Raised when the price returned by the Oracle is zero.*,
However, at least theoretically there could be a case where the oracle value is negative since `Decimal` is signed, right?

In that case wouldn't be better to define this error alluding to the price being strictly positive?

frame/clearing-house/src/tests/update_funding.rs
```diff
@@ -12,17 +7,25 @@ use crate::{
...
 use composable_traits::{
 	clearing_house::{ClearingHouse, Instruments},
 	time::{DurationSeconds, ONE_HOUR},
+	vamm::{AssetType, Vamm},
 };
 use frame_support::{assert_noop, assert_ok};
 use proptest::prelude::*;
 use sp_runtime::{FixedI128, FixedU128};
 
+use super::as_balance;
```
Cardosaum, line 27: Why not include `as_balance` in the `use crate::tests{...}` block?

(I think maybe what happened was that your IDE imported it automatically and placed the import here.)

frame/clearing-house/src/tests/update_funding.rs
```diff
@@ -38,6 +41,39 @@ fn get_position(account: &AccountId, market_id: &MarketId) -> Position {
...
+		VammPallet::set_next_twap_of(&vamm_id, Some(1.into()));
+		assert_ok!(<TestPallet as ClearingHouse>::update_funding(&market_id));
+
+		let market = get_market(&market_id);
+		// First update since market creation, so TWAP should be the same as current price
+		assert_eq!(market.last_oracle_price, 1.into());
+		assert_eq!(market.last_oracle_twap, market.last_oracle_price);
+		assert_eq!(market.last_oracle_ts, ONE_HOUR);
+		// Assert `Vamm::update_twap` was called
+		assert_eq!(VammPallet::get_twap(vamm_id, AssetType::Base).unwrap(), 1.into());
```
Cardosaum, line 73: The current implementation of `get_twap` for the Vamm simply return the twap stored in the runtime, without worrying to update it, which would require a separate call to `update_twap`. (twap updates happen automatically during swaps though). 

Would it be desirable to try to update the twap every time a call to `get_twap` happens?

Cardosaum (approved):
Great work! 

Just minor comments, but awesome job with this PR! :smile:

frame/clearing-house/src/lib.rs
0xangelo: Looking at the `Oracle` trait, which is where the Clearing House gets its prices from, `get_price` is returned as a `Balance`, thus it is nonnegative.

But at a higher level, all prices should be positive indeed. The error of the price being zero is just one instance of the oracle reading being invalid. So I agree with your suggestion

0xangelo: Specially because we may switch to an Oracle which uses signed quantities in the futures and still use the same error

frame/clearing-house/src/tests/update_funding.rs
0xangelo: Yeah, that's correct. Good catch

frame/clearing-house/src/tests/update_funding.rs
0xangelo: I'm testing that `update_funding` calls `update_twap` on the Vamm internally. If you see how I implemented `set_next_twap_of` in the mock Vamm, it's supposed to be followed by a call to `update_twap`.

Side comment: `update_funding` is the only place I call `update_twap` on the Vamm explicitly. For `open_position` and `close_position`, I count on the Vamm doing that automatically before each swap

Cardosaum (approved):

kevin-composable (approved):
I think this is the smallest PR I have seen by you - nice work as always

PR #1193, opened 2022-06-22
CU-2nvggn0 - Move math helpers to shared labs directory

Cardosaum:
## Issue
[CU-2nvggn0](https://app.clickup.com/t/2nvggn0)

## Description
This PR moves math helpers functions to a common `labs` directory, improving code reuse across Composable Labs pallets.

PR #1212, opened 2022-06-27
Escher bugfix/vamm twap reciprocal

Cardosaum:
## Issue
[CU-2pr8rzq](https://app.clickup.com/t/2pr8rzq)

## Description
Fix bug of twaps not being reciprocals of each other.

- Fix trait implementation
- Add tests ensuring twaps are always reciprocal

### DIFF
--- a/frame/clearing-house/src/lib.rs
+++ b/frame/clearing-house/src/lib.rs
@@ -135,8 +135,6 @@ mod mock;
 #[cfg(test)]
 mod tests;
 
-mod math;
-
 mod types;
 
 mod weights;
@@ -155,13 +153,13 @@ pub mod pallet {
 		Market, MarketConfig, Position,
 	};
 	use crate::{
-		math::{
-			self, FixedPointMath, FromBalance, IntoBalance, IntoDecimal, IntoSigned, UnsignedMath,
-		},
 		types::{AccountSummary, PositionInfo, BASIS_POINT_DENOMINATOR},
 		weights::WeightInfo,
 	};
 	use codec::FullCodec;
+	use composable_maths::labs::numbers::{
+		self, FixedPointMath, FromBalance, IntoBalance, IntoDecimal, IntoSigned, UnsignedMath,
+	};
 	use composable_traits::{
 		clearing_house::{ClearingHouse, Instruments},
 		defi::DeFiComposableConfig,
@@ -2150,8 +2148,7 @@ pub mod pallet {
 				};
 				let last_oracle_10bp =
 					last_oracle.try_div(&T::Decimal::saturating_from_integer(1000))?;
-				oracle = math::clip(
-					oracle,
+				oracle = oracle.clamp(
 					last_oracle.try_sub(&last_oracle_10bp)?,
 					last_oracle.try_sub(&last_oracle_10bp)?,
 				);
@@ -2165,7 +2162,7 @@ pub mod pallet {
 				let since_last = cmp::max(1, now.saturating_sub(market.last_oracle_ts));
 				let from_start = cmp::max(1, market.twap_period.saturating_sub(since_last));
 
-				let new_twap = math::weighted_average(
+				let new_twap = numbers::weighted_average(
 					&oracle,
 					&market.last_oracle_twap,
 					&T::Decimal::saturating_from_integer(since_last),
--- a/frame/composable-maths/src/labs/numbers.rs
+++ b/frame/composable-maths/src/labs/numbers.rs
@@ -7,16 +7,10 @@ use sp_runtime::{
 	ArithmeticError::{DivisionByZero, Overflow, Underflow},
 	FixedPointNumber, FixedPointOperand,
 };
-use sp_std::cmp;
 
 // -------------------------------------------------------------------------------------------------
 //                                          Functions
 // -------------------------------------------------------------------------------------------------
-
-/// Clips `value` between `lower` and `upper` (inclusive).
-pub fn clip<T: FixedPointNumber>(value: T, lower: T, upper: T) -> T {
-	cmp::min(cmp::max(lower, value), upper)
-}
 
 /// Computes a weighted average as `(a * wa + b * wb) / (wa + wb)`.
 pub fn weighted_average<T>(a: &T, b: &T, weight_a: &T, weight_b: &T) -> Result<T, ArithmeticError>
--- a/frame/vamm/src/lib.rs
+++ b/frame/vamm/src/lib.rs
@@ -742,9 +742,8 @@ pub mod pallet {
 
 			// Delegate update twap to internal functions.
 			let (base_twap, quote_twap) = match (base_twap, quote_twap) {
-				(Some(base_twap), Some(quote_twap)) => {
-					Self::do_update_twap(vamm_id, &mut vamm_state, base_twap, quote_twap, &None)?
-				},
+				(Some(base_twap), Some(quote_twap)) =>
+					Self::do_update_twap(vamm_id, &mut vamm_state, base_twap, quote_twap, &None)?,
 				_ => Self::update_vamm_twap(vamm_id, &mut vamm_state, &None)?,
 			};
 
@@ -1106,9 +1105,8 @@ pub mod pallet {
 		) -> Result<BalanceOf<T>, DispatchError> {
 			let quote_asset_reserve_change = match direction {
 				Direction::Add => quote_asset_reserve_before.try_sub(quote_asset_reserve_after)?,
-				Direction::Remove => {
-					quote_asset_reserve_after.try_sub(quote_asset_reserve_before)?
-				},
+				Direction::Remove =>
+					quote_asset_reserve_after.try_sub(quote_asset_reserve_before)?,
 			};
 
 			let quote_asset_amount =
--- a/frame/vamm/src/lib.rs
+++ b/frame/vamm/src/lib.rs
@@ -591,6 +591,7 @@ pub mod pallet {
 		///
 		/// # Runtime
 		/// `O(1)`
+		#[transactional]
 		fn get_price(
 			vamm_id: VammIdOf<T>,
 			asset_type: AssetType,
@@ -729,6 +730,7 @@ pub mod pallet {
 		///
 		/// # Runtime
 		/// `O(1)`
+		#[transactional]
 		fn update_twap(
 			vamm_id: VammIdOf<T>,
 			base_twap: Option<DecimalOf<T>>,
--- a/frame/vamm/src/lib.rs
+++ b/frame/vamm/src/lib.rs
@@ -129,7 +129,8 @@ pub mod pallet {
 	// ----------------------------------------------------------------------------------------------------
 
 	use codec::{Codec, FullCodec};
-	use composable_maths::labs::numbers::{IntoDecimal, UnsignedMath};
+	use composable_maths::labs::numbers::{FixedPointMath, IntoDecimal, UnsignedMath};
+	use composable_tests_helpers::test::helper::default_acceptable_computation_error;
 	use composable_traits::vamm::{
 		AssetType, Direction, MovePriceConfig, SwapConfig, SwapOutput, SwapSimulationConfig, Vamm,
 		VammConfig, MINIMUM_TWAP_PERIOD,
@@ -196,6 +197,8 @@ pub mod pallet {
 			+ Codec
 			+ Copy
 			+ From<u64>
+			+ From<u128>
+			+ Into<u128>
 			+ MaxEncodedLen
 			+ MaybeSerializeDeserialize
 			+ Ord
@@ -206,6 +209,8 @@ pub mod pallet {
 		/// Signed decimal fixed point number.
 		type Decimal: Default
 			+ FixedPointNumber<Inner = Self::Balance>
+			+ One
+			+ Zero
 			+ FullCodec
 			+ MaxEncodedLen
 			+ MaybeSerializeDeserialize
@@ -400,6 +405,9 @@ pub mod pallet {
 		/// Tried to update twap for an asset, but the desired new twap value is
 		/// zero.
 		NewTwapValueIsZero,
+		/// Tried to update twaps with values that are not reciprocal of one
+		/// another.
+		TwapsMustBeReciprocals,
 		/// Tried to create a vamm with a
 		/// [`twap_period`](VammState::twap_period) smaller than the
 		/// minimum allowed one specified by
@@ -525,7 +533,7 @@ pub mod pallet {
 				..Default::default()
 			};
 			let base_twap = Self::do_get_price(&tmp_vamm_state, AssetType::Base)?;
-			let quote_twap = Self::do_get_price(&tmp_vamm_state, AssetType::Quote)?;
+			let quote_twap = Self::reciprocal_twap(&base_twap)?;
 
 			VammCounter::<T>::try_mutate(|next_id| {
 				let id = *next_id;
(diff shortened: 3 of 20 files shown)


ComposableFiB0t:
Task linked: [CU-2pr8rzq Fix Update TWAP reciprocal values](https://app.clickup.com/t/2pr8rzq)

frame/vamm/src/tests/update_twap.rs
```diff
@@ -98,20 +121,34 @@ fn update_twap_fails_if_new_twap_is_zero() {
...
+	let new_twap_reciprocal = Some(TestPallet::reciprocal_twap(&new_twap.unwrap()).unwrap());
 	ExtBuilder { vamm_count: 1, vamms: vec![(0, vamm_state)] }
 		.build()
 		.execute_with(|| {
 			// For event emission
 			run_for_seconds(timestamp);
-
 			assert_noop!(
-				TestPallet::update_twap(0, new_twap, new_twap),
+				TestPallet::update_twap(0, new_twap, new_twap_reciprocal),
```
0xangelo, line 151: I think `base_twap` and `quote_twap` would be more descriptive here

frame/vamm/src/tests/update_twap.rs
```diff
@@ -147,22 +184,23 @@ fn update_twap_succeeds() {
...
-			run_for_seconds(1);
+			run_for_seconds(twap_update_delay(0));
 			assert_ok!(TestPallet::update_twap(0, None, new_twap));
 
-			run_for_seconds(1);
+			run_for_seconds(twap_update_delay(0));
 			assert_ok!(TestPallet::update_twap(0, new_twap, None));
 
-			run_for_seconds(1);
+			run_for_seconds(twap_update_delay(0));
```
0xangelo, line 203: It seems that these calls are advancing time to
```now + twap_update_delay = now + (vamm_state.twap_timestamp + vamm_state.twap_period + 1)```
since `run_for_seconds` increments the **current** timestamp by its argument. 

Are you sure you didn't mean to do only the following?
```now + vamm_state.twap_period + 1```
That would be `run_for_seconds(vamm_state.twap_period + 1)`

Or you could use a different function like `run_to_seconds(...)` that advances time **to** its argument, which I think is what you intended here.

File: frame/vamm/src/tests/mod.rs [hunk 3]
```diff
@@ -110,6 +118,14 @@ fn create_vamm(vamm_config: &VammConfig<Balance, Moment>) {
 	assert_ok!(TestPallet::create(vamm_config));
 }
 
+fn twap_update_delay(vamm_id: VammId) -> Moment {
+	let vamm_state = TestPallet::get_vamm(vamm_id).unwrap();
+	vamm_state
+		.twap_period
+		.saturating_add(vamm_state.twap_timestamp)
+		.saturating_add(1)
+}
```
0xangelo, line 127: Curious to know the reason behind the + 1 here

frame/vamm/src/tests/update_twap.rs
```diff
@@ -223,26 +262,62 @@ proptest! {
...
+				assert_ok!(
+					TestPallet::update_twap(0, Some(twap), None)
+				);
+
+				run_for_seconds(delay);
+				assert_ok!(
+					TestPallet::update_twap(0, None, Some(twap))
+				);
+
+				run_for_seconds(delay);
```
0xangelo, line 306: I think these may suffer from the same issue pointed out in the other comment

frame/vamm/src/tests/update_twap.rs
```diff
@@ -223,26 +262,62 @@ proptest! {
...
+	#[test]
+	fn should_succeed_computing_correct_reciprocal_twap_proptest(
+		twap in balance_range(),
+	) {
+		let twap_reciprocal: Decimal = Decimal::one().div(twap.into());
+		assert_ok!(
+			TestPallet::reciprocal_twap(&twap.into()),
+			twap_reciprocal
+		);
+	}
```
0xangelo, line 322: `FixedPointNumber` already has a method for computing `reciprocal`
https://paritytech.github.io/substrate/master/sp_arithmetic/fixed_point/trait.FixedPointNumber.html#method.reciprocal

frame/vamm/src/lib.rs
```diff
@@ -737,14 +748,21 @@ pub mod pallet {
...
+
+			// Handle optional values.
 			let (base_twap, quote_twap) = match (base_twap, quote_twap) {
-				(Some(base_twap), Some(quote_twap)) =>
-					Self::do_update_twap(vamm_id, &mut vamm_state, base_twap, quote_twap, &None)?,
-				_ => Self::update_vamm_twap(vamm_id, &mut vamm_state, &None)?,
+				(Some(base_twap), Some(quote_twap)) => (base_twap, quote_twap),
+				(Some(base_twap), None) => (base_twap, Self::reciprocal_twap(&base_twap)?),
+				(None, Some(quote_twap)) => (Self::reciprocal_twap(&quote_twap)?, quote_twap),
+				(None, None) => Self::compute_new_twaps(&vamm_state, &None)?,
```
0xangelo, line 759: Due to the complexity of input handling and given the fact that `FixedPointNumber` already has a convenient method to compute `reciprocal`s, I would consider restricting the function signature to take in only the `base_twap`.
This would simplify the input parsing and wouldn't burden the caller too much, since it may use `.reciprocal()` if it has the quote TWAP in hand before passing it into this function.

0xangelo: With this, you could also have one less error checking, since `TwapsMustBeReciprocals` wouldn't be needed with this change

frame/vamm/src/lib.rs
```diff
@@ -1143,15 +1214,29 @@ pub mod pallet {
...
-			// 3) Only update asset's twap if time has passed since last update.
+			// 4) Only update asset's twap if time has passed since last update
+			// and if the delay is greater than the required time between
+			// updates.
 			let now = Self::now(now);
-			ensure!(now > vamm_state.twap_timestamp, Error::<T>::AssetTwapTimestampIsMoreRecent);
+			ensure!(
+				now > vamm_state.twap_timestamp.try_add(&vamm_state.twap_period)?,
+				Error::<T>::AssetTwapTimestampIsMoreRecent
+			);
```
0xangelo, line 1235: Why would at least `twap_period` have to pass between updates? Wouldn't `weight_now` in line 1031 always be `twap_period` and `weight_last_twap` in line 1032 always be `1` if we force that?

For reference, this is how I use `twap_period` for the Oracle in the Clearing House
```
/// The reference time span used for weighting the EMA updates for the Oracle and Vamm TWAPs.
/// ```text
///                                                       twap_period
///                                        |---------------------------------------|
///                                         from_start          since_last
///                                        |-----------|---------------------------|
/// -------------------------------------------------------------------------------|
///                                        ^           ^                           ^
///                                      now -         |                          now
///                                   twap_period      |
///                                               last_twap_ts
/// ```
/// In the example above, the current price is given a weight of `since_last` and the last
/// TWAP, `from_start`. The new TWAP is then the weighted average of the two.
```
Where `since_last` and `from_start` would be equivalent to `weight_now` and `weight_last_timestamp` respectively

frame/vamm/src/tests/update_twap.rs
Cardosaum: Oh, okay :+1: 
I'll change that, thanks!

frame/vamm/src/tests/update_twap.rs
Cardosaum: Oh, thanks for pointing out!
Yeah I was using it like I use `run_to_block`. 
Fixed it :+1:

frame/vamm/src/tests/mod.rs
Cardosaum: For the caller to be able to update the twap successfully, it has to call it strictly after `twap_period`. As I'm using this function as a helper to jump to the next available time to update twaps in tests, I simply added this `+ 1` so I don't need to do so in the tests itself.

frame/vamm/src/tests/update_twap.rs
Cardosaum: yeah, I changed it too. 
Thanks for pointing out! :)

frame/vamm/src/tests/update_twap.rs
Cardosaum: Oh, I think I saw you using it in the CH some PRs ago but I totally forgot it! :see_no_evil: 
Thanks so much, that really helps!

frame/vamm/src/lib.rs
Cardosaum: Oh, yeah changing the function to take only `base_twap` really helps and simplifies a **lot**!
Thanks for this suggestion :)

I'm more prone to keep `TwapsMustBeReciprocals` though, because in the function `do_update_twap` we take both `base_twap` and `quote_twap`. There I think it's still important to have sanity checks before updating the runtime storage, as we don't want to modify the runtime with incorrect values by any means. (it would be just a matter of more caution actually, since these errors should not occur at all due to the way we're using the functions in the pallet.)

frame/vamm/src/lib.rs
Cardosaum: Oh, you're **totally** right! 
I confused it while doing other changes. 
Already reverted it to the way it was before, thanks!

frame/vamm/src/lib.rs
```diff
@@ -665,17 +661,18 @@ pub mod pallet {
 
 			match asset_type {
 				AssetType::Base => Ok(vamm_state.base_asset_twap),
-				AssetType::Quote => Ok(vamm_state.quote_asset_twap),
+				AssetType::Quote => Ok(Self::reciprocal_twap(&vamm_state.base_asset_twap)?),
```
0xangelo, line 664: could use `.reciprocal()` here

frame/composable-maths/src/labs/numbers.rs
```diff
@@ -57,20 +57,20 @@ pub trait UnsignedMath: CheckedAdd + CheckedDiv + CheckedMul + CheckedSub + Unsi
 }
 
 pub trait FixedPointMath: FixedPointNumber {
-	/// Like [`FixedPointNumber::checked_add`], but returning a [`Result`] with [`ArithmeticError`]
+	/// Like [`sp_runtime::traits::CheckedAdd`], but returning a [`Result`] with [`ArithmeticError`]
```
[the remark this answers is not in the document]
Cardosaum: It would be nice, indeed! :+1:

frame/vamm/src/lib.rs
0xangelo: Ah, I see. That makes sense. I would suggest creating a `TryReciprocal` trait that implements a `try_reciprocal` method returning the appropriate arithmetic error upon failure. What do you think?

frame/composable-maths/src/labs/numbers.rs
```diff
@@ -201,7 +205,7 @@ impl<T: FixedPointNumber> FixedPointMath for T {
 		//       1           1 | Overflow
 		//    -1/1           0 | DivisionByZero
 		if other.into_inner().is_zero() {
-			return Err(DivisionByZero)
+			return Err(DivisionByZero);
```
0xangelo, line 208: Curious to know if these were about a clippy warning or a compiler complaint

frame/composable-traits/src/assets.rs
```diff
@@ -23,7 +23,7 @@ impl BasicAssetMetadata {
 	}
 }
 
-#[derive(Decode, Encode, Debug, Clone, PartialEq)]
+#[derive(Decode, Encode, Debug, Clone, PartialEq, Eq)]
```
0xangelo, line 26: Curious to know if these were auto-fixes by `clippy` or manual additions by you

0xangelo (commented):
Awesome! You can ping me for an official approval after Kevin's approval

frame/composable-traits/src/assets.rs
Cardosaum: Auto-fixes when running `clippy --fix`.

I was curious about why adding this derive too, here is the [explanation that clippy gives](https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq)

frame/composable-maths/src/labs/numbers.rs
Cardosaum: actually it was added automatically, but I don't remember if it was due to `cargo fmt` or `cargo clippy`.

frame/clearing-house/src/lib.rs
```diff
@@ -2308,8 +2306,7 @@ pub mod pallet {
 				};
 				let last_oracle_10bp =
 					last_oracle.try_div(&T::Decimal::saturating_from_integer(1000))?;
-				oracle = math::clip(
-					oracle,
+				oracle = oracle.clamp(
```
kevin-composable, line 2309: same comment from PR #1193: I think `oracle_price` is more readable than `oracle`

kevin-composable (approved):
Everything looks fantastic! I am going to approve this one since my comment still stands on the other PR

frame/clearing-house/src/lib.rs
0xangelo: That's on me. I'll change it in the Clearing House pallet when I open a new PR

Cardosaum: @kevin-composable, regarding your question about where the function `clamp` is declared, [it is in the standard library, to every type that defines `Ord`.](https://doc.rust-lang.org/nightly/core/cmp/trait.Ord.html#method.clamp)

Cardosaum: I removed `math::clip` and opted to just use `clamp` because in that way we don't duplicate existing code. There are two points to consider, though: 
- In the latest version of this code block that is in `escher` one typo/bug is fixed (using `.try_sub()` for both parameters)
- This function may panic if `min > max`, so it's important to ensure that it would never happen. (currently not checked)

I can make this change in #1193 if you want, but I think it would be more suitable to @0xangelo in his next PR.

The reasoning for me to change the method in this code block is because I was moving some helper functions and removing unneeded code.

0xangelo: I already renamed `oracle` to `oracle_price` in my local branch, so you don't need to worry about that

Cardosaum: Just need to keep in mind that using `clamp` rather than `math::clip` might panic. 
Would you prefer to re-add the `math::clip` helper? 
Maybe adding it as a trait method rather than a helper function would be better. I can do it in a separate PR if you wish.
cc/ @0xangelo

0xangelo: I'll add a `TryClamp` trait in my next PR, sounds good?

0xangelo: That is, after you merge the math helpers PR #1193

0xangelo (approved):

Cardosaum:
bors try

Cardosaum:
bors try

frame/clearing-house/src/lib.rs
0xangelo: Done in #1284

kevin-composable (approved):

Cardosaum:
bors try

kevin-composable (approved):

0xangelo:
Once you fix the conflicts @Cardosaum , ping me for an approval

Cardosaum:
bors try

kevin-composable (approved):

0xangelo (approved):

PR #1216, opened 2022-06-27
Added benchmarks for liquidations pallet

mikolaichuk:
## Issue
https://app.clickup.com/t/32b5019

## Description
Halborn report:

> The liquidations pallet defines a sell function responsible for creating a Dutch Auction for Liquidating a position. The weight associated with calling this extrinsic is a constant value of 10000. Such configuration is invalid, as the sell function accepts a Vector as a parameter, which length is controlled by the user. Two issues originate from this fact. The first one is that regardless of the Vector size, the extrinsic will have a constant weight and would always be perceived as filling a constant space in the block. The second one is related to fees associated with calling the extrinsic. Currently, the sell function is internally creating a Dutch Auction. Anyone can call the sell auction, similarly to the fact that anyone can create their own Dutch Auction via the dutch-auction pallet. There is a correlation between weights and fees. With the current weight configuration for the sell function, it is cheaper to create a Dutch Auction via the liquidations pallet rather than via the dutch-auction itself. The prerequisite for this is that there is already a desired configuration defined in the liquidations pallet.

## Checklist

### DIFF
--- a/frame/dutch-auction/src/weights.rs
+++ b/frame/dutch-auction/src/weights.rs
@@ -19,7 +19,7 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	// Storage: Tokens Accounts (r:1 w:1)
 	// Storage: DutchAuction SellOrders (r:0 w:1)
 	fn ask() -> Weight {
-		(185_212_000 as Weight)
+		(19_571_000 as Weight)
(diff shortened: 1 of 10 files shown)


PoisonPhang (approved):
LGTM!

kevin-composable (approved):

PR #1220, opened 2022-06-28
[CU-2ecu5r4] Integrate bonded-finance total purchased bonds into subsquid

andresmechali:
## Issue
https://app.clickup.com/t/2ecu5r4

## Description
Keeps count of total purchased bonds

## Checklist

- ~~[ ] I have updated the cargo docs to reflect changes made by this PR _(if applicable)_~~
- ~~[ ] I have updated the `book/` to reflect changes made by this PR _(if applicable)_~~


ComposableFiB0t:
Task linked: [CU-2ecu5r4 Subsquid: integrated subsquid into bonded finance pallet](https://app.clickup.com/t/2ecu5r4)

KaiserKarel:
Probably needs a rebase @andresmechali

andresmechali:
> Probably needs a rebase @andresmechali

It was missing the rebase, but the linting errors are happening on other PRs as well so it might be something broken on main. I'll look into it

subsquid/src/pabloProcessor.ts
```diff
@@ -59,8 +59,8 @@ interface PoolCreatedEvent {
 }
 
 function getPoolCreatedEvent(event: PabloPoolCreatedEvent): PoolCreatedEvent {
-    if (event.isV2100) {
-        const {owner, poolId, assets} = event.asV2100;
```
KaiserKarel, line 63: Do we need to remove the old parsing logic?

andresmechali: @saadjhk what do you think about this?

subsquid/schema.graphql
```diff
@@ -66,4 +66,11 @@ type PabloPoolAsset @entity {
   blockNumber: BigInt!
   "Unix timestamp in ms"
   calculatedTimestamp: BigInt!
-}
\ No newline at end of file
+}
+
+type BondedFinanceBondOffer @entity {
+  "Total amount of purchased bonds"
```
vimukthi-git, line 72: Need to have an `ID!` field for this entity

subsquid/src/bondedFinanceProcessor.ts
```diff
@@ -0,0 +1,103 @@
...
+function getBeneficiary(ctx: EventHandlerContext): string {
+  const { beneficiary } = ctx.event.extrinsic?.args[0]?.value as Record<
+    string,
+    unknown
+  >;
+
+  return beneficiary as string;
+}
+
+export async function processNewOfferEvent(
```
vimukthi-git, line 55: Why do we need to process this event?

subsquid/test/bondedFinanceTest.ts
```diff
@@ -0,0 +1,256 @@
...
+  totalPurchasedStored: Record<string, BondedFinanceBondOffer>,
+  offerId: string,
+  purchased: bigint
+) {
+  const { event } = createNewBondEvent(offerId, purchased);
+
+  // Update the total purchased bonds
+  totalPurchased[offerId].purchased += purchased;
+
+  await processNewBondEvent(ctx, event);
```
vimukthi-git, line 141: I would move things above this line to the actual test, instead of this function as these are not assertions.

vimukthi-git (commented):
Looks good, some minor comments and a question though!

subsquid/schema.graphql
andresmechali: It is autogenerated. On the migration files (`/db/migrations`) you can see the following:
```
await db.query(`CREATE TABLE "bonded_finance_bond_offer" ("id" character varying NOT NULL, "purchased" numeric NOT NULL, "asset_id" numeric NOT NULL, "beneficiary" text NOT NULL, CONSTRAINT "PK_1a7a97e3d57a4ac842dc2ef48ba" PRIMARY KEY ("id"))`)
```

subsquid/test/bondedFinanceTest.ts
andresmechali: done 👍🏼

subsquid/schema.graphql
vimukthi-git: What is inserted as the ID when inserting data then?

subsquid/src/bondedFinanceProcessor.ts
vimukthi-git: Then I propose having another entity to track the bond offers without the attachment to beneficiary

andresmechali: We can do that but would there be any benefit of having the beneficiaries on a separate entity, given that the beneficiary does not change over time?

subsquid/schema.graphql
andresmechali: The `offerId`, which is unique.
```
new BondedFinanceBondOffer({
  id: offerId.toString(),
  purchased: BigInt(0),
  beneficiary,
})
```

Do you prefer a uuid instead?

subsquid/src/bondedFinanceProcessor.ts
vimukthi-git: Seems with this task we want to track both bond offers and bonds, which have a one to many relationship. With current structure it is not tracked correctly

subsquid/schema.graphql
andresmechali: Maybe I'm confused but isn't the beneficiary the account that receives the assets?
```
pub struct BondOffer<AccountId, AssetId, Balance, BlockNumber> {
	/// The account that will receive the locked assets.
	pub beneficiary: AccountId,
	...
}
```

vimukthi-git: If the need is only to track the beneficiary we need to make a change to the `NewOffer` event so that it contains the beneficiary field. Then we don't have a need to process `NewBond` event

andresmechali: The need is to be able to query the total amount of purchased bonds for a given `offerId`, with `beneficiary` being a nice to have additional information

subsquid/src/bondedFinanceProcessor.ts
andresmechali: @easteregg

subsquid/schema.graphql
vimukthi-git: We need to make sure that the entities here follow more or less the same relationships as the runtime so that there won't be a costly data migration every time FE requirements change.

subsquid/src/bondedFinanceProcessor.ts
easteregg: Individual bonds are going to play a role once VestingSchedules are trackable. 

Ultimately we would want to get access to the user's individual bond, and its claimable amount, and claim it. 

This might be far from fetching right now, but for now, if we could only get the Total Purchased amount for each offer, that's enough.

vimukthi-git: Cool, lets find out the exact requirement first. Otherwise current structure is ok as long as the need is to track only the fields coming from `BondOffer` entity in the runtime. We can add the beneficiary field to the `NewOffer` event also in the runtime so that a party which only needs beneficiary does not need to process `NewBond` event

subsquid/src/bondedFinanceProcessor.ts
```diff
@@ -0,0 +1,103 @@
...
+  const { offerId } = event.asLatest;
+  return { offerId };
+}
+
+/**
+ * Extracts beneficiary from a bond offer
+ * @param ctx
+ */
+function getBeneficiary(ctx: EventHandlerContext): string {
+  const { beneficiary } = ctx.event.extrinsic?.args[0]?.value as Record<
```
vimukthi-git, line 47: Are we sure that this is the `beneficiary` always?

andresmechali: Good question. It should, as `beneficiary` is indeed the first argument of the extrinsic.

subsquid/schema.graphql
andresmechali: I mark this a `resolved` as it overlaps with the other conversation. Feel free to un-resolve it.

subsquid/src/bondedFinanceProcessor.ts
vimukthi-git: Hmm I don't think that is the case according to the rust code `BondOffer.beneficiary` can be set to something other than the first argument(`from`) of the extrinsic

andresmechali: Ok, in that case I think the best way is to add it to the `NewOffer` event, which right now is only including the `offerId`. We could also obtain the beneficiary by querying the chain's state after we get the event. but that would be more error-prone in my opinion.

vimukthi-git: > in that case I think the best way is to add it to the NewOffer event

Yeah that's the right approach

vimukthi-git: Let's make a separate PR to make the rust updates.

andresmechali: For reference: https://github.com/ComposableFi/composable/pull/1247

subsquid/src/model/generated/pabloPool.model.ts
```diff
@@ -54,7 +54,7 @@ export class PabloPool {
   @Column_("numeric", {transformer: marshal.bigintTransformer, nullable: false})
   calculatedTimestamp!: bigint
 
-  @OneToMany_(() => PabloPoolAsset, e => e.pool, {eager: true})
```
vimukthi-git, line 57: looks like an unintended change, let's revert. Also let's run the unit tests locally for the subsquid stuff as they are not yet running automatically on CI

subsquid/src/pabloProcessor.ts
vimukthi-git: I think for now removing old parsing logic is ok as the runtime is yet to be released to a production chain that we need to keep backwards compatibility with.

vimukthi-git (approved):
LGTM

subsquid/src/model/generated/pabloPool.model.ts
andresmechali: This is automatically generated by Subsquid using TypeORM, so it's not something we should hard-code. As nothing changed on the schema regarding this type, it might be something that they changed internally

vimukthi-git: I know it's automatically generated, but I remember having to change to get data loaded for some logic for pablo. Better run all tests and verify.

easteregg (approved):

Dom-Roth (approved):

PR #1190, opened 2022-06-22
Fix dependency on assets list that exist out of scope of the react App

saadjhk:
## Issue
 Pablo modules use hardcoded list of assets.

## Description
Previously, all modules of pablo relied of asset information that existed outside the scope of react state. Due to this module were not flexbile enough to incorporate list of assets dynamically. Therefore, this PR aims to remove all such dependencies. 
Solution: use assets list only from the state (store) and a mocked assets list API response

## Checklist

- [x] Refactor Swaps

- [x] Refactor Auctions

- [x] Refactor Pools

- [x] Refactor Create Pool

- [x] Refactor Add Liquidity

- [x] Refactor Remove Liquidity

- [x] Merge DIdd work on Bonds

- [x] Integrate Pablo Bonds Deposit Form

- [ ] Integrate User Bond Positions

PR #1227, opened 2022-06-28
Auction Enhancement

saadjhk:
## Description
Refactored auctions to disable both inputs when output is being calculated and moved form specific logic from UI component to a custom hook.


saadjhk:
Merged in https://github.com/ComposableFi/composable/pull/1190

PR #1230, opened 2022-06-29
[Merged by Bors] - Price feed twap

vivekvpandya:
## Issue
https://app.clickup.com/t/2gg4b6r

## Description
Use subxt to subscribe TwapUpdated event from pablo pallet and then generate Feed notification from processed event.

## Checklist


utils/price-feed/src/feed/composable.rs
```diff
@@ -0,0 +1,91 @@
...
+	pub async fn start(
+		assets: &HashSet<(Asset, Asset)>,
+	) -> FeedResult<Feed<FeedIdentifier, Asset, TimeStampedPrice>> {
+		let (sink, source) = mpsc::channel(CHANNEL_BUFFER_SIZE);
+		// Notifiy feed started
+		sink.send(FeedNotification::Started { feed: FeedIdentifier::Composable })
+			.await
+			.map_err(|_| FeedError::ChannelIsBroken)?;
+		let api = ClientBuilder::new()
+			.build()
```
vivekvpandya, line 28: Perhaps I need a URL parameter to help specify non local host and non default port.

vimukthi-git: This should require the URL, otherwise how would the program know which RPC host to subscribe to?

vivekvpandya: Added.

vimukthi-git:
@vivekvpandya @benluelo looks like this branch has some sort rebase error?

benluelo:
I have it fixed locally, it will be fixed when I push my changes @vimukthi-git

.github/CODEOWNERS
```diff
@@ -39,4 +39,6 @@ integration-tests/runtime-tests/ @ComposableFi/testers
 .devcontainer/ @ComposableFi/codespacers
 
 # Oracle Setup Script
-scripts/oracle-setup @ComposableFi/parachain
\ No newline at end of file
+scripts/oracle-setup @ComposableFi/parachain
```
vimukthi-git, line 42:
```suggestion
scripts/oracle-setup @ComposableFi/parachain-finance 
```

vimukthi-git (approved):
LGTM, let's make sure this is tested locally before merge

Dom-Roth (approved):

vimukthi-git:
@dzmitry-lahoda 

> please consider: ..

@benluelo let's report these as separate issues(and handle) rather than extend this PR with all those additional stuff. We can prioritize them as we get time.

> these are out of consesus and require key trust or something...

We have a staking/slashing mechanism to provide game theoretic security to Oracles while also making manipulation less likely through the use of statistical median to accept the correct price. We also conduct security audits for Oracle pallet. In any case I don't think this PR is the place to raise security concerns about the Oracle design as it is just about the price server. You can possibly report an issue with the details of the exploit though, thanks :)

benluelo (approved):

vimukthi-git (approved):

PR #1242, opened 2022-07-02
xcvm: comments on protobuf is not ideal

dzmitry-lahoda:
for ordered execution and hashing of programs as per spec

### DIFF
--- a/lib/xcvm/protobuf/src/xcvm.proto
+++ b/lib/xcvm/protobuf/src/xcvm.proto
@@ -50,6 +50,9 @@ message Account {
 }
 
 message U128 {
+  // minimums:
+  // bytes - 1 byte type + 1 byte length + 1 byte value, but not 7 bit encoded, so always 2+ bytes (more than 65535 tokens)
+  // low + high - 2 byte types + 2 bytes values
   uint64 low = 1;
   uint64 high = 2;
 }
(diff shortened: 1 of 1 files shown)


lib/xcvm/protobuf/src/xcvm.proto
```diff
@@ -2,6 +2,12 @@ syntax = "proto3";
 
 package xcvm;
 
+/// proto buf tells that
+/// - any repeated field can be intemingled with not not, e.g. `asset A, account, asset B`can happen
+/// - also can happend repeats, which are really not, exampeles `transfer, transfer` and `asset A, asset A`
+/// so proto are not enforce inter field order niether exact once by specfication,
+/// so that should be specified on XCVM compatible protocol to use proper protoc
```
KaiserKarel, line 9: If the decoder implements the spec properly, it should be able to handle this case right? 

Basing my knowledge off [this](https://stackoverflow.com/questions/55063424/does-repeated-fields-in-protobuffers-keep-the-order-they-are-inserted).

KaiserKarel (approved):

KaiserKarel:
Probably requires a rebase

lib/xcvm/protobuf/src/xcvm.proto
dzmitry-lahoda: order of one field yes, but inter field is not guaranteed like in comment in above link https://stackoverflow.com/questions/55063424/does-repeated-fields-in-protobuffers-keep-the-order-they-are-inserted#comment96919403_55069102 

neither is guaranteed that field only encoded once even if singular.

so by spec proto is kind of of no ideal for linear interpretation with lowers possible memory usage. so if compilers used are more strict than spec, then that will work.

PR #1247, opened 2022-07-04
add beneficiary to NewOffer event

andresmechali:
## Issue
https://github.com/ComposableFi/composable/pull/1220#discussion_r912929099

## Description
Need to get bond beneficiary from every `NewOffer` event, to be able to include it on Subsquid

### DIFF
--- a/frame/bonded-finance/src/lib.rs
+++ b/frame/bonded-finance/src/lib.rs
@@ -94,7 +94,7 @@ pub mod pallet {
 	#[pallet::generate_deposit(pub(super) fn deposit_event)]
 	pub enum Event<T: Config> {
 		/// A new offer has been created.
-		NewOffer { offer_id: T::BondOfferId },
+		NewOffer { offer_id: T::BondOfferId, beneficiary: AccountIdOf<T> },
 		/// A new bond has been registered.
 		NewBond { offer_id: T::BondOfferId, who: AccountIdOf<T>, nb_of_bonds: BalanceOf<T> },
 		/// An offer has been cancelled by the `AdminOrigin`.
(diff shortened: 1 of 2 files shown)


Dom-Roth (dismissed):
LGTM!
Integration tests will need a small update to also check against this new result, but with the current launcher issues, this can be difficult to run and would likely be coded without actually running the test. 
Therefore we'll update them later.

frame/bonded-finance/src/lib.rs
```diff
@@ -310,7 +310,7 @@ pub mod pallet {
 				keep_alive,
 			)?;
 			BondOffers::<T>::insert(offer_id, (from.clone(), offer));
-			Self::deposit_event(Event::<T>::NewOffer { offer_id });
+			Self::deposit_event(Event::<T>::NewOffer { offer_id, beneficiary: from.clone() });
```
vimukthi-git, line 313:
```suggestion
			Self::deposit_event(Event::<T>::NewOffer { offer_id, beneficiary: offer.beneficiary.clone() });
```

frame/bonded-finance/src/tests.rs
```diff
@@ -59,7 +59,7 @@ proptest! {
 					  prop_assert_ok!(offer_id);
 					  let offer_id = offer_id.expect("impossible; qed");
 
-					  System::assert_last_event(Event::BondedFinance(crate::Event::NewOffer{ offer_id }));
+					  System::assert_last_event(Event::BondedFinance(crate::Event::NewOffer{ offer_id, beneficiary: ALICE }));
```
vimukthi-git, line 62: Better to test with an account different than the origin

```suggestion
					  System::assert_last_event(Event::BondedFinance(crate::Event::NewOffer{ offer_id, beneficiary: BOB }));
```

frame/bonded-finance/src/lib.rs
andresmechali: done

frame/bonded-finance/src/tests.rs
andresmechali: Good point. Fixed.

vimukthi-git (approved):
LGTM

vivekvpandya (approved):

vimukthi-git:
bors p=[2]

saadjhk:
@andresmechali i reckon we will have to create a new PR here as well

PR #1271, opened 2022-07-06
[Merged by Bors] - #CU-344fut5 make ability to set assets symbols

dzmitry-lahoda:

## Description
Proper register of CurrencyFactory pallet in runtime

### DIFF
--- a/runtime/dali/src/lib.rs
+++ b/runtime/dali/src/lib.rs
@@ -1207,7 +1207,7 @@ construct_runtime!(
 
 		Tokens: orml_tokens::{Pallet, Call, Storage, Event<T>} = 51,
 		Oracle: oracle::{Pallet, Call, Storage, Event<T>} = 52,
-		CurrencyFactory: currency_factory::{Pallet, Storage, Event<T>} = 53,
+		CurrencyFactory: currency_factory = 53,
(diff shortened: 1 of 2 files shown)


ComposableFiB0t:
Task linked: [CU-344fut5 Add extrinsic to upsert token symbol (impl is here, no extinsic)](https://app.clickup.com/t/344fut5)

runtime/dali/src/lib.rs
```diff
@@ -1207,7 +1207,7 @@ construct_runtime!(
 
 		Tokens: orml_tokens::{Pallet, Call, Storage, Event<T>} = 51,
 		Oracle: oracle::{Pallet, Call, Storage, Event<T>} = 52,
-		CurrencyFactory: currency_factory::{Pallet, Storage, Event<T>} = 53,
+		CurrencyFactory: currency_factory = 53,
```
KaiserKarel, line 1210: Changed with the new dot version?

KaiserKarel (approved):

runtime/dali/src/lib.rs
dzmitry-lahoda: we forget to add call. but with 9.22 it all auto.

PR #1282, opened 2022-07-07
[CU-33t4j5e] emit VestingScheduleAdded when vesting schedule is called from another pallet

andresmechali:
## Issue
(https://app.clickup.com/t/33t4j5e)

### DIFF
--- a/frame/vesting/src/lib.rs
+++ b/frame/vesting/src/lib.rs
@@ -148,7 +148,7 @@ pub mod module {
 	}
 
 	#[pallet::event]
-	#[pallet::generate_deposit(fn deposit_event)]
+	#[pallet::generate_deposit(pub fn deposit_event)]
 	pub enum Event<T: Config> {
 		/// Added new vesting schedule. \[from, to, schedule\]
 		VestingScheduleAdded {
@@ -319,8 +319,11 @@ impl<T: Config> VestedTransfer for Pallet<T> {
 
 		T::Currency::transfer(asset, from, to, schedule_amount)?;
 		T::Currency::set_lock(VESTING_LOCK_ID, asset, to, total_amount)?;
-		<VestingSchedules<T>>::try_append(to, asset, schedule)
+		<VestingSchedules<T>>::try_append(to, asset, &schedule)
 			.map_err(|_| Error::<T>::MaxVestingSchedulesExceeded)?;
+
+		Self::deposit_event(Event::VestingScheduleAdded { from: from.clone(), to: to.clone(), asset, schedule });
+
 		Ok(())
 	}
 }


ComposableFiB0t:
Task linked: [CU-33t4j5e Enable Pallet-to-Pallet calls should emit event when vesting schedule added](https://app.clickup.com/t/33t4j5e)

frame/bonded-finance/src/tests.rs
```diff
@@ -92,6 +95,46 @@ proptest! {
...
+			let mut offer = offer.clone();
+			offer.beneficiary = BOB;
+			System::set_block_number(1);
+			prop_assert_ok!(Tokens::mint_into(NATIVE_CURRENCY_ID, &ALICE, Stake::get()));
+			prop_assert_ok!(Tokens::mint_into(offer.reward.asset, &ALICE, offer.reward.amount));
+			let offer_id = BondedFinance::do_offer(&ALICE, offer.clone(), false);
+			prop_assert_ok!(offer_id);
+			let offer_id = offer_id.expect("impossible; qed");
+
+			prop_assert_ok!(BondedFinance::do_bond(offer_id, &BOB, 1, false));
```
dzmitry-lahoda, line 110: why you test bond offer created in vesting schdule created test?

frame/bonded-finance/src/tests.rs
```diff
@@ -92,6 +95,46 @@ proptest! {
...
+
+			let schedule = VestingSchedule {
+				window: BlockNumberBased { start: 1_u64, period: 1_u64 },
+				 period_count: 1_u32,
+				   per_period: 1000000_u128,
+			};
+
+			println!("{:?}", System::events());
```
dzmitry-lahoda, line 118: afaik we do not merge println in tests

frame/bonded-finance/src/tests.rs
vimukthi-git: Yeah this is not nice, we can just test even publishing in the vesting pallet

frame/bonded-finance/src/tests.rs
andresmechali: nope, it was temporarily there while still in draft.

frame/bonded-finance/src/tests.rs
andresmechali: The `vesting` pallet already has a test for `VestingScheduleAdded` which was passing, even though the event was not generated when bonding. Doesn't it make sense that we make sure in the bonded finance tests that this event is generated, or should we keep it strictly decoupled?

vimukthi-git: I guess the test you would want to do is a test that makes sure that 
```
<Self as VestedTransfer>::vested_transfer(asset, &from, &to, schedule)?;
```
emits the event in vesting pallet. That should work for all pallets who would depend on `VestedTransfer` trait, so that test naturally belongs in the vesting pallet.

andresmechali: fixed ✅

vimukthi-git (approved):
LGTM

vivekvpandya (approved):

PR #1284, opened 2022-07-07
CU-2na9jhm - Oracle validity checks

0xangelo:
## Issue
[CU-2na9jhm](https://app.clickup.com/t/2na9jhm)

## Description
Includes:
- refactor(CH): convert indentation to spaces
- refactor(CH): rename last_oracle to last_oracle_price
- refactor(CH): rename oracle to oracle_price
- [CU-2t866b0](https://app.clickup.com/t/2t866b0)

### DIFF
--- a/frame/clearing-house/src/lib.rs
+++ b/frame/clearing-house/src/lib.rs
@@ -159,7 +159,7 @@ pub mod pallet {
 			self, FixedPointMath, FromBalance, IntoBalance, IntoDecimal, IntoSigned, UnsignedMath,
 		},
 		types::{
-			AccountSummary, PositionInfo, TradeResponse, TraderPositionState,
+			AccountSummary, OracleStatus, PositionInfo, TradeResponse, TraderPositionState,
 			BASIS_POINT_DENOMINATOR,
 		},
 		weights::WeightInfo,
@@ -1123,7 +1123,8 @@ pub mod pallet {
 			}
 
 			// For checking oracle guard rails afterwards
-			let mark_index_divergence_before = Self::mark_index_divergence(&market)?;
+			let mark_index_divergence_before =
+				Self::mark_index_divergence(&market, &oracle_status.price)?;
 
 			let outstanding_profits =
 				Self::outstanding_profits(account_id, market_id).unwrap_or_else(Zero::zero);
@@ -1143,6 +1144,7 @@ pub mod pallet {
 
 			Self::check_oracle_guard_rails(
 				&market,
+				&oracle_status,
 				mark_index_divergence_before,
 				is_risk_increasing,
 			)?;
@@ -1208,7 +1210,8 @@ pub mod pallet {
 				}
 
 				// For checking oracle guard rails afterwards
-				let mark_index_divergence_before = Self::mark_index_divergence(&market)?;
+				let mark_index_divergence_before =
+					Self::mark_index_divergence(&market, &oracle_status.price)?;
 
 				let (base_swapped, entry_value, exit_value) = Self::do_close_position(
 					&mut positions,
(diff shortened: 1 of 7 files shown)


frame/clearing-house/src/lib.rs
```diff
@@ -1707,48 +1730,56 @@ pub mod pallet {
...
 			let was_mark_index_too_divergent =
 				Self::exceeds_max_price_divergence(mark_index_divergence_before);
-			let divergence = Self::mark_index_divergence(market)?;
+			let divergence = Self::mark_index_divergence(market, &oracle_status.price)?;
 			let is_mark_index_too_divergent = Self::exceeds_max_price_divergence(divergence);
 
 			// Block if mark-index divergence was pushed too far
 			ensure!(
-				was_mark_index_too_divergent || !is_mark_index_too_divergent,
+				!oracle_status.is_valid ||
```
Cardosaum, line 1745: I didn't get why here and in line `1751` a valid oracle status would throw an error.

frame/clearing-house/src/tests/close_position.rs
```diff
@@ -248,7 +251,7 @@ fn should_fail_if_pushes_index_mark_divergence_above_threshold() {
 		));
 
 		// Alice tries to close her position, but it fails because it pushes the mark price too
-		// below the index Closing tanks mark to 89% of previous price
+		// below the index. Closing tanks the mark price to 89% of previous one.
```
Cardosaum, line 254: "Closing tanks the mark price to 89% of *the* previous one." ? (Not sure about it though, I'm not good at grammar :smiling_face_with_tear: )

Cardosaum (commented):
Great work in this PR! 

I made some comments, but actually more related to my understanding about how the CH should operate. 

Awesome work overall!

frame/clearing-house/src/math.rs
```diff
@@ -26,6 +26,22 @@ where
...
+		Ok(if self < min {
+			min
+		} else if self > max {
+			max
+		} else {
+			self
+		})
+	}
+}
+
```
Cardosaum, line 44: Nice addition to `math.rs`!

0xangelo: What if I told most of it was GitHub copilot?

frame/clearing-house/src/tests/close_position.rs
0xangelo: You are right ser

frame/clearing-house/src/lib.rs
0xangelo: Another way to frame this check is
```rust
if oracle_status.is_valid && !was_mark_index_too_divergent && is_mark_index_too_divergent {
    return Err(/* ... */)
}
```
But here we use `ensure!`, so we need to invert things and use De Morgan's Laws.

0xangelo: I'll add a coment so that future maintainers can understand it

frame/clearing-house/src/lib.rs
```diff
@@ -1707,48 +1730,56 @@ pub mod pallet {
...
+					was_mark_index_too_divergent ||
+					!is_mark_index_too_divergent,
 				Error::<T>::OracleMarkTooDivergent
 			);
 
 			ensure!(
-				!is_risk_increasing ||
-					!is_mark_index_too_divergent ||
+				!oracle_status.is_valid ||
+					!is_risk_increasing || !is_mark_index_too_divergent ||
```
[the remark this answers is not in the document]
0xangelo: There's a slight difference in what they're checking for: 
- the previous one will let the resulting mark-index divergence be large if the previous one was already so
- the second will not let a risk-increasing trade increase the mark-index divergence to a value that's too large, regardless of its previous value

I think you have a point about redundancy. I'll see if there's a way to fuse the two

frame/clearing-house/src/math.rs
dzmitry-lahoda: i recall it can be simolified via .max  . bot sure so if need error here at all. why not just use .max directly?

0xangelo: The alternative was to use `.clamp`, but it panics if `min > max`, instead of giving an error. This implementation is just to make sure we don't shoot ourselves in the foot in the future.

kevin-composable (approved):

PR #1287, opened 2022-07-07
fixed vault's prop test

mikolaichuk:
## Issue
https://app.clickup.com/t/2tdx0bu

## Description
Fixed vault's prop test. Added additional check for the cases when return amount exceeds actual debt balance.

## Checklist

### DIFF
--- a/frame/lending/src/tests/vault.rs
+++ b/frame/lending/src/tests/vault.rs
@@ -1,7 +1,7 @@
 use super::prelude::*;
 use crate::tests::borrow;
 use codec::Decode;
-use composable_traits::vault::Vault as VaultTrate;
+use composable_traits::{lending::TotalDebtWithInterest, vault::Vault as VaultTrate};
 use frame_support::traits::{fungible::Mutate as FungibleMutateTrate, fungibles::Mutate};
 use sp_runtime::{traits::TrailingZeroInput, Perquintill};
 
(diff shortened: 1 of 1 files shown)


frame/lending/src/tests/vault.rs
```diff
@@ -158,6 +158,15 @@ proptest! {
 			//Now vault is unbalanced and should restore equilibrium state.
 			 while Lending::ensure_can_borrow_from_vault(&vault_id, &market_account).is_err() {
 				for borrower in &borrowers {
+					// Sometimes return_amount can exceed total debt.
+					// In such cases we have to repay actual debt balance.
+					let borrower_total_debt = <Lending as LendingTrait>::total_debt_with_interest(&market_id, &borrower).unwrap();
+					let mut return_amount = return_amount;
+					match borrower_total_debt {
```
dzmitry-lahoda, line 165: 1. may be if let and if will look better then match. you also may not need mut as can do let q = if let Some(a) and if a.b { a.b } else q
2. not sure about test in general. if user wants to return debt more than he ows, let him do it , so do not take whole amount? just part he ows? rigth? so it is not error to repay more, but protocol will not take more. 
3. how it appears that that stuff is less than borrowed. may be borrowed_amount_per_borrower is to return, not return amount?

benluelo (approved):

frame/lending/src/tests/vault.rs
mikolaichuk: 1. Fixed

benluelo (approved):

mikolaichuk:
bors merge

PR #1316, opened 2022-07-13
Update release

haroldsphinx:
## Issue

## Description

## Checklist


KaiserKarel (changes requested):
Why are we removing devnet-stage

KaiserKarel:
Also please describe why we are doing this PR using the description template.

.github/workflows/docker-release.yml
```diff
@@ -94,7 +96,7 @@ jobs:
     - name: Run ansible playbook
       working-directory: "./.maintain/playbooks"
       run: |
-        ansible-playbook -l _environment_testnet client-upgrade.yml -i gcp.yaml  --user runner -e 'ansible_python_interpreter=/usr/bin/python3'
+        ansible-playbook -l _environment_testnet client-upgrade.yml -i gcp.yaml  --user root -e 'ansible_python_interpreter=/usr/bin/python3'
```
KaiserKarel, line 99: Why root?

haroldsphinx: this is because of the authorized user we added to all the nodes for ssh, it can only be accessed by root user

KaiserKarel: Might be good to create a separate user and grant them access to that instead of using root?

PR #1319, opened 2022-07-13
Migrate pallet-ibc to child tries for ICS23

### LINKED ISSUE #951: Switch to child tries for pallet-ibc
We currently have to manually build a new root hash of data that technically already exists in the runtime, but this time using custom paths defined by the ibc spec.

Long term it would be better to use a pallet level child trie for this + add upstream (to substrate) the ability to define custom key paths for storage items.

This way, we can get rid of our very expensive `on_finalize` function in `pallet-ibc` and just read the child trie root of the module `on_finalize` and use that as the `commitment_root` value for the `BeefyConsensusState`. This also allows us just use state proofs for ibc.

seunlanlege:
closes #951


frame/ibc/src/lib.rs
```diff
@@ -209,13 +198,11 @@ pub mod pallet {
...
 	#[allow(clippy::disallowed_types)]
-	/// connection_id => ConnectionEnd
-	pub type Connections<T: Config> =
-		CountedStorageMap<_, Blake2_128Concat, Vec<u8>, Vec<u8>, ValueQuery>;
+	pub type ChannelCounter<T: Config> = StorageValue<_, u32, ValueQuery>;
 
 	#[pallet::storage]
 	#[allow(clippy::disallowed_types)]
-	pub type ChannelCounter<T: Config> = StorageValue<_, u32, ValueQuery>;
+	pub type PacketCounter<T: Config> = StorageValue<_, u32, ValueQuery>;
```
Wizdave97, line 205: This would probably not be needed, since we'll rewrite the `on_finalize` benchmark

frame/ibc/ibc-trait/src/lib.rs
```diff
@@ -228,11 +228,11 @@ pub fn client_type_from_bytes(client_type: Vec<u8>) -> Result<ClientType, Error>
 }
 
 /// Get trie key by applying the commitment prefix to the path and scale encoding the result
```
Wizdave97, line 230:
```suggestion
/// Get trie key by applying the commitment prefix to the path
```

Wizdave97 (changes requested):
Looks good, great work

seunlanlege:
This should be ready now, last step is to rewrite the benchmarking of `on_finalize`

Wizdave97 (approved):

runtime/dali/src/weights/pallet_ibc.rs
```diff
@@ -31,230 +31,213 @@ use sp_std::marker::PhantomData;
...
+			.saturating_add((26_000 as Weight).saturating_mul(c as Weight))
+			// Standard Error: 8_000
+			.saturating_add((24_000 as Weight).saturating_mul(d as Weight))
+			// Standard Error: 8_000
+			.saturating_add((18_000 as Weight).saturating_mul(e as Weight))
+			// Standard Error: 8_000
+			.saturating_add((37_000 as Weight).saturating_mul(f as Weight))
+			.saturating_add(T::DbWeight::get().reads(4 as Weight))
 			.saturating_add(T::DbWeight::get().writes(2 as Weight))
 	}
```
seunlanlege, line 220: chef's kiss

blasrodri (approved):

PR #1352, opened 2022-07-20
[Merged by Bors] - Remove unused CNAME and sudo.md in root

KaiserKarel:
## Issue
Seems we still had some unused files in the root, this cleans that up.


CNAME
```diff
@@ -1 +0,0 @@
-get.composable.finance
```
dzmitry-lahoda, line 1: it was github pages named with dns afaik and we had only script there.  may be should clean up DNS record @haroldsphinx

dzmitry-lahoda: and clean up github side setting for github pages too cc @KaiserKarel

dzmitry-lahoda (approved):

benluelo (approved):

CNAME
KaiserKarel: I think get.composable.finance was from a legacy script way, way back by Oleksii, so should be easy to remove now.
What do we need to remove pages side?

andor0 (approved):

PR #1354, opened 2022-07-20
[CU-2rb73xg] Integration Tests - Launch Plan Phase 1

Dom-Roth:
## Issue
https://app.clickup.com/t/2rb73xg

## Description
Integrated by adding the Tag `LAUNCH` to the test required by these.

## Checklist

- [x] Implemented all tests
- [x] Implemented all required test checks
- [x] Cleaned up all code
- [x] Added required documentation
- [x] Ran the whole test suite without any failures
- [x] Ran `npm run lint` without any warnings / errors
- [x] Created / Updated ClickUp report

### DIFF
--- a/code/integration-tests/runtime-tests/package.json
+++ b/code/integration-tests/runtime-tests/package.json
@@ -12,6 +12,7 @@
     "lint": "eslint 'test/**/*.ts' 'src/**/*.ts' --fix-dry-run ",
     "test": "ts-mocha --paths -p tsconfig.json",
     "test_short": "ts-mocha --paths -p tsconfig.json --grep 'SHORT'",
+    "test_launch": "ts-mocha --paths -p tsconfig.json --grep 'LAUNCH'",
     "gen": "npm run gen:defs && npm run gen:meta",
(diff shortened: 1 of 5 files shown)


ComposableFiB0t:
Task linked: [CU-2rb73xg Test Suite - Phase 1](https://app.clickup.com/t/2rb73xg)

Dom-Roth:
Setting back to draft for now. Because it is currently missing a test for this condition:
> Users can see their crowdloan but not claim

vimukthi-git (approved):
So I guess the actual tests for "PICA for gas" etc needs to be written later?

Dom-Roth:
> So I guess the actual tests for "PICA for gas" etc needs to be written later?

Yes, since all of the pallets that'll will additional tests are pallets that are on my "Tests about to check" list anyway. So i'll take care of them, when i take care of that.

The only exception to this is the additional crowdloan test. Since the crowdloan isn't on my check list, I'll take care of it within this PR as well.

kollegian (approved):

PR #1355, opened 2022-07-20
pallet cosmwasm

hussein-aitlahcen:


seunlanlege:
cosmwasm api over wasmi backend, fucking epic man

runtime/dali/src/lib.rs
```diff
@@ -1177,6 +1179,94 @@ impl pallet_ibc_ping::Config for Runtime {
 	type IbcHandler = Ibc;
 }
 
+/// Native <-> Cosmwasm account mapping
+pub struct AccountToAddr;
+impl Convert<alloc::string::String, Result<AccountId, ()>> for AccountToAddr {
+	fn convert(a: alloc::string::String) -> Result<AccountId, ()> {
+		match a.strip_prefix("0x") {
+			Some(account_id) =>
+				Ok(TryInto::<[u8; 32]>::try_into(hex::decode(account_id).map_err(|_| ())?)
```
benluelo, line 1188: I believe you should also be able to use `<[u8; 32]>::try_from(..)`

frame/cosmwasm/src/lib.rs
```diff
@@ -0,0 +1,1071 @@
...
+	/// The purpose of this nonce is just to make sure that contract trie are unique.
+	#[pallet::storage]
+	pub(crate) type CurrentNonce<T: Config> =
+		StorageValue<_, u64, ValueQuery, Nonce<ZeroInit, SafeIncrement>>;
+
+	/// A mapping between a contract and it's nonce.
+	/// The nonce is a monotonic counter incremented when the contract instantiate another contract.
+	#[pallet::storage]
+	pub(crate) type ContractNonce<T: Config> =
+		StorageMap<_, Identity, AccountIdOf<T>, u64, ValueQuery>;
```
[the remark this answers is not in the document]
hussein-aitlahcen: Yes, it start with 0

frame/cosmwasm/src/lib.rs
```diff
@@ -0,0 +1,1071 @@
...
+			);
+			let nonce =
+				CurrentNonce::<T>::increment().expect("2305843009 contracts per human; qed;");
+			let trie_id = Self::derive_contract_trie_id(&contract, nonce);
+			let contract_info =
+				ContractInfoOf::<T> { instantiator, code_id, trie_id, admin, label };
+			ContractToInfo::<T>::insert(&contract, &contract_info);
+			CodeIdToInfo::<T>::try_mutate(code_id, |entry| -> Result<(), Error<T>> {
+				let code_info = entry.as_mut().ok_or(Error::<T>::CodeNotFound)?;
+				code_info.refcount += 1;
```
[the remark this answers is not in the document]
hussein-aitlahcen: I doubt a refcount limit can be reached but I'll make it safe indeed :)

frame/cosmwasm/src/instrument.rs
```diff
@@ -0,0 +1,26 @@
+use codec::{Decode, Encode};
+use parity_wasm::elements::Module;
+use scale_info::TypeInfo;
+use wasm_instrument::gas_metering::{self, Rules};
+
+/// Errors likely to happen while instrumenting a code.
```
KaiserKarel, line 6: Are they `likely` to happen?

frame/cosmwasm/src/instrument.rs
```diff
@@ -0,0 +1,26 @@
...
+#[derive(Clone, Debug, Encode, Decode, TypeInfo)]
+pub enum InstrumentationError {
+	GasMeteringInjection,
+	StackHeightLimitingInjection,
```
KaiserKarel, line 10: These names are a bit technical, would like to see comments describing what is wrong with the module if these are encountered, or fields with more info

frame/cosmwasm/src/instrument.rs
```diff
@@ -0,0 +1,26 @@
...
+	gas_module_name: &str,
+	stack_limit: u32,
+	cost_rules: &impl Rules,
+) -> Result<Module, InstrumentationError> {
+	let gas_instrumented_module = gas_metering::inject(module, cost_rules, gas_module_name)
+		.map_err(|_| InstrumentationError::GasMeteringInjection)?;
+	let stack_and_gas_instrumented_module =
+		wasm_instrument::inject_stack_limiter(gas_instrumented_module, stack_limit)
+			.map_err(|_| InstrumentationError::StackHeightLimitingInjection)?;
+	Ok(stack_and_gas_instrumented_module)
```
KaiserKarel, line 25: Seems that some error contexts is lost here.

frame/cosmwasm/src/lib.rs
```diff
@@ -0,0 +1,1071 @@
...
+		/// CosmWasm export signatures.
+		///
+		/// * Emits an `Uploaded` event on success.
+		///
+		/// Arguments
+		///
+		/// - `origin` the original dispatching the extrinsic.
+		/// - `code` the actual wasm code.
+		#[transactional]
+		#[pallet::weight(0)]
```
KaiserKarel, line 351: Missing weightinfo

frame/cosmwasm/src/lib.rs
```diff
@@ -0,0 +1,1071 @@
...
+		/// Arguments
+		///
+		/// * `origin` the origin dispatching the extrinsic.
+		/// * `code_id` the unique code id generated when the code has been uploaded via [`upload`].
+		/// * `salt` the salt, usually used to instantiate the same contract multiple times.
+		/// * `funds` the assets transferred to the contract prior to calling it's `instantiate`
+		///   export.
+		/// * `gas` the maximum gas to use, the remaining is refunded at the end of the transaction.
+		#[transactional]
+		#[pallet::weight(T::WeightInfo::execute().saturating_add(*gas))]
```
KaiserKarel, line 413: Same here; charge weight based on funds?

frame/cosmwasm/src/lib.rs
```diff
@@ -0,0 +1,1071 @@
...
+			UncheckedFrom::unchecked_from(T::Hashing::hash(&data))
+		}
+
+		/// Deterministic contract trie id generation.
+		pub(crate) fn derive_contract_trie_id(
+			contract: &AccountIdOf<T>,
+			nonce: u64,
+		) -> ContractTrieIdOf<T> {
+			let data: Vec<_> =
+				contract.as_ref().iter().chain(&nonce.to_le_bytes()).cloned().collect();
```
KaiserKarel, line 451: nit: hash takes a reference, so I think we can avoid the clone + collect, and instead add the data to the hasher?

frame/cosmwasm/src/lib.rs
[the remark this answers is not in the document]
KaiserKarel: Even if it will only overflow in 10 billion years, auditors will still mark it as dangerous.

frame/cosmwasm/src/lib.rs
```diff
@@ -0,0 +1,1071 @@
...
+			admin: Option<AccountIdOf<T>>,
+			label: ContractLabelOf<T>,
+		) -> Result<(AccountIdOf<T>, ContractInfoOf<T>), Error<T>> {
+			let contract = Self::derive_contract_address(&instantiator, code_id, salt);
+			ensure!(
+				!ContractToInfo::<T>::contains_key(&contract),
+				Error::<T>::ContractAlreadyExists
+			);
+			let nonce =
+				CurrentNonce::<T>::increment().expect("2305843009 contracts per human; qed;");
```
KaiserKarel, line 472: Got some panic-ptsd; would prefer this to be an error being returned.

frame/cosmwasm/src/lib.rs
```diff
@@ -0,0 +1,1071 @@
...
+				Ok(())
+			})?;
+			Self::deposit_event(Event::<T>::Instantiated {
+				contract: contract.clone(),
+				info: contract_info.clone(),
+			});
+			Ok((contract, contract_info))
+		}
+
+		/// Create the VM shared state. Including readonly stack, VM depth, gas metering limits and
```
KaiserKarel, line 489:
```suggestion
		/// Create the shared VM state. Including readonly stack, VM depth, gas metering limits and
```

frame/cosmwasm/src/lib.rs
```diff
@@ -0,0 +1,1071 @@
...
+		/// code cache.
+		///
+		/// This state is shared accross all VMs (all contracts loaded within a single call) and is
+		/// used to optimize some operations as well as track shared state (readonly storage while
+		/// doing a `query` etc...)
+		pub(crate) fn do_create_vm_shared(gas: u64, readonly: bool) -> CosmwasmVMShared {
+			CosmwasmVMShared {
+				readonly: if readonly { 1 } else { 0 },
```
KaiserKarel, line 497: Should we change the definition of `CosmwasmVMShared` do have an enum field, and go for something like:

```
do_create_vm_shared(100, Mutability::ReadOnly)
vs
do_create_vm_shared(100, true)
```

frame/cosmwasm/src/lib.rs
```diff
@@ -0,0 +1,1071 @@
...
+		/// Arguments
+		///
+		/// * `origin` the origin dispatching the extrinsic.
+		/// * `code_id` the unique code id generated when the code has been uploaded via [`upload`].
+		/// * `salt` the salt, usually used to instantiate the same contract multiple times.
+		/// * `funds` the assets transferred to the contract prior to calling it's `instantiate`
+		///   export.
+		/// * `gas` the maximum gas to use, the remaining is refunded at the end of the transaction.
+		#[transactional]
+		#[pallet::weight(T::WeightInfo::instantiate().saturating_add(*gas))]
```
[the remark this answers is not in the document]
hussein-aitlahcen: i'm not sure I understand why?

frame/cosmwasm/src/lib.rs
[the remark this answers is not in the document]
hussein-aitlahcen: fixed

frame/cosmwasm/src/instrument.rs
hussein-aitlahcen: I'm going to log them, the issue is that we can't forward it to the user anyway

frame/cosmwasm/src/lib.rs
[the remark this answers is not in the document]
KaiserKarel: So if funds transferred is 0, we are not calling into the Assets pallet for the transfer; but when it is set, we make a transfer, so the weightinfo needs to take that into account

frame/cosmwasm/src/lib.rs
hussein-aitlahcen: The thing is the `CosmwasmVMSharted` struct contains a `readonly` counter (not a boolean).
I.e. if you have contract `A` and `B`, call `query` on `A`, the `readonly` counter is incremented/decremented before/after the `query` call. This avoid a situation where you  `query(A)` and within the query, `A` do a `query(B)` => 2 readonly depth, we can't set readonly = false when we exit `query(B)` or `query(A)` would be able to mutate the state after the call to `query(B)`.

hussein-aitlahcen: I'll still introduce an enum for the initial counter increment

frame/cosmwasm/src/instrument.rs
hussein-aitlahcen: We now log the error at the node level.

frame/cosmwasm/src/lib.rs
KaiserKarel: Or just the function definitions. If the counter makes sense inside the VM, let's keep it at that.

frame/cosmwasm/src/lib.rs
hussein-aitlahcen: The `Hasher` here is a trait that has a `hash` function only, operating on `[u8]`. Am I missing something?

frame/cosmwasm/src/lib.rs
hussein-aitlahcen: Yep!

frame/cosmwasm/src/lib.rs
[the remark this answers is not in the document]
hussein-aitlahcen: Ah, you mean the weight depend on the size of the funds map, yeah, I'll add that

runtime/dali/src/lib.rs
```diff
@@ -1068,6 +1072,94 @@ impl pallet_ibc_ping::Config for Runtime {
...
+	fn convert(CurrencyId(currency_id): CurrencyId) -> alloc::string::String {
+		alloc::format!("{}", currency_id)
+	}
+}
+
+parameter_types! {
+  pub const CosmwasmPalletId: PalletId = PalletId(*b"cosmwasm");
+  pub const ChainId: &'static str = "composable-network-dali";
+  pub const MaxFrames: u32 = 64;
+	pub const MaxCodeSize: u32 = 512 * 1024;
```
dzmitry-lahoda, line 1111: fmt

runtime/dali/src/lib.rs
```diff
@@ -1068,6 +1072,94 @@ impl pallet_ibc_ping::Config for Runtime {
...
+  pub const MaxInstrumentedCodeSize: u32 = 1024 * 1024;
```
dzmitry-lahoda, line 1112: interesting, what is  limit of trie entry in polkadot? may be they have limit too and can plug it?

also, size is in bytes or in kilobytes?

runtime/dali/src/lib.rs
```diff
@@ -1068,6 +1072,94 @@ impl pallet_ibc_ping::Config for Runtime {
...
+  pub const MaxMessageSize: u32 = 256 * 1024;
+  pub const MaxContractLabelSize: u32 = 64;
```
dzmitry-lahoda, line 1114: can use ConstU8<64> instead of macro -> more typed

runtime/dali/src/lib.rs
```diff
@@ -802,8 +804,10 @@ impl pallet_staking_rewards::Config for Runtime {
 pub struct BaseCallFilter;
 impl Contains<Call> for BaseCallFilter {
 	fn contains(call: &Call) -> bool {
-		!(call_filter::Pallet::<Runtime>::contains(call) ||
-			matches!(call, Call::Tokens(_) | Call::Indices(_) | Call::Treasury(_)))
+		if call_filter::Pallet::<Runtime>::contains(call) {
+			return false
+		}
+		!matches!(call, Call::Tokens(_) | Call::Indices(_) | Call::Democracy(_) | Call::Treasury(_))
```
dzmitry-lahoda, line 810: please return back filter. it is correct and does not prevents democracy

dzmitry-lahoda (changes requested):
please merge out democracy filter as it is on main

aeryz:
Btw since this PR requires nightly, we might wanna add `rustup default nightly` to both rust setup docs and init script.

frame/cosmwasm/src/lib.rs
```diff
@@ -0,0 +1,1091 @@
...
+		#[pallet::constant]
+		type CodeStackLimit: Get<u32>;
+
+		/// Price of a byte when uploading new code.
+		/// The price is expressed in [`Self::NativeAsset`].
+		/// This amount is reserved from the owner and released when the code is destroyed.
+		#[pallet::constant]
+		type CodeStorageByteDeposit: Get<u32>;
+
+		/// Price of writting a byte in the storage.
```
aeryz, line 257: `writting` -> `writing`

hussein-aitlahcen:
> Btw since this PR requires nightly, we might wanna add `rustup default nightly` to both rust setup docs and init script.

We'll wait for the nix pipeline for that

benluelo:
Since you're using nightly we should probably have a `rust-toolchain.toml` file pinning the version

frame/cosmwasm/src/lib.rs
```diff
@@ -0,0 +1,1091 @@
...
+
+		/// Max wasm stack size limit.
+		#[pallet::constant]
+		type CodeStackLimit: Get<u32>;
+
+		/// Price of a byte when uploading new code.
+		/// The price is expressed in [`Self::NativeAsset`].
+		/// This amount is reserved from the owner and released when the code is destroyed.
+		#[pallet::constant]
+		type CodeStorageByteDeposit: Get<u32>;
```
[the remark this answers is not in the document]
hussein-aitlahcen: The issue is that it doesn't implements MaxEncodedLen etc by default, which mean I'll need to introduce extra boilerplate to make this generic, which is probably not worth the effort

frame/cosmwasm/src/lib.rs
```diff
@@ -0,0 +1,1091 @@
...
+		#[transactional]
+		#[pallet::weight(T::WeightInfo::instantiate(funds.len()).saturating_add(*gas))]
+		pub fn instantiate(
+			origin: OriginFor<T>,
+			code_id: CosmwasmCodeId,
+			salt: ContractSaltOf<T>,
+			admin: Option<AccountIdOf<T>>,
+			label: ContractLabelOf<T>,
+			funds: FundsOf<T>,
+			gas: u64,
```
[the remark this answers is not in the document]
hussein-aitlahcen: This is a cosmwasm spec, the gas is u64 and not depending on the balance type. Probabably worth extracting it to a type alias though

hussein-aitlahcen:
> Since you're using nightly we should probably have a `rust-toolchain.toml` file pinning the version

Now that we have Nix the rust version is pinned in the flake, I just rebased and reflected the nightly use in the flake itself :). Which will also upgade the developer shell

aeryz (approved):

frame/cosmwasm/src/runtimes/wasmi.rs
```diff
@@ -0,0 +1,724 @@
...
+	/// Shared Gas metering.
+	pub gas: Gas,
+	/// Shared cache.
+	pub cache: CosmwasmVMCache,
+}
+
+impl CosmwasmVMShared {
+	/// Whether the storage is currently readonly.
+	pub fn storage_is_readonly(&self) -> bool {
+		self.storage_readonly_depth == 0
```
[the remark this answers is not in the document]
hussein-aitlahcen: Nice catch!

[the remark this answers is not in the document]
aeryz: I fixed this in the PR that is already merged to this branch so I'm resolving this.

KaiserKarel (approved):

dzmitry-lahoda (approved):

PR #1359, opened 2022-07-21
[Merged by Bors] - Ben/reward accumulation hook/#31md7rz

benluelo:
## TODO
- [x] Write tests for the accumulation hook
- [x] Document the current behaviour of the hook (i.e. how errors are handled)
  - The hook code seems fairly self-documenting, I can add a doc comment explaining the behaviour if necessary.
- [x] Come up with a better name for the `RewardAccumulationError` event?
  - Added `MaxRewardsAccumulated` event with the same fields as `RewardAccumulationError`, the latter only used in the case of `UnixTime` errors as of right now.
- [x] Review what data to include in the `RewardAccumulationError` event
  - `pool_id` and `asset_id` is enough to identify where the max rewards were hit.

## Checklist


ComposableFiB0t:
Task linked: [CU-31md7rz Add pool rate based rewards (automatic)](https://app.clickup.com/t/31md7rz)

frame/composable-traits/src/staking/mod.rs
```diff
@@ -102,7 +109,58 @@ pub struct RewardConfig<AssetId, Balance> {
...
+	pub reward_rate: RewardRate<Balance>,
+}
+
+#[derive(RuntimeDebug, PartialEq, Eq, Clone, MaxEncodedLen, Encode, Decode, TypeInfo)]
+pub struct RewardRate<Balance> {
+	/// The period that the rewards are handed out in, in seconds.
+	// REVIEW(benluelo): NonZeroU64?
+	pub period: u64,
+	/// The amount that is rewarded each period.
+	pub amount: Balance,
```
benluelo, line 121: This needs to be non-zero; should I use `Validated` for this?

frame/staking-rewards/src/lib.rs
```diff
@@ -599,8 +614,121 @@ pub mod pallet {
...
+					let updated_rewards = reward_pool
+						.rewards
+						.into_iter()
+						.map(|(asset_id, reward)| {
+							Self::reward_acumulation_hook_reward_update_calculation(
+								pool_id, asset_id, reward, now,
+							)
+						})
+						.collect::<BTreeMap<_, _>>()
+						.try_into()
```
andor0, line 705: `.into_iter() -> ... -> .try_into()` can be rewritten using [try_mutate()](https://paritytech.github.io/substrate/master/frame_support/struct.BoundedBTreeMap.html#method.try_mutate)

benluelo: I would still need to iterate over the map in the closure provided to try_mutate, this felt cleaner

benluelo: also see https://github.com/paritytech/substrate/pull/11869

andor0:
It would be nice if you write a benchmark for `reward_acumulation_hook_reward_update_calculation`.

benluelo:
> It would be nice if you write a benchmark for `reward_acumulation_hook_reward_update_calculation`.

I wrote this: https://github.com/ComposableFi/composable/pull/1359/files#diff-08e6e8c2efd3f6c119fe0ebc7e4944a806004bf3ae80c20f5f61dc26adb506e1R136-R148

frame/staking-rewards/src/benchmarking.rs
```diff
@@ -121,5 +133,19 @@ benchmarks! {
...
+			asset_id: 1_u128.into(),
+			total_rewards: 0_u128.into(),
+			total_dilution_adjustment: 0.into(),
+			max_rewards: 100.into(),
+			reward_rate: RewardRate::per_second(1.into(), 1.try_into_validated().unwrap()),
+			last_updated_timestamp: 0,
+			claimed_rewards: 0_u128.into()
+		};
+	}: {
+		let reward = Pallet::<T>::reward_acumulation_hook_reward_update_calculation(1.into(), 1.into(), reward, T::UnixTime::now().as_secs());
```
andor0, line 147: looks like, you added weights of `T::UnixTime::now()` (1 read from DB) in weights of `reward_acumulation_hook_reward_update_calculation`.

benluelo: Ah, I didn't realize that now() was a DB read. I'll extract it out, thanks!

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -505,19 +505,7 @@ Following algorithm should be part of the block hook in the pallet.
...
-    :rewards_pool = readStorage(assetId);
-    while (reward = rewards_pool.rewards.has_next) is (yes)
-        :assert max_reward not exceeded;
-        :reward.total_rewards += elapsed_time * reward.reward_rate;
-    endwhile (no)
-    :save(rewards_pool);
-endwhile (no)
-:last_update_timestamp = current_time;
-stop
+c
```
vimukthi-git, line 508: Mistake?

frame/composable-traits/src/staking/mod.rs
```diff
@@ -102,7 +109,58 @@ pub struct RewardConfig<AssetId, Balance> {
...
+}
+
+pub struct BoundedU64<const MIN: u64 = 0, const MAX: u64 = { u64::MAX }> {
```
vimukthi-git, line 124: Good one, but will the use of this cause any errors that are hard to understand?

frame/composable-traits/src/staking/mod.rs
```diff
@@ -102,7 +109,58 @@ pub struct RewardConfig<AssetId, Balance> {
...
+	// _marker: PhantomData<(MIN, MAX)>,
+}
+
+// TODO: TESTS!!!
+impl<const MIN: u64, const MAX: u64> Validate<u64, BoundedU64<MIN, MAX>> for BoundedU64<MIN, MAX> {
+	fn validate(input: u64) -> Result<u64, &'static str> {
+		if input > MAX {
+			Err("value too big")
```
vimukthi-git, line 132:
```suggestion
			Err("BoundedU64: value too big")
```
?

frame/composable-traits/src/staking/mod.rs
```diff
@@ -102,7 +109,58 @@ pub struct RewardConfig<AssetId, Balance> {
...
+		} else if input < MIN {
+			Err("value too small")
```
vimukthi-git, line 134:
```suggestion
			Err("BoundedU64: value too small")
```
?

frame/staking-rewards/src/lib.rs
```diff
@@ -599,8 +620,119 @@ pub mod pallet {
...
+				// hence the double iteration.
+				.collect::<Vec<_>>()
+				.into_iter()
+				.fold(T::WeightInfo::unix_time_now(), |total_weights, (pool_id, reward_pool)| {
+					// 128 bit platforms don't exist as of writing this so this cast should be ok
+					let amount_of_rewards_in_pool = reward_pool.rewards.len() as u64;
+
+					RewardPools::<T>::insert(pool_id, reward_pool);
+
+					total_weights +
```
vimukthi-git, line 729: IMO this code would be much more readable if we did the weight updates in a for or while loop with the actual logic.

frame/staking-rewards/src/benchmarking.rs
```diff
@@ -121,5 +133,27 @@ benchmarks! {
...
+			total_dilution_adjustment: 0.into(),
+			max_rewards: 100.into(),
+			reward_rate: RewardRate::per_second(6.into(), 10_000.try_into_validated().unwrap()),
+			last_updated_timestamp: now.as_secs(),
+			claimed_rewards: 0_u128.into()
+		};
+
+		let now = now + std::time::Duration::from_secs(12);
+	}: {
+		let reward = Pallet::<T>::reward_acumulation_hook_reward_update_calculation(1.into(), 1.into(), reward, now.as_secs());
```
vimukthi-git, line 151: Can we benchmark `acumulate_rewards_hook` instead?

frame/staking-rewards/src/lib.rs
```diff
@@ -599,8 +620,119 @@ pub mod pallet {
...
+		) -> (T::AssetId, Reward<T::AssetId, T::Balance>) {
+			let elapsed_time = match now.safe_sub(&reward.last_updated_timestamp) {
+				Ok(elapsed_time) => elapsed_time,
+				Err(_) => {
+					Self::deposit_event(Event::<T>::RewardAccumulationError { pool_id, asset_id });
+					return (asset_id, reward)
+				},
+			};
+
+			// SAFETY(benluelo): RewardRate::period is non-zero.
```
vimukthi-git, line 642: So we need safe division?

frame/staking-rewards/src/lib.rs
```diff
@@ -599,8 +620,119 @@ pub mod pallet {
...
+			let periods_surpassed = elapsed_time.div(&reward.reward_rate.period);
+
+			if periods_surpassed.is_zero() {
+				(asset_id, reward)
+			} else {
+				// NOTE(benluelo): used this instead of `safe_mul` since
+				// `T::Balance` can't be multiplied by a `u64` and the definition of
+				// `CheckedMul` constrains the `Rhs` type to `Self`, so `T::balance`
+				// can't `.safe_mul(u64)`
```
vimukthi-git, line 651: This reason by it self just not justify the use of `repeat` since we can safely convert u64 to balance given that our runtimes use u128 for it.

frame/staking-rewards/src/lib.rs
```diff
@@ -599,8 +620,119 @@ pub mod pallet {
...
+				//
+				// this also allows us to accumulate up to the maximum amount before
```
vimukthi-git, line 653: Isn't this also possible with a normal saturating multiplication?

frame/staking-rewards/src/lib.rs
```diff
@@ -599,8 +620,119 @@ pub mod pallet {
...
+								ControlFlow::Break(current_total_amount)
+							},
+						}
+					});
+
+				(
+					asset_id,
+					Reward {
+						total_rewards: match new_total_rewards {
+							ControlFlow::Continue(new_total_rewards) => new_total_rewards,
```
vimukthi-git, line 683: This contraption is super hard to understand tbh. What does it do?

frame/staking-rewards/src/lib.rs
```diff
@@ -599,8 +620,119 @@ pub mod pallet {
...
+				//
+				// this also allows us to accumulate up to the maximum amount before
+				// erroring, short-circuiting on any error found and returning the last known
+				// "good" value.
+				let new_total_rewards = iter::repeat(reward.reward_rate.amount)
+					// REVIEW(benluelo): this cast is *probably* safe, but
+					// should be verified somehow
+					.take(periods_surpassed as usize)
+					.try_fold(reward.total_rewards, |current_total_amount, reward_amount| {
+						match current_total_amount.safe_add(&reward_amount) {
```
vimukthi-git, line 661: Looks like we are implementing multiplication in a complicated way, can we simplify?

frame/staking-rewards/src/lib.rs
benluelo: No, we shouldn't need it since the period can't be zero. I haven't enforced this yet though, unsure whether `NonZeroU64` or a `Validated` type would be best

frame/composable-traits/src/staking/mod.rs
```diff
@@ -102,7 +109,58 @@ pub struct RewardConfig<AssetId, Balance> {
...
+		}
+	}
+}
+
+// TODO: TESTS!!!
+// TODO(benluelo): uom?
+impl<Balance> RewardRate<Balance> {
+	pub fn per_second(
+		amount: Balance,
+		seconds: Validated<u64, BoundedU64<1, { u64::MAX }>>,
```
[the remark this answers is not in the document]
benluelo: The idea was per second**s**, but I see how that could be confusing. Perhaps separate methods? These were somewhat of an experiment, I think it would be fine to just struct literal syntax instead of constructors

frame/staking-rewards/src/lib.rs
benluelo: for this and all related comments: https://github.com/ComposableFi/composable/pull/1359/commits/f45dc1dea2e0ba47ee2ba1facea41583eab6f639

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
benluelo: Yep, not sure how that happened. Fixed: https://github.com/ComposableFi/composable/pull/1359/commits/323f56203262bea61fa2fd252ca961553b535af3

frame/staking-rewards/src/lib.rs
```diff
@@ -599,8 +620,98 @@ pub mod pallet {
...
+						reward
+					} else {
+						let new_total_rewards = u128::from(periods_surpassed)
+							.saturating_mul(reward.reward_rate.amount.into());
+
+						let new_total_rewards = if new_total_rewards <= reward.max_rewards.into() {
+							new_total_rewards.into()
+						} else {
+							// saturate at max_rewards, but emit an error first
+							Self::deposit_event(Event::<T>::RewardAccumulationError {
```
vimukthi-git, line 649:
```suggestion
							Self::deposit_event(Event::<T>::MaxRewardAccumulated {
```

frame/staking-rewards/src/lib.rs
benluelo: proptest not required, see updated comment

frame/staking-rewards/src/lib.rs
benluelo: This got buried, I didn't see it. See https://github.com/ComposableFi/composable/pull/1359/commits/e9cc52cfc36892769e1a8a697bd77557559d7938

frame/staking-rewards/src/test/test_reward_accumulation_hook.rs
```diff
@@ -0,0 +1,418 @@
...
+				.unwrap(),
+				lock: default_lock_config(),
+			},
+		];
+
+		for cfg in cfgs {
+			StakingRewards::create_reward_pool(Origin::root(), cfg).unwrap();
+		}
+
+		fn check_rewards(expected: &[(Public, u128, &[(u128, u128)])]) {
```
vimukthi-git, line 182: Could you add a name comment to each anon field here? It would be easier to understand

frame/staking-rewards/src/test/test_reward_accumulation_hook.rs
```diff
@@ -0,0 +1,418 @@
...
+		// should report an error since the max was hit
+		assert_last_event::<Test>(Event::StakingRewards(
+			crate::Event::<Test>::MaxRewardsAccumulated { pool_id: POOL_ID, asset_id: PICA::ID },
+		));
+	})
+}
+
+#[test]
+// takes about 3 minutes to run
+fn test_acumulate_rewards_hook() {
```
vimukthi-git, line 84: Good test :)

frame/staking-rewards/src/lib.rs
```diff
@@ -599,8 +626,118 @@ pub mod pallet {
...
+				// "If you alter the map while doing this, you'll get undefined results."
+				// hence the double iteration.
+				.collect::<Vec<_>>();
+
+			let mut total_weight = unix_time_now_weight;
+
+			for (pool_id, reward_pool) in updated_pools {
+				// 128 bit platforms don't exist as of writing this so this usize -> u64 cast should
+				// be ok
+				let amount_of_rewards_in_pool = reward_pool.rewards.len() as u64;
```
vimukthi-git, line 729:
```suggestion
				let number_of_rewards_in_pool = reward_pool.rewards.len() as u64;
```

vimukthi-git (approved):
LGTM

frame/staking-rewards/src/lib.rs
```diff
@@ -599,8 +626,118 @@ pub mod pallet {
 					.try_insert(*asset_id, inflation)
 					.map_err(|_| Error::<T>::ReductionConfigProblem)?;
 			}
+
 			Ok((rewards_btree_map, reductions))
 		}
+
+		pub(crate) fn reward_acumulation_hook_reward_update_calculation(
```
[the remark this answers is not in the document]
benluelo: I prefixed the name with `reward_acumulation_hook` since the function should only be used within that hook (I had to separate it out for benchmarking) but I'm not partial to it

frame/staking-rewards/src/lib.rs
```diff
@@ -599,8 +626,118 @@ pub mod pallet {
...
+									asset_id: reward.asset_id,
+								});
+							}
+							reward.max_rewards
+						};
+
+						Reward {
+							total_rewards: new_total_rewards,
+							last_updated_timestamp: reward
+								.last_updated_timestamp
```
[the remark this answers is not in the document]
benluelo: Because there could be a situation where the reward period doesn't line up with the block time (think a period of 5 seconds, and a block time of 6 seconds); after 5 blocks (30 secs) you would expect the rewards to have accumulated 6 times, but it would only be 5 since 1 second would be "lost" each block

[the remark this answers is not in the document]
benluelo: This was more of an issue when working with arbitrary reward periods, but I think it's good to keep in for if/when we add more reward periods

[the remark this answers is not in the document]
benluelo: (basically the issue is `let periods_surpassed = elapsed_time.div(reward_rate_period_seconds);` is floor division)

frame/staking-rewards/src/test/test_reward_accumulation_hook.rs
benluelo: good idea, will add

frame/staking-rewards/src/benchmarking.rs
benluelo: Since the hook weight is completely dependant on the amount of pools and rewards, I figured this would give a more accurate weight

frame/staking-rewards/src/benchmarking.rs
```diff
@@ -121,5 +132,27 @@ benchmarks! {
...
+			asset_id: 1_u128.into(),
+			total_rewards: 0_u128.into(),
+			total_dilution_adjustment: 0.into(),
+			max_rewards: 1_000_000.into(),
+			reward_rate: RewardRate::per_second(10_000),
+			last_updated_timestamp: now,
+			claimed_rewards: 0_u128.into()
+		};
+
+		let now = now + 12;
```
andor0, line 148: what is 12?

andor0, line 148:
```suggestion
                let seconds_per_block = 12;
		let now = now + seconds_per_block;
```

andor0 (approved):

frame/staking-rewards/src/lib.rs
```diff
@@ -599,8 +626,118 @@ pub mod pallet {
...
+
+			for (pool_id, reward_pool) in updated_pools {
+				// 128 bit platforms don't exist as of writing this so this usize -> u64 cast should
+				// be ok
+				let number_of_rewards_in_pool = reward_pool.rewards.len() as u64;
+
+				RewardPools::<T>::insert(pool_id, reward_pool);
+
+				total_weight += (number_of_rewards_in_pool * T::WeightInfo::reward_acumulation_hook_reward_update_calculation()) +
+						// REVIEW(benluelo): I'm assuming that `StorageMap::iter` does one read per item, I could be wrong though
```
andor0, line 734: I looked pallets of substrate some time ago and came to the same conclusion.

benluelo: Coo, good to know. Updated the comment :+1:

frame/staking-rewards/src/benchmarking.rs
benluelo: https://github.com/ComposableFi/composable/pull/1359/commits/f8cf66ce8f912ed28eb17a5acb3c6f75ecc999b2

andor0 (approved):

KaiserKarel (approved):

vimukthi-git (approved):

andor0 (approved):

vimukthi-git (approved):
LGTM

vivekvpandya (approved):

benluelo:
Bors r+

Dom-Roth (approved):

vimukthi-git (approved):

PR #1360, opened 2022-07-22
[CU-345zb4r] add id to vesting schedules

andresmechali:
## Issue
https://app.clickup.com/t/345zb4r

## Description
This draft PR only comments on the changes that need to be done in order to support this feature

### DIFF
--- a/frame/bonded-finance/src/lib.rs
+++ b/frame/bonded-finance/src/lib.rs
@@ -69,6 +69,7 @@ pub mod pallet {
 		transactional, PalletId,
 	};
 	use frame_system::{ensure_signed, pallet_prelude::OriginFor};
+	use pallet_vesting::VestingScheduleCount;
 	use scale_info::TypeInfo;
 	use sp_runtime::{
 		helpers_128bit::multiply_by_rational,
@@ -355,13 +356,14 @@ pub mod pallet {
 							keep_alive,
 						)?;
 						let current_block = frame_system::Pallet::<T>::current_block_number();
+						let vesting_schedule_id = VestingScheduleCount::<T>::get().increment()?;
 						// Schedule the vesting of the reward.
 						T::Vesting::vested_transfer(
 							offer.reward.asset,
 							&offer_account,
 							from,
 							VestingSchedule {
-								/// vesting_schedule_id: /// TODO: generate random id
+								vesting_schedule_id,
 								window: BlockNumberBased {
 									start: current_block,
 									period: offer.reward.maturity,
@@ -370,6 +372,7 @@ pub mod pallet {
 								per_period: reward_share,
 							},
 						)?;
+						let vesting_schedule_id = VestingScheduleCount::<T>::get().increment()?;
 						match offer.maturity {
 							BondDuration::Finite { return_in } => {
 								// Schedule the return of the bonded amount
@@ -378,7 +381,7 @@ pub mod pallet {
 									&offer.beneficiary,
 									from,
 									VestingSchedule {
-										/// vesting_schedule_id: /// TODO: generate random id
+										vesting_schedule_id,
 										window: BlockNumberBased {
 											start: current_block,
 											period: return_in,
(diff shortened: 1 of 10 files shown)


ComposableFiB0t:
Task linked: [CU-345zb4r Add an ID to track vesting schedules individually](https://app.clickup.com/t/345zb4r)

frame/bonded-finance/src/lib.rs
```diff
@@ -361,6 +361,7 @@ pub mod pallet {
 							&offer_account,
 							from,
 							VestingSchedule {
+								/// vesting_schedule_id: /// TODO: generate random id
```
vimukthi-git, line 364: Could you check if there is a standard way to generate an ID for this in substrate?

frame/vesting/src/lib.rs
```diff
@@ -346,11 +354,14 @@ impl<T: Config> Pallet<T> {
 	}
 
 	/// Returns locked balance based on current block number.
+	/// TODO: add vesting_schedule_id parameter
```
vimukthi-git, line 357: IMO it might be better to just return a total balance and balances grouped by vesting_schedule_id

vimukthi-git (commented):
two questions

frame/vesting/src/lib.rs
andresmechali: Yes, this makes sense. Updated.

frame/bonded-finance/src/lib.rs
andresmechali: I just spoke with @vivekvpandya and we have 2 possible ways of handling this:
1. We replace the current storage map of `(AccountId, AssetId) -> Vec<VestingSchedule>` for something like `(AccountId, AssetId, count) -> VestingSchedule`, where `count` would be a `u128` that is incremented every time a `VestingSchedule` is added for a given pair of (`AccountId, AssetId`). We can then use `count` as a unique id for claiming rewards.

andresmechali: I just spoke with @vivekvpandya and we have 2 possible ways of handling this:

1. We replace the current storage map of `(AccountId, AssetId) -> Vec<VestingSchedule>` for something like `(AccountId, AssetId, count) -> VestingSchedule`, where `count` would be a `u128` that is incremented every time a `VestingSchedule` is added for a given pair of (`AccountId, AssetId`). We can then use `count` as a unique id for claiming rewards. This would require an extra storage of `(AccountId, AssetId) -> count`.

2. Similarly, we add a `count` id to the `VestingSchedule` structure, and we keep the `VestingSchedules` storage as it is, while also adding the extra storage of `(AccountId, AssetId) -> count`. This would keep things simpler IMO because we wouldn't need to refactor too much.

vimukthi-git: Tracking a count is ok, we can track it separately and use that as the ID inside the VestingSchedule struct.

> while also adding the extra storage of (AccountId, AssetId) -> count

I think this doesn't need to be a map no? We can just have a counter in storage.

andresmechali: Yes, right 👌🏼

frame/vesting/src/lib.rs
```diff
@@ -333,20 +396,45 @@ impl<T: Config> VestedTransfer for Pallet<T> {
 }
 
 impl<T: Config> Pallet<T> {
-	fn do_claim(who: &AccountIdOf<T>, asset: AssetIdOf<T>) -> Result<BalanceOf<T>, DispatchError> {
-		let locked = Self::locked_balance(who, asset);
+	fn do_claim(
+		who: &AccountIdOf<T>,
+		asset: AssetIdOf<T>,
+		vesting_schedule_id: Option<T::VestingScheduleId>,
```
vimukthi-git, line 402: We can keep the existing function and add this stuff as new function

vimukthi-git:
@andresmechali Let's close this if no longer useful

PR #1378, opened 2022-07-26
[Merged by Bors] - fixing call filter governance and setup

dzmitry-lahoda:

## Description
* call filter should not filter out democracy
* and it should have proper origin setup
* and picasso should have it in runtime


runtime/composable/src/lib.rs
```diff
@@ -749,6 +749,7 @@ pub struct BaseCallFilter;
 
 impl Contains<Call> for BaseCallFilter {
 	fn contains(call: &Call) -> bool {
+		// TODO: not up to date as whole composable setup
```
KaiserKarel, line 752: What do you mean by this

runtime/dali/src/lib.rs
```diff
@@ -798,24 +800,16 @@ impl pallet_staking_rewards::Config for Runtime {
...
 pub struct BaseCallFilter;
-
 impl Contains<Call> for BaseCallFilter {
 	fn contains(call: &Call) -> bool {
-		if call_filter::Pallet::<Runtime>::contains(call) {
-			return false
-		}
-		!matches!(call, Call::Tokens(_) | Call::Indices(_) | Call::Democracy(_) | Call::Treasury(_))
+		!(call_filter::Pallet::<Runtime>::contains(call) ||
+			matches!(call, Call::Tokens(_) | Call::Indices(_) | Call::Treasury(_)))
```
KaiserKarel, line 806: Should we add `Assets` to this and filter tokens?

KaiserKarel (approved):

runtime/dali/src/lib.rs
dzmitry-lahoda: i removed only democracy, other than that filter is same. it is clear why tokens are filtered out. assets should be not as i think. balances? may or may not be, anyway they sudo secured.

runtime/composable/src/lib.rs
dzmitry-lahoda: i do not updtae composable runtime. and not updatwd democracy and xcmp here too. and others do not update it too. so likely need at some point to recheck whole runtime and update.

alexkdefi (approved):
👍

andor0 (approved):

hussein-aitlahcen (approved):

PR #1392, opened 2022-07-29
[Merged by Bors] - feat(xcvm): introduce `xcvm-core`

hussein-aitlahcen:
- introduce basic generic instructions set
- generic, partially type-level, type-safe xcvm builder

### DIFF
--- a/xcvm/lib/core/Cargo.toml
+++ b/xcvm/lib/core/Cargo.toml
@@ -13,6 +13,7 @@ scale-info = { version = "2.1.1", default-features = false, features = [
 schemars = { version = "0.8.1", optional = true }
 serde = { version = "1", default-features = false, features = [ "derive", "alloc" ] }
 fixed = { version = "1.15", default-features = false }
+num = { version = "0.4", default-features = false }
 
 [features]
 default = []
--- a/xcvm/lib/core/Cargo.toml
+++ b/xcvm/lib/core/Cargo.toml
@@ -11,8 +11,7 @@ scale-info = { version = "2.1.1", default-features = false, features = [
   "derive",
 ] }
 schemars = { version = "0.8.1", optional = true }
-serde = { version = "1", default-features = false, features = ["derive"] }
-serde_json = { version = "1", default-features = false, features = ["alloc"] }
+serde = { version = "1", default-features = false, features = [ "derive", "alloc" ] }
 fixed = { version = "1.15", default-features = false }
 
 [features]
--- a/xcvm/lib/core/Cargo.toml
+++ b/xcvm/lib/core/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+authors = ["Composable Developers"]
+edition = "2021"
+homepage = "https://composable.finance"
+name = "xcvm-core"
+version = "0.1.0"
+
+[dependencies]
+codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false }
+scale-info = { version = "2.1.1", default-features = false, features = [
+  "derive",
+] }
+schemars = { version = "0.8.1", optional = true }
+serde = { version = "1", default-features = false, features = ["derive"] }
+serde_json = { version = "1", default-features = false, features = ["alloc"] }
+fixed = { version = "1.15", default-features = false }
+
+[features]
+default = []
+std = ["dep:schemars"]
--- a/xcvm/lib/core/src/abstraction.rs
+++ b/xcvm/lib/core/src/abstraction.rs
@@ -10,9 +10,9 @@ pub struct Zero;
 pub struct Succ<T>(PhantomData<T>);
 
 mod _priv {
-  pub trait Sealed {}
-  impl Sealed for super::Zero {}
-  impl<X> Sealed for super::Succ<X> {}
+	pub trait Sealed {}
+	impl Sealed for super::Zero {}
+	impl<X> Sealed for super::Succ<X> {}
 }
 
 pub trait Nat: _priv::Sealed {
--- a/xcvm/lib/core/src/abstraction.rs
+++ b/xcvm/lib/core/src/abstraction.rs
@@ -9,7 +9,13 @@ pub struct Zero;
 /// Inductive case, 1 + x
 pub struct Succ<T>(PhantomData<T>);
 
-pub trait Nat {
+mod _priv {
+  pub trait Sealed {}
+  impl Sealed for super::Zero {}
+  impl<X> Sealed for super::Succ<X> {}
+}
+
+pub trait Nat: _priv::Sealed {
 	const VALUE: u8;
 }
 
--- a/xcvm/lib/core/src/abstraction.rs
+++ b/xcvm/lib/core/src/abstraction.rs
@@ -1,11 +1,25 @@
+//! https://en.wikipedia.org/wiki/Peano_axioms
+//! https://wiki.haskell.org/Peano_numbers#:~:text=Peano%20numbers%20are%20a%20simple,arithmetic%20due%20to%20their%20simplicity.
 use core::marker::PhantomData;
 
 /// Dummy structure used for the compiler to infer the position inside a tuple.
 /// Base case, 0
 pub struct Zero;
 /// Dummy structure used for the compiler to infer the position inside a tuple.
-/// Inductive case, x + 1
+/// Inductive case, 1 + x
 pub struct Succ<T>(PhantomData<T>);
+
+pub trait Nat {
+	const VALUE: u8;
+}
+
+impl Nat for Zero {
+	const VALUE: u8 = 0;
+}
+
+impl<X: Nat> Nat for Succ<X> {
+	const VALUE: u8 = 1 + X::VALUE;
+}
 
 /// Compile time indexing of an element of type `T` inside a structure of type `U`
 pub trait IndexOf<T, U> {
@@ -14,13 +28,14 @@ pub trait IndexOf<T, U> {
 
 /// Base case
 impl<T, U> IndexOf<T, Zero> for (T, U) {
-	const INDEX: u8 = 0;
+	const INDEX: u8 = Zero::VALUE;
 }
 
 /// Inductive case
 impl<T, U, V, X> IndexOf<T, Succ<X>> for (U, V)
 where
+	X: Nat,
 	V: IndexOf<T, X>,
 {
-	const INDEX: u8 = 1 + V::INDEX;
+	const INDEX: u8 = <Succ<X>>::VALUE;
 }
--- a/xcvm/lib/core/src/abstraction.rs
+++ b/xcvm/lib/core/src/abstraction.rs
@@ -1,0 +1,26 @@
+use core::marker::PhantomData;
+
+/// Dummy structure used for the compiler to infer the position inside a tuple.
+/// Base case, 0
+pub struct Zero;
+/// Dummy structure used for the compiler to infer the position inside a tuple.
+/// Inductive case, x + 1
+pub struct Succ<T>(PhantomData<T>);
+
+/// Compile time indexing of an element of type `T` inside a structure of type `U`
+pub trait IndexOf<T, U> {
+	const INDEX: u8;
+}
+
+/// Base case
+impl<T, U> IndexOf<T, Zero> for (T, U) {
+	const INDEX: u8 = 0;
+}
+
+/// Inductive case
+impl<T, U, V, X> IndexOf<T, Succ<X>> for (U, V)
+where
+	V: IndexOf<T, X>,
+{
+	const INDEX: u8 = 1 + V::INDEX;
+}
--- a/xcvm/lib/core/src/asset.rs
+++ b/xcvm/lib/core/src/asset.rs
@@ -69,7 +69,7 @@ pub struct ETH;
 pub struct USDT;
 pub struct USDC;
 
-/// List of XCVM compatible networks.
+/// List of XCVM compatible assets.
 /// The order matter and must not be changed, adding a network on the right is safe.
 pub type Assets = (InvalidAsset, (PICA, (ETH, (USDT, (USDC, ())))));
 
@@ -151,7 +151,7 @@ impl Amount {
 				.saturating_mul(FixedU128::<U16>::from_num(*x as u128).saturating_div(FixedU128::<
 					U16,
 				>::from_num(
-					100
+					u32::MAX
 				)))
 				.to_num(),
 		}
--- a/xcvm/lib/core/src/asset.rs
+++ b/xcvm/lib/core/src/asset.rs
@@ -7,7 +7,8 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use crate::abstraction::IndexOf;
 
 /// Newtype for XCVM assets ID. Must be unique for each asset and must never change.
-/// This ID is an opaque, arbitrary type from the XCVM protocol and no assumption must be made on how it is computed.
+/// This ID is an opaque, arbitrary type from the XCVM protocol and no assumption must be made on
+/// how it is computed.
 #[cfg_attr(feature = "std", derive(schemars::JsonSchema))]
 #[derive(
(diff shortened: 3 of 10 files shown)


seunlanlege (approved):
pretty dope stuff

xcvm/lib/core/src/asset.rs
```diff
@@ -0,0 +1,214 @@
...
+
+pub struct InvalidAsset;
+pub struct PICA;
+pub struct ETH;
+pub struct USDT;
+pub struct USDC;
+
+/// List of XCVM compatible networks.
+/// The order matter and must not be changed, adding a network on the right is safe.
+pub type Assets = (InvalidAsset, (PICA, (ETH, (USDT, (USDC, ())))));
```
dzmitry-lahoda, line 73: afaik USDT is not network, but token on TRX. USDC is on ETH.

xcvm/lib/core/src/asset.rs
```diff
@@ -0,0 +1,214 @@
...
+	}
+}
+
+impl Amount {
+	pub fn apply(&self, value: u128) -> u128 {
+		match self {
+			Amount::Fixed(Displayed(x)) => *x,
+			Amount::Ratio(x) => FixedU128::<U16>::from_num(value)
+				.saturating_mul(FixedU128::<U16>::from_num(*x as u128).saturating_div(FixedU128::<
+					U16,
```
dzmitry-lahoda, line 152: just it get this, ratio only can represend 1/100? i mean if i seen 1000, I can tell 990 or 10, not 999 or 1? Either fee or expected return?

hussein-aitlahcen: Oh, it's legacy, must be u32::MAX

hussein-aitlahcen: Nice catch

xcvm/lib/core/src/instruction.rs
```diff
@@ -0,0 +1,34 @@
...
+	/// For most of the network in fact, we need to provide the target address along the payload,
+	/// which can be encoded inside this single payload.
+	#[serde(rename_all = "snake_case")]
+	Call { encoded: Payload },
+	/// Spawn a sub-program on the target `network`.
+	///
+	/// The program will be spawned with the desired [`Assets`].
+	/// The salt is used to track the program when events are dispatched in the network.
+	#[serde(rename_all = "snake_case")]
+	Spawn { network: Network, salt: Vec<u8>, assets: Assets, program: Program<VecDeque<Self>> },
```
dzmitry-lahoda, line 33: > program: Program<VecDeque<Self>>

so it is 

`program: Program<VecDeque<Instruction<Network, Payload, Account, Assets>>>`

which is tied to Self::Network only? Which is better named TargetNetwork? 

So how can express Network1 -> Network2 -> Network2? Network1 -> Network2 -> Network3? Network1 -> Network1 -> Network2? 

Not sure it will work.

xcvm/lib/core/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+}
+
+impl<CurrentNetwork, Account, Assets> ProgramBuilder<CurrentNetwork, Account, Assets>
+where
+	CurrentNetwork: Network,
+	CurrentNetwork::EncodedCall: Into<Vec<u8>>,
+{
+	#[inline]
+	pub fn new(tag: Option<Vec<u8>>) -> Self {
+		ProgramBuilder { tag, instructions: VecDeque::new(), _marker: PhantomData }
```
dzmitry-lahoda, line 45: afaik `ProgramBuilder { tag, ..Default::default() } could work, so not sure

xcvm/lib/core/src/instruction.rs
dzmitry-lahoda: you will convert builder.okey

xcvm/lib/core/src/program.rs
```diff
@@ -0,0 +1,12 @@
+use alloc::vec::Vec;
+use codec::{Decode, Encode};
+use scale_info::TypeInfo;
+use serde::{Deserialize, Serialize};
+
+#[cfg_attr(feature = "std", derive(schemars::JsonSchema))]
+#[derive(Clone, PartialEq, Eq, Debug, Encode, Decode, TypeInfo, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Program<Instructions> {
+	pub tag: Option<Vec<u8>>,
```
dzmitry-lahoda, line 10: it can be just tag: Vec<u8> ? may be. (can be changed later to 256 values bounded vec)

dzmitry-lahoda (approved):
types magic

xcvm/lib/core/src/lib.rs
benluelo: `ProgramBuilder` doesn't implement default, so it wouldn't work (you could do `		ProgramBuilder { tag, instructions: VecDeque::new(), _marker: Default::default() }` though, although that's probably not much of an improvement lol)

xcvm/lib/core/src/network.rs
```diff
@@ -0,0 +1,96 @@
...
+	Ord,
+	Debug,
+	Encode,
+	Decode,
+	TypeInfo,
+	Serialize,
+	Deserialize,
+)]
+#[repr(transparent)]
+pub struct NetworkID(pub u8);
```
benluelo, line 26: Nit: Should be `NetworkId`

xcvm/lib/core/src/network.rs
```diff
@@ -0,0 +1,96 @@
...
+pub struct InvalidNetwork;
+/// Composable Picasso (Kusama parachain)
+pub struct Picasso;
+/// Ethereum mainnet
+pub struct Ethereum;
+/// Juno (Cosmos) mainnet
+pub struct Juno;
+
+/// List of networks supported by XCVM.
+/// The order matter and must not be changed, adding a network on the right is safe.
```
benluelo, line 63:
```suggestion
/// The order matters and must not be changed, adding a network on the right is safe.
```

xcvm/lib/core/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+version = "0.1.0"
+
+[dependencies]
+codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false }
+scale-info = { version = "2.1.1", default-features = false, features = [
+  "derive",
+] }
+schemars = { version = "0.8.1", optional = true }
+serde = { version = "1", default-features = false, features = ["derive"] }
+serde_json = { version = "1", default-features = false, features = ["alloc"] }
```
benluelo, line 15: does serde need the alloc feature as well? or does serde_json:alloc bring in serde:alloc automatically

benluelo (approved):
🚀

xcvm/lib/core/Cargo.toml
hussein-aitlahcen: not sure about that, but isn't serde just here for the `Serialize` and `Deserialize` traits + derive?

benluelo: I was just referencing this: https://serde.rs/no-std.html, but it seems that serde_json/alloc brings in serde/alloc (https://github.com/serde-rs/json/blob/master/Cargo.toml#L54)

xcvm/lib/core/src/asset.rs
```diff
@@ -0,0 +1,215 @@
...
+		Displayed(x)
+	}
+}
+
+#[cfg_attr(feature = "std", derive(schemars::JsonSchema))]
+#[derive(
+	Clone, PartialEq, Eq, PartialOrd, Ord, Debug, Encode, Decode, TypeInfo, Serialize, Deserialize,
+)]
+#[serde(rename_all = "snake_case")]
+pub enum Amount {
```
aeryz, line 135: Implementing [zero trait](https://docs.rs/num/latest/num/traits/trait.Zero.html) for this might be a good idea for a quick zero check. Otherwise, you would have to use `match` and check the underlying values of the variants.

xcvm/lib/core/src/abstraction.rs
```diff
@@ -0,0 +1,26 @@
...
+	const INDEX: u8 = 0;
+}
+
+/// Inductive case
+impl<T, U, V, X> IndexOf<T, Succ<X>> for (U, V)
+where
+	V: IndexOf<T, X>,
+{
+	const INDEX: u8 = 1 + V::INDEX;
+}
```
andor0, line 26: [Peano numbers](https://wiki.haskell.org/Peano_numbers) can be added as docs

xcvm/lib/core/src/asset.rs
```diff
@@ -0,0 +1,215 @@
...
+	Ord,
+	Debug,
+	Encode,
+	Decode,
+	TypeInfo,
+	Serialize,
+	Deserialize,
+)]
+#[repr(transparent)]
+pub struct AssetID(pub u32);
```
aeryz, line 28: Asset id's are represented by `u128` in [XCVM spec](https://github.com/ComposableFi/composable/blob/main/xcvm/SPEC.md#asset-registries).

xcvm/lib/core/src/instruction.rs
```diff
@@ -0,0 +1,34 @@
...
+	///
+	/// On picasso, this will be a SCALE encoded dispatch call.
+	/// On ethereum, an ethereum ABI encoded call.
+	/// On cosmos, a raw json WasmMsg call.
+	///
+	/// Depending on the network, the payload might be more structured than the base call.
+	/// For most of the network in fact, we need to provide the target address along the payload,
+	/// which can be encoded inside this single payload.
+	#[serde(rename_all = "snake_case")]
+	Call { encoded: Payload },
```
aeryz, line 27: Not having `address` field here results in platform-specific implementation in the contract side. For example we expect
```json
{
    "address": "blabla",
    "payload": "{Json Encoded WasmMsg}"
}
```
in our cosmwasm interpreter contract. We might have optional `address` field, which would be deducted when serialized if there is no address specified so that we only pay for what we use.

xcvm/lib/core/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+mod program;
+mod protocol;
+
+use core::marker::PhantomData;
+
+pub use crate::{asset::*, instruction::*, network::*, program::*, protocol::*};
+use alloc::{collections::VecDeque, vec::Vec};
+
+#[inline]
+pub fn serialize_json<T: serde::Serialize>(
```
aeryz, line 18: Do we need to have this? Users can use `serde_json` or `serde-json-wasm` on their choice.

xcvm/lib/core/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+	program: &Program<T>,
+) -> Result<Vec<u8>, serde_json::Error> {
+	serde_json::to_vec(program)
+}
+
+#[inline]
+pub fn deserialize_json<T: serde::de::DeserializeOwned>(
```
aeryz, line 25: [see](https://github.com/ComposableFi/composable/pull/1392/files#r943853544)

xcvm/lib/core/src/asset.rs
```diff
@@ -0,0 +1,215 @@
...
+}
+
+pub struct InvalidAsset;
+pub struct PICA;
+pub struct ETH;
+pub struct USDT;
+pub struct USDC;
+
+/// List of XCVM compatible assets.
+/// The order matter and must not be changed, adding a network on the right is safe.
```
aeryz, line 73: nit: matters

xcvm/lib/core/src/instruction.rs
aeryz: Actually, current implementation of the interpreter accepts only `msg` field of `WasmMsg`. I think it should be the whole `WasmMsg` and it should only accept the `Execute` variant of `WasmMsg` which is:
```rust
pub enum WasmMsg {
    Execute {
        contract_addr: String,
        msg: Binary,
        funds: Vec,
    },
    ...
}
```
This already contains the `contract_addr` so we won't need a seperate address field.

aeryz: We now expect `CosmosMsg` in `Call` instruction so this is not a problem anymore :)

xcvm/lib/core/Cargo.toml
hussein-aitlahcen: Remove serde_json and added alllloc

xcvm/lib/core/src/asset.rs
hussein-aitlahcen: I've moved from Fixed/Ratio to a linear equation and implemented the according operations

xcvm/lib/core/src/lib.rs
hussein-aitlahcen: Removed

xcvm/lib/core/src/asset.rs
hussein-aitlahcen: Fixed

aeryz (approved):
LGTM!

xcvm/lib/core/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+  "derive",
+] }
+schemars = { version = "0.8.1", optional = true }
+serde = { version = "1", default-features = false, features = [ "derive", "alloc" ] }
+fixed = { version = "1.15", default-features = false }
+num = { version = "0.4", default-features = false }
+
+[features]
+default = []
+std = ["dep:schemars"]
```
KaiserKarel, line 20: I haven't seen this syntax yet though lol

hussein-aitlahcen: https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies

xcvm/lib/core/src/asset.rs
```diff
@@ -0,0 +1,262 @@
...
+
+pub struct InvalidAsset;
+pub struct PICA;
+pub struct ETH;
+pub struct USDT;
+pub struct USDC;
+
+/// List of XCVM compatible assets.
+/// /!\ The order matters and must not be changed, adding a network on the right is safe.
+pub type Assets = (InvalidAsset, (PICA, (ETH, (USDT, (USDC, ())))));
```
KaiserKarel, line 75: Macro to make this

```
assets![InvalidAsset, PICA, ETH, USDT, USDC]?
```

xcvm/lib/core/src/network.rs
```diff
@@ -0,0 +1,96 @@
...
+pub struct InvalidNetwork;
+/// Composable Picasso (Kusama parachain)
+pub struct Picasso;
+/// Ethereum mainnet
+pub struct Ethereum;
+/// Juno (Cosmos) mainnet
+pub struct Juno;
+
+/// List of networks supported by XCVM.
+/// /!\ The order matters and must not be changed, adding a network on the right is safe.
```
KaiserKarel, line 63:
```suggestion
// /!\ The order matters and must not be changed, adding a network on the right is safe.
```

Comments for maintainers should not be doc comments btw

xcvm/lib/core/src/network.rs
```diff
@@ -0,0 +1,96 @@
...
+pub type Networks = (InvalidNetwork, (Picasso, (Ethereum, (Juno, ()))));
+
+/// Type implement network must be part of [`Networks`], otherwise invalid.
```
KaiserKarel, line 66: Same here, describe what Network is first, and restrictions related to maintaining should be a regular comment.

xcvm/lib/core/src/asset.rs
hussein-aitlahcen: Not worth the overhead of instroducing an extra macro tbh, that will invalidate most IDE typechecking hints also

KaiserKarel (approved):

aeryz (approved):

cor (approved):
LGTM 🚀

andor0 (approved):

PR #1400, opened 2022-08-02
[Merged by Bors] - [CU-345zb4r] Andres/add id to vesting schedules

andresmechali:
## Issue
https://app.clickup.com/t/345zb4r

### DIFF
--- a/frame/composable-traits/src/vesting.rs
+++ b/frame/composable-traits/src/vesting.rs
@@ -3,6 +3,9 @@ use frame_support::pallet_prelude::*;
 use scale_info::TypeInfo;
 use sp_runtime::traits::AtLeast32Bit;
 
+#[cfg(feature = "std")]
+use serde::{Deserialize, Serialize};
+
 /// An object allowing us to transfer funds from one account to another in a vested fashion.
 pub trait VestedTransfer {
 	type AccountId;
@@ -50,6 +53,7 @@ pub enum VestingWindow<BlockNumber, Moment> {
 /// Benefits would be granted gradually, `per_period` amount every `window.period`
 /// of blocks after `window.start`.
 #[derive(Clone, Encode, Decode, PartialEq, Eq, RuntimeDebug, MaxEncodedLen, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 pub struct VestingSchedule<VestingScheduleId, BlockNumber, Moment, Balance: HasCompact> {
 	/// Vesting schedule id
 	pub vesting_schedule_id: VestingScheduleId,
--- a/frame/composable-traits/src/vesting.rs
+++ b/frame/composable-traits/src/vesting.rs
@@ -22,12 +22,7 @@ pub trait VestedTransfer {
 		asset: Self::AssetId,
 		from: &Self::AccountId,
 		to: &Self::AccountId,
-		schedule: VestingSchedule<
-			Self::VestingScheduleId,
-			Self::BlockNumber,
-			Self::Moment,
-			Self::Balance,
-		>,
+		schedule: VestingScheduleInput<Self::BlockNumber, Self::Moment, Self::Balance>,
 	) -> DispatchResult;
 }
 
@@ -66,6 +61,17 @@ pub struct VestingSchedule<VestingScheduleId, BlockNumber, Moment, Balance: HasC
 	pub per_period: Balance,
 	/// Amout already claimed
 	pub already_claimed: Balance,
+}
+
+#[derive(Clone, Encode, Decode, PartialEq, Eq, RuntimeDebug, MaxEncodedLen, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct VestingScheduleInput<BlockNumber, Moment, Balance: HasCompact> {
+	pub window: VestingWindow<BlockNumber, Moment>,
+	/// Number of vest
+	pub period_count: u32,
+	/// Amount of tokens to release per vest
+	#[codec(compact)]
+	pub per_period: Balance,
 }
 
 pub enum VestingWindowResult<BlockNumber, Moment> {
--- a/frame/composable-traits/src/vesting.rs
+++ b/frame/composable-traits/src/vesting.rs
@@ -1,7 +1,7 @@
 use codec::{HasCompact, MaxEncodedLen};
 use frame_support::pallet_prelude::*;
 use scale_info::TypeInfo;
-use sp_runtime::traits::AtLeast32Bit;
+use sp_runtime::traits::{AtLeast32Bit, Zero};
 
 #[cfg(feature = "std")]
 use serde::{Deserialize, Serialize};
@@ -15,7 +15,7 @@ pub trait VestedTransfer {
 	type Balance: HasCompact;
 	type MinVestedTransfer: Get<Self::Balance>;
 	type VestingScheduleId;
-	type VestingScheduleCount;
+	type VestingScheduleNonce;
 
 	/// Transfer `asset` from `from` to `to` vested based on `schedule`.
 	fn vested_transfer(
@@ -141,6 +141,19 @@ impl<
 		self.per_period
 			.checked_mul(&unrealized.into())
 			.expect("ensured non-overflow total amount; qed")
+	}
+
+	pub fn from_input(
+		vesting_schedule_id: VestingScheduleId,
+		vesting_schedule_input: VestingScheduleInput<BlockNumber, Moment, Balance>,
+	) -> VestingSchedule<VestingScheduleId, BlockNumber, Moment, Balance> {
+		VestingSchedule {
+			vesting_schedule_id,
+			window: vesting_schedule_input.window,
+			per_period: vesting_schedule_input.per_period,
+			period_count: vesting_schedule_input.period_count,
+			already_claimed: Zero::zero(),
+		}
 	}
 }
 
--- a/frame/composable-traits/src/vesting.rs
+++ b/frame/composable-traits/src/vesting.rs
@@ -149,6 +149,7 @@ mod tests {
 			window: MomentBased { start: 1u64, period: 10u64 },
 			period_count: 100,
 			per_period: 1u64,
+			already_claimed: 0_u64,
 		};
 		assert!(!vesting_schedule_time_based.is_zero_period());
 		vesting_schedule_time_based.window = MomentBased { start: 1, period: 0 };
@@ -159,6 +160,7 @@ mod tests {
 			window: BlockNumberBased { start: 1u64, period: 10u64 },
 			period_count: 100,
 			per_period: 1u64,
+			already_claimed: 0_u64,
 		};
 		assert!(!vesting_schedule_block_number_based.is_zero_period());
 		vesting_schedule_block_number_based.window = BlockNumberBased { start: 1, period: 0 };
@@ -172,6 +174,7 @@ mod tests {
 			window: MomentBased { start: 1u64, period: 10u64 },
 			period_count: 100,
 			per_period: 1u64,
+			already_claimed: 0_u64,
 		};
 		match vesting_schedule_time_based.end() {
 			None => {},
@@ -185,6 +188,7 @@ mod tests {
 			window: BlockNumberBased { start: 1u64, period: 10u64 },
 			period_count: 100,
 			per_period: 1u64,
+			already_claimed: 0_u64,
 		};
 		match vesting_schedule_block_number_based.end() {
 			None => {},
@@ -202,6 +206,7 @@ mod tests {
 			window: BlockNumberBased { start: 1u64, period: 10u64 },
 			period_count: 100,
 			per_period: 1u64,
+			already_claimed: 0_u64,
 		};
 		assert_eq!(vesting_schedule.total_amount().unwrap(), 100)
 	}
@@ -214,6 +219,7 @@ mod tests {
 			window: MomentBased { start: 1u64, period: 10u64 },
 			period_count: 100,
 			per_period: 1u64,
+			already_claimed: 0_u64,
 		};
 		assert_eq!(vesting_schedule_time_based.locked_amount(1, 1), 100);
 		assert_eq!(vesting_schedule_time_based.locked_amount(1, 11), 99);
@@ -224,6 +230,7 @@ mod tests {
 			window: BlockNumberBased { start: 1u64, period: 10u64 },
 			period_count: 100,
 			per_period: 1u64,
+			already_claimed: 0_u64,
 		};
 		assert_eq!(vesting_schedule_block_number_based.locked_amount(1, 1), 100);
 		assert_eq!(vesting_schedule_block_number_based.locked_amount(11, 1), 99);
--- a/frame/composable-traits/src/vesting.rs
+++ b/frame/composable-traits/src/vesting.rs
@@ -15,6 +15,7 @@ pub trait VestedTransfer {
 	type Balance: HasCompact;
 	type MinVestedTransfer: Get<Self::Balance>;
 	type VestingScheduleId;
+	type VestingScheduleCount;
 
 	/// Transfer `asset` from `from` to `to` vested based on `schedule`.
 	fn vested_transfer(
--- a/frame/composable-traits/src/vesting.rs
+++ b/frame/composable-traits/src/vesting.rs
@@ -44,12 +44,12 @@ pub enum VestingWindow<BlockNumber, Moment> {
 	},
 }
 
-/// VestingScheduleId type for claiming
+/// VestingScheduleId type for claiming.
 #[cfg_attr(feature = "std", derive(serde::Deserialize, serde::Serialize))]
 #[derive(Clone, Encode, Decode, PartialEq, Eq, RuntimeDebug, MaxEncodedLen, TypeInfo)]
-pub enum Schedules<Id> {
+pub enum Schedules<Id, MaxVestingSchedules> {
 	All,
-	One(Id),
+	Many(BoundedVec<Id, MaxVestingSchedules>),
 }
 
 /// The vesting schedule.
@@ -71,7 +71,7 @@ pub struct VestingSchedule<VestingScheduleId, BlockNumber, Moment, Balance: HasC
 	pub already_claimed: Balance,
 }
 
-/// Vesting schedule input
+/// Vesting schedule input, which is used to create a VestingSchedule.
 ///
 /// This is used for creating a VestingSchedule
 #[derive(Clone, Encode, Decode, PartialEq, Eq, RuntimeDebug, MaxEncodedLen, TypeInfo)]
--- a/frame/composable-traits/src/vesting.rs
+++ b/frame/composable-traits/src/vesting.rs
@@ -45,6 +45,8 @@ pub enum VestingWindow<BlockNumber, Moment> {
 /// of blocks after `window.start`.
 #[derive(Clone, Encode, Decode, PartialEq, Eq, RuntimeDebug, MaxEncodedLen, TypeInfo)]
 pub struct VestingSchedule<BlockNumber, Moment, Balance: HasCompact> {
+	/// Vesting schedule id
+	/// pub vesting_schedule_id: VestingScheduleId /// TODO: create VestingScheduleId type
 	pub window: VestingWindow<BlockNumber, Moment>,
 	/// Number of vest
 	pub period_count: u32,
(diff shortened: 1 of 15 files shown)


ComposableFiB0t:
Task linked: [CU-345zb4r Add an ID to track vesting schedules individually](https://app.clickup.com/t/345zb4r)

frame/vesting/src/lib.rs
```diff
@@ -127,6 +141,20 @@ pub mod module {
...
+			+ Eq
+			+ Debug
+			+ Zero
+			+ SafeAdd
+			+ One
+			+ Ord
+			+ FullCodec
+			+ MaxEncodedLen
+			+ MaybeSerializeDeserialize
+			+ TypeInfo;
```
benluelo, line 157: Consider using [this](https://docs.rs/frame-support/latest/frame_support/pallet_prelude/trait.Member.html) and then add any other bounds you need

frame/vesting/src/lib.rs
```diff
@@ -181,6 +217,13 @@ pub mod module {
 		ValueQuery,
 	>;
 
+	/// Counter used to uniquely identify vesting schedules within this pallet.
+	#[pallet::storage]
+	#[pallet::getter(fn vesting_schedules_count)]
+	#[allow(clippy::disallowed_types)] // nonce, ValueQuery is OK
+	pub type VestingScheduleCount<T: Config> =
```
benluelo, line 224: I would rename this to `VestingScheduleNonce` (or something similar)

frame/vesting/src/lib.rs
```diff
@@ -156,9 +186,15 @@ pub mod module {
...
 			schedule: VestingScheduleOf<T>,
+			vesting_schedule_id: T::VestingScheduleId,
 		},
 		/// Claimed vesting. \[who, locked_amount\]
-		Claimed { who: AccountIdOf<T>, asset: AssetIdOf<T>, locked_amount: BalanceOf<T> },
+		Claimed {
+			who: AccountIdOf<T>,
+			asset: AssetIdOf<T>,
+			locked_amount: BalanceOf<T>,
+			vesting_schedule_id: Option<T::VestingScheduleId>,
```
benluelo, line 196:
```suggestion
			who: AccountIdOf<T>,
			asset: AssetIdOf<T>,
			vesting_schedule_id: Option<T::VestingScheduleId>,
			locked_amount: BalanceOf<T>,
```

benluelo: Just a nit, but this order reads better since it goes identifier -> data (if that makes sense)

frame/vesting/src/lib.rs
```diff
@@ -326,45 +395,91 @@ impl<T: Config> VestedTransfer for Pallet<T> {
...
 		Ok(locked)
 	}
 
-	/// Returns locked balance based on current block number.
-	fn locked_balance(who: &AccountIdOf<T>, asset: AssetIdOf<T>) -> BalanceOf<T> {
+	/// Returns total locked balance and balance per vesting schedule, based on current block number
+	fn locked_balance(
+		who: &AccountIdOf<T>,
+		asset: AssetIdOf<T>,
+		vesting_schedule_id: Option<T::VestingScheduleId>,
```
benluelo, line 428: Consider a more strongly-typed enum here:

```rust
enum Schedules<Id> {
    All,
    One(Id),
}
```

benluelo: Could also consider a `BoundedVec<_>` to allow for claiming any amount at once, rather than just all or nothing

frame/vesting/src/lib.rs
```diff
@@ -300,21 +358,32 @@ impl<T: Config> VestedTransfer for Pallet<T> {
...
 		ensure!(from != to, Error::<T>::TryingToSelfVest);
 
+		let vesting_schedule_id = Self::VestingScheduleCount::increment()?; // VestingScheduleCount::<T>::increment()?;
+		let schedule: VestingScheduleOf<T> = VestingSchedule {
+			vesting_schedule_id,
+			window: schedule_input.window,
+			period_count: schedule_input.period_count,
+			per_period: schedule_input.per_period,
+			already_claimed: Zero::zero(),
+		};
```
benluelo, line 380: perhaps an associated function on `VestingSchedule`, `fn from_input(id, input)`? would make this cleaner and be useful in tests as well

frame/vesting/src/lib.rs
andresmechali: ✅

frame/vesting/src/lib.rs
andresmechali: ✅

frame/composable-traits/src/vesting.rs
```diff
@@ -39,12 +50,49 @@ pub enum VestingWindow<BlockNumber, Moment> {
...
+	Encode,
+	Decode,
+	PartialEqNoBound,
+	EqNoBound,
+	RuntimeDebugNoBound,
+	MaxEncodedLen,
+	TypeInfo,
+)]
+#[scale_info(skip_type_params(MaxVestingSchedules))]
+pub enum Schedules<Id: Clone + Eq + PartialEq + Debug, MaxVestingSchedules: Get<u32>> {
```
vimukthi-git, line 66:
```suggestion
pub enum VestingScheduleIdSet<Id: Clone + Eq + PartialEq + Debug, MaxVestingSchedules: Get<u32>> {
```

frame/vesting/src/lib.rs
```diff
@@ -300,72 +359,116 @@ impl<T: Config> VestedTransfer for Pallet<T> {
...
 	}
 }
 
 impl<T: Config> Pallet<T> {
-	fn do_claim(who: &AccountIdOf<T>, asset: AssetIdOf<T>) -> Result<BalanceOf<T>, DispatchError> {
-		let locked = Self::locked_balance(who, asset);
+	fn do_claim(
+		who: &AccountIdOf<T>,
+		asset: AssetIdOf<T>,
+		vesting_schedule_id: Schedules<T::VestingScheduleId, T::MaxVestingSchedules>, /* Option<T::VestingScheduleId>, */
```
vimukthi-git, line 408:
```suggestion
		vesting_schedule_ids: Schedules<T::VestingScheduleId, T::MaxVestingSchedules>, /* Option<T::VestingScheduleId>, */
```

frame/vesting/src/tests.rs
```diff
@@ -435,22 +512,31 @@ fn claim_works() {
...
 
 		System::set_block_number(21);
 		// claim more
-		assert_ok!(Vesting::claim(Origin::signed(BOB), MockCurrencyId::BTC));
+		assert_ok!(Vesting::claim(
+			Origin::signed(BOB),
+			MockCurrencyId::BTC,
+			VestingScheduleIdSet::All
+		));
+		println!("{:?}", VestingSchedules::<Runtime>::get(BOB, MockCurrencyId::BTC));
```
vimukthi-git, line 539: Need to verify that events are emitted with required fields

frame/vesting/src/lib.rs
```diff
@@ -282,12 +340,13 @@ pub mod module {
...
 			asset: AssetIdOf<T>,
+			vesting_schedule_id: VestingScheduleIdSet<T::VestingScheduleId, T::MaxVestingSchedules>,
 		) -> DispatchResult {
 			let _ = ensure_signed(origin)?;
 			let who = T::Lookup::lookup(dest)?;
-			let locked_amount = Self::do_claim(&who, asset)?;
+			let locked_amount = Self::do_claim(&who, asset, vesting_schedule_id.clone())?;
 
-			Self::deposit_event(Event::Claimed { who, asset, locked_amount });
+			Self::deposit_event(Event::Claimed { who, asset, locked_amount, vesting_schedule_id });
```
vimukthi-git, line 349: We could move this event into `do_claim`

vimukthi-git (approved):
LGTM :rock: 
.. minor comments

frame/vesting/src/lib.rs
```diff
@@ -300,72 +359,127 @@ impl<T: Config> VestedTransfer for Pallet<T> {
...
+					let ids_to_claim: Vec<_> = match vesting_schedule_ids {
+						VestingScheduleIdSet::All => schedules.keys().copied().collect(),
+						VestingScheduleIdSet::Many(ids) => ids.into_inner(),
+						VestingScheduleIdSet::One(id) => vec![id],
+					};
+
+					let mut ids_to_remove = vec![];
+
+					let mut total_locked_amount: BalanceOf<T> = Zero::zero();
+
```
vimukthi-git, line 441: IMO no need of this extra space :)

frame/vesting/src/lib.rs
```diff
@@ -300,72 +359,127 @@ impl<T: Config> VestedTransfer for Pallet<T> {
...
+						let total_amount = schedule.total_amount()?;
+						let available_amount = total_amount.safe_sub(&locked_amount)?;
+
+						// if the schedule id is specified, we update the claimed
+						// amount for the specified schedule
+						schedule.already_claimed = available_amount;
+
+						if locked_amount.is_zero() {
+							ids_to_remove.push(id_to_claim);
```
vimukthi-git, line 459: Why not just remove the `id_to_claim` from schedules here?

frame/vesting/src/lib.rs
```diff
@@ -300,72 +359,127 @@ impl<T: Config> VestedTransfer for Pallet<T> {
...
+	fn locked_balance(
+		who: &AccountIdOf<T>,
+		asset: AssetIdOf<T>,
+		vesting_schedule_ids: VestingScheduleIdSet<T::VestingScheduleId, T::MaxVestingSchedules>,
+	) -> Result<BalanceOf<T>, DispatchError> {
+		<VestingSchedules<T>>::try_mutate_exists(who, asset, |maybe_schedules| {
+			match maybe_schedules {
+				Some(schedules) => {
+					let ids_to_claim: Vec<_> = match vesting_schedule_ids {
+						VestingScheduleIdSet::All => schedules.keys().copied().collect(),
```
vimukthi-git, line 433: Nit: I think a nicer pattern here would be to do iterator syntax on the ids and map to a single small function that does the part inside the for loop from line 442.

frame/vesting/src/lib.rs
andresmechali: It's a bit tricky, because if we remove it here, then we lose track of the `already_claimed` balance for this schedule, and can not use it on the next loop. We can also accumulate all the balances from the removed schedules before removing and then subtract this from the `total_locked_amount`, but seems like the code would get harder to understand in the future.

vimukthi-git: So the goal seems to be to get the `total_locked_amount`. If I understand correctly the calculation for it would still be correct without the removed schedules as sum of `total_locked_amount` for removed ones is 0.

andresmechali: You are actually right. A previous implementation didn't allow me to do this, but the current one does. Well spotted! 👁️

frame/vesting/src/lib.rs
andresmechali: done ✅

vimukthi-git (approved):
LGTM

dzmitry-lahoda (approved):
Just question. Do we have any vesting on Picasso storage right now? If not - good. If yes - are changed backward compatible?

As I see it changed mostly to allow multi vestings and that it.

frame/composable-traits/src/vesting.rs
```diff
@@ -39,12 +50,50 @@ pub enum VestingWindow<BlockNumber, Moment> {
...
+	Decode,
+	PartialEqNoBound,
+	EqNoBound,
+	RuntimeDebugNoBound,
+	MaxEncodedLen,
+	TypeInfo,
+)]
+#[scale_info(skip_type_params(MaxVestingSchedules))]
+pub enum VestingScheduleIdSet<Id: Clone + Eq + PartialEq + Debug, MaxVestingSchedules: Get<u32>> {
+	All,
```
dzmitry-lahoda, line 67: please add doc comments for each entry

dzmitry-lahoda (changes requested):
please prove that claiming one schedule does not unlocks others prematurely

andresmechali:
> please prove that claiming one schedule does not unlocks others prematurely

This is proved in `claim_with_id_works`, which claims schedules one by one and asserts the total locked value each time.

frame/composable-traits/src/vesting.rs
andresmechali: ✅

andresmechali:
> Just question. Do we have any vesting on Picasso storage right now? If not - good. If yes - are changed backward compatible?
> 
> As I see it changed mostly to allow multi vestings and that it.

We currently don't have any

dzmitry-lahoda:
how claim works
1. claim from schedules we have
2. run over schedules all acc locked
3. set lock to updated

so for me is better
1. claim shedules we have
2. get lock and remove claimed
3. set new lock after sub

coder will be shorted, easier to read, faster to run imho

vimukthi-git:
Good observation @dzmitry-lahoda :)

We can save the second loop on `locked_balance` function by just subtracting the claimed amount from the total locked and setting the lock back to that value. Seems being able to reuse the same `locked_balance` function has caused the current code to be non-optimal.

dzmitry-lahoda (approved):

dzmitry-lahoda:
@andresmechali please consider https://github.com/ComposableFi/composable/pull/1400#issuecomment-1206532501

vimukthi-git (approved):
LGTM, few adjustments though

frame/vesting/src/lib.rs
```diff
@@ -300,72 +358,209 @@ impl<T: Config> VestedTransfer for Pallet<T> {
...
+			_ => Zero::zero(),
+		};
+
+		Ok(total_unclaimed)
+	}
+
+	/// Returns balance available to claim for a given account, asset and vesting schedules, based
+	/// on current block number
+	/// It also updates the already claimed balance and removes completely vested schedules
+	fn available_balance(
```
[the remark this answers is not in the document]
andresmechali: Well it's not exactly the `unlocked` balance, as it subtracts the balance that has already been claimed, leaving the "unlocked balance which has not been claimed yet, and therefore is available for claiming". I can call it "unlocked" if it is still clear what it represents though.

frame/vesting/src/lib.rs
```diff
@@ -300,72 +358,209 @@ impl<T: Config> VestedTransfer for Pallet<T> {
...
 		} else {
-			T::Currency::set_lock(VESTING_LOCK_ID, asset, who, locked)?;
+			T::Currency::set_lock(VESTING_LOCK_ID, asset, who, new_locked_amount)?;
 		}
-		Ok(locked)
+
+		Self::deposit_event(Event::Claimed {
+			who: who.clone(),
+			asset,
+			locked_amount: new_locked_amount,
```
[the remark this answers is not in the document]
andresmechali: Yes, I kept it as it was but I agree that it should also emit the claimed amount.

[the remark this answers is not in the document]
andresmechali: added ✅

frame/vesting/src/lib.rs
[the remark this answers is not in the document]
vimukthi-git: I also understood what you wrote above, that's why I suggested this name contrasting with the other function `locked_balance`.

vimukthi-git (approved):
LGTM, Remaining comments are not critical.

frame/vesting/src/lib.rs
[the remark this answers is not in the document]
benluelo: `unlocked_claimable_balance`?

frame/vesting/src/lib.rs
```diff
@@ -374,24 +618,34 @@ impl<T: Config> Pallet<T> {
 		asset: AssetIdOf<T>,
 		schedules: Vec<VestingScheduleOf<T>>,
 	) -> DispatchResult {
-		let bounded_schedules: BoundedVec<VestingScheduleOf<T>, T::MaxVestingSchedules> =
-			schedules.try_into().map_err(|_| Error::<T>::MaxVestingSchedulesExceeded)?;
-
 		// empty vesting schedules cleanup the storage and unlock the fund
-		if bounded_schedules.len().is_zero() {
+		if schedules.len().is_zero() {
```
benluelo, line 622: I believe you can use `.is_empty()` here

frame/vesting/src/lib.rs
```diff
@@ -374,24 +618,34 @@ impl<T: Config> Pallet<T> {
...
 			T::Currency::remove_lock(VESTING_LOCK_ID, asset, who)?;
 			return Ok(())
 		}
 
+		ensure!(
+			schedules.len() as u32 <= T::MaxVestingSchedules::get(),
+			Error::<T>::MaxVestingSchedulesExceeded
+		);
+
+		let bounded_schedules: BoundedBTreeMap<_, _, _> = schedules
```
benluelo, line 633:
```suggestion
		let bounded_schedules = schedules
```

benluelo: Type annotation is redundant as it's also written in the `.collect()` call

frame/vesting/src/lib.rs
```diff
@@ -300,72 +405,211 @@ impl<T: Config> VestedTransfer for Pallet<T> {
...
+		let vesting_schedule_id = Self::VestingScheduleNonce::increment()?;
+		let schedule = VestingSchedule::from_input(vesting_schedule_id, schedule_input);
+
 		let schedule_amount = ensure_valid_vesting_schedule::<T>(&schedule)?;
 
-		let total_amount = Self::locked_balance(to, asset)
-			.checked_add(&schedule_amount)
-			.ok_or(ArithmeticError::Overflow)?;
+		let locked =
+			Self::locked_balance(to, asset, VestingScheduleIdSet::All).unwrap_or(Zero::zero());
```
benluelo, line 426: use `unwrap_or_else` as per clippy

frame/vesting/src/lib.rs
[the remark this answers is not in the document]
andresmechali: I used to call it `claimable_balance` so Ben's suggestion is a good mix

frame/vesting/src/lib.rs
andresmechali: Doesn't work if I omit it (`cannot infer type`)

frame/vesting/src/lib.rs
```diff
@@ -236,31 +292,67 @@ pub mod module {
...
+		///
+		/// NOTE: This will unlock all schedules through the current block.
 		#[pallet::weight(<T as Config>::WeightInfo::vested_transfer())]
 		pub fn vested_transfer(
 			origin: OriginFor<T>,
 			from: <T::Lookup as StaticLookup>::Source,
 			beneficiary: <T::Lookup as StaticLookup>::Source,
 			asset: AssetIdOf<T>,
-			schedule: VestingScheduleOf<T>,
+			schedule_input: VestingScheduleInputOf<T>,
```
dzmitry-lahoda, line 337: `schedule_input: VestingScheduleInputOf`so you have var name `shedule_input` which can be pascal cased and prefixed with mod and you get VestingSchdeuleInput. So any naming you pick will bring zero information. Most extenrable naming and neutral is `input`, no prefix.

frame/vesting/src/lib.rs
```diff
@@ -300,72 +405,206 @@ impl<T: Config> VestedTransfer for Pallet<T> {
...
 	}
 }
 
 impl<T: Config> Pallet<T> {
-	fn do_claim(who: &AccountIdOf<T>, asset: AssetIdOf<T>) -> Result<BalanceOf<T>, DispatchError> {
-		let locked = Self::locked_balance(who, asset);
-		if locked.is_zero() {
+	fn do_claim(
+		who: &AccountIdOf<T>,
+		asset: AssetIdOf<T>,
```
dzmitry-lahoda, line 454: not sure if this API is good. you have to tell what assets you what to claim and what schedules. while claim all assists and claim all schedules is best thing. i may be claim with to parameters so it can be bot automated to some aggregated account.

dzmitry-lahoda (approved):

andor0 (approved):

frame/vesting/src/lib.rs
andresmechali: Yeah makes sense. Just renamed it to `vesting_info`, inspired by https://github.com/paritytech/substrate/blob/master/frame/vesting/src/lib.rs

vimukthi-git (approved):

Dom-Roth (approved):

andor0 (approved):

PR #1372, opened 2022-07-25
feat(vamm): Add `swap_simulation` functionality

Cardosaum:
## Issue
[CU-2ng5mmh](https://app.clickup.com/t/2ng5mmh)

## Description

This commit is a lengthy one, other changes were also made while adding
this new functionality. The following list enumerates some of the
noteworthy changes in the code base for this Pallet:

- The capability of simulating a new swap for the Vamm Pallet,
- Tests for the `swap_simulation` function,
- Helper Functions and Types were moved to specific inner modules,
- Documentation for all public functions, types and fields was added,
- Documentation now pointing to all the possible functions throwing a specific error,
- Changes in related Pallets due to modifications in the Vamm Pallet,
- Improvements in the pre-commit hooks.

PR #1401, opened 2022-08-02
refactor(vamm): Implement code review suggestions

Cardosaum:
## Issue
[CU-2tyk4k7](https://app.clickup.com/t/2tyk4k7)

## Description
This PR implement changes proposed in #1372. The list of major changes can be listed as:
- Assert runtime storage doesn't change in a `swap_simulation` test.
- Refactor tests for `swap` and `swap_simulation` functionality, using pre-defined constants over hard-coded values directly used in tests.
- Use an already existing method in `FixedPointNumber` over a custom function.
- Remove unneeded error variant.
- Implement `try_update_twap` function, fixing a hack in `swap` function.
- Use `with_context` blocks in tests, removing duplicated code and improving readability.
- Rename fallible traits in `composable-maths` to `Try`-variants, following rust code convention.
- Implement a new trait in `composable-maths` meant to convert `U256` values.
- Remove custom conversion functions in Vamm Pallet, using `composable-maths` equivalent methods.
- Improve `swap_simulation` documentation.
- Fix several broken links in modules `pallet-vamm`, `composable-traits`, `composable-maths` and `composable-support`
- Fix a bug in `compute_swap_base_asset` where removing an asset was not considered `negative`.


ComposableFiB0t:
Task linked: [CU-2tyk4k7 Code Review for Feature 8](https://app.clickup.com/t/2tyk4k7)

Cardosaum:
bors try

frame/vamm/src/helpers/math.rs
```diff
@@ -25,11 +29,9 @@ impl<T: Config> Pallet<T> {
...
 		ensure!(!quote.is_zero(), Error::<T>::QuoteAssetReserveIsZero);
 
-		let base_u256 = Self::balance_to_u256(base)?;
-		let quote_u256 = Self::balance_to_u256(quote)?;
-		let invariant = base_u256
-			.checked_mul(quote_u256)
-			.ok_or(Error::<T>::FailedToDeriveInvariantFromBaseAndQuoteAsset)?;
+		let base_u256 = base.into_u256();
+		let quote_u256 = quote.into_u256();
+		let invariant = base_u256.checked_mul(quote_u256).ok_or(Overflow)?;
```
0xangelo, line 34: Doesn't `.try_mul()` work in this case?

frame/vamm/src/helpers/twap.rs
```diff
@@ -5,29 +5,49 @@ use sp_runtime::traits::Saturating;
...
+	///
+	/// # Errors
+	///
+	/// * All errors returned by
+	/// [`sanity_check_before_update_twap`](
+	/// struct.Pallet.html#method.sanity_check_before_update_twap),
+	/// except
+	/// [`Error::<T>::AssetTwapTimestampIsMoreRecent`](
+	/// ../pallet/enum.Error.html#variant.AssetTwapTimestampIsMoreRecent).
+	#[transactional]
```
0xangelo, line 43: What's the Substrate version that we're using? There's already one which includes a transactional layer by default.

frame/vamm/src/helpers/twap.rs
```diff
@@ -5,29 +5,49 @@ use sp_runtime::traits::Saturating;
...
+	}
 
-		Ok(base_twap)
+	/// *Tries to* perform runtime storage changes, effectively updating the
+	/// asset twap.  Diferently than [`do_update_twap`](Self::do_update_twap) though, this variation
+	/// accepts one specific error to occur, therefore not honoring all
+	/// properties described in
+	/// [`update_twap`](../pallet/struct.Pallet.html#method.update_twap). The
+	/// property that can be violated using this variation is that the current
+	/// twap timestamp can in fact be more recent than the current time.
```
0xangelo, line 33: Is this the function used before each swap? In that case, shouldn't it fail silently, i.e., not raising an error but not modifying storage, when the current timestamp equals the timestamp of the last TWAP update?

frame/vamm/src/helpers/vamm.rs
```diff
@@ -11,8 +11,7 @@ impl<T: Config> Pallet<T> {
 	pub fn get_vamm_state(vamm_id: &T::VammId) -> Result<VammStateOf<T>, DispatchError> {
 		// Requested vamm must exists and be retrievable.
```
0xangelo, line 12: *'must exist'

frame/vamm/src/helpers/vamm.rs
```diff
@@ -11,8 +11,7 @@ impl<T: Config> Pallet<T> {
...
 		ensure!(VammMap::<T>::contains_key(vamm_id), Error::<T>::VammDoesNotExist);
-		let vamm_state = VammMap::<T>::get(vamm_id).ok_or(Error::<T>::FailToRetrieveVamm)?;
-		Ok(vamm_state)
+		VammMap::<T>::get(vamm_id).ok_or_else(|| Error::<T>::FailToRetrieveVamm.into())
```
0xangelo, line 14: Can't these be merged into one? Basically by keeping the second one only. I'm not aware of a scenario where the `ensure!` would succeed but the `.get()` wouldn't

frame/vamm/src/tests/swap.rs
```diff
@@ -440,19 +459,19 @@ proptest! {
...
-				let new_vamm_state = TestPallet::get_vamm(0).unwrap();
-				System::assert_has_event(
-					Event::UpdatedTwap { vamm_id: 0, base_twap: new_vamm_state.base_asset_twap }.into(),
-				);
+		with_existent_vamm_swap_contex(
+			vamm_state,
+			swap_config,
+			|swap_config| {
+				// For event emission.
+				run_for_seconds(vamm_state.twap_period);
```
0xangelo, line 467: Could perhaps try with several time periods instead of just `twap_period`, e.g., after `1..=X` seconds

0xangelo (approved):
Looks great! Mind renaming `calculate_exponential_moving_average` to `calculate_weighted_average`?

frame/composable-maths/src/dex/constant_product.rs
```diff
@@ -11,7 +11,7 @@ use sp_runtime::{
 	ArithmeticError, PerThing,
 };
 
-/// From https://balancer.fi/whitepaper.pdf, equation (2)
+/// From <https://balancer.fi/whitepaper.pdf>, equation (2)
```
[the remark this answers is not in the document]
Cardosaum: No, it was `cargo doc`. When running it, we get warnings about "raw links". The warning message advises enclosing the links in `<>` so that it becomes clickable.

frame/vamm/src/helpers/math.rs
Cardosaum: `.try_mul()` requires `UnsignedMath`, which is not implemented to `U256` yet.
(I'd like to implement it and then have a better way to work with `U256` values, but I'm still lacking the knowledge to do so. Gotta study more about Rust :books: )

frame/vamm/src/helpers/twap.rs
Cardosaum: Good catch!
Fixing it, sir.

frame/vamm/src/helpers/math.rs
kevin-composable: If you don't own `UnsignedMath` or `U256` you won't be able to implement `UnsignedMath` for `U256`

frame/vamm/src/helpers/twap.rs
```diff
@@ -5,29 +5,49 @@ use sp_runtime::traits::Saturating;
...
-		vamm_state.twap_timestamp = now;
-
-		// Update runtime storage.
-		VammMap::<T>::insert(&vamm_id, vamm_state);
+		Self::internal_update_twap(vamm_id, vamm_state, base_twap, now, false)
+	}
 
-		Ok(base_twap)
+	/// *Tries to* perform runtime storage changes, effectively updating the
+	/// asset twap.  Diferently than [`do_update_twap`](Self::do_update_twap) though, this variation
```
kevin-composable, line 28: Differently* or change this sentence to "Contrary to [`do_update_twap`] (Self::do_update_twap), this variation ..."

frame/vamm/src/helpers/twap.rs
```diff
@@ -53,6 +73,32 @@ impl<T: Config> Pallet<T> {
 		}
 	}
 
+	/// Efectively mutate runtime storage and
```
kevin-composable, line 76: Effectively*

frame/vamm/src/helpers/vamm.rs
kevin-composable: Thats a good catch - Error::<T>::FailToRetrieveVamm will never be thrown if it reaches that point

kevin-composable (approved):
I had a couple comments but nothing huge to comment on. With this PR and the past couple PRs you've done a really great job restructuring and cleaning up your code - Nice work Matheus!

frame/vamm/src/helpers/vamm.rs
Cardosaum: I see.... 
In that case maybe it would make more sense to keep `VammDoesNotExist` but delete the other error variant. (removing the `ensure!`), right?

frame/vamm/src/helpers/math.rs
Cardosaum: We do own `UnsignedMath` (`frame/composable-maths/src/labs/numbers.rs:45`), but the requirements for it are not satisfied by `U256`. I have to study more about how to implement it :sweat_smile:

frame/vamm/src/helpers/vamm.rs
Cardosaum: thanks!

frame/vamm/src/helpers/twap.rs
Cardosaum: Thanks! :)

frame/vamm/src/helpers/twap.rs
Cardosaum: Thx!

frame/vamm/src/tests/swap.rs
Cardosaum: Done :+1: 
Thanks for the suggestion! :)

Cardosaum:
bors try

kevin-composable (dismissed):

frame/vamm/src/tests/swap.rs
```diff
@@ -463,12 +488,23 @@ proptest! {
...
+							// Ensure twap was updated
+							let vamm_state_after = TestPallet::get_vamm(0).unwrap();
+							assert_ne!(vamm_state.twap_timestamp, vamm_state_after.twap_timestamp);
+						}
+					},
+					_ => {
+						if swap_output.is_ok() {
+							// Ensure twap was *not* updated
+							let vamm_state_after = TestPallet::get_vamm(0).unwrap();
+							assert_eq!(vamm_state.twap_timestamp, vamm_state_after.twap_timestamp);
```
0xangelo, line 505: Isn't `twap_period` only used to compute the weights in the weighted average? Why is it that if more than `twap_period` has passed, the TWAP isn't updated? The more time passes, the more aggressive we should be with the update (less weight to older TWAP value, more to current price value)

Cardosaum:
bors try

frame/vamm/src/tests/swap.rs
```diff
@@ -315,6 +316,29 @@ fn should_update_twap_when_removing_quote_asset() {
...
+		assert_ok!(TestPallet::swap(&swap_config));
+		let vamm_state_t1 = TestPallet::get_vamm(0).unwrap();
+		assert_eq!(vamm_state_t1.twap_timestamp, vamm_state_t0.twap_timestamp + delay);
+		assert_ne!(vamm_state_t0.twap_timestamp, vamm_state_t1.twap_timestamp);
+
+		// In the second swap the time didn't pass between operations, so we
+		// can't allow the twap value to be updated again.
+		assert_ok!(TestPallet::swap(&swap_config));
+		let vamm_state_t2 = TestPallet::get_vamm(0).unwrap();
+		assert_eq!(vamm_state_t1.twap_timestamp, vamm_state_t2.twap_timestamp);
```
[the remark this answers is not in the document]
Cardosaum: Good observation, thanks!

frame/vamm/src/tests/swap.rs
Cardosaum: You're correct.
Fixed it in recent commits 👍

frame/vamm/src/tests/swap.rs
```diff
@@ -332,13 +332,40 @@ fn should_not_update_twap_if_current_twap_timestamp_is_more_recent() {
 		assert_ne!(vamm_state_t0.twap_timestamp, vamm_state_t1.twap_timestamp);
 
 		// In the second swap the time didn't pass between operations, so we
-		// can't allow the twap value to be updated again.
+		// can't allow the twap timestamp nor the twap value to be updated
+		// again.
 		assert_ok!(TestPallet::swap(&swap_config));
 		let vamm_state_t2 = TestPallet::get_vamm(0).unwrap();
 		assert_eq!(vamm_state_t1.twap_timestamp, vamm_state_t2.twap_timestamp);
+		assert_eq!(vamm_state_t1.base_asset_twap, vamm_state_t2.base_asset_twap);
```
0xangelo, line 340: What happens if the price changes rapidly within the same block? Shouldn't we allow the TWAP to change a bit in that case? Otherwise, the TWAP in a block will be determined by the first swap in it, and subsequent swaps would be free to be as large as they please without changing the TWAP.

I have a feeling that's why other protocols have a minimum weight always for the latest price, even if little to no time has passed since the last swap. DM me if you want to discuss

```text
                                                       twap_period
                                        |---------------------------------------|
                                         from_start          since_last
                                        |-----------|---------------------------|
 -------------------------------------------------------------------------------|
                                        ^           ^                           ^
                                      now -         |                          now
                                   twap_period      |
                                               last_twap_ts
```

frame/vamm/src/tests/swap.rs
```diff
@@ -332,13 +332,40 @@ fn should_not_update_twap_if_current_twap_timestamp_is_more_recent() {
...
+fn should_update_twap_if_twap_period_has_passed(#[case] delta: Timestamp) {
+	with_swap_context(
+		TestVammConfig { base_asset_reserves: as_decimal_inner(42), ..Default::default() },
+		TestSwapConfig { input_amount: as_decimal_inner(30), ..Default::default() },
+		|_, swap_config| {
+			let vamm_state_t0 = TestPallet::get_vamm(0).unwrap();
+			run_for_seconds(delta.saturating_add(vamm_state_t0.twap_period));
+			assert_ok!(TestPallet::swap(&swap_config));
+			run_for_seconds(delta.saturating_add(vamm_state_t0.twap_period));
+			assert_ok!(TestPallet::swap(&swap_config));
```
0xangelo, line 361: Why does it have to a delta **plus** the 'twap period'? Again, I think the naming might be causing confusion here, `twap_period` is only used to calculate the weights in the weighted average between the current price and last TWAP, not as a reference for how much time should pass between swaps

0xangelo: Unless this test is supposed to check if the TWAP is updated *even if* more than 'twap_period' has passed.

0xangelo: My last question would be: why do 2 swaps here if you don't get the vAMM state in between them?

0xangelo (approved):
LGTM! 🚀

kevin-composable (approved):

PR #1421, opened 2022-08-05
#CU-3ajf8dm xcmp picasso ci tests

dzmitry-lahoda:
## Description
- we need to run picasso XCMP simulator on each run same way we run dali to prove XCMP


ComposableFiB0t:
Task linked: [CU-3ajf8dm Run simulator tests on Kusama as part of CI](https://app.clickup.com/t/3ajf8dm)

PR #1429, opened 2022-08-09
Extend Query with Program

KaiserKarel:
## Description
Extends the Query instruction with a return Program, to be executed on the origin chain once the Query is fulfilled.

## Checklist

### DIFF
--- a/xcvm/SPEC.md
+++ b/xcvm/SPEC.md
@@ -221,8 +221,6 @@ sequenceDiagram
     XCVM Interpreter->>Opaque Contract: Call
     XCVM Interpreter->>Gateway: Spawn
     XCVM Interpreter->>Gateway: Query
-    XCVM Interpreter->>XCVM Interpreter: Commit
-    XCVM Interpreter->>XCVM Interpreter: Abort
 ```
 
 ##### Call Instruction and Late Bindings
--- a/xcvm/SPEC.md
+++ b/xcvm/SPEC.md
@@ -204,7 +204,7 @@ Payload ::= bytes
 Bindings ::= [(u16, BindingValue)]
 BindingValue ::= Self | Relayer | Result | Balance | AssetId
 Spawn ::= Network BridgeSecurity Salt Program Assets
-Query ::= Network Salt
+Query ::= (Network Salt | Identity ) BridgeSecurity Program
 Account ::= bytes
 ```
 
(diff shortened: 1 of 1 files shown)


KaiserKarel:
An alternative method is to get rid of the query instruction, and have the router handle querying and sending back an answer:

The origin contract X on chain A send a [Spawn B [Call Y with chain_b_encoding(Address of X)]], contract Y on chain B does the query the user want and yield [Spawn A [Call X with chain_a_encoding(query_result)]] 
Contract Y on chain B has the interpreter address as it's the "sender" of the call, it also know the origin contract address

The downside of this is higher gas cost + a more complicated setup.

xcvm/SPEC.md
```diff
@@ -364,4 +350,38 @@ Spawn XYZ BridgeSecurity::Deterministic 0 [
...
+```solidity
+function uniswapV2Call(address sender, uint amount0, uint amount1, bytes calldata data) external;
+```
+
+To properly construct this call and return funds to the interpreter instance, the `sender` parameter needs to be set to the interpreter's address. Normally this is an easy operation, as you can simply call the contract and pass `self`. However, from another chain, this is more difficult, as the address of the instance is unknown. There are three approaches to figuring out the interpreter's address:
+
+#### CREATE2
+
+The EVM has a specific [instruction](https://docs.openzeppelin.com/cli/2.8/deploying-with-create2#create2) which can be used to determine what the address of a contract will be based on keccak(0xFF, sender, salt, bytecode). In our case, the sender is the `XCVM Router` for that specific chain and known beforehand, the salt is the `Account, Salt` pair from the spawn call, and the bytecode is known beforehand as well. This means that when constructing the encoded uniswapV2Call, we can compute the `to` parameter on the origin chain. `CREATE2` is not available everywhere, however.
+
```
dzmitry-lahoda, line 368: we can assume it will be? and support only these chains if main chains do support that? (EVM for Polygone/Ethereum/Polkadot)? So we can simplfify on target market

xcvm/SPEC.md
```diff
@@ -364,4 +350,38 @@ Spawn XYZ BridgeSecurity::Deterministic 0 [
...
+#### Remote Instantiation First
```
dzmitry-lahoda, line 369: I seems that either chain support stuff (create2) or relayer does stuff, so may omit third way to complicate programs for end users.

dzmitry-lahoda:
so either address determinisic or relayer makes it determined into special storage for xcvm program? so  need to complicate on chain program?

xcvm/SPEC.md
```diff
@@ -235,27 +241,7 @@ Amounts of assets can be specified using the `Balance` type. This allows foreign
...
-
-To inspect a Query result from an XCVM program, use `Call` to a contract with the capabilities to handle it properly.
-
-```
-Query Ethereum 1                    // Pauses execution, which will be resumed once the 
-                                    // result values have been loaded.
-Call 0xmy_contract_on_this_chain    // my_contract_on_this_chain can inspect the Result Register of his                             
-                                    // instance and do something with the actual result
-```
+Registers are persisted after `Program` invocations and `Query`able from other chains. Querying does not alter the registry state, so multiple `Query` invocations will always return the same data.
```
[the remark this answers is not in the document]
KaiserKarel: Ah let me elaborate:

On the origin chain, interpreter.result == (destination.ip, destination.result, destination.address)
On the destination chain, interpreter.result remains the same.

xcvm/SPEC.md
KaiserKarel: EVM does, Cosmos does not, so we have a significant market that needs the other two solutions.

xcvm/SPEC.md
KaiserKarel: We can't have the relayer change data, as we use trustless proofs for data authenticity.

KaiserKarel:
> so either address determinisic or relayer makes it determined into special storage for xcvm program? so need to complicate on chain program?

Yeah, that's your only solution to be honest.

xcvm/SPEC.md
dzmitry-lahoda: can contracts have `self` or `this`(and `parent` may be:) ) as special keyword, so they are resolved as stuff executed? instead of direct `query` in xcvm program code.

dzmitry-lahoda: i am no deep in the topic, but if relayer can set some data permission less and then it can be validated on chain, than it can set data. so kind if can change data in certain ways, not sure if that fits here.

KaiserKarel: solved this with late bindings.

KaiserKarel: solved this with late bindings.

dzmitry-lahoda (approved):

aeryz (approved):

PR #1436, opened 2022-08-10
feat(cosmwasm): cosmwasm contracts

aeryz:
* Add interpreter contract which is the actual xcvm interpreter
  instance.

* Add asset registry which stores the mapping of cw20 token instances.
  Interpreters will query this contract to get the token instance, then
  call that cw20 token instance by themselves(not through the asset
  registry).

* Add `cosmwasm-contracts` profile which contains the necessary options
  for cosmwasm production builds. Cosmwasm contracts should be built by
  using this profile.

* Patch `serde_json_wasm`. Original `serde_json_wasm` does not support
  (de)serialization of `BTreeMap`. This produces `Unreachable`
  instruction which makes the contract crash. However our version does
  support this.

**EDIT**
* We decided on accepting `CosmosMsg` instead of `WasmMsg::Execute` in the `Call` instruction. This way, we don't restrict users to just be able to execute contracts but do anything that cosmwasm supports. So this will be added to this PR as well.
- [x] Support `CosmosMsg` in `Call` instruction

## Issue
https://app.clickup.com/t/2mezh62

## Description
See commit message.

## Checklist


xcvm/cosmwasm/contracts/asset-registry/README.md
```diff
@@ -0,0 +1,3 @@
+# Asset Registry Contract
+
+// TODO
```
KaiserKarel, line 3: Would prefer the writeup to be part of this PR

xcvm/cosmwasm/contracts/asset-registry/examples/schema.rs
```diff
@@ -0,0 +1,17 @@
...
+fn main() {
+    let mut out_dir = current_dir().unwrap();
+    out_dir.push("schema");
+    create_dir_all(&out_dir).unwrap();
+    remove_schemas(&out_dir).unwrap();
+
+    export_schema(&schema_for!(InstantiateMsg), &out_dir);
+    export_schema(&schema_for!(ExecuteMsg), &out_dir);
+    export_schema(&schema_for!(QueryMsg), &out_dir);
+}
```
KaiserKarel, line 17: Why is this in a `examples` directory?

xcvm/cosmwasm/contracts/asset-registry/src/contract.rs
```diff
@@ -0,0 +1,168 @@
+use std::collections::BTreeMap;
+
+#[cfg(not(feature = "library"))]
+use cosmwasm_std::entry_point;
+use cosmwasm_std::{to_binary, Binary, Deps, DepsMut, Env, MessageInfo, Response, StdResult};
+
```
KaiserKarel, line 6: nit: I personally prefer no linebreaks between imports, as then they will be sorted by rustfmt

xcvm/cosmwasm/contracts/asset-registry/src/contract.rs
```diff
@@ -0,0 +1,168 @@
...
+		.keys(deps.storage, None, None, cosmwasm_std::Order::Ascending)
+		.collect::<Result<Vec<_>, _>>()?
+	{
+		ASSETS.remove(deps.storage, key);
+	}
+
+	for (asset_id, contract_addr) in assets {
+		let addr = deps.api.addr_validate(&contract_addr)?;
+		ASSETS.save(deps.storage, asset_id.parse::<XcvmAssetId>().unwrap(), &addr)?;
+	}
```
KaiserKarel, line 58: Is there a way to not need to iterate over every key?

xcvm/cosmwasm/contracts/asset-registry/src/contract.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
+	match msg {
+		QueryMsg::GetAssetContract(token_id) => to_binary(&query_asset_contract(deps, token_id)?),
+	}
+}
+
+pub fn handle_set_assets(
```
KaiserKarel, line 43: Might be good to have an add/remove asset too? Incremental changes

xcvm/cosmwasm/contracts/asset-registry/src/helpers.rs
```diff
@@ -0,0 +1,27 @@
...
+
+/// AssetRegistryContract is a wrapper around Addr that provides helpers
+/// for working with this as a library.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct AssetRegistryContract(pub Addr);
+
+impl AssetRegistryContract {
+    pub fn addr(&self) -> Addr {
+        self.0.clone()
+    }
```
KaiserKarel, line 16: nit: why not `&Addr`, and let the caller clone?

xcvm/cosmwasm/contracts/asset-registry/src/contract.rs
aeryz: There isn't unfortunately. We tried removing the whole assets storage but there is no such function exists. I think this is because it stores ASSETS related things with assets prefix and cannot remove all of them at a constant time.

xcvm/cosmwasm/contracts/interpreter/README.md
```diff
@@ -0,0 +1,7 @@
+# Name Service
+
+The goal of the application you are building is to let users buy names and to set a value these names resolve to.
+The owner of a given name will be the current highest bidder. In this section, you will learn how these simple
+ requirements translate to application design.
+
+Here is the tutorial for this application: [tutorial](https://docs.cosmwasm.com/tutorials/name-service/intro)
```
KaiserKarel, line 7: Old Readme?

xcvm/cosmwasm/contracts/interpreter/examples/schema.rs
```diff
@@ -0,0 +1,15 @@
...
+
+fn main() {
+    let mut out_dir = current_dir().unwrap();
+    out_dir.push("schema");
+    create_dir_all(&out_dir).unwrap();
+    remove_schemas(&out_dir).unwrap();
+
+    export_schema(&schema_for!(InstantiateMsg), &out_dir);
+    export_schema(&schema_for!(ExecuteMsg), &out_dir);
+}
```
KaiserKarel, line 15: Same, why in `example` dir?

xcvm/cosmwasm/contracts/interpreter/src/msg.rs
```diff
@@ -0,0 +1,24 @@
...
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct InstantiateMsg {
+    pub registry_address: String,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum ExecuteMsg {
+    Execute { program: XCVMProgram },
```
KaiserKarel, line 23: Are we also going to create an entrypoint for serialized programs?

xcvm/lib/core/src/program.rs
```diff
@@ -4,7 +4,9 @@ use scale_info::TypeInfo;
 use serde::{Deserialize, Serialize};
 
 #[cfg_attr(feature = "std", derive(schemars::JsonSchema))]
-#[derive(Clone, PartialEq, Eq, Debug, Encode, Decode, TypeInfo, Serialize, Deserialize)]
+#[derive(
+	Clone, PartialEq, Eq, Debug, Encode, Decode, TypeInfo, Serialize, Deserialize
+)]
 #[serde(rename_all = "camelCase")]
 pub struct Program<Instructions> {
 	pub tag: Option<Vec<u8>>,
```
KaiserKarel, line 12: Can we avoid the Option here? Isn't an empty tag semantically the same?

xcvm/cosmwasm/contracts/interpreter/src/msg.rs
aeryz: We can have ExecuteRaw or ExecuteEncoded. What do you think?

xcvm/cosmwasm/contracts/asset-registry/examples/schema.rs
aeryz: So that it can easily be run to create schemas. I would prefer this to be in scripts directory alongside with a script to build and run. But I didn't want to change this since all of the contracts that are provided and the template use this convention. We can have a script to generate schema for all or specified contracts if you want.

xcvm/cosmwasm/contracts/interpreter/examples/schema.rs
aeryz: [see](https://github.com/ComposableFi/composable/pull/1436#discussion_r943152450)

xcvm/lib/core/src/program.rs
aeryz: We can I guess. Empty vec will be serialized into `[]` and `None` to `null`. But let me check if we do any kind of stuff based on the None check on tag.

xcvm/cosmwasm/contracts/asset-registry/src/contract.rs
aeryz: We discussed this with @hussein-aitlahcen and decided on having only `set` function. What do you think @hussein-aitlahcen?

xcvm/cosmwasm/contracts/interpreter/src/contract.rs
```diff
@@ -0,0 +1,274 @@
...
+				interpret_spawn(network, salt, assets, program, response),
+			Instruction::Transfer { to, assets } =>
+				interpret_transfer(&mut deps, to, assets, response),
+		}?;
+	}
+	Ok(response)
+}
+
+pub fn interpret_call(encoded: Vec<u8>, response: Response) -> Result<Response, ContractError> {
+	let cosmos_msg: CosmosMsg =
```
aeryz, line 65: This is actually pretty cool. We used to expect payload for `WasmMsg::Execute`. And we were using a custom payload format such as:
```rust
struct Payload {
    address: String,
    msg: String
}
```
which is not very nice. Now we expect a `CosmosMsg`. This means that users can do any kind of stuff that cosmwasm environment support such as transfer native tokens, instantiate token, custom messages, etc.

xcvm/lib/core/src/program.rs
hussein-aitlahcen: Has been updated on latest core

xcvm/cosmwasm/contracts/asset-registry/examples/schema.rs
hussein-aitlahcen: This is following CosmWasm way of easily exporting schemas

xcvm/cosmwasm/contracts/asset-registry/src/contract.rs
hussein-aitlahcen: Let's keep it like this and optimize it later

xcvm/cosmwasm/contracts/interpreter/src/msg.rs
hussein-aitlahcen: Probably cleaner to have `Execute { program: EncodedProgram} enum EncodedProgram { Typed(XCVMProgram), Serialized(Binary) }`?

xcvm/cosmwasm/contracts/interpreter/src/contract.rs
```diff
@@ -0,0 +1,288 @@
...
+			if amount.slope.0 == 0 {
+				amount.intercept.0
+			} else {
+				let query_msg = Cw20QueryMsg::Balance { address: to.clone() };
+				let response: BalanceResponse =
+					deps.querier.query(&QueryRequest::Wasm(WasmQuery::Smart {
+						contract_addr: cw20_address.addr.clone().into_string(),
+						msg: to_binary(&query_msg)?,
+					}))?;
+				amount.apply(response.balance.into())
```
hussein-aitlahcen, line 126: Why are we checking the slope instead of applying `apply` in both cases?

xcvm/cosmwasm/contracts/interpreter/src/contract.rs
```diff
@@ -0,0 +1,288 @@
...
+
+		let cw20_address: GetAssetContractResponse = deps.querier.query(
+			&WasmQuery::Smart { contract_addr: registry_addr.clone(), msg: to_binary(&query_msg)? }
+				.into(),
+		)?;
+		let contract = Cw20Contract(cw20_address.addr.clone());
+
+		if amount.is_zero() {
+			return Err(ContractError::ZeroTransferAmount)
+		}
```
hussein-aitlahcen, line 114: I don't think it should be an error but rather a NOOP

xcvm/cosmwasm/contracts/router/src/contract.rs
```diff
@@ -0,0 +1,161 @@
...
+				admin: Some(env.contract.address.clone().into_string()),
+				code_id: interpreter_code_id,
+				msg: to_binary(&InterpreterInstantiateMsg {
+					registry_address: registry_address.into_string(),
+					network_id,
+					user_id: user_id.clone(),
+				})?,
+				funds: vec![],
+				label: format!("xcvm-interpreter-{}", network_id.0), /* TODO(aeryz): juno doesn't
+				                                                      * allow empty label */
```
hussein-aitlahcen, line 76: remove todo and make the label `xcvm-interpreter-{hash(origin)}`

xcvm/cosmwasm/contracts/router/src/contract.rs
```diff
@@ -0,0 +1,161 @@
...
+		let contract = Cw20Contract(cw20_address.addr.clone());
+
+		if amount.intercept.0 == 0 {
+			continue
+		}
+
+		response = response.add_message(contract.call(Cw20ExecuteMsg::Transfer {
+			recipient: interpreter_address.clone(),
+			amount: amount.intercept.0.into(),
+		})?);
```
hussein-aitlahcen, line 118: For the funds moved to the router, we should override the generic type to be `Displayed<u128>` instead of `Amount` as you are always expecting an absolute amount

hussein-aitlahcen: a.k.a. `Funds<Displayed<u128>>`

hussein-aitlahcen (changes requested):

xcvm/cosmwasm/contracts/interpreter/src/contract.rs
aeryz: Yeah actually this is unnecessary.

dzmitry-lahoda:
https://github.com/Wasmswap/wasmswap-contracts/issues/68

dzmitry-lahoda:
@aeryz may you fix tests so we catch bug here, not when running e2e tests?

aeryz:
> tests?

Note that these are unit tests. Since there is no vm, we only test if correct submessages are returned, values are stored/modified correctly etc. If there is a problem with the vm, there still be some bugs in execution.

dzmitry-lahoda:
@aeryz   unit test are failing. how can we be sure that we can have working xcvm programs for demo when doing end to end runs?  neirher we can merge into main.

aeryz:
> @aeryz unit test are failing. how can we be sure that we can have working xcvm programs for demo when doing end to end runs? neirher we can merge into main.

Some changes that we made for relayer broke the tests apparently, fixed all and completed the router's execution tests.

aeryz:
@hussein-aitlahcen @dzmitry-lahoda Guys let's not forget that unit tests are not enough. We need to verify that our contracts interact with each other correctly, and behave the same way on both Cosmos-based chains and Picasso. To ensure this and test the flow (on a specific chain without the relayer), we should have a set of contracts for testing purposes like [this one](https://github.com/CosmWasm/cosmwasm/tree/main/contracts/crypto-verify). This is not urgent right now, but it should be done after the demo.

dzmitry-lahoda:
i am doing e2e semi automated swap call via xcvm, i guess with nix run and our exsiting tests it will take easy to get into our existing automation for picasso

xcvm/lib/core/src/lib.rs
```diff
@@ -72,7 +73,7 @@ where
 	}
 
 	#[inline]
-	pub fn call<T>(self, protocol: T) -> Result<Self, T::Error>
+	pub fn call_raw_protocol<T>(self, protocol: T) -> Result<Self, T::Error>
```
hussein-aitlahcen, line 76: Why renaming this???

dzmitry-lahoda: because there is call_raw. and this is also call_RAW_protocol. 

likely we will have call_lazy and call_LAZY_protocol. so it will be just call, no lazy word. 

so to avoid confusion to split raw and lazy non raw calls. right?

dzmitry-lahoda: so Mars will do Mars protocol with only ToRaw<Network> trait. Later the may add ToRaw<Polygone>. 
And later ToXcvm trait.

so call_raw_protocol requires ToRaw,

while call_procol requires ToXcvm

dzmitry-lahoda: ok I revert back, but will request the change to have consistent naming.

hussein-aitlahcen (approved):
:rocket:

dzmitry-lahoda (approved):

scripts/style.sh
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env bash
```
dzmitry-lahoda, line 1: not sure. we though it would be nix. `nix run .#fmt`

code/xcvm/cosmwasm/contracts/router/Developing.md
```diff
@@ -0,0 +1,104 @@
...
+
+If you have recently created a contract with this template, you probably could use some
+help on how to build and test the contract, as well as prepare it for production. This
+file attempts to provide a brief overview, assuming you have installed a recent
+version of Rust already (eg. 1.58.1+).
+
+## Prerequisites
+
+Before starting, make sure you have [rustup](https://rustup.rs/) along with a
+recent `rustc` and `cargo` version installed. Currently, we are testing on 1.58.1+.
```
dzmitry-lahoda, line 11: not sure if we testing on this version so.

aeryz: Yeah, this document comes from the template itself so better to remove it completely.

scripts/style.sh
aeryz: yes, I messed up in the rebase, I'll remove this

code/Cargo.toml
```diff
@@ -117,171 +128,171 @@ xcm-builder = { git = "https://github.com/ComposableFi/polkadot", branch = "rele
...
-sp-version-proc-macro = { git = "https://github.com/ComposableFi/substrate", rev = "f709a3d3b8f116f8b7e92ec56bc1ca5a0409eaa5" }
-sp-wasm-interface = { git = "https://github.com/ComposableFi/substrate", rev = "f709a3d3b8f116f8b7e92ec56bc1ca5a0409eaa5" }
-substrate-build-script-utils = { git = "https://github.com/ComposableFi/substrate", rev = "f709a3d3b8f116f8b7e92ec56bc1ca5a0409eaa5" }
-substrate-frame-rpc-system = { git = "https://github.com/ComposableFi/substrate", rev = "f709a3d3b8f116f8b7e92ec56bc1ca5a0409eaa5" }
-substrate-prometheus-endpoint = { git = "https://github.com/ComposableFi/substrate", rev = "f709a3d3b8f116f8b7e92ec56bc1ca5a0409eaa5" }
-substrate-state-trie-migration-rpc = { git = "https://github.com/ComposableFi/substrate", rev = "f709a3d3b8f116f8b7e92ec56bc1ca5a0409eaa5" }
-substrate-wasm-builder = { git = "https://github.com/ComposableFi/substrate", rev = "f709a3d3b8f116f8b7e92ec56bc1ca5a0409eaa5" }
-try-runtime-cli = { git = "https://github.com/ComposableFi/substrate", rev = "f709a3d3b8f116f8b7e92ec56bc1ca5a0409eaa5" }
+beefy-gadget = { git = "https://github.com/ComposableFi/substrate", rev = "fd75550a935d69b6a4ee4f8a48767557b4c9e801" }
```
dzmitry-lahoda, line 131: not sure. any reason changing substrate dependency? i just guess you do not wanted to do that

code/parachain/frame/cosmwasm/runtime-api/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+targets = ["x86_64-unknown-linux-gnu"]
+
+[dependencies]
+codec = { default-features = false, features = [
+  "derive",
+], package = "parity-scale-codec", version = "3.0.0" }
+sp-api = { default-features = false, git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.27" }
+sp-std = { default-features = false, git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.27" }
+
+# REVIEW: Does the runtime API need features?
```
dzmitry-lahoda, line 19: yes.

code/parachain/frame/cosmwasm/runtime-api/src/lib.rs
```diff
@@ -0,0 +1,37 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+#![allow(clippy::too_many_arguments)]
+#![allow(clippy::unnecessary_mut_passed)]
+
+use codec::Codec;
+use sp_std::collections::btree_map::BTreeMap;
+#[cfg(not(feature = "std"))]
+use sp_std::vec::Vec;
+
+// Cosmwasm Runtime API declaration. Implemented for each runtime at
```
dzmitry-lahoda, line 10: afaik picasso and composable have no cosmwasm so cannot be implemented. or these are todo!()?

code/parachain/runtime/dali/src/lib.rs
```diff
@@ -1542,6 +1542,45 @@ impl_runtime_apis! {
...
+			cosmwasm::instantiate::<Runtime>(
+				instantiator,
+				code_id,
+				salt,
+				admin,
+				label,
+				funds,
+				gas,
+				message
+			).map_err(|err| alloc::format!("{:?}", err).into_bytes())
```
dzmitry-lahoda, line 1580: sp_std::format?

code/xcvm/cosmwasm/contracts/interpreter/src/contract.rs
```diff
@@ -0,0 +1,428 @@
...
+) -> Result<Response, StdError> {
+	let registry_address = deps.api.addr_validate(&msg.registry_address)?;
+	let config =
+		Config { registry_address, network_id: msg.network_id, user_id: msg.user_id.clone() };
+	CONFIG.save(deps.storage, &config)?;
+
+	Ok(Response::new().add_event(
+		Event::new("xcvm.interpreter.instantiated").add_attribute(
+			"data",
+			to_binary(&(msg.network_id.0, msg.user_id))?.to_base64().as_str(),
```
dzmitry-lahoda, line 34: not for now, but seems it is common practice to add several attributes without join and encode. at least mars does that

code/xcvm/cosmwasm/contracts/interpreter/src/contract.rs
```diff
@@ -0,0 +1,428 @@
...
+
+	Ok(response.add_event(
+		Event::new("xcvm.interpreter.spawn")
+			.add_attribute(
+				"origin_network_id",
+				serde_json_wasm::to_string(&config.network_id.0)
+					.map_err(|_| ContractError::DataSerializationError)?,
+			)
+			.add_attribute(
+				"origin_user_id",
```
dzmitry-lahoda, line 174: here we do not glue user and network. add 2 attributes. while in interpreter we do. feels incosistent.

code/xcvm/cosmwasm/contracts/router/src/contract.rs
```diff
@@ -0,0 +1,368 @@
...
+			let instantiate_msg: CosmosMsg = WasmMsg::Instantiate {
+				admin: Some(env.contract.address.clone().into_string()),
+				code_id: interpreter_code_id,
+				msg: to_binary(&InterpreterInstantiateMsg {
+					registry_address: registry_address.into_string(),
+					network_id,
+					user_id: user_id.clone(),
+				})?,
+				funds: vec![],
+				label: format!("xcvm-interpreter-{}-{}", network_id.0, hex::encode(&user_id)),
```
dzmitry-lahoda, line 75: may be spec should be updated. we have 4 specs now already. and yet there is on team. not in this PR and may be not you. just mention. will not bother with this further.

dzmitry-lahoda (approved):

code/xcvm/cosmwasm/contracts/interpreter/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+library = []
+
+[dependencies]
+cosmwasm-std = { version = "1.0.0" }
+cw-storage-plus = "0.13.2"
+cw20 = "0.13.2"
+num = "0.4"
+schemars = "0.8.8"
+serde = { version = "1.0.137", default-features = false, features = ["derive"] }
+serde-json-wasm = { git = "https://github.com/hussein-aitlahcen/serde-json-wasm", rev = "1608a13d2a2ba90605d9626a51ff6667aca5a2d6" }
```
dzmitry-lahoda, line 30: not sure. as per spec protobuf and json formats has no dictionaries. that is only view side interpretation. in proto files maps are not really maps too. 

so could just do serde interpretation of maps on client side with out fork.

I just want to say that adding map to low level embeeded json likely will not merge map interpretation into core. really protobuf and json have only vectors of tuples - bags, not sets.

aeryz: I haven't use it but seen this: https://developers.google.com/protocol-buffers/docs/proto#maps.

code/Cargo.toml
aeryz: I do actually. The thing is we were having a problem related to substrate. Which is also a known bug in original substrate repo. So I cherry-picked the fix and upgrade the version here.

code/xcvm/cosmwasm/contracts/interpreter/src/contract.rs
aeryz: Can you show me the part where we glue them?

code/xcvm/cosmwasm/contracts/router/src/contract.rs
aeryz: This is a cosmwasm-specific way of handling internal contracts which is very specific for the general spec. We could state this in somewhere if anyone needs it but not sure they will.

code/xcvm/cosmwasm/contracts/interpreter/src/contract.rs
aeryz: Oh, now I see. Good point. But the reason why we do this like this is only for internal use. We use this `data` attribute when handling the reply in router contract. So this won't be exposed to users or the frontend.

code/xcvm/cosmwasm/contracts/interpreter/src/contract.rs
aeryz: Good point but as I mentioned in the other comment, this is exposed to the relayer/frontend (outer world you can say). But the one we use them together is only for internal use of router. So that is the most performant way of doing that. Because we don't have to iterate for two different attributes. But we find one and just decode the raw data. All of the exposed parts use separate attributes for these. So it is consistent.

code/xcvm/cosmwasm/contracts/interpreter/Cargo.toml
aeryz: We talked to @hussein-aitlahcen and it is doable if your key is a simple type like string or int. So in our case, we can use it without a problem.

code/parachain/runtime/dali/src/lib.rs
aeryz: they say that it is preferable to use `format!` instead in their docs. https://docs.rs/sp-std/latest/sp_std/fmt/fn.format.html

benluelo: I believe dzmitry was referring to using `sp_std::format!` instead of `alloc::format!`, but it doesn't really matter since sp_std is just reexports of core/alloc/std.

benluelo (approved):

code/xcvm/cosmwasm/contracts/interpreter/Cargo.toml
dzmitry-lahoda: language has map. but if you will send several same keys it still will be up to spec. so they tell have map. but there is not restriction to have map on message format. example, such messages likely will fail evm parser if it linear (not sure it is)

code/parachain/runtime/dali/src/lib.rs
dzmitry-lahoda: yes. just we use sp_std a lot and it maps to whatever it maps. so any part of code is portable.

dzmitry-lahoda:
```
Module [sp_std](https://docs.rs/sp-std/latest/sp_std/index.html)::[fmt](https://docs.rs/sp-std/latest/sp_std/fmt/index.html#)Copy item path
1.0.0[[−]](javascript:void(0))[[src]](https://doc.rust-lang.org/nightly/src/alloc/lib.rs.html#196)
Utilities for formatting and printing Strings.

This module contains the runtime support for the [format!](https://doc.rust-lang.org/nightly/alloc/macro.format.html) syntax extension. This macro is implemented in the compiler to emit calls to this module in order to format arguments at runtime into strings.

[Usage](https://docs.rs/sp-std/latest/sp_std/fmt/index.html#usage)
The [format!](https://doc.rust-lang.org/nightly/alloc/macro.format.html) macro is intended to be familiar to those coming from C’s printf/fprintf functions or Python’s str.format function.

Some examples of the [format!](https://doc.rust-lang.org/nightly/alloc/macro.format.html) extension are:

format!("Hello");                 // => "Hello"
format!("Hello, {}!", "world");   // => "Hello, world!"
format!("The number is {}", 1);   // => "The number is 1"
format!("{:?}", (3, 4));          // => "(3, 4)"
format!("{value}", value=4);      // => "4"
let people = "Rustaceans";
format!("Hello {people}!");       // => "Hello Rustaceans!"
format!("{} {}", 1, 2);           // => "1 2"
format!("{:04}", 42);             // => "0042" with leading zeros
format!("{:#?}", (100, 200));     // => "(
                                  //       100,
                                  //       200,
                                  //     )"
From these, you can see that the first argument is a format string. It is required by the compiler for this to be a string literal; it cannot be a variable passed in (in order to perform validity checking). The compiler will then parse the format string and determine if the list of arguments provided is suitable to pass to this format string.

To convert a single value to a string, use the [to_string](https://doc.rust-lang.org/nightly/alloc/string/trait.ToString.html#tymethod.to_string) method. This will use the [Display](https://docs.rs/sp-std/latest/sp_std/fmt/trait.Display.html) formatting trait.
```

PR #1437, opened 2022-08-10
Add global-registry.md

KaiserKarel:
## Description
Adds the initial specification for asset registries.

### DIFF
--- a/xcvm/architecture/asset-registry/global-registry.md
+++ b/xcvm/architecture/asset-registry/global-registry.md
@@ -1,0 +1,68 @@
+# Global Asset Registry
+
+Within `XCVM` programs, assets are identified by a global identifier, which is the same within each chain. Bitcoin for example is identified by `1`.
+
+```
+AssetId ::= u64
+```
+
+The XCVM interpreter performs a look-up when executing instructions such as `Spawn` and `Transfer` to translate the local identifier to a global one. Local identifiers are chain specific, such as `address` in EVM chains. There is a two-way supported lookup:
+
+```
+AssetsById ::= AssetId -> LocalId
+IdByAssets ::= LocalId -> AssetId
+```
+
+## Querying the Registry
+
+When constructing `Call`s, we often need to convert `LocalId`s to `LocalId`s based on the `NetworkId`. Even though USDC is an erc20 token on many chains, the actual `address` is different on each chain. For this purpose, the registry also supports converting by NetworkId:
+
+```
+AssetsByIdAndNetwork ::= AssetId NetworkId -> LocalId
+LocalIdOn ::= LocalId NetworkId -> LocalId
+```
+
+These mappings should be used to ensure that the following program is constructed properly:
+
+```
+Spawn XYZ [}
+    Call { LocalId }
+] { {AssetId}: {Amount} }
+```
+
+Here the constructor needs to ensure that `AssetId` maps to the `LocalId` on chain `XYZ`. This can be hardcoded, or achieved by an alternative registry, but more robustly by interacting with our registry, to guarantee that the mapping is the same.
+
+### Queryable Data
+
+On-chain, we store as little data as possible, thus per asset the only values stored are:
+
+```
+AssetData ::= AssetId LocalId Decimals
+Decimals ::= u8
+```
+
+Where `Decimals` can be used to identify the decimal point location. This is especially important when bridging between fastly different ecosystems, where for example 1 USDC could be expressed as `100` in one ecosystem, and `1_000_000` in another.
+
+## Updating the Registry
+
+The Asset Registry is always updated from the Composable Parachain, and then updates are propagated over `XCVM` to connected chains.
+
+```mermaid
+sequenceDiagram
+
+participant P as Proposer
+participant PR as Parachain Registry
+participant G as Governance
+participant X as XCVM
+
+P->>PR: Propose Registry Update
+G->>PR: Approve Update
+PR->>X: Propagate Update
+```
+
+This means that the most current source of truth (SoT) is the Composable Parachain state.
+
+In case of emergencies, asset registries on connected chains can be updated directly. These updates can be disputed and must then be settled by governance. Some examples of emergencies:
+
+- Block reorderings changing `LocalIds`.
+- Protocol errors (bugs).
--- a/xcvm/architecture/asset-registry/global-registry.md
+++ b/xcvm/architecture/asset-registry/global-registry.md
@@ -41,7 +41,7 @@ AssetData ::= AssetId LocalId Decimals
 Decimals ::= u8
 ```
 
-Where `Decimals` can be used to identify the decimal point location. This is especially important when bridging between fastly different ecosystems, where for example 1 USDC could be expressed as `100` in one ecosystem, and `1_000_000` in another.
+Where `Decimals` can be used to identify the decimal point location. This is especially important when bridging between vastly different ecosystems, where for example 1 USDC could be expressed as `100` in one ecosystem, and `1_000_000` in another.
 
 ## Updating the Registry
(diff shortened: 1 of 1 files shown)


xcvm/architecture/asset-registry/global-registry.md
```diff
@@ -0,0 +1,68 @@
...
+AssetsById ::= AssetId -> LocalId
+IdByAssets ::= LocalId -> AssetId
+```
+
+## Querying the Registry
+
+When constructing `Call`s, we often need to convert `LocalId`s to `LocalId`s based on the `NetworkId`. Even though USDC is an erc20 token on many chains, the actual `address` is different on each chain. For this purpose, the registry also supports converting by NetworkId:
+
+```
+AssetsByIdAndNetwork ::= AssetId NetworkId -> LocalId
```
aeryz, line 21: Since there are separate registries running on each network, why do we use network id? Let's say that a token's global identifier is 1, this will be resolved in chain A as something (some address) and chain B as something else. Could you tell a use case so that I can understand?

KaiserKarel: I'll rephrase this section, but the use case is actually:

```
Spawn XYZ [}
    Call { LocalId }
] { {AssetId}: {Amount} }
```

If we're transferring `AssetId`, to then do `Call` on the other chain, and Call is a uniswap trade call, what asset id do we need to use when constructing the `Call` on the origin chain? (The one that AssetId maps to on the destination chain).  So the contract will want to do something like:

```
/// called by a user
fn lp_on_fantom(assetId: AssetId, amount: Balance) {
    let id_on_fantom = xcvm.assets.get(Fantom::NetworkId, assetId).unwrap();
    let program = spawn
        .on(Fantom, (assetId, amount))
        .call(Univ2::lp(0x0000, id_on_fantom, amount)) // Here we cannot use AssetId, but need a LocalId
}
```

aeryz: Makes sense, I thought we were gonna use network id for transfer instruction as well.

dzmitry-lahoda: > Call { LocalId }

Not sure why we use local calls?  Can we specify global ids as XCVM app developers, and then interpreters maps behind the scenes?

xcvm/architecture/asset-registry/global-registry.md
```diff
@@ -0,0 +1,68 @@
...
+### Queryable Data
+
+On-chain, we store as little data as possible, thus per asset the only values stored are:
+
+```
+AssetData ::= AssetId LocalId Decimals
+Decimals ::= u8
+```
+
+Where `Decimals` can be used to identify the decimal point location. This is especially important when bridging between fastly different ecosystems, where for example 1 USDC could be expressed as `100` in one ecosystem, and `1_000_000` in another.
```
dzmitry-lahoda, line 44: vastly

xcvm/architecture/asset-registry/global-registry.md
```diff
@@ -0,0 +1,68 @@
...
+Spawn XYZ [}
+    Call { LocalId }
+] { {AssetId}: {Amount} }
+```
+
+Here the constructor needs to ensure that `AssetId` maps to the `LocalId` on chain `XYZ`. This can be hardcoded, or achieved by an alternative registry, but more robustly by interacting with our registry, to guarantee that the mapping is the same.
+
+### Queryable Data
+
+On-chain, we store as little data as possible, thus per asset the only values stored are:
```
dzmitry-lahoda, line 37: also symbols are not executed on network, yet they are good for debugging, coding (name convention), governed and security mechanism. so may be remove `as little as possible` just state what we do.

xcvm/architecture/asset-registry/global-registry.md
```diff
@@ -0,0 +1,68 @@
...
+
+```
+AssetData ::= AssetId LocalId Decimals
```
dzmitry-lahoda, line 40: it can be useful to disable mapping. and consider mappings immutable. so that same id is never mapped to different token.

xcvm/architecture/asset-registry/global-registry.md
KaiserKarel: So the problem is that we cannot do late-bindings, as Call { LocalId } is opague. To the interpreter its just bytes and we cannot figure out what we should put there. We could change the type of a Call payload to:

```
Call ::= bytes [(Index, Value)]
Value ::= AssetId | Self
```

And we define the handling of call to be that the interpreter inserts the mapped Values at index.

xcvm/architecture/asset-registry/global-registry.md
KaiserKarel: Tokens sometimes migrate between contracts :(

xcvm/architecture/asset-registry/global-registry.md
KaiserKarel: Note that encoding becomes a lot more expensive this way.

dzmitry-lahoda (approved):

xcvm/architecture/asset-registry/global-registry.md
KaiserKarel: It might actually not technically be more expensive, but we'd probably need a different codegen tool; as for example ethers.rs will not support this.

KaiserKarel:
@Dom-Roth do you happen to know why the runtime test fails?

xcvm/architecture/asset-registry/global-registry.md
```diff
@@ -0,0 +1,68 @@
...
+```
+AssetsByIdAndNetwork ::= AssetId NetworkId -> LocalId
+LocalIdOn ::= LocalId NetworkId -> LocalId
+```
+
+These mappings should be used to ensure that the following program is constructed properly:
+
+```
+Spawn XYZ [}
+    Call { LocalId }
```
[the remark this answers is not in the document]
dzmitry-lahoda: > This can be hardcoded, or achieved by an alternative registry, but more robustly by interacting with our registry,

I think this is best and default. Trying to map LocalIdOn is future improvement? Can we retain only global AssetIds? For now.

dzmitry-lahoda (approved):

dzmitry-lahoda:
@KaiserKarel I think first release can be simplified by removid Local+ Network <=> Local mapping and removing Decimals for assets we handle first.

xcvm/architecture/asset-registry/global-registry.md
```diff
@@ -0,0 +1,68 @@
...
+The XCVM interpreter performs a look-up when executing instructions such as `Spawn` and `Transfer` to translate the local identifier to a global one. Local identifiers are chain specific, such as `address` in EVM chains. There is a two-way supported lookup:
+
+```
+AssetsById ::= AssetId -> LocalId
+IdByAssets ::= LocalId -> AssetId
+```
+
+## Querying the Registry
+
+When constructing `Call`s, we often need to convert `LocalId`s to `LocalId`s based on the `NetworkId`. Even though USDC is an erc20 token on many chains, the actual `address` is different on each chain. For this purpose, the registry also supports converting by NetworkId:
```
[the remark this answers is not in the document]
aeryz: So the thing is lets say you have the same erc20 tokens on two different chains. Although the token is the same token, since the chain is different, their `LocalId` which is `Address` in this case will be different. So this is a way to identify a specific token on specific chain.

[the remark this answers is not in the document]
dzmitry-lahoda: it leads many to many maps `each chain to each chain`(many to many on each chain). while `global asset id` to `this`(local)  chain asset address/id makes map one to many on each chain.

PR #1447, opened 2022-08-15
#CU-3ajfdfq nix karura without preopened channels

dzmitry-lahoda:
## Description
We need karura run with and without preopened channels, also we we need picass+kusama only or with karura. So we have builder. And test on it.


ComposableFiB0t:
Task linked: [CU-3ajfdfq Run Picasso and Karura chain spec locally no open channel by default](https://app.clickup.com/t/3ajfdfq)

PR #1466, opened 2022-08-18
[Merged by Bors] - #CU-2mezh63 junod client for cosmos for xcvm

dzmitry-lahoda:


cor (approved):
LGTM

flake.nix
```diff
@@ -458,61 +478,85 @@
...
-                rnix-lsp
-                rust-nightly
-                taplo
-                wasm-optimizer
-                xorriso
-                zlib.out
-              ];
-              NIX_PATH = "nixpkgs=${pkgs.path}";
+
+            junod = pkgs.callPackage ./.nix/junod.nix { };
```
cor, line 508: I think you accidentally added this as a devshell instead of as a package

cor (dismissed):
one fix needed

hussein-aitlahcen (approved):

flake.nix
```diff
@@ -458,61 +478,85 @@
...
               buildInputs = with packages; [ rust-nightly ];
               NIX_PATH = "nixpkgs=${pkgs.path}";
             });
 
+            developers-xcvm = developers-minimal.overrideAttrs (base: {
+              buildInputs = with packages;
+                base.buildInputs ++
+                [
+                  junod
+                  devnet-dali
```
cor, line 548: I'm not sure if we need to include junod and devnet-dali in the developer shells. I think these will only be used by other `nix run ...` commands

flake.nix
```diff
@@ -120,45 +131,53 @@
...
+                  let
+                    customFilter = name: type:
+                      (
+                        !(type == "directory"
+                        && builtins.elem (baseNameOf name) dir-blacklist)
+                      )
+                      && (
+                        !(type == "file"
+                        && builtins.elem (baseNameOf name) file-blacklist)
+                      );
```
cor, line 174: Please change to
```
!(type == directory) && builtins.elem (baseNameOf name) (dir-blacklist ++ file-blacklist)
```

cor (changes requested):

flake.nix
dzmitry-lahoda: bad merge

flake.nix
dzmitry-lahoda: i will use junod for sure. 

why you think that only nix run will use it?

hussein-aitlahcen (approved):

KaiserKarel (approved):

PR #1467, opened 2022-08-18
[CU-3axvr5v] Picasso overview stats - Subsquid integration

andresmechali:
On draft until Firesquid is merged

## Issue
https://app.clickup.com/t/3axvr5v

Changes include:

https://app.zeplin.io/project/61e6d8b9c4b6e686f1213af7/screen/624f3ad8f924c961e249d699

and

image

Note: This does not include TVL, which depends on staking and fNFTs.


ComposableFiB0t:
Task linked: [CU-3axvr5v Picasso Stats FE Integrations Subsquid Requirements](https://app.clickup.com/t/3axvr5v)

subsquid/schema.graphql
```diff
@@ -96,10 +93,100 @@ type VestingSchedule @entity {
...
+  STAKING_REWARDS_SPLIT_POSITION,
+  BONDED_FINANCE_NEW_OFFER,
+  BONDED_FINANCE_NEW_BOND,
+  BONDED_FINANCE_OFFER_CANCELLED,
+  VESTING_SCHEDULES_VESTING_SCHEDULE_ADDED,
+  VESTING_SCHEDULES_CLAIMED,
+  VESTING_SCHEDULES_VESTING_SCHEDULES_UPDATED,
+}
+
+type PicassoTransaction @entity {
```
vimukthi-git, line 127: This does not look like the most optimal or reusable way to handle the transactions in a generic way for both Picasso and Pablo.

subsquid/schema.graphql
```diff
@@ -96,10 +93,100 @@ type VestingSchedule @entity {
...
+  id: ID!
+  "ID of the event associated with this activity"
+  eventId: String!
+  "ID of the asset"
+  asset: Asset!
+  price: BigInt!
+  timestamp: BigInt!
+}
+
+type PicassoStakingPosition @entity {
```
vimukthi-git, line 172: We can model a single staking position and have them distinguished by the staked assetId instead of creating a data structure specific to one asset type

andresmechali: Yeah, the thing is I want to store them in USD so that we have a good representation of historical data. I will think of a better way to make it more flexible.

subsquid/schema.graphql
andresmechali: Yes, that's true. I did it because there was already a `PabloTransaction` but now I've unified them in a single `Transaction` entity, which has an optional `PabloTransaction` field with the relevant information for Pablo.

subsquid/db/migrations/1662371228302-Init.js
```diff
@@ -0,0 +1,105 @@
+module.exports = class Init1662371228302 {
+  name = "Init1662371228302";
+
+  async up(db) {
+    await db.query(
```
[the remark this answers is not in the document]
andresmechali: This is an auto-generated file so I don't think I have much power here. If I change it, it will need to be manually done every time the schema changes. Anyway, this is just about the creation of the tables so it should only run once and with no data right?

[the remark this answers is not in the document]
andresmechali: This is the current ER diagram:

er-diagram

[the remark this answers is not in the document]
easteregg: This is a great convo. in the latest Firesquid implementation, we've generated everything from scratch and migrated (inside docker). 

But this approach is certainly significant when it comes to staging/production. 

I couldn't find the deploy script for this, but we will lose data if we generate the migration file from scratch every time we release it.

subsquid/schema.graphql
easteregg: I also think that `USD` is mera here. it would be nice to add a currency field to this to declare the price in that currency.

subsquid/schema.graphql
```diff
@@ -96,10 +68,136 @@ type VestingSchedule @entity {
...
+  positionId: String!
+  transactionId: String!
+  poolId: String!
+  amount: BigInt!
+  owner: String!
+  startTimestamp: BigInt!
+  endTimestamp: BigInt
+}
+
+type PicassoPool @entity {
```
easteregg, line 128: (Question) What is the use case for this?

andresmechali: Good question. When a pool is created we can store the asset and any other information that might not be always available, so we can use it later on. For example when a user stakes on a pool, we can get the corresponding asset, which will in turn be used to query the current price from the oracle.

subsquid/db/migrations/1662371228302-Init.js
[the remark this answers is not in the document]
andresmechali: Screen Shot 2022-09-06 at 17 05 16

PR #1473, opened 2022-08-18
Ben/rewards pots/#2tyky2z

benluelo:
TODO:

- [x] add more tests
- [x] add benchmarks

Any `dbg!`s/ commented out code will be removed once I finish writing tests
I will run the benchmarks in a followup PR


ComposableFiB0t:
Task linked: [CU-2tyky2z Allow setting reward allocation by governance instead of inflation for pool rewards](https://app.clickup.com/t/2tyky2z)

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
```diff
@@ -498,8 +498,34 @@ endwhile (no)
...
-==== Update Reward Pool
+===== Update Reward Allocation Per Pool
+
+Each reward pool would have its own reward pot account.
+
+* sub-accounts for the rewards pots,
+** A prefix is needed to differentiate between the pallet "hot wallet" which stores the claimable funds and the "cold wallet" which transferred provided treasury funds.
+... Hot wallet prefix: "Hot"
+... Cold wallet prefix: "Cold"
+* Lock the assets in the cold wallet so that funds can be transferred only when unlocked (defensive programming)
```
dzmitry-lahoda, line 511: not sure. you can lock funds without having hot or cold. via lock. it will lock some amount. also there is reserve. i mean not sure why 2 wallets are needed? what mechanics it helps, except defensive programming?

dzmitry-lahoda (changes requested):
you complicate things, you can just lock funds. example - vesting pallet. there are funds, but these are locked.

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
benluelo: it provides a distinction between the claimable funds and the currently locked funds - an extra layer between the users/claiming and the locked/held funds.

benluelo: if you think that just locking the funds in one wallet is fine, then we can discuss it  with @vimukthi-git

vimukthi-git: I'm ok with just locking in a single account, cold/hot wallet concept is not needed for derived pallet accounts as they accessed only programmatically as I understand more.

vimukthi-git (commented):
Let's just go with the solution of locking funds in a single account for each pool

rfcs/0005-pablo-distribution-assets/0005-pablo-distribution.adoc
benluelo: Updated in https://github.com/ComposableFi/composable/pull/1473/commits/fcaddfc09bfea06e1f463ab55ea340f1afd47c12

frame/staking-rewards/src/lib.rs
```diff
@@ -1051,58 +1101,103 @@ fn update_rewards_pool<T: Config>(
...
-		Err(_) => Err(RewardAccumulationCalculationError::BackToTheFuture(reward)),
+		} else {
+			// at this point, reward.total_rewards is known to be equal to max_rewards which means
+			// that the max rewards was hit previously
+			Err(RewardAccumulationCalculationError::MaxRewardsAccumulatedPreviously)
+		}
+	}
+}
+
+/// Releases and transfers the specified amount of funds from the cold wallet to the hot wallet.
```
vimukthi-git, line 1170:
```suggestion
/// Releases the specified amount of funds from in the reward pool pallet account so that it's claimable.
```

vimukthi-git (approved):
Need to update the design doc to match the latest consensus not to use hot/cold wallets

frame/staking-rewards/src/lib.rs
```diff
@@ -440,6 +462,21 @@ pub mod pallet {
 			T::RewardPoolUpdateOrigin::ensure_origin(origin)?;
 			update_rewards_pool::<T>(pool_id, reward_updates)
 		}
+
+		/// Add funds to the reward pool's rewards pot for the specified asset.
+		///
+		/// Emits `RewardsPotIncreased` when successful.
+		#[pallet::weight(10_000)]
```
andor0, line 469:
```suggestion
		#[pallet::weight(T::WeightInfo::add_to_rewards_pot())]
```

frame/staking-rewards/src/lib.rs
```diff
@@ -840,80 +878,65 @@ pub mod pallet {
...
+					Self::reward_accumulation_hook_reward_update_calculation(
+						pool_id,
+						reward,
+						now_seconds,
+					);
+				}
 
-				RewardPools::<T>::insert(pool_id, reward_pool);
+				// 128 bit platforms don't exist as of writing this so this usize -> u64 cast
+				// should be ok
```
andor0, line 930: ` reward_pool.rewards` is limited `T::MaxRewardConfigsPerPool`, [T::MaxRewardConfigsPerPool](https://github.com/ComposableFi/composable/blob/main/frame/staking-rewards/src/lib.rs#L270) is `u32`.

frame/staking-rewards/src/lib.rs
benluelo: Docs have been updated :+1:

frame/staking-rewards/src/lib.rs
```diff
@@ -1051,58 +1116,121 @@ fn update_rewards_pool<T: Config>(
...
+		let total_locked_rewards: u128 =
+			T::Assets::balance_on_hold(reward.asset_id, pool_account).into();
+
+		// the maximum amount repayable given the reward rate.
+		// i.e. if total locked is 50, and the reward rate is 15, then this would be 3
+		//
+		// SAFETY(benluelo): Usage of Div::div: Integer division can only fail if rhs == 0, and
+		// reward.reward_rate.amount is known to be non-zero here as per the check at the beginning
+		// of this function
+		let maximum_releasable_periods = total_locked_rewards.div(reward.reward_rate.amount.into());
```
vimukthi-git, line 1151: Not sure why we need this, Can't we do,
```
newly_accumulated_rewards = min(periods_surpassed * reward.reward_rate.amount, total_locked_rewards)
```
To get the unlockable amount? Period related calculations seem redundant to me.

frame/staking-rewards/src/lib.rs
benluelo: good point, I've updated the comment :+1:

frame/staking-rewards/src/lib.rs
benluelo: we need to keep the accumulated amount in sync with the period, otherwise we need to track partially rewarded periods somewhere

vimukthi-git: I see, just wondering then wouldn't it be better then to calculate and set the `last_updated_time` to be in the past to match 
`(periods_surpassed * reward.reward_rate.amount - total_locked_rewards) / reward_rate` ? Then we only need to do period related calculations in case the amount to be unlocked is larger than `total_locked_rewards`.

frame/staking-rewards/src/lib.rs
```diff
@@ -1051,58 +1118,121 @@ fn update_rewards_pool<T: Config>(
...
+			newly_accumulated_rewards.saturating_add(reward.total_rewards.into());
+
+		// u64::MAX is roughly 584.9 billion years in the future, so saturating at that should be ok
+		let last_updated_timestamp = reward.last_updated_timestamp.saturating_add(
+			reward_rate_period_seconds
+				.get()
+				.saturating_mul(releasable_periods_surpassed.saturated_into::<u64>()),
+		);
+
+		if new_total_rewards <= reward.max_rewards.into() {
```
vimukthi-git, line 1172:
```suggestion
		if new_total_rewards <= (reward.max_rewards + reward.total_dilution_adjustment).into() {
```

frame/staking-rewards/src/lib.rs
```diff
@@ -1051,58 +1118,121 @@ fn update_rewards_pool<T: Config>(
...
+				)
+				.expect("funds should be avaliable to release based on previous check; qed;");
+
+				reward.total_rewards = new_total_rewards.into();
+				reward.last_updated_timestamp = last_updated_timestamp;
+				Ok(())
+			} else {
+				Err(RewardAccumulationCalculationError::RewardsPotEmpty)
+			}
+		} else if reward.total_rewards < reward.max_rewards {
```
vimukthi-git, line 1189:
```suggestion
		} else if reward.total_rewards < (reward.max_rewards + reward.total_dilution_adjustment) {
```

frame/staking-rewards/src/lib.rs
vimukthi-git: It's ok to add a TODO for simplifying this and do it in a follow up PR :)

frame/staking-rewards/src/lib.rs
benluelo: I added the dilution adjustment to the total rewards since max_rewards is documented as the total + dilution adjustment

dzmitry-lahoda (approved):

frame/staking-rewards/src/lib.rs
```diff
@@ -1051,58 +1118,128 @@ fn update_rewards_pool<T: Config>(
...
-		Err(_) => Err(RewardAccumulationCalculationError::BackToTheFuture(reward)),
+		} else if reward.total_rewards.saturating_add(reward.total_dilution_adjustment) <
+			reward.max_rewards
+		{
+			// if the new total rewards are less than or equal to the max rewards AND the current
+			// total rewards are less than the max rewards (i.e. the newly accumulated rewards is
+			// less than the the amount that would be accumulated based on the periods surpassed),
+			// then release *up to* the max rewards
+			//
+			// REVIEW(benluelo): Should max_rewards be max_periods instead? Currently, the
```
vimukthi-git, line 1200: Given how this is evolving, my idea is that we can completely remove the `max_rewards` field. Having already a reward pot means that there is already a way to control the amounts being rewarded. We can make it in the future so that if the reward pot account is empty the pool stops being active.

vimukthi-git (commented):
I will regenerate the md and approve this

frame/staking-rewards/src/lib.rs
benluelo: I think that would make more sense - would we want to then restrict who can add to the pool?

benluelo: (I will create a ticket to review this)

vimukthi-git: It should be ok to have anyone add to the pool and keep the pool alive actually, that's basically a gauge/bribe protocol :)

vimukthi-git (approved):
LGTM

vimukthi-git:
@Mergifyio refresh

benluelo:
@Mergifyio requeue

PR #1478, opened 2022-08-19
[CU-3baupye] emit PriceChanged event when price changes for selected assets

andresmechali:
## Issue
https://app.clickup.com/t/3baupye


ComposableFiB0t:
Task linked: [CU-3baupye Emit `PriceChanged` event on `Oracle` pallet](https://app.clickup.com/t/3baupye)

frame/oracle/src/lib.rs
```diff
@@ -981,6 +989,14 @@ pub mod pallet {
 					PrePrices::<T>::remove(asset_id);
 
 					Self::handle_payout(&pre_prices, price, asset_id, &asset_info)?;
+
+					// Emit `PriceChanged` event when PICA or PABLO prices have changed.
+					if price != last_price &&
+						vec![CurrencyId::PICA, CurrencyId::PBLO]
```
vimukthi-git, line 995: We can set this as an attribute/flag of `asset_info` and based on that emit the event rather than hard coding

frame/oracle/src/tests.rs
```diff
@@ -1368,6 +1368,70 @@ fn on_init() {
...
+		// Set series of EQUAL prices for PICA into storage.
+		//				 -----
+		for _ in 0..3 {
+			let price = 100_u128;
+			add_price_storage(price, 1, account_1, 2);
+		}
+
+		// Set and initialize block 4.
+		System::set_block_number(4);
+		Oracle::on_initialize(4);
```
benluelo, line 1422: Consider using `process_and_progress_blocks` (`frame/composable-tests-helpers/src/test/block.rs`)

frame/oracle/src/lib.rs
vimukthi-git: The consideration is whether we want the event emission for a particular asset type to be configurable with an extrinsic call or runtime upgrade. I think in this case as assets are configured with a extrinsic it makes sense to use the same way to configure whether events are emitted for them. So let's go with a new flag on `asset_info`.

vimukthi-git (approved):
LGTM

frame/oracle/src/lib.rs
```diff
@@ -347,6 +348,8 @@ pub mod pallet {
 		RewardingAdjustment(T::Moment),
 		/// Answer from oracle removed for staleness. \[oracle_address, price\]
 		AnswerPruned(T::AccountId, T::PriceValue),
+		/// Price changed by oracle \[asset_id, price\]
```
blasrodri, line 351: why the "\" in comments?

andresmechali: Just following previous code convention. I assume it's escape the `[` for some reason.

benluelo: rustdoc tries to read `[content]` as a link and removes the brackets @blasrodri

vimukthi-git (approved):
LGTM

benluelo (approved):

Dom-Roth (approved):

Juliansv (approved):
LGTM

easteregg (approved):

Dom-Roth (approved):

KaiserKarel (approved):

PR #1491, opened 2022-08-23
#WIP #CU-3bf7dy0 xcvm cosmwasm genesis state and contract application script

dzmitry-lahoda:

- ensure we are test env
- get alice and validator keys
- manually applied and cp genesis + keys
- reference prebuild wasms (use their docker to build wasmswap)
- put these into container and into junod client
- script apply contract


dzmitry-lahoda:
@Mergyfyio rebase

PR #1497, opened 2022-08-24
chore: remove paths-ignore

KaiserKarel:
Some CI steps steps were being skipped because of a yaml rule. Now that we are using nix, we're letting nix handle caching etc.

Signed-off-by: Karel L. Kubat <k.l.kubat@gmail.com>


dzmitry-lahoda (approved):

alexkdefi (approved):
👍

KaiserKarel:
bors cancel

PR #1501, opened 2022-08-24
[CU-2vj3uvt] [CU-2t22kwn] Introduce Asset IDs for financial NFT as well as the shares of the pool

PoisonPhang:
## Issue
[CU-2vj3uvt](https://app.clickup.com/t/2vj3uvt)
[CU-2t22kwn](https://app.clickup.com/t/2t22kwn)

## Description

- [x] Introduce shared asset ID (xTokens)
- [x] Introduce fNFT Collection ID
- [X] Add range for xTokens to Currency Factory
- [x] Derive default value for fNFT Stake Collection
- [x] Add currency IDs for CU-2t22kwn


ComposableFiB0t:
Task linked: [CU-2vj3uvt [Staking-rewards] Introduce asset ids for financial NFT as well as the shares of the pool](https://app.clickup.com/t/2vj3uvt)

frame/pablo/src/lib.rs
```diff
@@ -685,6 +685,9 @@ pub mod pallet {
 				end_block,
 				reward_configs,
 				lock,
+				share_asset_id: primitives::currency::CurrencyId::xPBLO.0.into(),
+				// TODO: Derive default fNFT collection ID for staking
+				financial_nft_asset_id: todo!(),
```
vimukthi-git, line 690: 1. We can create a new range for fNFT asset IDs as well
2. We can use a special one in the range for the PBLO and PICA stake fNFTs

frame/composable-traits/src/dex.rs
```diff
@@ -32,6 +32,9 @@ pub trait Amm {
 
 	fn lp_token(pool_id: Self::PoolId) -> Result<Self::AssetId, DispatchError>;
 
+	/// Returns the xToken given to users when they stake
+	fn x_token(pool_id: Self::PoolId) -> Result<Self::AssetId, DispatchError>;
```
vimukthi-git, line 36: Why is this needed in the dex trait?

frame/pablo/src/lib.rs
PoisonPhang: 4a92e26f93c539223ab07f4cc29e85a06a6ef284

frame/composable-traits/src/dex.rs
PoisonPhang: It doesn't, tho I might consider moving the new functions to their own traits so they can be tested more easily
e065b5f9ac97e99d3f08d2cc5189c0d1d3c0b2e4

frame/composable-traits/src/staking/mod.rs
```diff
@@ -179,12 +180,13 @@ pub enum RewardPoolConfiguration<AccountId, AssetId, BlockNumber, RewardConfigs,
 		reward_configs: RewardConfigs,
 		// possible lock config for this reward
 		lock: LockConfig<DurationPresets>,
+
 		// TODO (vim): Introduce asset ids for financial NFT as well as the shares of the pool
```
vimukthi-git, line 184: Let's remove the TODO comments as they are addressed now

frame/composable-traits/src/staking/mod.rs
```diff
@@ -133,12 +133,13 @@ pub struct RewardPool<AccountId, AssetId, Balance, BlockNumber, DurationPresets,
 
 	// possible lock config for this pool
 	pub lock: LockConfig<DurationPresets>,
+
 	// TODO (vim): Introduce asset ids for financial NFT as well as the shares of the pool
```
vimukthi-git, line 137: Let's remove the TODO comments as they are addressed now

frame/staking-rewards/src/lib.rs
```diff
@@ -308,9 +308,21 @@ pub mod pallet {
 		#[pallet::constant]
 		type PicaAssetId: Get<Self::AssetId>;
 
+		#[pallet::constant]
+		type XPicaAssetId: Get<Self::AssetId>;
```
vimukthi-git, line 312: Not something we need to address in this PR, but we need to think of how to remove runtime specificity from the pallet

runtime/primitives/src/currency.rs
```diff
@@ -112,11 +113,41 @@ impl CurrencyId {
...
+		#[allow(non_upper_case_globals)]
+		pub const wBTC: CurrencyId = CurrencyId(132);
+		/// Wrapped ETH
+		#[allow(non_upper_case_globals)]
+		pub const wETH: CurrencyId = CurrencyId(133);
+		/// Acala Dollar
+		#[allow(non_upper_case_globals)]
+		pub const aUSD: CurrencyId = CurrencyId(134);
+
+		// Staked asset xTokens (1001 - 100_000_000_000)
```
vimukthi-git, line 132: What does the `100_000_000_000` in this comment mean?

vimukthi-git (approved):
No need to adjust the design docs with this PR as it's not a major algo update(Just need a few data structures to be updated). We can do a separate PR for that(if needed) to unblock anything depending on this PR :)

runtime/primitives/src/currency.rs
PoisonPhang: The upper limit to the rage available to this type of token. This can be changed, I chose it because this is at the bottom of the set ranges in CurrencyFactory.

frame/composable-traits/src/staking/mod.rs
PoisonPhang: bc781fa156cd6b8918e008183db87107308cfa2b

runtime/primitives/src/currency.rs
vimukthi-git: hmm but allowing that upper bound would mix the two ranges for x tokens and fnft right?

vimukthi-git (approved):

runtime/primitives/src/currency.rs
PoisonPhang: You're right. Good catch. I forgot to update this
0d21e689588a4796cfe0fab8b91f86e450360276

frame/composable-traits/src/currency.rs
```diff
@@ -31,6 +31,8 @@ impl RangeId {
 	pub const TOKENS: RangeId = RangeId(1);
 	pub const FOREIGN_ASSETS: RangeId = RangeId(2);
 	pub const IBC_ASSETS: RangeId = RangeId(3);
+	pub const FNFT_ASSETS: RangeId = RangeId(4);
+	pub const XTOKEN_ASSETS: RangeId = RangeId(6);
```
benluelo, line 35: What about 5?

benluelo (approved):

runtime/primitives/src/currency.rs
```diff
@@ -112,11 +113,41 @@ impl CurrencyId {
...
+		#[allow(non_upper_case_globals)]
+		pub const wBTC: CurrencyId = CurrencyId(132);
+		/// Wrapped ETH
+		#[allow(non_upper_case_globals)]
+		pub const wETH: CurrencyId = CurrencyId(133);
+		/// Acala Dollar
+		#[allow(non_upper_case_globals)]
+		pub const aUSD: CurrencyId = CurrencyId(134);
+
+		// Staked asset xTokens (1001 - 2000)
```
vimukthi-git, line 132: This and the staking NFTs grow in parallel so it makes sense to allow both of them to grow at the same rate and size. May be let's set aside 100k ids for both? (Being over optimistic I know ;)

PoisonPhang: These ranges are only for the preset fNFT IDs that are declared in our runtime code. The ID range for user-created ones have a massive range declared in Currency factory.
Do we still want the 100k range for the hard-coded ones?

frame/composable-traits/src/currency.rs
PoisonPhang: :sweat_smile: 
7317b8273beb7e5666f9cabd15b587548978e30d

frame/staking-rewards/src/lib.rs
PoisonPhang: I will move this conversation to slack so the PR can be merged

PR #1504, opened 2022-08-25
[cu-2w8zbnh] Make RewardPoolId same as the staked assetId

vimukthi-git:
Refer issue

### DIFF
--- a/frame/staking-rewards/src/test/test_reward_accumulation_hook.rs
+++ b/frame/staking-rewards/src/test/test_reward_accumulation_hook.rs
@@ -77,7 +77,7 @@ fn test_reward_update_calculation() {
 			] {
 				assert_no_event::<Test>(Event::StakingRewards(
 					crate::Event::<Test>::RewardAccumulationHookError {
-						pool_id,
+						pool_id: PICA::ID,
 						asset_id: PICA::ID,
 						error,
 					},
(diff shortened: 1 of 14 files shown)


ComposableFiB0t:
Task linked: [CU-2w8zbnh Add PICA and PBLO stake pools as part of genesis configuration of the staking rewards pallet and not create in Pablo pallet at pool creation time](https://app.clickup.com/t/2w8zbnh)

benluelo (approved):

frame/staking-rewards/src/test/test_reward_accumulation_hook.rs
```diff
@@ -505,7 +505,7 @@ fn test_accumulate_rewards_hook() {
 			]);
 
 			check_events([crate::Event::<Test>::MaxRewardsAccumulated {
-				pool_id: bobs_pool_id,
+				pool_id: C::ID,
```
benluelo, line 508: was this change required? (curious why there were changes in `check_events` and not in `check_rewards`)

vimukthi-git: It's because we no longer require an explicit `pool_id` to verify rewards, staked asset ID(`pool_asset_id`) is already the pool id. Made a change to get rid of it in the test anyway

benluelo (approved):

andor0 (approved):

hussein-aitlahcen (approved):

PR #1505, opened 2022-08-25
Add RELEASE.md

KaiserKarel:
We had already agreed on this flow, but not explicitly documented anywhere.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,49 @@
+# Releasing
+
+This repository uses a branch-based strategy to manage releases.
+
+> **Note**
+> This strategy has been introduced later on in this repository's lifetime, so not all releases have an associated branch.
+
+## Naming scheme
+
+A release branch is identified as follows:
+
+`release-{VERSION}{CANDIDATE}`
+
+where `VERSION` is a semantic version in the form of `vx.y.z`, and `CANDIDATE` is the optional `-rc` tag, indicating that it is a release candidate.
+
+- `release-v3.2.1-rc` is the candidate to become `release-v3.2.1`.
+
+## Creating a release
+
+1. Create a PR to the branch of `release-vx.y.z-rc` from `main`, where the version is the next semantic version of the repository depending on changes.
+
+```
+release-vx.y.z-rc <- main
+```
+
+2. Automation will test the release candidate and run integration tests for this release specifically.
+
+    - If fixes need to be applied to this release branch; PR the fixes to main, and cherry-pick the commits into the release candidate.
+
+3. Once the release candidate has been approved by QA and auditors, create a PR from the candidate to the actual release:
+
+```
+release-vx.y.z <- release-vx.y.z-rc
+```
+
+Automation will test this PR again, although we do not expect any failures at this point.
+
+4. Once `3.` is merged, automation will generate artifacts and create a Github release.
+
+## Backporting fixes
+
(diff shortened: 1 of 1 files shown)


vimukthi-git (approved):

RELEASE.md
```diff
@@ -0,0 +1,49 @@
...
+
+1. Create a PR to the branch of `release-vx.y.z-rc` from `main`, where the version is the next semantic version of the repository depending on changes.
+
+```
+release-vx.y.z-rc <- main
+```
+
+2. Automation will test the release candidate and run integration tests for this release specifically.
+
+    - If fixes need to be applied to this release branch; PR the fixes to main, and cherry-pick the commits into the release candidate.
```
dzmitry-lahoda, line 28: If foo bar, please read `## Backporting fixes` ?

RELEASE.md
```diff
@@ -0,0 +1,49 @@
...
+
+3. Once the release candidate has been approved by QA and auditors, create a PR from the candidate to the actual release:
```
dzmitry-lahoda, line 30: Just to be clear, auditors will approve branch or some diff? acala produced DIFF of all they have, so you can approve diff.

Also is this release just for wasm or it is release to be deployed and upgraded? 

Is version is for node or runtime? Acala has no version for wasm, just bump +1.

RELEASE.md
```diff
@@ -0,0 +1,49 @@
+# Releasing
```
dzmitry-lahoda, line 1: generally makes sence, so can split out quality gates as reference to other doc, because depending on what is released it may or may have various gates 

https://github.com/ComposableFi/composable/blob/main/rfcs/0006-picasso-runtime-upgrades.md

for me release is git commit accepted as releasable and relevant wasm +reference from it + diff (with accompanied release notes). + binary node built from commit.  likely node only needs semantic versioning, not sure about wasm. also there is sdk. but imagine we have repo builds, then git commit is root of all, semantic versions are just comments (release notes = notes + version comment)

RELEASE.md
```diff
@@ -0,0 +1,49 @@
...
+
+`release-{VERSION}{CANDIDATE}`
+
+where `VERSION` is a semantic version in the form of `vx.y.z`, and `CANDIDATE` is the optional `-rc` tag, indicating that it is a release candidate.
+
+- `release-v3.2.1-rc` is the candidate to become `release-v3.2.1`.
+
+## Creating a release
+
+1. Create a PR to the branch of `release-vx.y.z-rc` from `main`, where the version is the next semantic version of the repository depending on changes.
```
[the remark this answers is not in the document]
dzmitry-lahoda: so may be just tags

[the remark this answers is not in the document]
KaiserKarel: Actually we can protect branches by regex. Currently any branch matching `*release*` is protected

RELEASE.md
KaiserKarel: Version for the entire repo; so internal consistency of FE is taken into account too.

RELEASE.md
KaiserKarel: For just the parachain related stuff, that is correct. Note that the actual version of the runtime is determined by the `impl_version` and the `spec_version`. Ideally we stop hardcoding those, and generate them during build time using the semantic version.

RELEASE.md
KaiserKarel: I'm just putting this here twice, as this is often overlooked.

KaiserKarel:
@Mergifyio requeue

PR #1515, opened 2022-08-25
Have substrate-react package choose between injected extensions

saadjhk:
## Issue
Have substrate-react package choose between the extensions that are installed in users browser, previously polkadot-js/extension-dapp would select all wallets, which is definitely not the best user experience.

## Description
- Use API directly exposed by injected objects.

## Checklist

- [x] Test transactions functionality
- [x] Test accounts separation
- [x] Should work with Next JS
- [x] Enable disconnect functionality
- [x] Cache last connected wallet
- [x] Update `extension-dapp` package


frontend/packages/substrate-react/src/dotsama/DotSamaContext.tsx
```diff
@@ -77,47 +102,61 @@ export const DotSamaContextProvider = ({
...
         setParachainProviders((s) => {
-          s[chainId].accounts = accounts.map((x, i) => ({
-            address: x.address,
-            name: x.meta.name ?? i.toFixed(),
-          }));
-          return { ... s };
+          s[chainId].accounts = (accounts as InjectedAccountWithMeta[]).map(
+            (x, i) => ({
+              address: x.address,
+              name: x.meta.name ?? i.toFixed(),
```
easteregg, line 156: Question: instead of index as fallback, do you think it would be better for end user if we use the shortened address instead? something like:

`5Grwva..KutQY`

frontend/packages/substrate-react/src/dotsama/hooks/useEagerConnect.ts
```diff
@@ -18,7 +18,11 @@ export const useEagerConnect = (chainId: ParachainId): boolean => {
 
   useEffect(() => {
     if (parachainApi !== undefined && activate !== undefined && !hasTriedEagerConnect && extensionStatus === "initializing") {
-      activate(false);
+      // activate();
+      const usedWallet = localStorage.getItem("wallet-id");
+      if (usedWallet && usedWallet === "talisman" || usedWallet === "polkadot-js") {
```
easteregg, line 23: Possible to use enum or const for the strings?

e.g. 

`usedWallet === WALLETS.Talisman`

frontend/apps/pablo/defi/hooks/swaps/usePabloSwap.ts
```diff
@@ -1,75 +1,101 @@
...
 
-                    let pair = {
-                        base: baseAssetId,
-                        quote: quoteAssetId,
-                    };
+  const pabloSwap = useCallback(async (): Promise<string> => {
+    return new Promise(async (res, rej) => {
+      if (
+        parachainApi &&
+        signer !== undefined &&
```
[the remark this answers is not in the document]
saadjhk: @easteregg i think so as well, however it'll require changes in many other places as well. I will create a task for this and create a dedicated PR once we merge this one. Wdyt?

frontend/packages/substrate-react/src/dotsama/DotSamaContext.tsx
saadjhk: @easteregg can be done, it will look much better than a number.

frontend/packages/substrate-react/src/dotsama/hooks/useEagerConnect.ts
saadjhk: Sure

vimukthi-git (approved):

andresmechali (approved):

PR #1524, opened 2022-08-26
fix(cosmwasm): crypto host functions

aeryz:

## Description
It turns out that substrate's crypto functions and cosmwasm's crypto function are expecting signatures with different formats. The main difference is cosmwasm expects recovery param (recovery id in substrate) and the actual signature separately in `ecdsa(secp256k1)`. But substrate expects recovery param to be added to the end of the signature.

## Checklist

### DIFF
--- a/frame/cosmwasm/src/runtimes/wasmi.rs
+++ b/frame/cosmwasm/src/runtimes/wasmi.rs
@@ -24,6 +24,8 @@ use sp_std::{collections::btree_map::BTreeMap, vec::Vec};
 use wasmi::CanResume;
 use wasmi_validation::{validate_module, PlainValidator};
 
+const SUBSTRATE_ECDSA_SIGNATURE_LEN: usize = 65;
+
 #[derive(Debug)]
 pub enum CosmwasmVMError<T: Config> {
 	Interpreter(wasmi::Error),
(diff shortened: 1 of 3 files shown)


benluelo (approved):

frame/cosmwasm/src/runtimes/wasmi.rs
```diff
@@ -312,18 +314,31 @@ impl<'a, T: Config> VMBase for CosmwasmVM<'a, T> {
...
+		let message_hash = match message_hash.try_into() {
 			Ok(message_hash) => message_hash,
 			Err(_) => return Ok(Err(())),
 		};
 
+		let signature = {
+			// Since we fill `signature_inner` with `recovery_param`, when 64 bytes are written
+			// the final byte will be the `recovery_param`.
+			let mut signature_inner = [recovery_param; ECDSA_SIGNATURE_LEN];
+			signature_inner[..ECDSA_SIGNATURE_LEN - 1].copy_from_slice(signature);
```
dzmitry-lahoda, line 338: so this is SUBSTRATE_ECDSA_SIGNATURE_LEN

dzmitry-lahoda (approved):

hussein-aitlahcen (approved):

frame/cosmwasm/src/runtimes/wasmi.rs
aeryz: Yes, it would be better to rename to that.

hussein-aitlahcen (approved):

PR #1529, opened 2022-08-28
Refactor unlocked_claimable_balance function to use try_fold

andresmechali:


frame/vesting/src/lib.rs
```diff
@@ -567,20 +567,29 @@ impl<T: Config> Pallet<T> {
...
 	> {
-		let mut claims_per_schedule: BoundedBTreeMap<
-			T::VestingScheduleId,
-			BalanceOf<T>,
-			T::MaxVestingSchedules,
-		> = BoundedBTreeMap::new();
-
 		<VestingSchedules<T>>::try_mutate_exists(who, asset, |maybe_schedules| {
-			let mut total_balance_to_claim: BalanceOf<T> = Zero::zero();
 			match maybe_schedules {
```
benluelo, line 571: this `match` could be `let schedules = maybe_schedules.ok_or(Error::<T>::VestingScheduleNotFound)?`

frame/vesting/src/lib.rs
```diff
@@ -567,20 +567,29 @@ impl<T: Config> Pallet<T> {
...
+				Some(schedules) =>
+					vesting_schedule_ids.into_all_ids(schedules).iter().try_fold::<_, _, Result<
+						(
+							BalanceOf<T>,
+							BoundedBTreeMap<
+								T::VestingScheduleId,
+								BalanceOf<T>,
+								T::MaxVestingSchedules,
+							>,
+						),
```
benluelo, line 581: This type annotation is reduntant - type inference should be able to pick it up (see following suggestion)

frame/vesting/src/lib.rs
```diff
@@ -567,20 +567,29 @@ impl<T: Config> Pallet<T> {
...
+						DispatchError,
+					>>(
+						(
+							Zero::zero(),
```
benluelo, line 585:
```suggestion
							BalanceOf::<T>::zero()
```

benluelo (commented):
LGTM, just a few things :+1:

frame/vesting/src/lib.rs
andresmechali: Right, with this change the type was properly inferred.

frame/vesting/src/lib.rs
```diff
@@ -610,16 +619,13 @@ impl<T: Config> Pallet<T> {
...
 							if locked_amount.is_zero() {
-								// remove fully claimed schedules
+								// Remove fully claimed schedules
 								schedules.remove(id_to_claim);
-							}
-
-							Ok(())
-						})
-						.collect::<Result<Vec<_>, DispatchError>>()?;
+							};
```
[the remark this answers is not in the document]
andresmechali: I think it's fine to do it at the end, as it uses the schedule before.

PoisonPhang (approved):

dzmitry-lahoda (approved):

PR #1531, opened 2022-08-28
feat(ci): ensure all checks pass before integration tests

hussein-aitlahcen:


dzmitry-lahoda (changes requested):
test should start asap its artifact read, if your want to make gate job you can do it separately

dzmitry-lahoda:
I mean we do not need to slow down most of builds. E.g. integration tests do not need linting. Let waste some CPU in parallel.

dzmitry-lahoda:
image

so we do all builds in parallel (sure so devnet can be only after nodes builds), 

but have one big gate on publish - deploy devnets, publis docker images, UI, etc.

dzmitry-lahoda:
closing this in favor if single effects gate. for now we will do it manually in yaml - define order in gate. so when and if hydra is used - then it will be more easy to build validate and auto paralyze DAG cc @cor

dzmitry-lahoda:
will add more stuctured gates in next PRs

PR #1534, opened 2022-08-29
nixify frontend-pablo and frontend-picasso

cor:
## Description
adds frontend-pablo and frontend-picasso to the devnet-up

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -34,8 +34,8 @@
       inputs.nixpkgs.follows = "nixpkgs";
     };
   };
-  outputs =
-    { self, nixpkgs, crane, flake-utils, rust-overlay, npm-buildpackage, arion-src }:
+  outputs = { self, nixpkgs, crane, flake-utils, rust-overlay, npm-buildpackage
+    , arion-src }:
     let
       # https://cloud.google.com/iam/docs/creating-managing-service-account-keys
       # or just use GOOGLE_APPLICATION_CREDENTIALS env as path to file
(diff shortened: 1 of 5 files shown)


hussein-aitlahcen (approved):

flake.nix
```diff
@@ -468,7 +468,37 @@
               '';
             };
 
+            frontend-static = let bp = pkgs.callPackage npm-buildpackage { };
+            in bp.buildYarnPackage {
+              nativeBuildInputs = [ pkgs.pkg-config pkgs.vips pkgs.python3 ];
+              src = ./frontend;
+              yarnBuildMore = "yarn export";
```
dzmitry-lahoda, line 475: just double check, there was not special yarn nix pacakges? so we use npm. i have not google so.

flake.nix
```diff
@@ -468,7 +468,37 @@
...
+              text = ''
+                miniserve -p 8002 --spa --index index.html ${frontend-static}/pablo
+              '';
+            };
+
+            frontend-picasso-server = pkgs.writeShellApplication {
+              name = "frontend-picasso-server";
+              runtimeInputs = [ pkgs.miniserve ];
+              text = ''
+                miniserve -p 8003 --spa --index index.html ${frontend-static}/picasso
```
dzmitry-lahoda, line 497: so we handle that as native and as docker.

dzmitry-lahoda (approved):

flake.nix
cor: No we're using the yarn.lock now combined with serokell's buildYarnPackage which should have a good enough reproducibility guarantee as i believe it stores hashes of all of the dependencies. It is true that there are also node packages in nixpkgs,which is an option we can consider in the future, but probably not what we want because we want to still let our fe devs be able to use `yarn` directly.

hussein-aitlahcen (approved):

cor:
@mergify requeue

dzmitry-lahoda (approved):

cor:
@mergify queue

PR #1546, opened 2022-08-30
feat(nix): avoid volumes for better macos support

hussein-aitlahcen:


.nix/services/hasura.nix
```diff
@@ -1,6 +1,30 @@
...
+    files = pkgs.linkFarm "context" [
+      {
+        name = "metadata";
+        path = metadata;
+      }
+      {
+        name = "Dockerfile";
+        path = pkgs.writeText "Dockerfile" ''
+          FROM hasura/graphql-engine:v2.12.0-beta.1.cli-migrations-v3
+          COPY metadata /hasura-metadata
```
dzmitry-lahoda, line 13: https://github.com/NixOS/nixpkgs/blob/578f231a53bfbedfad4b4cc17547f67ed99a72ba/pkgs/build-support/docker/examples.nix#L27 ?

dzmitry-lahoda: I guess metadata can be made other derivation which is mounted here?

dzmitry-lahoda (approved):

.nix/services/hasura.nix
hussein-aitlahcen: it's arion gere

hussein-aitlahcen: Here*

dzmitry-lahoda: oh, my bad, sorry :)

dzmitry-lahoda: i though they integrated, likely in future:)

hussein-aitlahcen: Yes hopefully :). In any case even if we copy store paths you need a way to have the path being copied in a target if the tool has no env variable to set an arbitrary one

cor (approved):
LGTM

PR #1550, opened 2022-08-31
convert_and_multiply_by_rational vault pallet refactor

Matteocolella1995:
## Issue
The new `multiply_by_rational_with_rounding` function introduced in polkadot 0.9.27 returns `None` when the divisor is 0 or when an `Overflow` occurs. I have modified the `convert_and_multiply_by_rational` to reproduce the old `multiply_by_rational` behavior, adding a fourth parameter to decide which type of `Rounding` to apply. 

## Description
I have chosen to do an `Rounding::Up` for the functions `do_calculate_lp_tokens_to_mint` because more `lp_tokens` would provide a (slightly) better representation of the collateral deposited. For the functions `do_amount_of_lp_token_for_added_liquidity` and `do_lp_share_value`, since they are one the inverse of the other, I have opted for opposite roundings, giving the `Down` rounding to the `do_lp_share_value` since it's usually used before burning `lp_tokens` (same point of the `do_calculate_lp_tokens_to_mint` function).

I have also reverted the check on the test `vault_withdraw_with_zero_lp_issued_fails_to_burn` from `Overflow` to `InsufficientLpTokens`

## Checklist

### DIFF
--- a/frame/composable-traits/src/vesting.rs
+++ b/frame/composable-traits/src/vesting.rs
@@ -105,7 +105,7 @@ pub struct VestingSchedule<VestingScheduleId, BlockNumber, Moment, Balance: HasC
 	/// Amount of tokens to release per vest
 	#[codec(compact)]
 	pub per_period: Balance,
-	/// Amout already claimed
+	/// Amount already claimed
 	pub already_claimed: Balance,
 }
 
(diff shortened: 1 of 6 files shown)


Matteocolella1995:
I have merged with the recent main changes again (yesterday's changes) to fix the `udeps` error, but cannot fix the spellchecks check unfortunately.

PoisonPhang (approved):

kevin-composable (commented):
Everything looks good; I am commenting instead of approving to give time for the devs behind this pallet to also review the changes. Thanks for fixing this Matteo

PR #1571, opened 2022-09-05
feat(cosmwasm): add `hash(pretty_json(init_msg))` to contract derivation

aeryz:

## Issue

## Description

## Checklist


frame/cosmwasm/src/lib.rs
```diff
@@ -482,11 +485,18 @@ pub mod pallet {
 			salt: &[u8],
 			admin: Option<AccountIdOf<T>>,
 			label: ContractLabelOf<T>,
+			message: &[u8],
 		) -> Result<(AccountIdOf<T>, ContractInfoOf<T>), Error<T>> {
+			let pretty_message = {
+				let value: serde_json::Value =
+					serde_json::from_slice(message).map_err(|_| Error::<T>::InvalidJsonMessage)?;
+				serde_json::to_vec_pretty(&value).map_err(|_| Error::<T>::InvalidJsonMessage)?
```
hussein-aitlahcen, line 493: If `to_vec` is minified, perhaps it's better to favor it as knowing what pretty mean is difficult, especially if you want to reproduce

hussein-aitlahcen (changes requested):

frame/cosmwasm/src/lib.rs
```diff
@@ -482,11 +485,18 @@ pub mod pallet {
...
+			};
 			let code_hash = CodeIdToInfo::<T>::get(code_id)
 				.ok_or(Error::<T>::CodeNotFound)?
 				.pristine_code_hash;
```
hussein-aitlahcen, line 497: Can we first check if the code is present before hashing the input message please?

frame/cosmwasm/src/lib.rs
aeryz:
```
{ "asd": [1   , 2] }
{ "asd": [1, 2] }
```
I tried using `to_vec` in both of these jsons and it produce the same result. So I'm changing this to `to_vec`.

KaiserKarel: What would happen if someone uploads a contract that is not using JSON for message passing? Perhaps from the pallets point of view, we should just hash the bytes, instead of trying to cleverly figure out intend.

aeryz: It won't work if users try to pass a data that is not JSON. Because cosmwasm uses JSON encoding for messages. If they pass, they will get a parsing error. So we are sure that the message is always JSON.

aeryz: Btw I am not talking about the error that we return which is `InvalidJsonMessage` but they would get parsing error internally anyways.

KaiserKarel: Yeah, so I am wondering if it is wise to put this type of validation at the pallet level, or to remain unopinionated about encodings here.

aeryz: If we don't want that, we would need to rely on users to provide the exact same data to get the same hash. Eg. garbage whitespaces in the message will change the hash.

KaiserKarel: Garbage goes in, garbage goes out ;) I'd leave it up to the user to properly instantiate

hussein-aitlahcen: Yeah we can have the user pass this as salt basically

hussein-aitlahcen: Ok let's stay agnostic to the message. @aeryz I think you can theoretically build a contract that does not use JSON

hussein-aitlahcen: A.k.a. the spec state that the host<->contract interactions are json based, not the message passed from the user. In fact, if we create a wasm blob with correct exports, you could pipe any message

aeryz: Let's be agnostic. But I'm not sure how we can use another encoding in contracts. 

Because for the entry points like `instantiate`, `execute`, etc., we use `entry_point` proc macro. [see](https://github.com/CosmWasm/cosmwasm/blob/6082e8a35a193f1365cb3367f77bd87c593a7ae4/contracts/staking/src/contract.rs#L19)

This converts the `instantiate` call for eg. to call the internal `do_instantiate` call of `cosmwasm-std`. [see](https://github.com/CosmWasm/cosmwasm/blob/6082e8a35a193f1365cb3367f77bd87c593a7ae4/packages/derive/src/lib.rs#L73)

And this function internally parses the message as JSON. [see](https://github.com/CosmWasm/cosmwasm/blob/6082e8a35a193f1365cb3367f77bd87c593a7ae4/packages/std/src/exports.rs#L408)

And here is the `from_slice` call. [see](https://github.com/CosmWasm/cosmwasm/blob/6082e8a35a193f1365cb3367f77bd87c593a7ae4/packages/std/src/serde.rs#L12)

code/parachain/frame/cosmwasm/src/lib.rs
```diff
@@ -170,13 +170,13 @@ pub mod pallet {
...
 		StackOverflow,
 		NotEnoughFundsForUpload,
-		ContractNonceOverflow,
 		NonceOverflow,
 		RefcountOverflow,
 		VMDepthOverflow,
 		SignatureVerificationError,
 		IteratorIdOverflow,
 		IteratorNotFound,
+		InvalidJsonMessage,
```
hussein-aitlahcen, line 179: This can be removed I guess?

hussein-aitlahcen (approved):

KaiserKarel (approved):

PR #1574, opened 2022-09-05
Add DEPENDENCIES.md

KaiserKarel:


hussein-aitlahcen (approved):

vimukthi-git (approved):
LGTM

dzmitry-lahoda:
Just for the info, I do not use this guidance as just cannot by design https://github.com/ComposableFi/composable/commit/a718eafa93e9ead2a9b576b9e4ef1e7005b06e2e . Would be happy to chat about each point so if there are really ways to work according it and I just do not know.

PR #1577, opened 2022-09-06
Refactor `pallet-cosmwasm` to move runtime logic to the pallet from wasmi

aeryz:
Runtime logic is moved from `wasmi` to `pallet` for the purpose of having runtime forward as much as it can.


hussein-aitlahcen (approved):

frame/cosmwasm/src/runtimes/wasmi.rs
```diff
@@ -285,6 +282,7 @@ impl<'a, T: Config> VMBase for CosmwasmVM<'a, T> {
 	}
 
 	fn addr_validate(&mut self, input: &str) -> Result<Result<(), Self::Error>, Self::Error> {
+		log::debug!(target: "runtime::contracts", "addr_validate");
```
dzmitry-lahoda, line 285: please make it trace (lower level)

dzmitry-lahoda (approved):

frame/cosmwasm/src/runtimes/wasmi.rs
aeryz: Can we merge this rn like this then if we find it suitable change to trace? Because we have too many debug's like this and if we change this, we need to change them all. If we do that, my other two PR's might have too many conflicts. And btw I find using debug suitable here since it shows us the execution flow of a contract but not function specifics (like a variable's value in a loop). I mean there is no algorithm specific logs. wdyt?

JurbinM (approved):

PR #1582, opened 2022-09-06
test again other chain on rococo before do it on kusama

dzmitry-lahoda:

## Description

* improved xcmp message so in case of errors assets are put on account of parachain
* test on rococo with 2048 chain

image


dzmitry-lahoda:
image

dzmitry-lahoda:
image

PoisonPhang (approved):

PR #1583, opened 2022-09-06
Feat/xcvm evm

kkx:
working in progress xcvm evm implementation


xcvm/evm/src/Gateway.sol
```diff
@@ -0,0 +1,159 @@
...
+        }
+    }
+
+    function runProgram(
+        Origin memory origin,
+        bytes calldata program,
+        address[] memory _assets,
+        uint256[] memory _amounts
+    ) external payable onlyBridge {
+        // a program is a result of spawm function, pull the assets from the bridge to the interpreter
```
aeryz, line 119:
```suggestion
        // a program is a result of spawn function, pull the assets from the bridge to the interpreter
```

xcvm/evm/src/Interpreter.sol
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * @title Interpreter
+ * @notice Custom interpreter
+ */
+contract Interpreter is IInterpreter{
+    enum OPERATION {
+        NONE,
+        TRANSFER,
+        SPAWM,
```
aeryz, line 15:
```suggestion
        SPAWN,
```

xcvm/evm/src/Interpreter.sol
```diff
@@ -0,0 +1,113 @@
...
+                (success, pos, field, _type) = ProtobufLib.decode_key(
+                    pos,
+                    program
+                );
+                (success, pos, val) = ProtobufLib.decode_uint32(
+                    pos,
+                    program
+                );
+                require(success, "decode key failed");
+            } else if (instruction == uint64(OPERATION.SPAWM)) {
```
aeryz, line 99:
```suggestion
            } else if (instruction == uint64(OPERATION.SPAWN)) {
```

aeryz (changes requested):

code/xcvm/SPEC.md
```diff
@@ -200,12 +200,15 @@ Absolute     ::= u128
 Unit         ::= u128 Ratio
 Ratio        ::= u128 u128
 Account      ::= bytes
-Assets       ::= { AssetId : Balance }
+AssetAmount  ::= assetId Ratio | Unit
```
aeryz, line 203:
```suggestion
AssetAmount  ::= AssetId Ratio | Unit 
```

code/xcvm/SPEC.md
```diff
@@ -234,6 +237,8 @@ sequenceDiagram
 
 The call instruction supports bindings values on the executing side of the program by specifying the `Bindings`. This allows us to construct a program that uses data only available on the executing side. For example, the swap call of the following smart contract snippet expects a `to` address to receive the funds after a trade.
 
+```
```
aeryz, line 240:
```suggestion

```

code/xcvm/SPEC.md
```diff
@@ -249,9 +254,9 @@ Call {
...
 
-Besides accessing the `Self` register, `BindingValue` allows for lazy lookups of AssetId conversions, by using `BindingValue::AssetId(GlobalId)`, or lazily converting decimal points depending on the chain using the `Balance` type.
+Besides accessing the `Self` register, `BindingValue` allows for lazy lookups of AssetId conversions, by using `BindingValue::AssetId(GlobalId)`, or lazily converting amount value depending on the chain using the `AssetAmount` type.
 
-Bindings do not support non byte aligned encodings.
+Bindings do not support non byte aligned encodings. To correctly bind values in the execution layer, the binding values should be ordered by their position index(u16) respectively since to process them interpreters will iterate over the bindings in sequence, after each iteration, the original payload will be updated with the new binding value and the next binding value position will be shifted by the length of the binding value.
```
aeryz, line 259:
```suggestion
Bindings do not support non-byte-aligned encodings. To correctly bind values in the execution layer, the binding values should be ordered by their position index(u16) respectively. Because to process them, interpreters will iterate over the bindings in sequence. After each iteration, the original payload will be updated with the new `BindingValue` and the next `BindingValue` position will be shifted by the length of the `BindingValue`. 
```

code/xcvm/evm/src/Gateway.sol
```diff
@@ -0,0 +1,145 @@
...
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+import "./Interpreter.sol";
+import "./interfaces/IGateway.sol";
+
+contract Gateway is Ownable, IGateway {
+    mapping(uint256 => mapping(bytes => address)) public userInterpreter;
+
+    mapping(address => Bridge) public bridgesInfo;
+    // TODO ? do we have only one bridgge per network and security
```
aeryz, line 15:
```suggestion
    // TODO ? do we have only one bridge per network and security
```

code/xcvm/evm/src/Gateway.sol
```diff
@@ -0,0 +1,145 @@
...
+        require(bridgesInfo[bridgeAddress].security != BridgeSecurity(0), "Gateway: bridge already disabled");
+        delete bridges[bridgesInfo[bridgeAddress].networkId][bridgesInfo[bridgeAddress].security];
+        bridgesInfo[bridgeAddress].security = BridgeSecurity(0);
+        bridgesInfo[bridgeAddress].networkId = 0;
+    }
+
+    //// TODO ? is the bridge who's gonna to provide internetwork assets transfer?
+    function _provisionAssets(
+        address payable interpreterAddress,
+        address[] memory erc20AssetList,
```
aeryz, line 81: If I am not wrong, if the data is read-only, using `callback` instead of `memory` uses less gas. (This applies to other places that uses `memory` in read-only case)

code/xcvm/evm/src/Gateway.sol
```diff
@@ -0,0 +1,145 @@
...
+    ) internal {
+        require(
+            erc20AssetList.length == amounts.length,
+            "Gateway: asset list size shuold be equal to amount list size"
+        );
+        if (msg.value > 0) {
+            bool sent = interpreterAddress.send(msg.value);
+            require(sent, "Failed to send Ether");
+        }
+        for (uint256 i = 0; i < erc20AssetList.length; i++) {
```
aeryz, line 92: I guess `++i` is more gas efficient than `i++`. https://www.reddit.com/r/ethdev/comments/tcwspw/i_vs_i_gas_efficiency/

aeryz: And also `uint256 i` is already `0`.

code/xcvm/evm/src/Gateway.sol
```diff
@@ -0,0 +1,145 @@
...
+
+    function unregisterAsset(uint128 assetId) external onlyOwner {
+        delete assets[assetId];
+    }
+
+    function registerBridge(
+        address bridgeAddress,
+        BridgeSecurity security,
+        uint256 networkId
+    ) external onlyOwner {
```
aeryz, line 61: SPEC states that there can be multiple admins of the contract. Router is the default owner and other owners are configurable. I think we should modify `onlyOwner` in this case.

code/xcvm/evm/src/Gateway.sol
```diff
@@ -0,0 +1,145 @@
+// SPDX-License-Identifier: MIT
```
aeryz, line 1: In Cosmwasm contracts, we use `Router` to handle `Gateway`'s and `Router`'s responsibilities. I think we should settle on whether we use `Gateway` or `Router`.

aeryz (changes requested):

code/xcvm/evm/src/interfaces/IInterpreter.sol
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.14;
```
aeryz, line 2: Either we should have `pragma solidity 0.8.14` or `pragma solidiy ^0.8.14` I guess.

code/xcvm/evm/src/interfaces/IGateway.sol
```diff
@@ -0,0 +1,37 @@
...
+pragma experimental ABIEncoderV2;
+
+interface IGateway {
+    struct Bridge {
+        uint256 networkId;
+        BridgeSecurity security;
+    }
+
+    enum BridgeSecurity {
+        Disabiled,
```
aeryz, line 13:
```suggestion
        Disabled,
```

code/xcvm/xcvm-typescript-sdk/src/interpreter.proto
```diff
@@ -0,0 +1,145 @@
...
+
+package interpreter;
+
+message Uint128 {
+  uint64 highBits = 1;
+  uint64 lowBits = 2;
+}
+
+message Program {
+  Instructions instructions = 1;
```
aeryz, line 11: We also need tag here

code/xcvm/xcvm-typescript-sdk/src/interpreter.proto
```diff
@@ -0,0 +1,145 @@
...
+  Network network = 1;
+  BridgeSecurity security = 2;
+  Salt salt = 3;
+  Program program = 4;
+  repeated Asset assets = 5;
+}
+
+message Query {
+  Network network = 1;
+  Salt salt = 3;
```
aeryz, line 137:
```suggestion
  Salt salt = 2;
```

aeryz (changes requested):

code/xcvm/evm/src/Interpreter.sol
```diff
@@ -0,0 +1,573 @@
...
+    }
+
+    function _getMessageLength(bytes calldata program, uint64 pos) private returns (uint64 size, uint64 newPos) {
+        // reading instruction message
+        bool success;
+        (success, newPos, size) = ProtobufLib.decode_embedded_message(pos, program);
+        require(success, "decode embedded message failed");
+    }
+
+    function _handelAccount(bytes calldata program, uint64 pos) internal returns (address account, uint64 newPos) {
```
aeryz, line 86:
```suggestion
    function _handleAccount(bytes calldata program, uint64 pos) internal returns (address account, uint64 newPos) {
```

code/xcvm/evm/src/Gateway.sol
aeryz: gotcha thanks

code/xcvm/xcvm-typescript-sdk/src/interpreter.proto
aeryz: tag ::= bytes

code/xcvm/xcvm-typescript-sdk/src/interpreter.proto
```diff
@@ -0,0 +1,145 @@
...
+message Transfer {
+  oneof account_type{
+    Account account = 1;
+    Relayer relayer = 2;
+  }
+  repeated Asset assets = 3;
+}
+
+message Salt {
+  uint64 salt = 1;
```
aeryz, line 113:
```suggestion
  bytes salt = 1;
```

aeryz (changes requested):
We can also change `AssetId` to `u64` instead of `u128` in the `SPEC.md`.

code/xcvm/evm/src/Gateway.sol
aeryz: It's a good point, I think we should have a set of owners defined by the router but we should ask @KaiserKarel

dzmitry-lahoda:
> AssetId to u64 instead of u128 in the SPEC.md.

It could be awesome if assetId to stay u128 and limitations of protobuff not to limit spec. That will allow to maintain some assets ids on Picasso to be same as on Xcvm for sure. And make XCVM compatible with GeneralIndex(u128) in XCM.

code/xcvm/evm/src/Gateway.sol
aeryz: Currently, what I do is router sets the owners while instantiating an interpreter. But I'm not sure if there is something like `instantiate` in solidity but maybe something like `constructor`. And then there is no way to change the owners. Maybe we need that though idk.

code/xcvm/evm/src/Gateway.sol
KaiserKarel: We might want to merge the router and gateway contract as they are global singletons anyway?

code/xcvm/evm/src/Gateway.sol
KaiserKarel: yeah, this is something that the SPEC currently does not cover at all. We need to prioritize this asap.

aeryz:
@kkx as we discussed with @hussein-aitlahcen and @KaiserKarel, we will have a dummy ownership which preset the owners to have router and then there will be `add_owners` `remove_owners` functions

aeryz:
> > @kkx as we discussed with @hussein-aitlahcen and @KaiserKarel, we will have a dummy ownership which preset the owners to have router and then there will be `add_owners` `remove_owners` functions
> 
> added functions to be called only by owners, but I think this part is not yet well defined. since we have initialized the router as the first owner, at least we need to have certain functions from the router to call these functions, but even if we have those functions what would be the required permission to do that? Should be called from the original account through the relayer? if so should we add new instructions to xcvm?

I think adding a new owner should be done through relayer at first. We don't need an extra permission mechanism for that. Because we will call the same interpreter only if the origin caller is the same. Because remember, we instantiate and call interpreters based on the origin network-userid. So the flow will be like:
1. Instantiate a new interpreter.
2. Add owners through the relayer.
3. Now owners can do anything without the router.

If we want more complex and programmable ownership mechanism, we might wanna embed ownership methods to the XCVM programs by creating custom instructions. But it is for the next time. For this iteration, we just wanna have a pretty basic, yet usable ownership mechanism. And then we will build upon that based on our thoughts and feedbacks.

code/xcvm/evm/src/Gateway.sol
aeryz: Explaining the flow here based on the conversations here and slack to be able to resolve this:

KaiserKarel:
We can achieve adding additional owners with the Call instruction, late bound using the Self register to obtain the address of the newly instantiated interpreter. No need for the relayer to participate.

code/xcvm/evm/src/Interpreter.sol
```diff
@@ -0,0 +1,595 @@
...
+        (
+            uint64 pos,
+            uint256 maxPos,
+            uint256 networkId,
+            IRouter.BridgeSecurity security,
+            bytes memory salt,
+            bytes memory spawnedProgram
+        ) = _handleSpawnParams(program, pos);
+        address bridgeAddress = IRouter(routerAddress).getBridge(networkId, security);
+        // TODO The fund should be pulled by the Bridge or sent from here??? which is more secure??
```
hussein-aitlahcen, line 484: I'd say we transfer to the bridge after having applied the xcvm transfer fees. Let's open a ticket if it's not already done :). In practice, this will be an IBC bridge, hence w'ell be able to call a function to dispatch a packet

code/xcvm/evm/src/Router.sol
```diff
@@ -0,0 +1,149 @@
...
+        require(
+            erc20AssetList.length == amounts.length,
+            "Router: asset list size should be equal to amount list size"
+        );
+        if (msg.value > 0) {
+            bool sent = interpreterAddress.send(msg.value);
+            require(sent, "Failed to send Ether");
+        }
+        for (uint256 i = 0; i < erc20AssetList.length; i++) {
+            IERC20(erc20AssetList[i]).transferFrom(msg.sender, interpreterAddress, amounts[i]);
```
hussein-aitlahcen, line 97: Unless I am missing something, this code assume that the ETH will be sent to this contract while the ERC tokens kept on the sender account, probably nice to explain that (its some sort of gas optimisation right, unless allowing this contract to transfer on behalf of the bridge is not more efficient than a normal transfer ,in which case its probably better to do a 2 step transfer here: relay -> bridge -> interpreter)

hussein-aitlahcen (approved):

.config/dictionaries/technical.dic
```diff
@@ -266,6 +282,8 @@ rebond
 reduxjs
 registryaddr
 rekt
+Remappings
+remappings
```
Martin-Kiss, line 286:
```suggestion
remappings
```
Nit redundant

Martin-Kiss (approved):

hussein-aitlahcen:
@mergifyio requeue

PR #1604, opened 2022-09-08
[CU-3dfhr3n] Integration tests/subsquid

Dom-Roth:
## Issue
https://app.clickup.com/t/3dfhr3n

## Description
- Adds new dependency to integration tests: @apollo/client & @apollo/server

## Checklist

~~- [ ] I have updated the cargo docs to reflect changes made by this PR _(if applicable)_~~
~~- [ ] I have updated the `book/` to reflect changes made by this PR _(if applicable)_~~


ComposableFiB0t:
Task linked: [CU-3dfhr3n Subsquid Integration tests](https://app.clickup.com/t/3dfhr3n)

code/integration-tests/runtime-tests/src/utils/subsquid/apollo/queries/bondedFinance.ts
```diff
@@ -0,0 +1,15 @@
+import { gql as gqlCore, gql as gqlReact } from "@apollo/client/core";
+
+const isFrontend = false;
+const gql = isFrontend ? gqlCore : gqlReact;
```
andresmechali, line 4: I like this approach but we need to agree with FE on how to centralize the gql queries. Also, we can probably do this only once by exporting `gql` from some dedicated file that only imports both versions, and exports the right one, and then we import from this file every time and avoid repeating this logic.

code/integration-tests/runtime-tests/package.json
```diff
@@ -29,43 +29,47 @@
...
+    "tsconfig-paths": "^4.1.0",
+    "typescript": "^4.8.2"
   },
   "engines": {
     "node": ">=12.0.0"
   },
   "dependencies": {
-    "@polkadot/api": "^8.3.2",
+    "@apollo/client": "^3.6.9",
+    "@apollo/server": "^4.0.0-alpha.10",
```
andresmechali, line 49: I think apollo server is no longer needed

code/integration-tests/runtime-tests/src/utils/subsquid/apollo/queries/vestingSchedule.ts
```diff
@@ -0,0 +1,13 @@
+import { gql } from "@apollo/client";
+
+export const GET_VESTING_SCHEDULE_BY_ADDRESS = gql`
```
andresmechali, line 3: is this being used?

andresmechali (commented):
Good! I think we should split this PR in 2, as most of the changes are not related directly to subsquid but are fixes to the broken integration tests.

PR #1605, opened 2022-09-08
fix: incorrect binary name for composable-node

cor:


aeryz (approved):

andor0 (approved):

KaiserKarel (approved):

hussein-aitlahcen (changes requested):
Wait

hussein-aitlahcen:
Let's check all occurence of the binary usage please

hussein-aitlahcen (approved):

dzmitry-lahoda (approved):

dzmitry-lahoda:
acala and basilisk and polkadot and statemine do not include node suffix into name

cor:
> acala and basilisk and polkadot and statemine do not include node suffix into name

@dzmitry-lahoda yes this is true. The reason why I settled on composable-node for now is because it is a bit clearer within the context of our nix flake, since we have a lot of packages and I don't want there to be ambiguity. We can consider renaming it in a later commit if we really want to remove the suffix

cor:
@mergify requeue

cor:
@mergify update

PR #1614, opened 2022-09-08
cor/fmt in nix

cor:
Already opening this PR in order to test


flake.nix
```diff
@@ -460,6 +460,35 @@
...
+                nixfmt
+                coreutils
+                rust-nightly
+                taplo
+                nodePackages.prettier
+              ];
+
+              text = ''
+                  # .nix 
+                	find . -name "*.nix" -type f -exec nixfmt {} \;
```
dzmitry-lahoda, line 476: nix checked else where afaik. and used xarga i stead of exec. as xargs more composable and no need detals of adhoc exec

dzmitry-lahoda (approved):

flake.nix
cor: Oh yeah good point, will convert to xargs

KaiserKarel (approved):

flake.nix
cor: fixed

cor:
@mergify requeue

PR #1631, opened 2022-09-12
fix instance id in SplitPosition event

andresmechali:
Fix typo in SplitPosition event

### DIFF
--- a/code/parachain/frame/staking-rewards/src/lib.rs
+++ b/code/parachain/frame/staking-rewards/src/lib.rs
@@ -1034,7 +1034,7 @@ pub mod pallet {
 			Self::deposit_event(Event::<T>::SplitPosition {
 				positions: vec![
 					(*fnft_collection_id, *fnft_instance_id, old_position_stake),
-					(*fnft_collection_id, *fnft_instance_id, new_stake.stake),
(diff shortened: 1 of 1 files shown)


code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1034,7 +1034,7 @@ pub mod pallet {
 			Self::deposit_event(Event::<T>::SplitPosition {
 				positions: vec![
 					(*fnft_collection_id, *fnft_instance_id, old_position_stake),
-					(*fnft_collection_id, *fnft_instance_id, new_stake.stake),
+					(*fnft_collection_id, new_fnft_instance_id, new_stake.stake),
```
vimukthi-git, line 1037: Are there no tests failing for this? We need to make sure that there is a test assertion for this

andresmechali: That test is currently ignored, apparently while something is being solved.

vimukthi-git: I see, I think @PoisonPhang is planning to fix that soon. Let's merge this then.

vimukthi-git (approved):

benluelo (approved):

PR #1661, opened 2022-09-13
Fix Firesquid Processor

saadjhk:
## Issue
* With the updated schema a couple of event listeners break on main

## Description
* Adding Fixes for most of identified bugs in this PR


andresmechali (approved):
🚀

andresmechali (approved):

PR #1662, opened 2022-09-13
feat: docker-wipe-system

cor:
I've had multiple people come to me reporting unexpected behavior with running our devnet because of docker container/image/volume caches. This PR adds a nix package that wipes your entire docker system in case any of this occurs.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -490,6 +490,15 @@
               '';
             };
 
+            docker-wipe-system =
+              pkgs.writeShellScriptBin "docker-wipe-system" ''
+                echo "Wiping all docker containers, images, and volumes";
+                docker stop $(docker ps -q)
+                docker system prune -f
+                docker rmi -f $(docker images -a -q)    
+                docker volume prune -f
+              '';
+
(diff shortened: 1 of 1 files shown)


dzmitry-lahoda (approved):
system prune superset of volume and other? also may join script with && so that if container stops then script stops or set exist on error and other strict stuff. add more to nix.

dzmitry-lahoda:
nounset errexit in shell or join it &&

hussein-aitlahcen (approved):

dzmitry-lahoda:
i also do, nix store  clean and cargo clean global of old git clones and target clean local. and tmp clean up.

cor:
> system prune superset of volume and other? also may join script with && so that if container stops then script stops or set exist on error and other strict stuff. add more to nix. 

@dzmitry-lahoda I specifically did not do this because the first command might fail and the rest should continue. For example the first one errors with code 1 when no containers are running, but it should still try to prune all images and volumes

cor:
@mergify requeue

PR #1667, opened 2022-09-14
[CU-2vj3vge] Avoid the possibility of collision in creating asset_accounts

PoisonPhang:
## Issue
[[CU-2vj3vge]](https://app.clickup.com/t/2vj3vge)

## Description

- [X] Added CurrencyFactory to pallet fNFT
- [X] Moved CurrencyFactory trait from generic types to associated types
- [X] Avoided collisions in `fn asset_account` by using protocol assetId instead of unique assetId
- [ ] Improved fNFT storage with protocol assetId


ComposableFiB0t:
Task linked: [CU-2vj3vge [fNFT] Avoid the possibility of collision in creating asset_accounts](https://app.clickup.com/t/2vj3vge)

code/parachain/frame/assets/src/mocks.rs
```diff
@@ -66,13 +66,23 @@ parameter_types! {
 
 pub struct CurrencyIdGenerator;
 
-impl CurrencyFactory<AssetId, Balance> for CurrencyIdGenerator {
+impl CurrencyFactory for CurrencyIdGenerator {
```
PoisonPhang, line 69: Use associated types in method signatures

code/parachain/frame/currency-factory/src/lib.rs
```diff
@@ -188,6 +191,11 @@ pub mod pallet {
 					.into())
 			}
 		}
+
+		fn unique_asset_id_to_protocol_asset_id(unique_asset_id: Self::AssetId) -> u32 {
```
PoisonPhang, line 195: Add test for this

code/parachain/frame/assets/src/mocks.rs
PoisonPhang: 8dd94b35a2f2060afd222d45bb6a0c1dde2d9521

code/parachain/frame/currency-factory/src/lib.rs
PoisonPhang: 6f2f879bd9804031e175102f405a461e336bc5ea

code/parachain/frame/fnft/src/test/mock.rs
```diff
@@ -8,14 +8,21 @@ use frame_support::{
...
 use sp_runtime::{
 	testing::Header,
 	traits::{BlakeTwo256, IdentityLookup},
 };
 
 type UncheckedExtrinsic = frame_system::mocking::MockUncheckedExtrinsic<MockRuntime>;
+type AccountId = u128;
+type Balance = u128;
 type Block = frame_system::mocking::MockBlock<MockRuntime>;
+type CurrencyId = u128;
```
benluelo, line 25: I would just use `u128` directly in the configs, type aliases can be confusing at use sites because they can get confused with newtypes (while not providing the safety of newtypes)

code/parachain/frame/fnft/src/lib.rs
```diff
@@ -498,14 +530,24 @@ pub mod pallet {
...
-		///   with collection id type u128 and instance id type u128 this definitely collides
-		///   because the seed is longer than the account ID
 		fn asset_account(
 			collection: &Self::CollectionId,
 			instance: &Self::ItemId,
-		) -> AccountIdOf<T> {
-			T::PalletId::get().into_sub_account_truncating((collection, instance))
+		) -> Result<AccountIdOf<T>, DispatchError> {
+			// If seed is already stored
+			let seed = match FinancialNftAssetAccountSeeds::<T>::try_get(collection, instance) {
```
dzmitry-lahoda, line 538: not sure. 

if collection. could you just `seed = hash(collection_od, instance_id)`?

code/parachain/frame/composable-traits/src/currency.rs
```diff
@@ -15,18 +15,23 @@ pub type Exponent = u32;
 /// Creates a new asset, compatible with [`MultiCurrency`](https://docs.rs/orml-traits/0.4.0/orml_traits/currency/trait.MultiCurrency.html).
 /// The implementor should ensure that a new `CurrencyId` is created and collisions are avoided.
 /// Is about Local assets representations. These may differ remotely.
-pub trait CurrencyFactory<AssetId, Balance> {
+pub trait CurrencyFactory {
+	type AssetId;
```
dzmitry-lahoda, line 19: I see. Impementation side needs more typing, but tests look cleaner.

code/parachain/frame/fnft/src/lib.rs
PoisonPhang: If the output of the hash is bigger than `u32`, I'd be concerned by the probability of a partial collision of the full hash being a full collision for the account.
However, it's unlikely. I'll examine a few hashing algorithms that we already have supplied and see if any would work well.
This would be a simpler solution.

code/parachain/frame/fnft/src/test/mock.rs
PoisonPhang: I understand where your coming from, but this also makes updating the types easier. There are certainly tradeoffs either way.

code/parachain/frame/fnft/src/lib.rs
PoisonPhang: Gonna try and improve this hash function for variable output and move it into a more general support file for use elsewhere. Seems useful.
ebc768226d872b0921902a95b416f3c7a6132ee1

dzmitry-lahoda (approved):

dzmitry-lahoda:
storage map already uses hash. so really key of double storage map and trie prefox of.pallet likely already computed and dorm something i tell.  shawn tabrizi has nice write up about hashes. so if consider the impl of keys in storage it like already hash. so you may. ot need see

code/parachain/frame/fnft/src/lib.rs
```diff
@@ -57,6 +57,7 @@ pub mod pallet {
 		},
 		PalletId,
 	};
+	use multihash::{Blake2b256, Hasher};
```
vimukthi-git, line 60: Is this not supported by a core library already? Let's not introduce new dependencies if they already exists in the core, specially cryptography stuff

PoisonPhang: 2b4b66d3fe39fff98c6cb34c748a5710f7ac616f

code/parachain/frame/fnft/src/lib.rs
PoisonPhang: 2b4b66d3fe39fff98c6cb34c748a5710f7ac616f

code/parachain/frame/fnft/Cargo.toml
```diff
@@ -31,6 +31,8 @@ sp-std = { default-features = false, git = "https://github.com/paritytech/substr
 composable-support = { path = "../composable-support", default-features = false }
 composable-traits = { path = "../composable-traits", default-features = false }
 
+multihash = { version = "0.16.3", default-features = false, features = ["multihash-impl", "blake2b"] }
```
PoisonPhang, line 34: Remove

PoisonPhang: 1cf22d6a5f4072739da9de4543aabf939180ae43

code/parachain/frame/fnft/src/lib.rs
```diff
@@ -471,14 +473,15 @@ pub mod pallet {
...
 		fn asset_account(
 			collection: &Self::CollectionId,
 			instance: &Self::ItemId,
 		) -> AccountIdOf<T> {
-			T::PalletId::get().into_sub_account_truncating((collection, instance))
+			// `into_sub_account_truncating()` gives us 20 bytes of space to create a seed.
+			// `blake2_128()` returns 16 bytes of data. Opperation will not create collisions.
+			T::PalletId::get().into_sub_account_truncating(sp_io::hashing::blake2_128(
+				&(collection, instance).encode(),
+			))
```
PoisonPhang, line 484: The problem I ran into early with using the (u32, u64) was that our AccountId type in the test was u128 and not the standard AccountId32 we use in runtime. I'm changing this, so we don't have to call an expensive hash function here.

PoisonPhang: 9cfa1ab7effdedfe8c3a9da11155cf68c37779d2

code/parachain/frame/fnft/src/test/mod.rs
```diff
@@ -1,12 +1,14 @@
...
 
 /// Various helpers used throughout this test suite.
 pub(crate) mod prelude;
 
-const ALICE: u128 = 0;
-const BOB: u128 = 1;
-const CHARLIE: u128 = 2;
+const ALICE: AccountId32 = AccountId32::new([0_u8; 32]);
+const BOB: AccountId32 = AccountId32::new([1_u8; 32]);
+const CHARLIE: AccountId32 = AccountId32::new([2_u8; 32]);
```
benluelo, line 11: `[1_u8; 32]` creates an array `[1, 1, 1, 1, 1, 1, ..]`, which is probably not what you want

PoisonPhang: What would you suggest?

dzmitry-lahoda:
@PoisonPhang may be just Account32::from(https://github.com/paritytech/substrate/blob/c24431eb6a95197550b30715a4c124be1aea79de/frame/support/src/storage/mod.rs#L67)

code/parachain/frame/fnft/src/lib.rs
```diff
@@ -471,14 +475,14 @@ pub mod pallet {
...
-		/// TODO (vim): Assess the probability of collision in generating accounts
-		///   with collection id type u128 and instance id type u128 this definitely collides
-		///   because the seed is longer than the account ID
 		fn asset_account(
 			collection: &Self::CollectionId,
 			instance: &Self::ItemId,
 		) -> AccountIdOf<T> {
-			T::PalletId::get().into_sub_account_truncating((collection, instance))
+			let protocol_collection_id =
+				T::CurrencyFactory::unique_asset_id_to_protocol_asset_id(*collection);
```
vimukthi-git, line 483: What makes this collision free?

PoisonPhang: The size of `protocol_collection_id` is `u32` (4 bytes), and the size of `instance` is `u64` (8 bytes). 
We have 20 bytes of space to create a unique seed, so the combination of both of these values puts us comfortably under that limit.

vimukthi-git: I see, as explained offline, if we can find a way to enforce these sizes remain within bounds at compile time it would be great.

PoisonPhang: After a conversation with Antonio, I've decided to go back to hashing
91811fae61299f12d3d06253da223a4762fed424

code/parachain/frame/fnft/src/test/mod.rs
PoisonPhang: 58666b7da195c5369514f752e12be256e6ec4b29

benluelo (approved):

vimukthi-git (approved):
LGTM, Good work in doing the research :)

code/parachain/frame/fnft/src/lib.rs
```diff
@@ -471,14 +475,17 @@ pub mod pallet {
...
 		fn asset_account(
 			collection: &Self::CollectionId,
 			instance: &Self::ItemId,
 		) -> AccountIdOf<T> {
-			T::PalletId::get().into_sub_account_truncating((collection, instance))
+			// `into_sub_account_truncating()` gives us 20 bytes of space to create a seed.
+			// `blake2_256()` returns 32 bytes of data, however, BLAKE2 already truncates its
+			// results. Truncating this to 20 bytes puts us at the recomended output length for
+			// BLAKE2.
+			T::PalletId::get().into_sub_account_truncating(sp_io::hashing::blake2_256(
```
dzmitry-lahoda, line 486: not sure why you hash if you already have hash for `(collection, instance)` ? weird

PoisonPhang: Does [hashed_key_for](https://paritytech.github.io/substrate/master/frame_support/pallet_prelude/struct.StorageDoubleMap.html#method.hashed_key_for) not recompute the hash each time anyway? @dzmitry-lahoda

PoisonPhang: Also, that key would be prefixed, which would be even more bytes to be concerned about creating collisions for. We'd *probably* still be safe, but not with the same probability

dzmitry-lahoda: it stores key. not rehashes each time.

dzmitry-lahoda: not sure, 
in both cases hash in put `collection, instance`. given same amount of data given hashes are good enough you will not doo better. 

PalletId ads it name as prefix to account to any hash you create. so in any case hash_out_of_curve(palletid, collection, instance) = account_id.

So in case of existing case, there is no compute and easy migration rules.

PoisonPhang: Yes, this also recomputes the hash and includes extra data that we would have to trim.
See underlying function [here](https://github.com/paritytech/substrate/blob/a3aa0e8f7bc8b10b981eb634f9e257f075e9aa15/frame/support/src/storage/generator/map.rs#L66~L80)

dzmitry-lahoda: holly...

dzmitry-lahoda: so when you when you check if item exists you get hash. and can reuse? so really you still better as only 1 hash i stead of 2. so hash can be captured from earliercomputstion. a d weird that by spec it stores hash and binary data glued. issue is that there is no reverse map.

dzmitry-lahoda: do you check item exists before call to asset_account? which would calculate hash? so really can just cal ulate hash and check it exists and then do.subaccount.

also, it adds storage prefix which may be improves hash collision?

PoisonPhang: > do you check item exists before call to asset_account? which would calculate hash? so really can just cal ulate hash and check it exists and then do.subaccount.
No, I'm not assuming that the fNFT exist before this function is called. I only assume the `collection` and `instance` are potentially valid.
> PalletId ads it name as prefix to account to any hash you create. so in any case hash_out_of_curve(palletid, collection, instance) = account_id.
Not necessarily. 
Currently `hash_out_of_curve(palletIdType, palletId, hash(collection, instance)) => accountId`.
What I think you're recommending would result in `hash_out_of_curve(palletTypeId, palletId, storageKey) => accountId`. Where `storageKey` is `palletId(8-byte) + blake128(collection)(16byte) + collection(16-byte) + blake128(instance)(16-byte) + instance(8-byte)`.
So, we would only capture `palletId` and part of `blake128(collection)` within the 20 bytes we have. I think that would guarantee collisions.

dzmitry-lahoda: [you no need to call into_account. you can just do Account32::from](https://github.com/paritytech/substrate/blob/20b5aac7e09e8b25abaa48c6781e653b36ba3ca0/client/transaction-pool/src/graph/pool.rs#L467)

dzmitry-lahoda: or other form of from. so what Pallet::into_subaccount does?

dzmitry-lahoda: just do not do it

dzmitry-lahoda: so if you have 32 bit and hash into 64 bit. and then hash into 32 back it still will be good if hash is good. so it should not loose quality if it expands and then contracts back.

dzmitry-lahoda: > storageKey is palletId(8-byte) + blake128(collection)(16byte) + collection(16-byte) + blake128(instance)(16-byte) + instance(8-byte)

that is not true

dzmitry-lahoda:
```
	fn storage_map_final_key<KeyArg>(key: KeyArg) -> Vec<u8>
	where
		KeyArg: EncodeLike<K>,
	{
		let storage_prefix = storage_prefix(Self::module_prefix(), Self::storage_prefix());
		let key_hashed = key.borrow().using_encoded(Self::Hasher::hash);

		let mut final_key = Vec::with_capacity(storage_prefix.len() + key_hashed.as_ref().len());

		final_key.extend_from_slice(&storage_prefix);
		final_key.extend_from_slice(key_hashed.as_ref());

		final_key
	}
```

dzmitry-lahoda: Account::from(final_key) where final_key is storage_prefix + key_hashed. where storage_prefix = hash(pallet + storage) and key_hashed = hash(collection + item id).

account::from(hash(hash(pallet, storage), hash(collection, id))) ?

dzmitry-lahoda: @PoisonPhang I bet 10 PICA Account32 is 32 byte and storage_map_final_key returns 32 byte.  So you do not even need to hash and have account id = key => can check existence easy when needed without storing reverse map on top (as will happen with suggested approach)

dzmitry-lahoda: and you as conflict free as substrate itself conflic free as it uses it for KV in rocksdb - I still suggest you to read Tabrizi blog and watch video))))

PoisonPhang: So, you're suggesting we skip the [into_sub_account_truncating](https://paritytech.github.io/substrate/master/sp_runtime/traits/trait.AccountIdConversion.html#tymethod.into_sub_account_truncating) and derive an account from the full storage key? I misunderstood
This feels byzantine AF, but it could work.

> that is not true

I think it is true. I sent the function for a StorageMap before, but we use a StorageDoubleMap which uses [this function](https://github.com/paritytech/substrate/blob/a3aa0e8f7bc8b10b981eb634f9e257f075e9aa15/frame/support/src/storage/generator/double_map.rs#L92~L111) and currently use Blake2 128 Concat for the keys. 

Concat outputs blake2_128(value) + value

PoisonPhang: > and you as conflict free as substrate itself conflic free as it uses it for KV in rocksdb - I still suggest you to read Tabrizi blog and watch video))))

I got this lecture live from Shawn himself. What you're proposing should work, it just feels byzantine. We can try it tho.

PoisonPhang: > I think it is true.

In the context of what I assumed before, using `into_sub_`

dzmitry-lahoda: > [this function](https://github.com/paritytech/substrate/blob/a3aa0e8f7bc8b10b981eb634f9e257f075e9aa15/frame/support/src/storage/generator/double_map.rs#L92~L111),

id does `hash(hash(pallet, storage), hash(collection, id)`

parity says that if you control ids, via incrementing, it is "good" to use weaker hashes like 64 bit. 

so instead of 128 + 128 + 128 you can use do 128(I assume pallet + storage hash 128 which may be less) + 64 + 64. 
if prefix is 64 , can do 64 + 128 + 64.  

`into_sub_` hashes hash. so it does not gets new information into it. not sure why it should be considered more collision resistant if there is no new if collision resistance = length of output + information in input(you do not seed changes which fight againts hash for collision) + hash function quality i guess. not expert :) . i guess all are good at case. hash(hash(x)) may not be needed.

PoisonPhang: Sir, I'll take my 10 PICA @dzmitry-lahoda 
```rust
fn asset_account(
			collection: &Self::CollectionId,
			instance: &Self::ItemId,
		) -> AccountIdOf<T> {
			println!("{:?}", Instance::<T>::hashed_key_for(collection, instance).len());
			// `into_sub_account_truncating()` gives us 20 bytes of space to create a seed.
			// `blake2_256()` returns 32 bytes of data, however, BLAKE2 already truncates its
			// results. Truncating this to 20 bytes puts us at the recomended output length for
			// BLAKE2.
			T::PalletId::get()
				.into_sub_account_truncating(sp_io::hashing::blake2_256(&(1).encode()))
		}
```
```
88
88
thread 'test::nonfungibles::transfer::roundtrip' panicked at 'called `Result::unwrap()` on an `Err` value: Module(ModuleError { index: 4, error: [4, 0, 0, 0], message: Some("Duplicate") })', parachain/frame/fnft/src/test/prelude.rs:98:74
```

PoisonPhang: This may change if we use a weaker hash function as you stated, blake2_128_concat is overkill in this situation.
> parity says that if you control ids, via incrementing, it is "good" to use weaker hashes like 64 bit.

vimukthi-git (approved):
LGTM, The current solution satisfies the requirement. Thanks!

dzmitry-lahoda:
@vimukthi-git I am not sure. It still may conflict and it is bad design from Parity. Really no conflict can be if not have item counter as fFNT id. But only item account from init data (block, creator, random, pallet prefix, storage name, global counter of items, collection id) hashed.

So you will never be able to access it by item id number.  Only via account. 

And another point - fNFT accounts must pay ED to to assets as all other usual accounts. When fNFT deleted. Pallet can be reported about dangling accounts and will be given ED for that. Called crunk. There should be no immediate delete of accounts on burn. Again, becasue of fNFT is used as storage for Staking, not other way.

So not sure if that is not done Staking will pass audit.

dzmitry-lahoda (approved):

vimukthi-git:
> So you will never be able to access it by item id number. Only via account.

We could in fact think of making the `ItemId` associated type an `AccountId` and just using it for asset account.

>  fNFT accounts must pay ED to to assets as all other usual accounts

Yes, this should be taken account in creating staking positions by users.

> What are requirements? ..

As you stated sir :) , that's the only requirement we are trying to address in this PR. We don't want to blow up PRs with refactoring that do not provide a good ROI.

PR #1668, opened 2022-09-15
Ben/refactor reward and reward pools

benluelo:
https://app.clickup.com/t/2yktfk7
https://app.clickup.com/t/2yz3j4r
https://app.clickup.com/t/2vxvem4
https://app.clickup.com/t/2yktg9r

Also some general refactoring that came up as I was doing the above tasks.

NOTE: Tests currently don't compile, I will fix the errors on Friday.


code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -715,75 +711,121 @@ pub mod pallet {
...
+					T::FinancialNft::get_next_nft_id(&rewards_pool.fnft_collection_id)?;
+				let fnft_account = T::FinancialNft::asset_account(
+					&rewards_pool.fnft_collection_id,
+					&fnft_instance_id,
+				);
 
-			rewards_pool.total_shares = rewards_pool.total_shares.safe_add(&boosted_amount)?;
-			rewards_pool.rewards = rewards;
+				// REVIEW(benluelo): Is this fallible? What does it mean if this fails?
+				rewards_pool.total_shares = rewards_pool.total_shares.safe_add(&boosted_amount)?;
```
PoisonPhang, line 772: Only for overflow safety

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -921,108 +990,94 @@ pub mod pallet {
...
-				stake: left_from_one_ratio.mul_floor(old_position.stake),
-				share: left_from_one_ratio.mul_floor(old_position.share),
-				fnft_instance_id: new_fnft_instance_id,
-				..old_position
-			};
+					let rewards_pool = RewardPools::<T>::get(old_position.reward_pool_id)
+						.ok_or(Error::<T>::RewardsPoolNotFound)?;
+
+					// REVIEW(benluelo): Make this one operation? (Would require a change in the
+					// fNFT interface)
```
PoisonPhang, line 1027: Perhaps we should create an ergonomic interface ontop of fNFT for use in other pallets while leaving the base fNFT pallet as is.

code/parachain/frame/staking-rewards/src/test/mod.rs
```diff
@@ -954,6 +958,8 @@ fn balance(asset_id: u128, account: &Public) -> u128 {
 	)
 }
 
+// TODO(benluelo): REMOVE THIS! Storage should NOT be directly modified in tests. Use the
+// appropriate extrinsics and exposed pallet interfaces.
```
PoisonPhang, line 962: +1
(Feel free to mark resolved)

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: the issue is that if this does overflow then i'm not sure what that means for the pool - using safe addition here prevents it from crashing the node, but the pool might not be in a valid state at that point

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: I think that would definitely be the best bet, @vimukthi-git and i briefly talked about this recently. The main issue is i'm not sure what the interface should be (pallet, trait, etc)

code/parachain/frame/composable-traits/src/staking/mod.rs
```diff
@@ -41,6 +42,38 @@ pub struct Reward<Balance> {
...
+			RewardConfigType::Earnings() => RewardType::Earnings(),
+			RewardConfigType::RateBased(rate_based_config) =>
+				RewardType::RateBased(RateBasedReward::from_config(rate_based_config, now_seconds)),
+		}
+	}
+}
+
+#[derive(RuntimeDebug, PartialEq, Eq, Clone, MaxEncodedLen, Encode, Decode, TypeInfo)]
+pub enum RewardConfigType<Balance> {
+	// REVIEW(benluelo): Consider renaming. Perhaps `ExternallyFunded` or something similar?
```
vimukthi-git, line 72: We can also call this `ProtocolDistribution` https://www.investopedia.com/terms/d/distribution.asp

code/parachain/frame/composable-traits/src/staking/mod.rs
```diff
@@ -71,16 +104,20 @@ impl RewardRatePeriod {
 }
 
 /// A reward update states the new reward and reward_rate for a given asset
+// REVIEW(benluelo): Make this an enum with variants per reward type?
```
vimukthi-git, line 107: Sounds like it would increase the complexity too much.

code/parachain/frame/composable-traits/src/staking/mod.rs
```diff
@@ -130,15 +164,15 @@ pub struct RewardPool<
 	pub share_asset_id: AssetId,
 
 	// Asset ID (collection ID) of the financial NFTs issued for staking positions of this pool
-	pub financial_nft_asset_id: AssetId,
+	pub fnft_collection_id: AssetId,
```
vimukthi-git, line 167:
```suggestion
	pub fnft_asset_id: AssetId,
```

code/parachain/frame/pablo/src/lib.rs
```diff
@@ -1331,6 +1337,9 @@ pub mod pallet {
 		}
 
 		#[transactional]
+		// FIXME(benluelo): This function gets the pool from storage, then calls
```
vimukthi-git, line 1340: Does this actually cause any issue?

code/parachain/frame/pablo/src/lib.rs
```diff
@@ -828,11 +832,13 @@ pub mod pallet {
 				T::Assets::transfer(fees.asset_id, who, owner, fees.owner_fee, false)?;
 			}
 			if !fees.protocol_fee.is_zero() {
-				T::ProtocolStaking::transfer_reward(
+				T::ProtocolStaking::transfer_earnings(
 					who,
 					&T::PbloAssetId::get(),
 					fees.asset_id,
 					fees.protocol_fee,
+					// REVIEW(benluelo): Should this be kept alive?
```
vimukthi-git, line 840: We shouldn't generally as the account is known in advance and more we keep alive the more storage the chain uses

code/parachain/frame/pablo/src/lib.rs
benluelo: It causes 2 extra database reads, which will increase the weight of the call significantly

code/parachain/frame/composable-traits/src/staking/mod.rs
benluelo: The current model only allows us to update a reward to rate based, if we define an enum (even with only one variant for now) it would allow us to reuse the same extrinsic for updating a reward to any type

code/parachain/frame/composable-traits/src/staking/mod.rs
benluelo: https://github.com/ComposableFi/composable/pull/1668/commits/cc9ca2596457b95ddbee6e9e541c914b52dbc3a8

code/parachain/frame/composable-traits/src/staking/mod.rs
vimukthi-git: I see, we don't really know the requirements for the future reward types to know if they need updates, so I think we can keep things as is for now :)

code/parachain/frame/composable-traits/src/staking/reward/rate_based.rs
```diff
@@ -0,0 +1,99 @@
+use codec::{Decode, Encode, MaxEncodedLen};
```
vimukthi-git, line 1: Let's not nest things unnecessarily without any of us having a deep understanding about how the protocol is going to evolve(We haven't launched).  As zen of python states, `Flat is better than nested`.

code/parachain/frame/composable-traits/src/staking/mod.rs
```diff
@@ -254,21 +128,14 @@ pub trait ProtocolStaking {
...
-	fn accumulate_reward(
-		pool: &Self::RewardPoolId,
-		reward_currency: Self::AssetId,
-		reward_increment: Self::Balance,
-	) -> DispatchResult;
-
 	/// Transfers rewards `from` to pool.
 	/// If may be bigger than total shares.
-	fn transfer_reward(
+	fn transfer_earnings(
```
vimukthi-git, line 133:
```suggestion
	fn transfer_protocol_distributions(
```
As per new naming

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -715,75 +721,119 @@ pub mod pallet {
 			duration_preset: DurationSeconds,
 			keep_alive: bool,
 		) -> Result<Self::PositionId, DispatchError> {
-			let mut rewards_pool =
-				RewardPools::<T>::try_get(pool_id).map_err(|_| Error::<T>::RewardsPoolNotFound)?;
-
-			let reward_multiplier = Self::reward_multiplier(&rewards_pool, duration_preset)
-				.ok_or(Error::<T>::NoDurationPresetsConfigured)?;
+			RewardPools::<T>::try_mutate(pool_id, |maybe_rewards_pool| {
```
vimukthi-git, line 724: What is the advantage of this restructuring? This actually reduces readability for me.

code/parachain/frame/staking-rewards/src/lib.rs
vimukthi-git: This does not ignore the overflow, just returns an `ArithmeticError`

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -793,56 +843,83 @@ pub mod pallet {
...
-					T::Assets::can_withdraw(rewards_pool.asset_id, who, amount),
-					WithdrawConsequence::Success
-				),
-				Error::<T>::NotEnoughAssets
-			);
+				RewardPools::<T>::try_mutate(stake.staked_asset_id, |maybe_rewards_pool| {
+					let rewards_pool =
+						maybe_rewards_pool.as_mut().ok_or(Error::<T>::RewardsPoolNotFound)?;
 
-			let boosted_amount = Self::boosted_amount(reward_multiplier, amount);
```
vimukthi-git, line 810: Same: Not sure what the advantage of this nesting is

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -715,75 +721,119 @@ pub mod pallet {
...
+				);
 
-			ensure!(
-				matches!(
-					T::Assets::can_withdraw(rewards_pool.asset_id, who, amount),
-					WithdrawConsequence::Success
-				),
-				Error::<T>::NotEnoughAssets
-			);
+				// REVIEW(benluelo): What does "boosted amount" mean?
```
vimukthi-git, line 737: This is one of the places where it's sort of obvious what it means. Based on the duration locked the staker gets additional shares in the pool

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -852,198 +929,198 @@ pub mod pallet {
...
+				Ok(())
+			})
 		}
 
 		#[transactional]
 		fn split(
 			who: &Self::AccountId,
-			(fnft_collection_id, fnft_instance_id): &Self::PositionId,
+			(fnft_collection_id, old_fnft_instance_id): &Self::PositionId,
+			// REVIEW(benluelo): Rename this argument so it's clearer what the split is doing;
```
vimukthi-git, line 1014: I would not change this, it just means "split by the given ratio". There is no keep or remove here. Both kept and owned act the same other than the size "ratio"

code/parachain/frame/staking-rewards/src/lib.rs
vimukthi-git: We want to keep `mint_into` as a standard interface as used by Substrate so let's keep this as is :)

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -852,198 +929,198 @@ pub mod pallet {
...
-			T::FinancialNft::burn(fnft_collection_id, fnft_instance_id, Some(who))?;
+				// Burn the financial NFT and the shares it holds
+				if is_early_unlock {
+					// If there is no penalty then there is nothing to burn as it will all have been
+					// transferred back to the staker. burn_from isn't a noop if the amount to burn
+					// is 0, hence the check
+					T::Assets::burn_from(
+						stake.staked_asset_id,
+						&fnft_asset_account,
+						stake.amount.defensive_saturating_sub(staked_amount_returned_to_staker),
```
vimukthi-git, line 994: SafeSub?

vimukthi-git (changes requested):
This PR is too large and does too many changes at the same time (hence hard to review), let's make a focused separate PRs on following items
- reward configuration and data structure changes
- staking business logic changes
- refactorings

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: but is this state reachable? should it be bounded somehow? what would it mean if the total shares hits the max?

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: `try_mutate` is the standard way to update something in storage, rather than reading and overwriting.

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: `staked_amount_returned_to_staker` is always <= stake.amount, so this should never fail - I thought I had added a comment explaining that but I must have forgot. I will add an explanation :+1:

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: perhaps this is just me, but it's not super clear which position gets which part of the ratio (without looking into the implementation). I have no problem leaving it as-is, I only left the comment to spark discussion about it and see if we could come up with a name that described the functionality better

code/parachain/frame/composable-traits/src/staking/reward/rate_based.rs
benluelo: I personally prefer to separate models (and their implementations) based on behaviour/ uses - they can always be reexported at the top level (which I should have done, not sure why I didn't :sweat_smile:) so changing the internal structure of the library/ module isn't a breaking change.

I can revert this to what it was before, with `lock` and `math` being the only submodules.

code/parachain/frame/staking-rewards/src/lib.rs
vimukthi-git: Looking at the  storage impl it's basically the same steps for both. So I actually prefer the previous version here as it is less nested and easier to read.

code/parachain/frame/staking-rewards/src/lib.rs
vimukthi-git: This could be reached depending on the config and boosting etc. I guess. So it's ok to assume the bound is the maximum allowed here by the type IMO as we are not going to make a logical error anyway.

code/parachain/frame/staking-rewards/src/lib.rs
vimukthi-git: Which position gets which ratio is not important from business logic as they are both owned by the same user. There is no advantage to existing position being the smaller or the large position out of the two.

> I have no problem leaving it as-is, I only left the comment to spark discussion

I understand, that's good :) , but let's not leave the comments without resolution in the code as that would confuse anyone else looking at the code.

benluelo: > they are both owned by the same user
good point, I guess it doesn't really matter :smile:

I'll remove the comment :+1:

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: read/overwrite also introduces room for bugs (overwriting the wrong key, forgetting to insert at the end, inserting at the wrong time, etc etc), and while these would ideally be caught by tests, I prefer to not rely on tests to verify correctness. 

I also personally prefer the nested `try_mutate`s and I actually find it to be more readable as it describes the functionality better (mutate being one operation vs read/overwrite being two). When I see `Storage::{try_}mutate`, I know what the behaviour is - find the value in storage and mutate it somehow; whereas when I see a `get` or `try_get`, I don't know if it's just a read or if it's intended to be overwritten until I've read the entire function body.

benluelo: Also, I would like to eventually move the actual functionality/ trait impls out of the pallet itself, which would help reduce the indentation level

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: So the slashed shares should be sent to the treasury?

vimukthi-git: NOT "shares" (shares are not slashed) but the staked asset "amounts" that slashed are supposed to go to treasury

code/parachain/frame/staking-rewards/src/lib.rs
vimukthi-git: > read/overwrite also introduces room for bugs (overwriting the wrong key, forgetting to insert at the end, inserting at the wrong time, etc etc) ..

I'm not convinced by this argument tbh. Those are standard things in normal code bases verified by business logic tests. Let's keep things flat as much as possible. If there is a way to use "try_mutate" without entire function body being inside the closure that's also fine though.

code/parachain/frame/composable-traits/src/staking/reward/rate_based.rs
vimukthi-git: yes, that preference is natural but that is a temptation we need to work against :) The problem is that it results in unnatural module boundaries/public interfaces when done without a deep understanding of the domain(not just in rust). Good refactorings are based on the business/technical merits rather than a shallow understanding of the current state (this is including me right now about staking).

vimukthi-git: > I can revert this to what it was before, with lock and math being the only submodules.

`lock` and `math` are not supposed to be separate modules either. Just previous state of the code

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: sounds good, i'll document the limit as such

code/parachain/frame/composable-traits/src/staking/reward/rate_based.rs
benluelo: Fair enough, I'll revert the changes to what they were before and then I can remove the other submodules in a separate PR

code/parachain/frame/pablo/src/lib.rs
benluelo: I'm not sure -`disburse_fees` is called in `exchange`, which in turn called in `buy` which is called with `ensure_signed(origin)?` as `who`, and the `buy` extrinsic (and `Amm::buy`) takes a `keep_alive` parameter which is passed to `exchange` but not `disburse_fees`

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -793,56 +826,89 @@ pub mod pallet {
...
+
+					T::Assets::transfer(
+						stake.staked_asset_id,
+						who,
+						&fnft_asset_account,
+						amount,
+						keep_alive,
+					)?;
+
+					T::Assets::mint_into(rewards_pool.share_asset_id, &fnft_asset_account, amount)?;
```
benluelo, line 901: FIX: lock both of these

PR #1669, opened 2022-09-15
Pablo vault feat/additional functionality for instrumental

belousm:
## Issue
[2z4fw6z](https://app.clickup.com/t/2z4fw6z)
[2z4fww1](https://app.clickup.com/t/2z4fww1)

#### Pablo part
- *Add `get_price_of_lp_token` function to Pablo*
#### Vault part
- *Add `get_strategies ` function to Vault*

### DIFF
--- a/frame/composable-traits/src/dex.rs
+++ b/frame/composable-traits/src/dex.rs
@@ -178,6 +178,9 @@ pub trait Amm {
 		min_receive: Self::Balance,
 		keep_alive: bool,
 	) -> Result<Self::Balance, DispatchError>;
+
+	/// Get current price of LP token.
+	fn get_price_of_lp_token(pool_id: Self::PoolId) -> Result<Self::Balance, DispatchError>;
 }
 
 #[derive(Encode, Decode, MaxEncodedLen, TypeInfo, Clone, PartialEq, Eq, Copy, RuntimeDebug)]
--- a/frame/composable-traits/src/vault.rs
+++ b/frame/composable-traits/src/vault.rs
@@ -3,6 +3,7 @@ use frame_support::{
 	pallet_prelude::*,
 	sp_runtime::Perquintill,
 	sp_std::{collections::btree_map::BTreeMap, fmt::Debug},
+	storage::PrefixIterator,
 };
 use scale_info::TypeInfo;
 
@@ -25,6 +26,18 @@ pub enum FundsAvailability<Balance> {
 	MustLiquidate,
 	/// When there are no balance that can be withdrawn or deposit and don't need to be liquidated.
 	None,
+}
+
+#[derive(Copy, Clone, Encode, Decode, MaxEncodedLen, Default, Debug, PartialEq, TypeInfo)]
+pub struct StrategyOverview<Balance> {
+	// The allocation of this strategy
+	pub allocation: Perquintill,
+	/// The reported balance of the strategy
+	pub balance: Balance,
+	/// Sum of all withdrawn funds.
+	pub lifetime_withdrawn: Balance,
+	/// Sum of all deposited funds.
+	pub lifetime_deposited: Balance,
 }
 
 #[derive(Copy, Clone, Encode, Decode, Debug, PartialEq, Eq, MaxEncodedLen, TypeInfo)]
@@ -119,6 +132,11 @@ pub trait Vault {
 		vault_id: &Self::VaultId,
 		asset_amount: Self::Balance,
 	) -> Result<Self::Balance, DispatchError>;
+
+	/// Return all strategies and their balances related to Vault.
+	fn get_strategies(
+		vault_id: &Self::VaultId,
+	) -> Result<PrefixIterator<(Self::AccountId, StrategyOverview<Self::Balance>)>, DispatchError>;
 }
 
 /// CapabilityVault exposes functionalities for stopping and limiting vault functionality.
--- a/frame/composable-traits/src/vault.rs
+++ b/frame/composable-traits/src/vault.rs
@@ -28,7 +28,7 @@ pub enum FundsAvailability<Balance> {
 	None,
 }
 
-#[derive(Copy, Clone, Encode, Decode, MaxEncodedLen, Default, Debug, PartialEq, TypeInfo)]
+#[derive(Copy, Clone, Encode, Decode, MaxEncodedLen, Default, Debug, PartialEq, TypeInfo, Eq)]
 pub struct StrategyOverview<Balance> {
 	/// The allocation of this strategy.
 	pub allocation: Perquintill,
--- a/frame/composable-traits/src/vault.rs
+++ b/frame/composable-traits/src/vault.rs
@@ -30,9 +30,9 @@ pub enum FundsAvailability<Balance> {
 
 #[derive(Copy, Clone, Encode, Decode, MaxEncodedLen, Default, Debug, PartialEq, TypeInfo)]
 pub struct StrategyOverview<Balance> {
-	// The allocation of this strategy
+	/// The allocation of this strategy.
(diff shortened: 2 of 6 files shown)


frame/composable-traits/src/vault.rs
```diff
@@ -27,6 +28,18 @@ pub enum FundsAvailability<Balance> {
 	None,
 }
 
+#[derive(Copy, Clone, Encode, Decode, MaxEncodedLen, Default, Debug, PartialEq, TypeInfo)]
```
PoisonPhang, line 31: Let's go ahead and add `Eq` to this while making changes.

frame/pablo/src/lib.rs
```diff
@@ -966,6 +966,39 @@ pub mod pallet {
...
+			match pool {
+				PoolConfiguration::StableSwap(StableSwapPoolInfo { pair, lp_token, .. }) => {
+					let pool_base_aum =
+						T::Convert::convert(T::Assets::balance(pair.base, &pool_account));
+					let pool_quote_aum =
+						T::Convert::convert(T::Assets::balance(pair.quote, &pool_account));
+					let lp_total_issuance =
+						T::Convert::convert(T::Assets::total_issuance(lp_token));
+					let pool_general_aum = pool_base_aum.safe_add(&pool_quote_aum)?;
+					Ok(T::Convert::convert(pool_general_aum.safe_div(&lp_total_issuance)?))
```
PoisonPhang, line 981: NIT:
Duplicate code can be moved to a new function

frame/composable-traits/src/vault.rs
belousm: Resolved

frame/pablo/src/lib.rs
belousm: Resolved

frame/composable-traits/src/vault.rs
```diff
@@ -27,6 +28,18 @@ pub enum FundsAvailability<Balance> {
...
+	/// The allocation of this strategy.
+	pub allocation: Perquintill,
+	/// The reported balance of the strategy.
+	pub balance: Balance,
+	/// Sum of all withdrawn funds.
+	pub lifetime_withdrawn: Balance,
+	/// Sum of all deposited funds.
+	pub lifetime_deposited: Balance,
+}
+
```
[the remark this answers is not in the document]
belousm: @PoisonPhang Because I created `get_strategies` function which use [`StrategyOverview`](https://github.com/ComposableFi/composable/blob/9f3b4c6e4c1ad70b58b4fd956c815c79eb2234a7/frame/composable-traits/src/vault.rs#L137-L139) in its annotation. So, I was needed to have this struct in the same file.

frame/composable-traits/src/dex.rs
```diff
@@ -178,6 +178,9 @@ pub trait Amm {
 		min_receive: Self::Balance,
 		keep_alive: bool,
 	) -> Result<Self::Balance, DispatchError>;
+
+	/// Get current price of LP token.
+	fn get_price_of_lp_token(pool_id: Self::PoolId) -> Result<Self::Balance, DispatchError>;
```
saruman9, line 183: Shouldn't we add this function to the `runtime` (`dali`), calculate weights for it?

P.S. The same goes for `update_accounts_deposited_one_asset_storage` function actually.

frame/composable-traits/src/vault.rs
```diff
@@ -27,6 +28,18 @@ pub enum FundsAvailability<Balance> {
 	None,
 }
 
+#[derive(Copy, Clone, Encode, Decode, MaxEncodedLen, Default, Debug, PartialEq, TypeInfo, Eq)]
+pub struct StrategyOverview<Balance> {
```
saruman9, line 32: I do not know how the relationship between these files is, but `StrategyOverview` appears in the following files:
- [`composable_api.rs`](https://github.com/ComposableFi/composable/blob/2805639cd55364dee41217bf68dd28b1de489aad/utils/price-feed/src/feed/composable_api.rs),
- [`composablejs` directory](https://github.com/ComposableFi/composable/tree/pablo-feat/constant_product_zapping/composablejs),
- [`integration-tests` directory](https://github.com/ComposableFi/composable/tree/pablo-feat/constant_product_zapping/integration-tests/runtime-tests),
- [`frontend` directory](https://github.com/ComposableFi/composable/tree/pablo-feat/constant_product_zapping/frontend/packages/defi-interfaces).

frame/composable-traits/src/vault.rs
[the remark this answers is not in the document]
saruman9: Maybe my question will seem stupid, but for what reason we can't use the structure from `models.rs` file?

saruman9 (changes requested):

frame/composable-traits/src/vault.rs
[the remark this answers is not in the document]
saruman9: Maybe my question will seem stupid, but for what reason we can't use the structure from `models.rs` file?

frame/composable-traits/src/dex.rs
belousm: To be honest, I don't know.

frame/composable-traits/src/vault.rs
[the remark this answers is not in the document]
belousm: In the course of my work with other pallets, I noticed that the structures used in the traits are also created from the same file. Therefore, I decided to do so in order to maintain some generality.

frame/composable-traits/src/vault.rs
benluelo: This was moved here from the vault pallet, so it already exists

frame/composable-traits/src/dex.rs
benluelo: Only extrinsics need weights, as far as i can tell this isn't called as an extrinsic

saruman9: Then be free to close (resolve) the thread if this function does not plan to be an extrinsic.

frame/composable-traits/src/vault.rs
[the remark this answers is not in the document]
belousm: @PoisonPhang Also the `vault pallet` itself has a dependency on `composable-traits`, and then it turns out there will be a cycle if I try to import the `models.rs` file into a `composable-traits`.

frame/pablo/src/lib.rs
```diff
@@ -966,6 +966,22 @@ pub mod pallet {
...
+			let pool = Self::get_pool(pool_id)?;
+			let pool_account = Self::account_id(&pool_id);
+			match pool {
+				PoolConfiguration::StableSwap(StableSwapPoolInfo { pair, lp_token, .. }) =>
+					calculate_price_of_lp_token::<T>(pair, lp_token, &pool_account),
+				PoolConfiguration::ConstantProduct(ConstantProductPoolInfo {
+					pair,
+					lp_token,
+					..
+				}) => calculate_price_of_lp_token::<T>(pair, lp_token, &pool_account),
```
saruman9, line 979:
```suggestion
				PoolConfiguration::StableSwap(StableSwapPoolInfo { pair, lp_token, .. }) |
				PoolConfiguration::ConstantProduct(ConstantProductPoolInfo {
					pair,
					lp_token,
					..
				}) => calculate_price_of_lp_token::<T>(pair, lp_token, &pool_account),
```
You can merge arms, please check the suggestion before accepting.

saruman9 (approved):
LGTM, but please find out if something needs to be changed [here](https://github.com/ComposableFi/composable/pull/1669#discussion_r972843241) due to the fact that `StrategyOverview` has changed its location before merging.

PR #1682, opened 2022-09-19
[CU-2z0jkuq] Move mdbook check to nix

andor0:
## Issue
https://app.clickup.com/t/2z0jkuq


flake.nix
```diff
@@ -805,6 +805,24 @@
...
+            mdbook-check = stdenv.mkDerivation {
+              name = "mdbook-check";
+              dontUnpack = true;
+              buildInputs = [ all-directories-and-files mdbook ];
+              installPhase = ''
+                mkdir -p $out/book
+                chmod 777 $out/book
+                mdbook --version
+                cd ${all-directories-and-files}/book
+                TMPDIR=$out/book mdbook build --dest-dir=$out/book 2>&1 | tee $out/log
```
dzmitry-lahoda, line 817: please leave comment why we do not use check

dzmitry-lahoda (approved):

flake.nix
andor0: done.

PR #1685, opened 2022-09-19
removed codespacers and made nix owners of dev tools

dzmitry-lahoda:

## Description

codespace MUST be in sync with nix shells/apps/packages/ports/envs and fulfills same function.
so naturally that is nix


cor (approved):
LGTM

KaiserKarel (approved):

PR #1695, opened 2022-09-20
#ibc #xcvm some diagrams

dzmitry-lahoda:
image

image

image

image

## Description
- IBC handshake
- IBC deployment 
- styling
- Assets/byog with IBC

### DIFF
--- a/.config/dictionaries/terminology.dic
+++ b/.config/dictionaries/terminology.dic
@@ -6,7 +6,7 @@ wococo
 westend
 dotsama
 karura
-bitfrost
+bifrost
 orml
 zombienet
 polkascan
--- a/.config/dictionaries/terminology.dic
+++ b/.config/dictionaries/terminology.dic
@@ -6,6 +6,7 @@ wococo
 westend
 dotsama
 karura
+bitfrost
 orml
 zombienet
 polkascan
@@ -378,3 +379,4 @@ xbtc
 pica
 pblo
 picapblo # PICA/PBLO
+vksm
--- a/docs/byog.md
+++ b/docs/byog.md
@@ -33,7 +33,7 @@ Currently, non-native currencies are mapped to native with some configurable rat
 
 ## How much does my transaction cost in Picasso?
 
-When a transaction is executed, its `weight` is known and roughly equal to the computational resources it consumes. Computational resources are benchmarked and updated with each runtime upgrade.
+When a transaction is executed, its `weight` is known and roughly equals the computational resources it consumes. Computational resources are benchmarked and updated with each runtime upgrade.
 
 Weight is converted into an appropriate amount of PICA by the polynomial formula.
 
--- a/docs/byog.md
+++ b/docs/byog.md
@@ -16,7 +16,7 @@ A runtime configuration allows mapping amount of execution which can be bought f
 
 ## Current state
 
-Any cross-chain XCMP transactions can choose what assets to use to pay for the transaction.
+Any cross-chain XCMP transactions can choose which assets to use to pay for the transaction.
 
 Direct transaction can add header to pick up `payment` asset or accounts can configure default [payment asset](https://github.com/paritytech/substrate/discussions/12055).
 
--- a/docs/byog.md
+++ b/docs/byog.md
@@ -8,9 +8,15 @@ Such mechanisms are built into the runtime on Picasso and [other Dotsama paracha
 
 `On ramps` and [meta (gas less) transactions](https://docs.polygon.technology/docs/category/meta-transactions) handle such possibilities on other chains.
 
+## Context
+
+Users pay fees in the runtime's native currency for the execution, allowing the runtime to earn and prevents spam and DDoS attacks. BYOG extend payments to registered non native currencies.
+
+A runtime configuration allows mapping amount of execution which can be bought for amount of native currency.
+
 ## Current state
 
-Any cross-chain (XCMP) transactions can choose what assets to use to pay for the transaction.
+Any cross-chain XCMP transactions can choose what assets to use to pay for the transaction.
 
 Direct transaction can add header to pick up `payment` asset or accounts can configure default [payment asset](https://github.com/paritytech/substrate/discussions/12055).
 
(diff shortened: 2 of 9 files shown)


docs/byog.md
```diff
@@ -22,13 +28,17 @@ Minting, burns, and airdrops are out of the scope of this feature.
...
 Fees are small now, but later we will have a way to show the exact fee for a transaction paid in any possible asset.
 
-## How much does my transaction cost?
+**Example**
+Currently, non-native currencies are mapped to native with some configurable ratio which in turn is mapped to a weight.
 
-When a transaction is executed, its `weight` is known and roughly equal to the computational resources it consumes.
+## How much does my transaction cost in Picasso?
+
+When a transaction is executed, its `weight` is known and roughly equal to the computational resources it consumes. Computational resources are benchmarked and updated with each runtime upgrade.
```
andor0, line 36:
```suggestion
When a transaction is executed, its `weight` is known and roughly equals the computational resources it consumes. Computational resources are benchmarked and updated with each runtime upgrade.
```

andor0 (approved):

docs/byog.md
```diff
@@ -8,9 +8,15 @@ Such mechanisms are built into the runtime on Picasso and [other Dotsama paracha
 
 `On ramps` and [meta (gas less) transactions](https://docs.polygon.technology/docs/category/meta-transactions) handle such possibilities on other chains.
 
+## Context
+
+Users pay fees in the runtime's native currency for the execution, allowing the runtime to earn and prevents spam and DDoS attacks. BYOG extend payments to registered non native currencies.
+
+A runtime configuration allows mapping amount of execution which can be bought for amount of native currency.
```
Martin-Kiss, line 15: What does "amount of execution" mean here? Is this referring to the number of computational resources needed to execute?

docs/byog.md
```diff
@@ -8,9 +8,15 @@ Such mechanisms are built into the runtime on Picasso and [other Dotsama paracha
...
+
 ## Current state
 
-Any cross-chain (XCMP) transactions can choose what assets to use to pay for the transaction.
+Any cross-chain XCMP transactions can choose what assets to use to pay for the transaction.
```
Martin-Kiss, line 19:
```suggestion
Any cross-chain XCMP transactions can choose which assets to use to pay for the transaction.
```
Nit

docs/currency.md
```diff
@@ -89,26 +91,19 @@ Remote currencies usually have their respective bridge identifiers attached in t
...
-A runtime configuration allows mapping amount of execution which can be bought for amount of native currency.
-
-This parachain uses pre-calculated weights to estimate the hardware costs of execution. An amount of weight a native currency can buy is a non-linear dynamic function from currency amount to weight. As the usage of this chain increases towards maximum capacity, the price of a unit of weight increases as well. 
+See [BYOG](./byog.md) for details.
 
 Cross-chain interactions over bridges may involve metered networks which allow setting limits on execution price and calculating the possible number of execution resources in the runtime. For example, Ethereum.
 
-
-Within the protocols built on top of the parachain, there are "native" protocol currencies that are used for governance and revenue distribution, but _cannot_ be used to pay for execution.
+Within the protocols built on top of the parachain, there are "native" protocol currencies that are used for governance and revenue distribution, these may be use to pay gas fees of configured that way.
```
Martin-Kiss, line 100:
```suggestion
Within the protocols built on top of the parachain, there are "native" protocol currencies that are used for governance and revenue distribution, these may be used to pay gas fees if configured that way.
```
Nit expression

docs/currency.md
```diff
@@ -89,26 +91,19 @@ Remote currencies usually have their respective bridge identifiers attached in t
...
 
 ### Bridged gas fee (`bring your own gas`)
 
-Users who want to have assets on our chain can either be minted with some native currency governance or be allowed to pay for cross-chain transactions in some non-native currency.
-
-Currently, non-native currencies are mapped to native with some configurable ratio which in turn is mapped to a weight.
+See [BYOG](./byog.md) for details.
 
-Transactions issued by users are likely to swap some remote currencies into native on transfer, to enact the execution of local transactions. 
+Users who want to have assets on our chain can either be minted with some native currency governance or be allowed to pay for cross-chain transactions in some non-native currency.
```
Martin-Kiss, line 106:
```suggestion
Users who want to have assets on our chain can either mint with some native currency governance or be allowed to pay for cross-chain transactions in some non-native currency.
```
Nit expression

dzmitry-lahoda: so in this case they do not not mint. but some sudo/protocol/admin mints to them (so user are passibe and minted to them. not sure how make it clear. but they do not mint.

docs/byog.md
dzmitry-lahoda: yes.

dzmitry-lahoda: will update in next PR were clarification on price covertion will be made

docs/currency.md
dzmitry-lahoda: hm, not that true too. he may be not requested, but that happend by some protocol. request can be one of these, or airdrop, etc. so `on chain protocol mints`

cor:
I cannot find the referenced ICS 11 at https://github.com/cosmos/ibc, do you have a reference for me?

docs/currency.md
Martin-Kiss: > hm, not that true too. he may be not requested, but that happend by some protocol. request can be one of these, or airdrop, etc. so `on chain protocol mints`

Reading this, my understanding is: 
"Users can take certain actions that trigger some on-chain protocol to mint assets to the user for native currency."

If users don't mint themself and don't request some entity to mint, I imagine users take some action for a protocol to react by minting to the user, is that correct?

dzmitry-lahoda: > Users can take certain actions that trigger some on-chain protocol to mint assets to the user for native currency

They can participate in offchain crowloans or other marketing, and they may be rewarded with amounts by us.

dzmitry-lahoda: > If users don't mint themself and don't request some entity to mint, I imagine users take some action for a protocol to react by minting to the user, is that correct?
don't request some entity to mint = take some action for a protocol  = bring some proof so some on chain entity that that request is allowed to execute and it will mint. so it is same. you prove to protocol to do some action, and that action can mint you or else account PICA. can be airdrop, LBP, crowdloans, XCVM swaps, sudo asset calls.

Martin-Kiss:
```suggestion
Users who want to have assets on our chain can either be issued native currency by some on-chain protocol or be allowed to pay for cross-chain transactions in some non-native currency.
```
Is this accurate?

dzmitry-lahoda: yes.

blasrodri:
@dzmitry-lahoda great visualizations! We're going with `ics10-grandpa` instead of `ics11-beefy` for now, until we can fully support beefy.
And I'm double checking whether the relayer also listens to the relay chain or not.

blasrodri:
> I cannot find the referenced ICS 11 at https://github.com/cosmos/ibc, do you have a reference for me?

https://github.com/ComposableFi/centauri/tree/master/light-clients/

Wizdave97:
> @dzmitry-lahoda great visualizations! We're going with `ics10-grandpa` instead of `ics11-beefy` for now, until we can fully support beefy. And I'm double checking whether the relayer also listens to the relay chain or not.

The relayer listens for finality events from the relay chain, and gets all other data like IbcEvents, packets from the parachain, but it only does the latter after receiving a finality notification.

Martin-Kiss (approved):

.config/dictionaries/terminology.dic
```diff
@@ -6,6 +6,7 @@ wococo
 westend
 dotsama
 karura
+bitfrost
```
KaiserKarel, line 9: bit or bifrost?

andor0 (approved):

andor0 (approved):

KaiserKarel (approved):

PR #1703, opened 2022-09-21
[CU-3bug4nt] Add source to HistoricalLockedValue

andresmechali:
## Issue
https://app.clickup.com/t/3bug4nt

## Description
This PR adds a field to the HistoricalLockedValue entity so that it can differentiate between the source of the locked values. It also adds a resolver for Picasso stats.


ComposableFiB0t:
Task linked: [CU-3bug4nt Aggregate Total Value Locked for Staking Rewards Pools](https://app.clickup.com/t/3bug4nt)

subsquid/src/processor.ts
```diff
@@ -145,4 +145,11 @@ processor.addEventHandler(
 
 processor.addEventHandler("Oracle.PriceChanged", processOraclePriceChanged);
 
+processor.addEventHandler("*", async (ctx) => {
+  const signer = ctx?.event?.extrinsic?.signature?.address;
```
saadjhk, line 149: @andresmechali `ctx?.event?.extrinsic?.signature?.address` i have added a fallback case in `saveEvent` however we need to identify what causes this be different in different cases

saadjhk (approved):

easteregg (approved):

andresmechali:
@mergify refresh

PR #1711, opened 2022-09-22
improve containers builds and some rename to ci commands

dzmitry-lahoda:

## Description
- we want spell, do not care it uses cspell
- publish containers from main or branches with proper tags
- baked some defaults into devcontainer


.github/workflows/build-test-deploy.yml
```diff
@@ -832,8 +830,18 @@ jobs:
...
           cachix watch-exec --jobs 16 --compression-level $CACHIX_COMPRESSION_LEVEL composable-community nix -- build .#devnet-container --no-update-lock-file --show-trace -L
           docker load --input ./result
-          docker tag composable-devnet-container:latest "${{ env.DOCKER_REGISTRY_NAME }}/composable-sandbox:latest"
-          docker push "${{ env.DOCKER_REGISTRY_NAME }}/composable-sandbox:latest"
+          # update soon https://github.com/actions/checkout/issues/760
+          git config --global --add safe.directory /__w/composable/composable
+          CONTAINER_NAME=composable-sandbox
+          SHA256=$(sha256sum ./result | cut --delimiter " " --fields 1)
+          COMMIT_SHA=$(git rev-parse HEAD)
+          DOCKER_IMAGE_HASH=$(docker images | tail --lines=1 | tr --squeeze-repeats  ' ' |  cut --delimiter " " --fields 3)
```
andor0, line 838: there is `composable-devnet-container:latest` docker image when the docker image is built.
so, you just can use `composable-devnet-container:latest` instead of `DOCKER_IMAGE_HASH`.

.github/workflows/build-test-deploy.yml
```diff
@@ -832,8 +830,18 @@ jobs:
...
+          docker image tag $DOCKER_IMAGE_HASH "${{ env.DOCKER_REGISTRY_NAME }}/${CONTAINER_NAME}:${COMMIT_SHA}"
+          docker image tag $DOCKER_IMAGE_HASH "${{ env.DOCKER_REGISTRY_NAME }}/${CONTAINER_NAME}:${SHA256}"
```
andor0, line 840: where are they used?

dzmitry-lahoda: you mean sha256?

dzmitry-lahoda: in nix

.github/workflows/build-test-deploy.yml
dzmitry-lahoda: i am worry that can change nix container name. then this thing will fail.
while i rely on fact that system is clean, and has only one just built container. that also can fail. 

not sure what is better. may be need to change to name based approach.

andor0: > i am worry that can change nix container name

why we can want to do it? we can just add a new tag  `composable-devnet-container:latest` that we need.

imho, `composable-devnet-container:latest` is read simler that `DOCKER_IMAGE_HASH=$(docker images | tail --lines=1 | tr --squeeze-repeats  ' ' |  cut --delimiter " " --fields 3)`

.github/workflows/build-test-deploy.yml
andor0: ahh, ok.

.github/workflows/build-test-deploy.yml
```diff
@@ -832,8 +830,18 @@ jobs:
...
-          docker push "${{ env.DOCKER_REGISTRY_NAME }}/composable-sandbox:latest"
+          # update soon https://github.com/actions/checkout/issues/760
+          git config --global --add safe.directory /__w/composable/composable
+          CONTAINER_NAME=composable-sandbox
+          SHA256=$(sha256sum ./result | cut --delimiter " " --fields 1)
+          COMMIT_SHA=$(git rev-parse HEAD)
+          LOCAL_DOCKER_IMAGE_NAME=composable-devnet-container:latest
+          docker image tag $LOCAL_DOCKER_IMAGE_NAME "${{ env.DOCKER_REGISTRY_NAME }}/${CONTAINER_NAME}:${COMMIT_SHA}"
+          docker image tag $LOCAL_DOCKER_IMAGE_NAME "${{ env.DOCKER_REGISTRY_NAME }}/${CONTAINER_NAME}:${SHA256}"
+          if [ $(git symbolic-ref HEAD) = "refs/heads/main" ]; then
```
dzmitry-lahoda, line 841: @easteregg

andor0 (approved):

PoisonPhang (approved):

andor0 (approved):

PoisonPhang (approved):

flake.nix
```diff
@@ -890,10 +892,9 @@
               cargoExtraArgs = "--benches --all --features runtime-benchmarks";
             });
 
-            cspell-check = stdenv.mkDerivation {
+            spell-check = stdenv.mkDerivation {
               name = "cspell-check";
```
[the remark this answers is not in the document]
dzmitry-lahoda: will resolve it as naming is consistent with what what desired and really happens. interally we build drv of cspell, and use it to `spell check`.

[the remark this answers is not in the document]
KaiserKarel: agree, the action is different from the implementation

dzmitry-lahoda:
@Mergufyio refresh

PR #1709, opened 2022-09-21
[cu-2vj3vmv] FinancialNftProtocol Impl + Extend TODOs + Fix Split + minor cleanups

vimukthi-git:
Rudimentary FinancialNftProtocol Impl

PR #1725, opened 2022-09-22
Refactor split implementation and update test helpers

benluelo:
Part 1/n of splitting up #1668

In this PR:

- Updates & fixes to `split` ~~(waiting on #1709 to merge first)~~
- Updates/ upgrades to test helpers

### DIFF
--- a/code/parachain/frame/staking-rewards/src/benchmarking.rs
+++ b/code/parachain/frame/staking-rewards/src/benchmarking.rs
@@ -1,12 +1,8 @@
 //! Benchmarks
-use core::ops::Mul;
-
 use crate::*;
 
 use composable_support::validation::TryIntoValidated;
-use composable_tests_helpers::test::{
-	block::process_and_progress_blocks, helper::assert_extrinsic_event_with,
-};
+use composable_tests_helpers::test::helper::assert_extrinsic_event_with;
 use composable_traits::{
 	staking::{
 		lock::LockConfig, RewardConfig, RewardPoolConfiguration::RewardRateBasedIncentive,
@@ -34,7 +30,7 @@ fn get_reward_pool<T: Config>(
 	owner: T::AccountId,
 	reward_count: u32,
 ) -> RewardPoolConfigurationOf<T> {
-	let pool_init_config = RewardRateBasedIncentive {
+	RewardRateBasedIncentive {
 		owner,
 		asset_id: BASE_ASSET_ID.into(),
 		start_block: 2_u128.saturated_into(),
@@ -43,8 +39,7 @@ fn get_reward_pool<T: Config>(
 		lock: lock_config::<T>(),
 		share_asset_id: X_ASSET_ID.into(),
 		financial_nft_asset_id: STAKING_FNFT_COLLECTION_ID.into(),
-	};
-	pool_init_config
+	}
 }
 
 fn lock_config<T: Config>() -> LockConfig<T::MaxStakingDurationPresets> {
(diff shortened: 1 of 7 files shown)


code/parachain/frame/staking-rewards/src/benchmarking.rs
```diff
@@ -159,21 +172,35 @@ benchmarks! {
...
-			100_000_000.into(),
-			ONE_HOUR,
-		).unwrap();
-
-		let ratio =  Permill::from_rational(1_u32,7_u32).try_into_validated().unwrap();
+		let instance_id = assert_extrinsic_event_with::<T, <T as frame_system::Config>::Event, crate::Event::<T>, _, _, _>(
+			Pallet::<T>::stake(
+				RawOrigin::Signed(user.clone()).into(),
+				BASE_ASSET_ID.into(),
+				// PICA::units(1_000).into(),
```
PoisonPhang, line 182: Remove?

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1060,21 +1053,39 @@ pub mod pallet {
...
 
+			dbg!();
+			// lock assets on new account
+			T::Assets::set_lock(
+				T::LockId::get(),
+				asset_id,
+				new_fnft_asset_account,
+				new_account_amount,
+			)?;
+			dbg!();
```
PoisonPhang, line 1088: Random debugs left dangling

PoisonPhang (approved):
LGTM
Minor NIT comments about debugging artifacts

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -918,111 +922,100 @@ pub mod pallet {
...
+				|maybe_old_position| {
+					let existing_position =
+						maybe_old_position.as_mut().ok_or(Error::<T>::StakeNotFound)?;
+
+					let left_from_one_ratio = ratio.left_from_one();
+
+					// create the new position first, before mutating the old position
+					// mul_ceil is used for the new position, and mul_floor for the old position,
+					// that way any rounding is accounted for.
+					let new_stake = left_from_one_ratio.mul_ceil(existing_position.stake);
```
vimukthi-git, line 944: We need a failing test that captures the problem fixed by doing floor/ceil

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -918,111 +922,100 @@ pub mod pallet {
...
-			let mut old_position_stake = BalanceOf::<T>::zero();
-			let mut old_position =
-				Stakes::<T>::try_mutate(fnft_collection_id, fnft_instance_id, |maybe_stake| {
-					let stake = maybe_stake.as_mut().ok_or(Error::<T>::StakeNotFound)?;
-
-					let old_value = stake.clone();
-					stake.stake = ratio.mul_floor(stake.stake);
-					stake.share = ratio.mul_floor(stake.share);
+		) -> Result<Self::PositionId, DispatchError> {
+			let (new_fnft_instance_id, new_position) = Stakes::<T>::try_mutate(
```
vimukthi-git, line 932: I don't see a value in changing the structure here, the previous code was better

vimukthi-git (changes requested):
Let's not mix very low priority refactorings that may not be valid with actual functionality fixes. It makes it hard to approve.

code/parachain/frame/staking-rewards/src/lib.rs
PoisonPhang: I'd argue this implementation looks cleaner. However, I'm not sure which one is more correct in terms of what this function returns. Who's the end user of these return values, and what do they want from the values returned?

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1085,20 +1079,35 @@ pub mod pallet {
...
 		pub(crate) fn split_lock(
 			asset_id: T::AssetId,
-			existing_account_amount: T::Balance,
+			old_fnft_asset_account: &T::AccountId,
+			new_fnft_asset_account: &T::AccountId,
+			old_account_amount: T::Balance,
 			new_account_amount: T::Balance,
-			existing_asset_account: &T::AccountId,
-			new_asset_account: &T::AccountId,
 		) -> DispatchResult {
```
PoisonPhang, line 1086: I like the new `split_lock` method, which cleans up the calling method quite a bit. However, I think we should use the terms "existing" and "new" as using "old" and "new" infers that the old is being wholly replaced, which is not happening.

benluelo: agreed - updated in https://github.com/ComposableFi/composable/commit/c9c49555f0eec267ead832ab544f9acd662a0534

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: https://github.com/ComposableFi/composable/commit/913dcfc8974ec2e1acb355eba5634e85d3b3dc70

code/parachain/frame/staking-rewards/src/benchmarking.rs
benluelo: https://github.com/ComposableFi/composable/commit/d61f1bf9347cddc1bc0cdc7adf810f7e0b55bee8

benluelo:
closed in favor off #1751

PR #1734, opened 2022-09-23
feat(nix): initial darwin compatibility

hussein-aitlahcen:


dzmitry-lahoda:
@hussein-aitlahcen can we close it as per discussion?

dzmitry-lahoda:
Closing as confirmed in slack we would not have darwin for now

PR #1741, opened 2022-09-23
[CU-2yyykk2] Added ensure for non-zero assetIds

PoisonPhang:
## Issue
[[CU-2yyykk2]](https://app.clickup.com/t/2yyykk2)

## Description
Decided to place the check inside of `create_staking_pool` to avoid unwrapping the pool config at the `pallet::call` level.

### DIFF
--- a/code/parachain/frame/staking-rewards/src/lib.rs
+++ b/code/parachain/frame/staking-rewards/src/lib.rs
@@ -203,6 +203,8 @@ pub mod pallet {
 	pub enum Error<T> {
 		/// Error when creating reward configs.
 		RewardConfigProblem,
+		/// AssetId is invalid, asset IDs must be graeter than 0
+		InvalidAssetId,
 		/// Reward pool already exists
 		RewardsPoolAlreadyExists,
 		/// No duration presets configured.
@@ -267,7 +269,8 @@ pub mod pallet {
 			+ Ord
 			+ From<u128>
 			+ Into<u128>
-			+ Copy;
+			+ Copy
+			+ Zero;
 
 		// REVIEW(benluelo): Mutate::CollectionId type?
 		type FinancialNft: NonFungiblesMutate<AccountIdOf<Self>>
@@ -637,6 +640,11 @@ pub mod pallet {
 					share_asset_id,
 					financial_nft_asset_id,
 				} => {
+					// AssetIds must be greater than 0
+					ensure!(!pool_asset.is_zero(), Error::<T>::InvalidAssetId);
+					ensure!(!share_asset_id.is_zero(), Error::<T>::InvalidAssetId);
+					ensure!(!financial_nft_asset_id.is_zero(), Error::<T>::InvalidAssetId);
+
 					// now < start_block < end_block
 					ensure!(
 						// Exclusively greater than to prevent errors/attacks
--- a/code/parachain/frame/staking-rewards/src/lib.rs
+++ b/code/parachain/frame/staking-rewards/src/lib.rs
@@ -203,7 +203,7 @@ pub mod pallet {
 	pub enum Error<T> {
 		/// Error when creating reward configs.
 		RewardConfigProblem,
-		/// AssetId is invalid, asset IDs must be graeter than 0
+		/// AssetId is invalid, asset IDs must be greater than 0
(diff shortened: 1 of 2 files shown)


ComposableFiB0t:
Task linked: [CU-2yyykk2 Staking Rewards - Pool can be created with `shareAssetId` & `financialNftAssetId` == 0](https://app.clickup.com/t/2yyykk2)

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -637,6 +640,11 @@ pub mod pallet {
 					share_asset_id,
 					financial_nft_asset_id,
 				} => {
+					// AssetIds must be greater than 0
+					ensure!(!pool_asset.is_zero(), Error::<T>::InvalidAssetId);
```
dzmitry-lahoda, line 644: can we use Validated?

code/parachain/frame/staking-rewards/src/test/mod.rs
```diff
@@ -85,6 +85,81 @@ fn test_create_reward_pool_invalid_end_block() {
...
+	});
+}
+
+#[test]
+fn create_staking_reward_pool_should_fail_when_share_asset_id_is_zero() {
+	new_test_ext().execute_with(|| {
+		System::set_block_number(1);
+
+		assert_err!(
+			StakingRewards::create_reward_pool(
```
dzmitry-lahoda, line 119: typical example where types would no need to write test. either use Validated OR create AssetId type with one line to serde back and check it is non null from some NonZero128. it would take several times less lines but universally reusable. and also reusable in arbitrary becase it will never generate 1. cc @benluelo

benluelo: Yep, agreed - we should be using `Validated` here

vimukthi-git (approved):
LGTM, though adding `Validated` would be nicer if the error messages works for the PolkadotJs integration correctly

PoisonPhang:
After talking to the rest of the Parachain finance team this morning, we've decided not to move this to Validated yet.
Currently, `Validated` does not properly emit errors to the runtime due to an oversight in its design.
Eventually, it would be good to move this, and other checks, to validated but we can't be doing that until we are receiving accurate errors.

code/parachain/frame/staking-rewards/src/test/mod.rs
PoisonPhang: After talking to the rest of the Parachain finance team this morning, we've decided not to move this to Validated yet.
Currently, Validated does not properly emit errors to the runtime due to an oversight in its design.
Eventually, it would be good to move this, and other checks, to validated but we can't be doing that until we are receiving accurate errors.

code/parachain/frame/staking-rewards/src/lib.rs
PoisonPhang: After talking to the rest of the Parachain finance team this morning, we've decided not to move this to Validated yet.
Currently, Validated does not properly emit errors to the runtime due to an oversight in its design.
Eventually, it would be good to move this, and other checks, to validated but we can't be doing that until we are receiving accurate errors.

dzmitry-lahoda:
@PoisonPhang there is no oversight. There is regression from Parity. Happend after 0.9.24 migration. Bad errors are not only from it, but from XCMP too. Same thing. Parity will fix it.

benluelo (approved):

dzmitry-lahoda:
not tracked. they enabled tx vodation by serde it. and in case if it fails, they panic. same happens with xcm and other protocola if message you path will not serde. https://composablefinance.slack.com/archives/C02GHMR9EF4/p1661433712074589?thread_ts=1660738175.670269&cid=C02GHMR9EF4

dzmitry-lahoda:
fix is trivial, but amount of churn is exorbital.

dzmitry-lahoda:
not tracked so. crazy they made such at first hand. likely they were lazy. 100÷ it will not go into their 1.0)))

dzmitry-lahoda:
nice. additionlly, i guess there is check is lower level also erroring that asset is wrong. so 2 have double layers of check with huge amount of code.

PR #1749, opened 2022-09-26
feat(devnet): persistent devnet with dispatch workflow

hussein-aitlahcen:
Introduce persistent devnet deployable using a dispatch workflow. Deploying the persistent devnet will always overwrite the existing one.


.github/workflows/build-test-deploy.yml
```diff
@@ -1101,26 +1101,6 @@ jobs:
           GCP_DEVNET_SERVICE_ACCOUNT: ${{ secrets.GCP_DEVNET_SERVICE_ACCOUNT }}
           GCP_DEVNET_SERVICE_ACCOUNT_KEY: ${{ secrets.GCP_DEVNET_SERVICE_ACCOUNT_KEY }}
 
-          # TODO: if not main -> dry run else deploy
```
dzmitry-lahoda, line 1104: not sure, it is still actual. it should check on branch before main deploy. maintaining things other way is pain.

dzmitry-lahoda:
can use https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule and 
do 
if main and push -> each build deploy
if cron or named trigger -> deploy staging/persistent
else do check

dzmitry-lahoda:
another issue. will it deploy only devnet, but what about UI. it should be coreleaed for shared playground imho

andor0 (approved):

.github/workflows/build-test-deploy.yml
```diff
@@ -1053,25 +1050,13 @@ jobs:
...
```
dzmitry-lahoda, line 1056: nixops scripts basically unmaintainable imho now, at least for me until something is done:
1. there is not main, there is no dryrun
2. there is not deep """typecheck""" of nix 
2.1 part started to be possible because of no mac
2.2. part was deleted with flake utils hook to allow nixops to be under system (so you could evaluate dummy package wrapping nixops and dump it). as nixops is root now (after removal of flake utils hook) not sure how to check it
3. sre did not gave me access of any kind to gce to have ability to map dynamic lazy not typed language to real data
4. i do not have experience with nixops

just FYI that that comment (and others removed earlier) were about maintain ability which is none for me now

.github/workflows/build-test-deploy.yml
```diff
@@ -1237,12 +1222,12 @@ jobs:
             nix run "github:ComposableFi/composable/${{ github.event.pull_request.head.sha }}#composable-node" -L
             ```
 
-            #### Spin up a local devnet (Dali + Rococo):
+            #### Spin up a local devnet (Rococo + Dali + Karura + Statemine):
```
dzmitry-lahoda, line 1225: that may be named just Release, because later there will be Hyperspace, and Picasso running at same time.

.github/workflows/build-test-deploy.yml
```diff
@@ -1237,12 +1222,12 @@ jobs:
...
             ```bash
-            nix run "github:ComposableFi/composable/${{ github.event.pull_request.head.sha }}#devnet-up" -L --option sandbox relaxed --show-trace
+            nix run "github:ComposableFi/composable/${{ github.event.pull_request.head.sha }}#devnet" -L --option sandbox relaxed --show-trace
             ```
 
-            #### Spin up a local XCVM devnet (Dali + Rococo + Juno):
+            #### Spin up a local XCVM devnet (Rococo + Dali + Juno):
```
dzmitry-lahoda, line 1230: This can be called just XCVM - same reason - soon it will be Eht, Aptos, etc.

.github/workflows/persistent-devnet.yml
```diff
@@ -0,0 +1,108 @@
...
+jobs:
+  cache-devnet:
+    name: "Cache Persistent Devnet"
+    runs-on:
+      - self-hosted
+      - linux
+      - x64
+      - sre
+    concurrency:
+       group: cache-devnet-persistent
```
dzmitry-lahoda, line 27: group may conflict with global group until prefixed by workflow (like in build editor)

dzmitry-lahoda: so you cannot run you own nixops from branch in parallel

.github/workflows/persistent-devnet.yml
```diff
@@ -0,0 +1,108 @@
...
+          nix-channel --add ${{ env.NIX_NIXPKGS_CHANNEL }} nixpkgs
+          nix-channel --update
+          nix-env -iA nixpkgs.cachix
+        authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+        name: ${{  env.CACHIX_COMPOSABLE }}
+    - name: "Build & Upload"
+      run: |
+        cachix watch-exec --jobs 16 --compression-level $CACHIX_COMPRESSION_LEVEL composable-community nix -- build .#devnet-persistent-program --no-update-lock-file --show-trace -L
+
+  deploy-devnet:
```
dzmitry-lahoda, line 52: name of job is different from display name - makes hard to navigate in ci jobs

.github/workflows/persistent-devnet.yml
```diff
@@ -0,0 +1,108 @@
...
+
+env:
+  NIX_NIXPKGS_CHANNEL: https://nixos.org/channels/nixpkgs-22.05-darwin
+  CACHIX_COMPOSABLE: composable-community
+  CACHIX_COMPRESSION_LEVEL: 3
+  NIXOPS_STATE_URL: gs://composable-state
+  NIXOPS_STATE_FILE: persistent.deployment.nixops
+
+jobs:
+  cache-devnet:
```
dzmitry-lahoda, line 19: name of jobs is different from display name . hard to navigate. also Persistent is already in root namespace, so having that repeated in name is not needed (same as in rust you do not do mod FooBar and then doo FooBar.FooBarMyStruct)

.github/workflows/persistent-devnet.yml
```diff
@@ -0,0 +1,108 @@
+name: Deploy Persistent Devnet
```
dzmitry-lahoda, line 1: display name of is different from file name - makes harder to navigate ci

.github/workflows/persistent-devnet.yml
```diff
@@ -0,0 +1,108 @@
...
+        then
+          gsutil cp $NIXOPS_STATE_URL/$NIXOPS_STATE_FILE $NIXOPS_STATE
+        else
+          cachix watch-exec --jobs 16 --compression-level $CACHIX_COMPRESSION_LEVEL composable-community nixops -- create --deployment devnet-gce --show-trace --option narinfo-cache-negative-ttl 0
+        fi
+      env:
+        NIXOPS_STATE: ${{ runner.temp }}/${{ env.NIXOPS_STATE_FILE }}
+    - name: Deploy
+      run: |
+        # For the love of God can anyone tell me why github runners are not stateless.
```
dzmitry-lahoda, line 98: i have not seen stat in their runners - at least not state which cannot be fixed from within job.

otherway comment is not needed

.nix/devnet-gce.nix
```diff
@@ -16,82 +15,78 @@ in {
...
-        tags = [ "http" "https" ];
+  "${machine-name}" = args@{ pkgs, resources, ... }:
+    ({
+      deployment = {
+        targetEnv = "gce";
+        gce = (gce-input // {
+          inherit region;
+          machineName = machine-name;
+          network = resources.gceNetworks.composable-devnet;
+          instanceType = "n2-standard-4";
```
dzmitry-lahoda, line 26: if we consider having ibc and xcvm stuff soon here to play with, `n2-standard-8` is safer not to worry in future. also it is safer for demos and people plaiting around. if this is machine all in one networks and ui and indexers.

.nix/devnet.nix
```diff
@@ -1,18 +1,124 @@
...
+              '';
+            };
+          };
+          mkDomain = name: attrs: {
+            "${name}" = {
+              enableACME = true;
+              forceSSL = true;
+            } // attrs;
+          };
+          # I agree, the hardcoded ports are a shame
```
dzmitry-lahoda, line 89: added to list of maintainability. can we have github or clickyp task about this? ports are use to wire IBC. so it will extend work needed to have IBC on persisten devnet.

dzmitry-lahoda: comment can be removed

composablejs/packages/types/src/cosmwasm/definitions.ts
```diff
@@ -0,0 +1,20 @@
+export default {
+  rpc: {
+  },
+  types: {
+    PalletCosmwasmEntryPoint: {
+      _enum: {
+		    Instantiate: "()",
```
dzmitry-lahoda, line 7: may fail pass ci formatter check

flake.nix
```diff
@@ -485,6 +483,55 @@
               ```
             '';
           };
+
+          frontend-static = mkFrontendStatic {
+            subsquidEndpoint = "http://localhost:4350/graphql";
+            picassoEndpoint = "ws://localhost:9988";
+            kusamaEndpoint = "ws://localhost:9944";
```
dzmitry-lahoda, line 490: it is rococo endpoint

flake.nix
```diff
@@ -1033,17 +1049,41 @@
...
+            in writeShellApplication {
+              name = "run-rococo-dali-karura";
+              text = ''
+                cat ${config-file}
+                rm -rf /tmp/polkadot-launch
+                ${packages.polkadot-launch}/bin/polkadot-launch ${config-file} --verbose
+              '';
+            };
+
+            devnet-rococo-picasso-karura-statemine = let
```
dzmitry-lahoda, line 1061: devnet-all? or devnet-release or devnet-persistent? assuming list will grow with other things . so it can just devnet-<name where it publushed>

.github/workflows/build-test-deploy.yml
hussein-aitlahcen: NixOps is not going to be maintained once the SRE team take the lead on deploying our artifacts, this is basically glue to have a devnet running before we have a proper testnet

.github/workflows/persistent-devnet.yml
hussein-aitlahcen: There is state for known hosts when you ssh. If the IP change it break.

.nix/devnet.nix
hussein-aitlahcen: It should not be maintained

flake.nix
hussein-aitlahcen: It can be kusama or even polkadot, must probably be renamed relaychain but this how FE named it

.nix/devnet-gce.nix
hussein-aitlahcen: This should not be maintained by the dev team and propely moved off to SRE/DevOps to deploy machines etc via a more stable tool like terraform

.github/workflows/persistent-devnet.yml
hussein-aitlahcen: It's done on purpose because the state file is shared and this tag must then be unique if the state file is used by another workflow

vimukthi-git (approved):

dzmitry-lahoda (approved):

PR #1750, opened 2022-09-26
[CU-2z0k75r] Move reward ammount balance check to `do_initialize`

PoisonPhang:
## Issue
[[CU-2z0k75r]](https://app.clickup.com/t/2z0k75r)

## Description
Credits to @benluelo for the `check_event` helper function

### DIFF
--- a/code/parachain/frame/crowdloan-rewards/src/lib.rs
+++ b/code/parachain/frame/crowdloan-rewards/src/lib.rs
@@ -113,6 +113,10 @@ pub mod pallet {
 			remote_account: RemoteAccountOf<T>,
 			reward_account: T::AccountId,
 		},
+		/// The crowdloan was successfully initialized, but with excess funds that won't be claimed
+		OverFunded {
+			excess_funds: T::Balance,
+		},
 	}
 
 	#[pallet::error]
@@ -173,6 +177,10 @@ pub mod pallet {
 		/// The upfront liquidity unlocked at first claim.
 		#[pallet::constant]
 		type InitialPayment: Get<Perbill>;
+
+		/// The percentage of excess funds required to trigger the `OverFunded` event.
+		#[pallet::constant]
+		type OverFundedThreshold: Get<Perbill>;
 
 		/// The time you have to wait to unlock another part of your reward.
 		#[pallet::constant]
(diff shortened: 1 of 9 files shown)


ComposableFiB0t:
Task linked: [CU-2z0k75r [Crowdloan Rewards] Move reward ammount balance check to `do_initialize`](https://app.clickup.com/t/2z0k75r)

code/parachain/frame/crowdloan-rewards/src/lib.rs
```diff
@@ -307,12 +315,34 @@ pub mod pallet {
...
 		pub(crate) fn do_initialize(at: MomentOf<T>) -> DispatchResult {
 			ensure!(!VestingTimeStart::<T>::exists(), Error::<T>::AlreadyInitialized);
+
 			let now = T::Time::now();
 			ensure!(at >= now, Error::<T>::BackToTheFuture);
+
+			let available_funds = T::RewardAsset::balance(&Self::account_id());
+			let total_rewards = TotalRewards::<T>::get();
+			ensure!(available_funds >= total_rewards, Error::<T>::RewardsNotFunded);
+			let excess_funds = available_funds - total_rewards;
```
benluelo, line 337:
```suggestion
            let excess_funds = available_funds.checked_sub(total_rewards).ok_or(Error::<T>::RewardsNotFunded)?;
```

benluelo: Not sure if this is better, but it removes the `-` and the invariant required to use it safely (just makes it the same operation)

benluelo (approved):

code/parachain/frame/crowdloan-rewards/src/lib.rs
PoisonPhang: Good catch. The actions themselves were originally separate steps, but can efficiently be taken together

code/parachain/frame/crowdloan-rewards/src/lib.rs
```diff
@@ -307,12 +315,35 @@ pub mod pallet {
...
 			ensure!(at >= now, Error::<T>::BackToTheFuture);
+
+			let available_funds = T::RewardAsset::balance(&Self::account_id());
+			let total_rewards = TotalRewards::<T>::get();
+			let excess_funds = available_funds
+				.checked_sub(&total_rewards)
+				.ok_or(Error::<T>::RewardsNotFunded)?;
+
+			if excess_funds > T::OverFundedThreshold::get().mul_floor(total_rewards) {
+				Self::deposit_event(Event::OverFunded { excess_funds })
```
dzmitry-lahoda, line 341: so it will spam continuously in case of issue. seem better place is logs.

dzmitry-lahoda (approved):

code/parachain/frame/crowdloan-rewards/src/lib.rs
benluelo: This isn't a hook, it only emits when the extrinsic is called

dzmitry-lahoda: oh. fun. then why not error. nwmd

dzmitry-lahoda: if that is not background, why not error? just qurious - may be that is some patter to follow

benluelo: Because it's not an error to have it over funded, this event is just a notification really

dzmitry-lahoda: okey. i have arguments it should not be here than. but whatever:)

PoisonPhang:
@Mergifyio refresh

PR #1756, opened 2022-09-27
[CU-302tgzn] Add a building of composable-centauri docker image

andor0:
## Issue
https://app.clickup.com/t/302tgzn


flake.nix
```diff
@@ -616,6 +649,14 @@
               chain-spec = "dali-dev";
             }).script;
 
+            # Dali Centauri devnet
+            devnet-dali-centauri = (callPackage mk-devnet {
```
dzmitry-lahoda, line 653: can we prefix this like

bridge-devnet-dali

flake.nix
```diff
@@ -644,6 +685,28 @@
               '';
             };
 
+            # Dali Centauri devnet container
+            centauri-devnet-container = dockerTools.buildImage {
```
dzmitry-lahoda, line 689: bridge-devnet-dali

dzmitry-lahoda (commented):
I think we should prefix temporal forks by team names which requested that. centauri prefix is okey too. so I would prefer bridge.

flake.nix
andor0: fixed

flake.nix
andor0: fixed

andor0:
> I think we should prefix temporal forks by team names which requested that. centauri prefix is okey too. so I would prefer bridge.

I added `bridge` prefix.

haroldsphinx (approved):

blasrodri (approved):

PR #1759, opened 2022-09-27
fix(nix-build): fixes subxt URL

RomarQ:
## Description

I am using `nix (Nix) 2.6.1` and was getting the following error `error: 'https://github.com/paritytech//subxt' is not a valid URL` when running `nix build ".#composable-node"`.

This commit fixes the issue on my side.


RomarQ:
Updating nix to `2.11.1` seems to have fixed the issue.

PR #1784, opened 2022-09-29
Simulating simnode test to debug issues [Ignore]

haroldsphinx:


dzmitry-lahoda (approved):

dzmitry-lahoda:
@haroldsphinx is this branch where you reproduced sim node panic issue?

PR #1803, opened 2022-09-30
[CU-30eeta4] Update paths in the workflow to update weight files

andor0:
## Issue
https://app.clickup.com/t/30eeta4


andor0:
> Should we add the needs-benchmarks label to update the benchmarks? They're quite outdated

It makes sense. Added `needs-benchmarks` label.

dzmitry-lahoda:
would be resoanble to move script into relevant folder of code. beacause it benches only substrate runtimes, not anything else. it is not global by local. it would be monorepo approach.

dzmitry-lahoda (approved):

dzmitry-lahoda:
thins what would be in multirepo? this script would be located into code. so monorepo is multirepo when people can think multi when needed:)

benluelo (approved):

PoisonPhang (approved):

dzmitry-lahoda:
could we run same script for bench checks and bench runs. so parameters passed with different count? that way check runs alwasy, and run never broken.

andor0:
> would be resoanble to move script into relevant folder of code. beacause it benches only substrate runtimes, not anything else. it is not global by local. it would be monorepo approach.

done.

haroldsphinx (approved):

dzmitry-lahoda (approved):

hussein-aitlahcen (approved):

andor0:
@mergify refresh

andor0:
@mergify refresh

PR #1840, opened 2022-10-04
[CU-33e4tdu] Crowdloan Rewards - Lock Reward Asset When Configured

PoisonPhang:
## Issue
[[CU-33e4tdu]](https://app.clickup.com/t/33e4tdu)

## Description
Depending on the runtime configuration, Crowdloan Rewards will lock the reward asset on claiming. This is configured by the `LockByDefault` value.

Once the time comes to remove locks, locks can be removed by the new `unlock_rewards_for` provided the accounts that need unlocking.
To ensure we're able to fit transactions in blocks. This is provided a Vector of accounts and can be called multiple times to eventually unlock all accounts.

> Do we really need this flag? Given that there is also RemoveRewardLocks which can be set/reset at genesis?

Changing the crowdloans genesis config was not an option

### DIFF
--- a/code/parachain/frame/crowdloan-rewards/src/lib.rs
+++ b/code/parachain/frame/crowdloan-rewards/src/lib.rs
@@ -517,7 +517,7 @@ pub mod pallet {
 			})
 		}
 
-		/// Sets `RemoveRewardLocks`, removes `RewardAsset` locks on provided accounts, emmits
+		/// Sets `RemoveRewardLocks`, removes `RewardAsset` locks on provided accounts, emits
 		/// `RewardsUnlocked`.
 		fn do_unlock(reward_accounts: Vec<T::AccountId>) {
 			RemoveRewardLocks::<T>::put(());
--- a/code/parachain/frame/crowdloan-rewards/src/lib.rs
+++ b/code/parachain/frame/crowdloan-rewards/src/lib.rs
@@ -80,7 +80,8 @@ pub mod pallet {
 		pallet_prelude::*,
 		traits::{
 			fungible::{Inspect, Mutate, Transfer},
-			Time,
+			tokens::WithdrawReasons,
+			LockIdentifier, LockableCurrency, Time,
 		},
 		transactional, PalletId,
 	};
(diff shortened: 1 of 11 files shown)


ComposableFiB0t:
Task linked: [CU-33e4tdu Crowdloan rewards pallet - config changes (lock PICA)](https://app.clickup.com/t/33e4tdu)

code/parachain/frame/crowdloan-rewards/src/lib.rs
```diff
@@ -199,6 +203,14 @@ pub mod pallet {
...
 		#[pallet::constant]
 		type PalletId: Get<PalletId>;
+
+		/// The unique identifier for locks maintained by this pallet.
+		#[pallet::constant]
+		type LockId: Get<LockIdentifier>;
+
+		/// If claimed amounts should be locked by the pallet
+		#[pallet::constant]
+		type LockByDefault: Get<bool>;
```
vimukthi-git, line 213: Do we really need this flag? Given that there is also `RemoveRewardLocks` which can be set/reset at genesis?

code/parachain/frame/crowdloan-rewards/src/lib.rs
```diff
@@ -457,6 +489,16 @@ pub mod pallet {
 					let available_to_claim = should_have_claimed.saturating_sub(reward.claimed);
 					ensure!(available_to_claim > T::Balance::zero(), Error::<T>::NothingToClaim);
 
+					reward.claimed = available_to_claim.saturating_add(reward.claimed);
+					if T::LockByDefault::get() && !RemoveRewardLocks::<T>::exists() {
```
vimukthi-git, line 493: Seems this condition is too verbose given my previous comment :thinking:

vimukthi-git (approved):
LGTM, minor comment

code/parachain/frame/crowdloan-rewards/src/lib.rs
PoisonPhang: I was under the impression that changing the crowdloans genesis config was not an option. Is this not the case?
@vimukthi-git

vimukthi-git: Yeah, I remember now, keep forgetting that detail :sweat_smile:

benluelo (approved):

PR #1841, opened 2022-10-04
Use firebase for documentation deployments.

haroldsphinx:
- firebase deploy

### DIFF
--- a/.github/workflows/docs-firebase-deploy.yml
+++ b/.github/workflows/docs-firebase-deploy.yml
@@ -1,0 +1,40 @@
+name: Deploy Documentation
+
+on:
+  push: 
+    branches: 
+      - main
+
+env:
+  FIREBASE_TOKEN: ${{ secrets.FIREBASE_TOKEN }}
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2     
+
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '16.4.0'
+
+      - name: Install Rust
+        run: |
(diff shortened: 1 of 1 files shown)


dzmitry-lahoda (approved):

Martin-Kiss (approved):

.github/workflows/docs-firebase-deploy.yml
```diff
@@ -0,0 +1,43 @@
+name: Firebase Deploy Gcloud
```
KaiserKarel, line 1:
```suggestion
name: Deploy Documentation
```

.github/workflows/docs-firebase-deploy.yml
```diff
@@ -0,0 +1,43 @@
...
+        run: |
+         rustup set profile minimal
+         rustup toolchain install stable
+         rustup default stable
+
+      - name: Install mdbook
+        run: |
+          mkdir bin
+          curl -sSL https://github.com/rust-lang/mdBook/releases/download/v0.4.14/mdbook-v0.4.14-x86_64-unknown-linux-gnu.tar.gz | tar -xz --directory=bin
+          echo "$(pwd)/bin" >> $GITHUB_PATH
```
KaiserKarel, line 32: We have/should have a nix step for this; we never manually install rust in CI, as that will lead to different versions.

cor (changes requested):
Just run `nix build ".#composable-book"` instead of installing all of this manual tooling

cor:
Not using Nix. Closing this one in favor of including it in the next Docusaurus PR

PR #1844, opened 2022-10-05
fixing statemine  and acala builds

dzmitry-lahoda:
## Issue
* seems statemine is not well buildable under crane nor rustPlatform. 

## Description
* so used raw build for statemine
* fixed acala hash (some how hash changed without change in commit -> some git hooks or git updates are no deterministic)
* made some clarification on shared `substrate` based network build deps

### DIFF
--- a/.nix/polkadot-bin.nix
+++ b/.nix/polkadot-bin.nix
@@ -17,7 +17,7 @@ rustPlatform.buildRustPackage rec {
   cargoSha256 = "sha256-6y+WK2k1rhqMxMjEJhzJ26WDMKZjXQ+q3ca2hbbeLvA=";
   meta = { mainProgram = "polkadot"; };
 
-  # next for substrate builds with runtime wasm generated crates
+  # substrate-attrs-node-with-attrs
   __noChroot = true;
   doCheck = false;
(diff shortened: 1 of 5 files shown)


easteregg (approved):
Thanks Dzmitry

.nix/polkadot-bin.nix
```diff
@@ -24,5 +29,4 @@ rustPlatform.buildRustPackage rec {
   LIBCLANG_PATH = "${llvmPackages.libclang.lib}/lib";
   PROTOC = "${protobuf}/bin/protoc";
   ROCKSDB_LIB_DIR = "${rocksdb}/lib";
-  meta = { mainProgram = "polkadot"; };
```
hussein-aitlahcen, line 27: Please keep the metadata

dzmitry-lahoda: the name is not shared across chains, so i moved it up(L18). so it would be easier to replace all common shared attrs later

andor0 (approved):

KaiserKarel (approved):

KaiserKarel (approved):

KaiserKarel (approved):

PR #1847, opened 2022-10-05
Add late-binding to XCVM SDK and CosmWasm interpreter

aeryz:
## Description
Late-binding is now supported. For V0, we go with the indexed method but the sdk makes it really easy to define a late call as it is explained in detail in docs of `cosmwasm.rs`.

## Checklist


code/xcvm/cosmwasm/contracts/router/src/contract.rs
```diff
@@ -179,38 +278,75 @@ fn handle_instantiate_reply(deps: DepsMut, msg: Reply) -> StdResult<Response> {
 		)?)?
 	};
 
-	INTERPRETERS.save(deps.storage, (router_reply.0, router_reply.1), &interpreter_address)?;
+	match INTERPRETERS.load(deps.storage, (router_reply.0, router_reply.1.clone())) {
```
hussein-aitlahcen, line 281: It's clearer to deconstruct a tuple than using indexing like instead of `(x.0, x.1)` deconstruct at the function def with `(meaningful_name, meanigful_name_2): (A, B)`

code/xcvm/lib/core/src/bridge.rs
```diff
@@ -0,0 +1,73 @@
...
+	Debug,
+	Encode,
+	Decode,
+	TypeInfo,
+	Serialize,
+	Deserialize,
+)]
+#[repr(i32)]
+pub enum BridgeSecurity {
+	None,
```
hussein-aitlahcen, line 25: Please remove this variant as its isomorphic to `Option<BridgeSecurity>`

code/xcvm/lib/proto/src/lib.rs
```diff
@@ -0,0 +1,452 @@
+#![no_std]
```
hussein-aitlahcen, line 1: Are we using this mapping instead of `proto::X` etc? It's better to avoid having a dep to protobuf in the contracts themselves, we need to map from the proto representation of a program to the base, Rust one

code/xcvm/cosmwasm/contracts/asset-registry/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
```
hussein-aitlahcen, line 1: I believe we need to update the registry to handle native assets as well, currently we only handle CW20 assets right? Would be nice to move to a `enum AssetReference { Native(Denom), Virtual(Address) }`

hussein-aitlahcen (changes requested):

code/xcvm/lib/core/src/bridge.rs
aeryz: It is more convenient to have None variant here though because it makes it super convenient to compare bridge securities since they can be represented as i32

code/xcvm/lib/proto/src/lib.rs
aeryz: We can possibly convert every proto type to our SDK types but I don't think we should pay for this. It will unnecessarily increase the gas price and we already convert to rust types if we want to use any sdk function.

code/xcvm/lib/proto/src/lib.rs
```diff
@@ -0,0 +1,452 @@
...
+
+// XCVM types to Protobuf conversion
+
+impl From<Amount> for Balance {
+	fn from(amount: Amount) -> Self {
+		// Note that although functionally nothing changes, there is no guarantee of getting the
+		// same protobuf when you convert protobuf to XCVM types and convert back again. Because
+		// `intercept = 0 & ratio = 0` is always converted to `Absolute`. But this can be also
+		// expressed with `Ratio` and `Unit` as well. Also, since the ratio is expanded to use
+		// denomitor `MAX_PARTS`, it also won't be the same.
```
Martin-Kiss, line 281:
```suggestion
		// denominator `MAX_PARTS`, it also won't be the same.
```
Nit spelling, failing spellcheck CI

code/xcvm/lib/proto/src/lib.rs
```diff
@@ -0,0 +1,452 @@
...
+}
+
+impl TryFrom<Balance> for Amount {
+	type Error = ();
+
+	fn try_from(balance: Balance) -> core::result::Result<Self, Self::Error> {
+		let balance_type = balance.balance_type.ok_or(())?;
+		let wrap = |num: u128| -> FixedU128<U16> { FixedU128::wrapping_from_num(num) };
+		// TODO(aeryz): This can be a helper function in SDK so that users won't
+		// necesarilly need to know how the ratio is handled in our SDK
```
Martin-Kiss, line 239:
```suggestion
		// necessarily need to know how the ratio is handled in our SDK
```
Nit typo, failing spellcheck CI

.config/dictionaries/technical.dic
```diff
@@ -358,6 +358,7 @@ USDCUSDT
 usermod
 vadimcn
 vsicons
+wasmd
```
Martin-Kiss, line 361:
```suggestion
wasmd
protos
```
Addition to the dictionary.

code/xcvm/lib/proto/src/lib.rs
hussein-aitlahcen: You cannot be sure that the cost will be very high after optimization and this will lead to the contract knowing implementation details of the serialization and also force you to unwrap or error almost everywhere because of how protobuf is  encoded. Let's move off using protobuf types directly

Martin-Kiss (approved):

hussein-aitlahcen (approved):

PR #1848, opened 2022-10-05
[CU-https://app.clickup.com/t/302tgzn] Add a building of composable-centauri docker image

andor0:
## Issue
https://app.clickup.com/t/302tgzn

### DIFF
--- a/scripts/polkadot-launch/rococo-local-bridge-dali-dev.nix
+++ b/scripts/polkadot-launch/rococo-local-bridge-dali-dev.nix
@@ -1,0 +1,68 @@
+# definition of parachain
+# TODO: replace with zombienet
+# because it allows to specify more things and tests
+# more structured and portable and officially endorsed by parity
+# so with nix it is easier to build own (nix+curl+websocket)
+
+{ pkgs, polkadot-bin, composable-bin }:
+with pkgs;
+let builder = pkgs.callPackage ./network-builder.nix { };
+in {
+  result = builder.mk-shared-security-network {
+    relaychain = {
+      bin = "${polkadot-bin}/bin/polkadot";
+      chain = "rococo-local";
+      port = 30444;
+      wsPort = 9944;
+      count = 5;
+      flags = [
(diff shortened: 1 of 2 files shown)


scripts/polkadot-launch/rococo-local-bridge-dali-dev.nix
```diff
@@ -0,0 +1,68 @@
...
+        "--beefy"
+      ];
+    };
+    parachains = [
+      {
+        id = 2001;
+        port = 31200;
+        wsPort = 9988;
+        count = 3;
+        chain = "dali-2001";
```
dzmitry-lahoda, line 33: would be simpler if one network will be 2087 as it was

dzmitry-lahoda (approved):

scripts/polkadot-launch/rococo-local-bridge-dali-dev.nix
andor0: The bridge team asked so.

andor0:
This is duplicate of https://github.com/ComposableFi/composable/pull/1849.

scripts/polkadot-launch/rococo-local-bridge-dali-dev.nix
dzmitry-lahoda: i see. so -bridge- is infix in file name. while in it is prefix (good) in flake. so naming all bridge team with bridge- prefix could be good. 

pablo team cannot use 2001 as all tooling ant rest rely on 2087.

andor0: > i see. so -bridge- is infix in file name. while in it is prefix (good) in flake. so naming all bridge team with bridge- prefix could be good.

fixed in https://github.com/ComposableFi/composable/pull/1849

> pablo team cannot use 2001 as all tooling ant rest rely on 2087.

perhaps, we will change para_id to 2087 in the future.

PR #1851, opened 2022-10-05
[CU-3070g6a] validate minimum staking position when staking and splitting

andresmechali:
## Issue
https://app.clickup.com/t/3070g6a

## MISSING
Review the default minimum value for staking

### DIFF
--- a/code/parachain/frame/composable-traits/src/staking/mod.rs
+++ b/code/parachain/frame/composable-traits/src/staking/mod.rs
@@ -137,6 +137,9 @@ pub struct RewardPool<
 
 	// Asset ID (collection ID) of the financial NFTs issued for staking positions of this pool
 	pub financial_nft_asset_id: AssetId,
+
+	/// Minimum amount to be staked.
+	pub minimum_staking_amount: Balance,
 }
 
 /// Default transfer limit on new asset added as rewards.
@@ -196,6 +199,9 @@ pub enum RewardPoolConfiguration<
 
 		// Asset ID (collection ID) of the financial NFTs issued for staking positions of this pool
 		financial_nft_asset_id: AssetId,
+
+		// Minimum amount to be staked
+		minimum_staking_amount: Balance,
 	},
 }
 
--- a/code/parachain/frame/pablo/src/lib.rs
+++ b/code/parachain/frame/pablo/src/lib.rs
@@ -676,7 +676,7 @@ pub mod pallet {
 			// NOTE(connor): `start_block` must greater than current block
 			let start_block = frame_system::Pallet::<T>::current_block_number() + 1_u32.into();
 			let end_block = start_block + five_years_block.into();
-			let minimum_staking_amount: T::Balance = T::Convert::convert(10_000_u128); // TODO: review this amount
+			let minimum_staking_amount: T::Balance = T::Convert::convert(10_000_u128);
 
 			Ok(RewardPoolConfiguration::RewardRateBasedIncentive {
 				owner: Self::account_id(pool_id),
--- a/code/parachain/frame/pablo/src/lib.rs
+++ b/code/parachain/frame/pablo/src/lib.rs
@@ -676,6 +676,7 @@ pub mod pallet {
 			// NOTE(connor): `start_block` must greater than current block
 			let start_block = frame_system::Pallet::<T>::current_block_number() + 1_u32.into();
 			let end_block = start_block + five_years_block.into();
+			let minimum_staking_amount: T::Balance = T::Convert::convert(10_000_u128); // TODO: review this amount
 
 			Ok(RewardPoolConfiguration::RewardRateBasedIncentive {
 				owner: Self::account_id(pool_id),
@@ -686,6 +687,7 @@ pub mod pallet {
 				lock,
 				share_asset_id: Self::get_x_token_from_pool(*pool_id)?,
 				financial_nft_asset_id: Self::get_financial_nft_from_pool(*pool_id)?,
+				minimum_staking_amount,
 			})
 		}
 
--- a/code/parachain/frame/pablo/src/lib.rs
+++ b/code/parachain/frame/pablo/src/lib.rs
@@ -676,7 +676,7 @@ pub mod pallet {
 			// NOTE(connor): `start_block` must greater than current block
 			let start_block = frame_system::Pallet::<T>::current_block_number() + 1_u32.into();
 			let end_block = start_block + five_years_block.into();
-			let minimum_staking_amount: T::Balance = T::Convert::convert(10_000_u128);
+			let minimum_staking_amount: T::Balance = T::Convert::convert(2_000_000_u128);
 
 			Ok(RewardPoolConfiguration::RewardRateBasedIncentive {
 				owner: Self::account_id(pool_id),
--- a/code/parachain/frame/staking-rewards/src/lib.rs
+++ b/code/parachain/frame/staking-rewards/src/lib.rs
@@ -254,6 +254,8 @@ pub mod pallet {
 		SlashedAmountTooLow,
 		/// Staked amount is less than the minimum staking amount for the pool.
 		StakedAmountTooLow,
+		/// Staked amount after split is less than the minimum staking amount for the pool.
+		StakedAmountTooLowAfterSplit,
 	}
 
 	pub(crate) type AssetIdOf<T> = <T as Config>::AssetId;
@@ -1039,11 +1041,11 @@ pub mod pallet {
 					ensure!(
 						ratio.mul_floor(existing_position.stake) >=
 							rewards_pool.minimum_staking_amount,
-						Error::<T>::StakedAmountTooLow
+						Error::<T>::StakedAmountTooLowAfterSplit
 					);
 					ensure!(
 						new_stake >= rewards_pool.minimum_staking_amount,
-						Error::<T>::StakedAmountTooLow
+						Error::<T>::StakedAmountTooLowAfterSplit
 					);
 
 					let new_reductions = {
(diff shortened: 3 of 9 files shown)


ComposableFiB0t:
Task linked: [CU-3070g6a Validate minimum staking position(amount) when staking](https://app.clickup.com/t/3070g6a)

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1025,6 +1032,20 @@ pub mod pallet {
...
+
+					ensure!(
+						ratio.mul_floor(existing_position.stake) >=
+							rewards_pool.minimum_staking_amount,
+						Error::<T>::StakedAmountTooLow
+					);
+					ensure!(
+						new_stake >= rewards_pool.minimum_staking_amount,
+						Error::<T>::StakedAmountTooLow
+					);
```
benluelo, line 1047: We could consider a different error here perhaps - something like `StakedAmountTooLowAfterSplit`?

benluelo (commented):
We also need to ensure that the minimum staked amount is >= ED when slashed, this should be done at pool creation

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: We could also add it to the docs on `StakedAmountTooLow` that it applies to both staking and splitting

andresmechali: Whatever option you prefer it's fine for me, let me know and I change it

andresmechali: For now I'll go with the first one, as a more specific seems always better

benluelo: I prefer the more specific error personally, so that works for me :+1:

code/parachain/frame/staking-rewards/src/test/mod.rs
```diff
@@ -271,6 +277,101 @@ fn stake_in_case_of_low_balance_should_not_work() {
...
+
+		assert_noop!(
+			StakingRewards::split(
+				Origin::signed(staker),
+				STAKING_FNFT_COLLECTION_ID,
+				original_fnft_instance_id,
+				ratio.try_into_validated().unwrap(),
+			),
+			crate::Error::<Test>::StakedAmountTooLowAfterSplit
+		);
```
PoisonPhang, line 371: NIT
These are three different test conditions, lets split them into their respective tests

code/parachain/frame/staking-rewards/src/test/mod.rs
```diff
@@ -271,6 +277,101 @@ fn stake_in_case_of_low_balance_should_not_work() {
...
+		});
+
+		process_and_progress_blocks::<StakingRewards, Test>(1);
+
+		let amount = 15_000;
+		let original_fnft_instance_id =
+			stake_and_assert::<Test, runtime::Event>(ALICE, PICA::ID, amount, ONE_MINUTE);
+
+		// Original stake is less than the minimum and new stake is greater.
+		let ratio = Permill::from_parts(333_334);
```
benluelo, line 326: you can also use ::from_rational, it's a bit simpler

PoisonPhang (approved):
LGTM!
Functionally looks correct, some NIT comments.

Product would have the most insight on if the default amount (10k) is correct.

vimukthi-git (approved):
LGTM

code/parachain/frame/composable-traits/src/staking/mod.rs
```diff
@@ -196,6 +199,9 @@ pub enum RewardPoolConfiguration<
 
 		// Asset ID (collection ID) of the financial NFTs issued for staking positions of this pool
 		financial_nft_asset_id: AssetId,
+
+		// Minimum amount to be staked
+		minimum_staking_amount: Balance,
```
benluelo, line 204: I'm wondering if we could/should make this optional? If a pool doesn't want to set a minimum staked amount, it would just default to the requiring `slash(staked amount) >= ED`

benluelo: but if a pool wants to have the minimum stake larger than that, then it could be configured as such

PoisonPhang: I think this would be a good change in terms of flexibility. Better to plan for it now then need to change it later
+1

vimukthi-git: the number of staked positions that can be created by an attacker(or just an annoying user) increases for the same amount when min stake is low.

benluelo: perhaps we should consider limiting the number of times a user can stake then?

vimukthi-git: That's not going to stop the storage growth for the system since a user can create any number of accounts they want

benluelo: Then should we limit the amount of stakes per pool then?

benluelo: Also, a minimum amount will be different per asset - even if we  set it to something arbitrarily high, any bad actor with enough funds could blow up our storage. (also 10k of one asset is different than 10k of another asset, so we can't just pass a maximum through the Config either)

vimukthi-git: That's going to limit the number of staking opportunities for a token holder, say a PICA holder. Ultimately the limit we want to enforce is the number of staked positions possible per supply of the token staked.

code/parachain/frame/staking-rewards/src/test/mod.rs
andresmechali: Seems like too repetitive to have 3 tests here, as they are basically the same but just making sure that under any non desirable scenario the operation will fail. Any better way of doing it without creating multiple tests? Maybe iteration over a small array?

code/parachain/frame/composable-traits/src/staking/mod.rs
vimukthi-git: >a minimum amount will be different per asset

This is why we are making it a configurable parameter instead of hard coding. We will decide these at the time of the launch ideally together with other params.

benluelo: Which works for pools created by governance, but if/when we allow pools to be created  by anyone, we will need to revisit this (perhaps enforcing `minimum stake > X % of total_issuance`, or something similar)

vimukthi-git: Agree, these are economic simulation level stuff, that we haven't got to yet. Ideally with Pablo launch we can decide on token prices better. But until that time any thing is just relative and subject to human judgement

vimukthi-git: Even Polkadot adjusts this value routinely

code/parachain/frame/staking-rewards/src/test/mod.rs
PoisonPhang: Prop testing with set values is an option but may be outside the scope of this PR.
In general, even if tests are repetitive in structure, it's good to have them split up if they are testing different cases. This will make debugging quicker if one but not all cases begin failing.
Up to you though, still a NIT comment.

benluelo (approved):

code/parachain/frame/pablo/src/lib.rs
```diff
@@ -676,6 +676,7 @@ pub mod pallet {
 			// NOTE(connor): `start_block` must greater than current block
 			let start_block = frame_system::Pallet::<T>::current_block_number() + 1_u32.into();
 			let end_block = start_block + five_years_block.into();
+			let minimum_staking_amount: T::Balance = T::Convert::convert(10_000_u128);
```
KaiserKarel, line 679: Can this be a documented const/passed through the config?

KaiserKarel (approved):

code/parachain/frame/pablo/src/lib.rs
benluelo: All of the default poop values will be passed in somehow through the config, this is just a temporary value until the requirements are reviewed with product

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -701,15 +709,34 @@ pub mod pallet {
 
 					let now_seconds = T::UnixTime::now().as_secs();
 
+					let existential_deposit = T::ExistentialDeposits::get(&pool_asset);
+
+					println!("existential_deposit, {:?}", existential_deposit);
```
dzmitry-lahoda, line 714: no sure, may avoid having print in this code?

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1025,6 +1055,21 @@ pub mod pallet {
 					// position, that way any rounding is accounted for.
 					let new_stake = left_from_one_ratio.mul_ceil(existing_position.stake);
 					let new_share = left_from_one_ratio.mul_ceil(existing_position.share);
+
+					let rewards_pool = RewardPools::<T>::get(existing_position.reward_pool_id)
```
dzmitry-lahoda, line 1059: More robust solution with less code in transactional extrinsic would be fail on save to storage. So any attempt to save stake with low amount could be done via one API call, regardless if that was split, stake, unstake. Error will be StakeBalanceTooLow or something. `StakedAmountTooLowAfterSplit` encodes 2 things, `she called split and get to low balance`. But she already know what tx she called. Error is not common path and increased weight in case of error is fine because it will not be larger then when success happens.

So I want to tell is that code is more complicated then it can be.

code/parachain/frame/staking-rewards/src/lib.rs
andresmechali: yes, I was just debugging something but no print will get merged

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: I think this current implementation is fine for now, we can look into possible alternatives in the future

PR #1856, opened 2022-10-05
Add migrate support to `pallet-cosmwasm` and implement migrate entrypoint on contracts

aeryz:
## Description
* Add `migrate` extrinsic for contract migration.
* Add `migrate` entrypoint to all of the cosmwasm contracts as well as CONTRACT_VERSION and NAMES to make sure that migration is securely done.
* Add `clear_admin` extrinsic to cover `ClearAdmin` functionality in cosmwasm.
* Add `update_admin` extrinsic to cover `UpdateAdmin` functionality in cosmwasm.
* Add `verify`s to extrinsic's benchmarks to test them during benchmarking.
* Use `CodeIdentifier` instead of code id to make users able to use code hash if they want to for example batch instantiate + execute or migrate calls. This way they would be able to identify the code in-front.

## Checklist


code/parachain/frame/cosmwasm/src/lib.rs
```diff
@@ -507,6 +555,58 @@ pub mod pallet {
...
+				Ok(())
+			})?;
+
+			Self::deposit_event(Event::<T>::Migrated {
+				contract: contract.clone(),
+				to: new_code_id,
+			});
+
+			Ok(contract_info)
+		}
```
KaiserKarel, line 608: Nit: this pattern allows for incorrectly migrating a contract. Instead we could use a Migrator struct, using Typestate to ensure that `migrate_phase1` is called before a contract is migrated.

KaiserKarel, line 608: Nit: this pattern allows for incorrectly migrating a contract. Instead we could use a Migrator struct, using Typestate to ensure that `migrate_phase1` is called before a contract is migrated.

KaiserKarel: In general, let's avoid patterns where foo_1 needs to be called before foo, and encode that logic in the type system.

KaiserKarel (changes requested):
Still needs benchmarks

code/xcvm/cosmwasm/contracts/asset-registry/src/contract.rs
```diff
@@ -32,6 +38,12 @@ pub fn execute(
 	}
 }
 
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn migrate(deps: DepsMut, _env: Env, _msg: MigrateMsg) -> Result<Response, ContractError> {
```
dzmitry-lahoda, line 42: please add /// doc to it.

dzmitry-lahoda (approved):

KaiserKarel (approved):

code/parachain/frame/cosmwasm/src/lib.rs
```diff
@@ -674,21 +749,70 @@ pub mod pallet {
...
+			if info.code_id != code_id {
+				// Increase the refcount of `new_code_id`.
+				CodeIdToInfo::<T>::try_mutate_exists(code_id, |entry| -> Result<(), Error<T>> {
+					let code_info = entry.as_mut().ok_or(Error::<T>::CodeNotFound)?;
+					code_info.refcount =
+						code_info.refcount.checked_add(1).ok_or(Error::<T>::RefcountOverflow)?;
+					Ok(())
+				})?;
+
+				// Modify the exising `code_id`'s states and unreserve the bonded funds.
```
Martin-Kiss, line 776:
```suggestion
				// Modify the existing `code_id`'s states and unreserve the bonded funds.
```
Nit spelling; failing spellcheck CI

PR #1857, opened 2022-10-05
[CU-3153c9f] Moved fNFT collection and share asset IDs into staking

PoisonPhang:
## Issue
[[CU-3153c9f]](https://app.clickup.com/t/3153c9f)

## Description
Had to change `FinancialNftProtocl::value_of` in staking because it incorrectly queried pools by fNFT collection ID rather than stake asset ID.

Added function to `CurrencyFactory` trait `expect_create()`, which creates an asset ID without a transactional layer so it can be used within genesis builds of other pallets.

NOTE: All tests that rely on `with_stake` and use its optional `claim` are incorrect and should be rewritten with correct values. This is outside the scope of this PR.

### DIFF
--- a/code/parachain/frame/composable-traits/src/fnft.rs
+++ b/code/parachain/frame/composable-traits/src/fnft.rs
@@ -56,6 +56,7 @@ pub trait FinancialNftProtocol {
 	/// - collection: id of the financial NFT collection issued/used by the protocol.
 	/// TODO (vim): Think how to represent the difference between assets and liabilities
 	fn value_of(
+		pool_id: &Self::AssetId,
 		collection: &Self::AssetId,
 		instance: &Self::ItemId,
 	) -> Result<Vec<(Self::AssetId, Self::Balance)>, DispatchError>;
--- a/code/parachain/frame/assets/src/mocks.rs
+++ b/code/parachain/frame/assets/src/mocks.rs
@@ -74,6 +74,10 @@ impl CurrencyFactory for CurrencyIdGenerator {
 		Ok(1)
 	}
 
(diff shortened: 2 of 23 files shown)


ComposableFiB0t:
Task linked: [CU-3153c9f Move fNFT and xToken ID aquesition into staking rewards](https://app.clickup.com/t/3153c9f)

code/parachain/frame/composable-traits/src/fnft.rs
```diff
@@ -56,6 +56,7 @@ pub trait FinancialNftProtocol {
 	/// - collection: id of the financial NFT collection issued/used by the protocol.
 	/// TODO (vim): Think how to represent the difference between assets and liabilities
 	fn value_of(
+		pool_id: &Self::AssetId,
```
vimukthi-git, line 59: This must not be required, breaks the interface if third parties who has no idea about the pools to find a pool id

vimukthi-git (changes requested):

vimukthi-git:
> Had to change FinancialNftProtocl::value_of in staking because it incorrectly queried pools by fNFT collection ID rather than stake asset ID.

Yeah definitely a bug though, the current change is incorrect, we need to somehow map fnft collection id to pool and execute the operation

code/parachain/frame/composable-traits/src/fnft.rs
PoisonPhang: I see; I'll make a workaround so the interface can still be neutral to the underlying design

PoisonPhang:
On hold while other tasks are prioritized.

PoisonPhang:
PR no longer relevant

PR #1859, opened 2022-10-06
run separate simnode job on managed runner

dzmitry-lahoda:

## Issue
- simnode on hosted runner fails often for more then half year and lately often with not fix
- failed on my recent devnet jobs each time

## Description
- it was proven that hosted runner is approx same fast to run simnode as non hosted in shared nixified job
- so replacing on separate job too
 
## Checklist

- ci works same way as before

### DIFF
--- a/.github/workflows/simnode.yml
+++ b/.github/workflows/simnode.yml
@@ -10,11 +10,7 @@ on:
 
 jobs:
   simnode:
-    runs-on:
-      - self-hosted
-      - linux
-      - x64
-      - sre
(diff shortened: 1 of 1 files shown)


alexkdefi (approved):
👍

KaiserKarel (approved):

dzmitry-lahoda:
@Mergifyio refresh

dzmitry-lahoda:
Something aborted mergifyio check on merge.

PR #1916, opened 2022-10-08
Refactor assertions into trait

benluelo:
Steps towards a better testing framework


vimukthi-git (approved):

code/parachain/frame/composable-tests-helpers/src/test/helper.rs
```diff
@@ -41,186 +41,187 @@ pub fn default_acceptable_computation_error(x: u128, y: u128) -> Result<(), Fixe
...
+			frame_system::Pallet::<Self>::events().pop().expect("No events present!");
+
+		match event.clone().try_into() {
+			Ok(pallet_event) => match f(pallet_event.clone()) {
+				Some(r) => r,
+				None => panic!("expected event was not found; found {pallet_event:#?}"),
+			},
+			Err(_) => panic!(
+				r#"last event was not from this pallet
 found {event:#?}"#
```
PoisonPhang, line 112: NIT
Indenting on this looks funky
(and on other uses of `r#".."#)

code/parachain/frame/staking-rewards/src/benchmarking.rs
```diff
@@ -198,26 +207,11 @@ benchmarks! {
...
-					reward_multiplier,
-					keep_alive
-				} => Some(fnft_instance_id),
-				_ => None,
-			}
+		let instance_id = stake_and_assert::<T, <T as frame_system::Config>::Event>(
+			user.clone(),
+			BASE_ASSET_ID.into(),
+			100_000_000.into(),
+			ONE_HOUR,
```
PoisonPhang, line 214: +1

PoisonPhang (approved):
LGTM! :rocket: 
Minor nit on formatting, but that might be `cargo fmt`. Good work, nice refactoring as well.

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -37,7 +37,7 @@
 #[cfg(any(feature = "runtime-benchmarks", test))]
 mod benchmarking;
 mod prelude;
-#[cfg(test)]
+#[cfg(any(feature = "runtime-benchmarks", test))]
```
PoisonPhang, line 40: This seems to be breaking benchmarks. What was the reasoning for this being added?

benluelo: It actually fixed benchmarks locally for me, very strange. I'll review it again

code/parachain/frame/composable-tests-helpers/src/test/helper.rs
benluelo: With raw strings, leading spaces are included in the string itself, so there's not much I can do about it other than make them all start with a newline like this:

```rust
let s = r#"
Blah
Blah
Blah
"#;

(Which I actually prefer and try do do, I'll update any that aren't formatted that way 👍)

File: code/parachain/frame/staking-rewards/src/test/test_update_reward_pools.rs [hunk 4]
```diff
@@ -1,7 +1,7 @@
 use composable_tests_helpers::test::{
```
dzmitry-lahoda, line 1: test/test_ and inside it [test]. not sure if need to say test so many times?

code/parachain/frame/staking-rewards/src/lib.rs
dzmitry-lahoda: modification is correct afaik.

dzmitry-lahoda (approved):
nice

code/parachain/frame/staking-rewards/src/test/test_update_reward_pools.rs
benluelo: I didn't write this original crate, I just went with the existing structure. I'm planning to restructure the `composable-tests-helpers` crate in the future though, since the current layout is definitely not ideal

code/parachain/frame/composable-tests-helpers/src/test/helper.rs
benluelo: I think I've fixed all of these, lmk if I missed any!

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: These should be fixed now - just have to wait for benchmarks to run to be sure :+1:

PR #1923, opened 2022-10-10
[Hyperspace] Fishing mode

### LINKED ISSUE #1870: Implement fishing mode

vmarkushin:
Closes ComposableFi/centauri#75


code/centauri/hyperspace/Cargo.toml
```diff
@@ -60,11 +60,42 @@ state-machine = { package = "sp-state-machine", git = "https://github.com/parity
...
+#sp-panic-handler = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.27" }
+#sp-runtime = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.27" }
+#sp-runtime-interface = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.27" }
+#sp-runtime-interface-proc-macro = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.27" }
+#sp-state-machine = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.27" }
+#sp-std = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.27" }
+#sp-storage = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.27" }
+#sp-tracing = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.27" }
+#sp-trie = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.27" }
+#sp-wasm-interface = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.27" }
```
seunlanlege, line 101:
```suggestion
```

code/centauri/hyperspace/parachain/src/parachain.rs
```diff
@@ -1 +1,55 @@
...
+	tip: u128,
+	asset_id: Option<u128>,
+}
+
+pub type SignedExtra = (Era, CheckNonce, ChargeAssetTxPayment);
+
+pub struct UncheckedExtrinsic<T: Config> {
+	pub signature: Option<(<T as Config>::Address, sp_runtime::MultiSignature, SignedExtra)>,
+	pub function: Call,
+}
```
seunlanlege, line 20: Why not use the one in `sp_runtime`

code/centauri/hyperspace/src/lib.rs
```diff
@@ -40,6 +40,70 @@ where
...
+			result = chain_a_client_updates.next() => {
+				match result.transpose()? {
+					// stream closed
+					None => break,
+					Some((transaction_id, events)) => {
+						for (i, event) in events.into_iter().enumerate() {
+							if let IbcEvent::UpdateClient(client_update) = event {
+								if *client_update.client_id() != chain_b.client_id() {
+									continue;
+								}
```
seunlanlege, line 65: much cleaner to do this at the point where the stream is initialized

code/centauri/hyperspace/parachain/src/parachain.rs
vmarkushin: @seunlanlege I tried this option at first, but it requires the `SignedExtension` trait to be implemented for `SignedExtra`. If it's fine to have a blank implementation of it, I can use it

code/centauri/hyperspace/primitives/src/lib.rs
```diff
@@ -344,6 +362,18 @@ pub trait Chain: IbcProvider + KeyProvider + Send + Sync {
 	/// This should be used to submit new messages [`Vec<Any>`] from a counterparty chain to this
 	/// chain.
 	async fn submit(&self, messages: Vec<Any>) -> Result<(), Self::Error>;
+
+	/// Returns an [`AnyClientMessage`] that triggered an [`IbcEvent`] at index `event_index`
+	/// on this chain.
+	///
+	/// Parameters `host_block_hash` and `transaction_id` identify the transaction that triggered
+	/// the event.
+	async fn query_client_message(
```
Wizdave97, line 371: A transaction id as a hash would be a more generic  interface than a `u32`

code/centauri/hyperspace/parachain/src/chain.rs
```diff
@@ -176,4 +395,50 @@ where
 
 		Ok(())
 	}
+
+	async fn query_client_message(
```
Wizdave97, line 399: @seunlanlege  could we just store the client update headers offchain, so they can be retrieved is an easier manner.

seunlanlege: Already discussed with @vmarkushin , we'll use tx querying as that is the most chain agnostic solution.

code/parachain/frame/ibc/src/lib.rs
```diff
@@ -406,9 +406,7 @@ pub mod pallet {
 	#[pallet::generate_deposit(pub (super) fn deposit_event)]
 	pub enum Event<T: Config> {
 		/// Events emitted by the ibc subsystem
-		Events { events: Vec<events::IbcEvent> },
-		/// Errors emitted by the ibc subsystem
-		Errors { errors: Vec<errors::IbcError> },
+		Events { events: Vec<Result<events::IbcEvent, errors::IbcError>> },
```
Wizdave97, line 409: You need to modify the `block_events` runtime_api implementation in the runtime file after this change.

code/centauri/hyperspace/Cargo.toml
```diff
@@ -61,3 +61,4 @@ sp-keyring = { git = "https://github.com/paritytech/substrate", branch = "polkad
 
 [features]
 parachain = ["dep:parachain", "subxt"]
+testing = []
```
seunlanlege, line 64:
```suggestion
```

code/centauri/algorithms/grandpa/prover/src/lib.rs
```diff
@@ -237,7 +236,7 @@ where
 			.ok_or_else(|| anyhow!("Failed to fetch previous finalized hash + 1"))?;
 
 		let mut unknown_headers = vec![];
-		for height in previous_finalized_height..=latest_finalized_height {
+		for height in client_state.latest_relay_height..=latest_finalized_height {
```
Wizdave97, line 239: Why this change, could cause us to miss some parachain updates?

code/centauri/algorithms/grandpa/prover/src/lib.rs
```diff
@@ -310,30 +310,30 @@ where
...
-			// in the special case where there's no parachain headers, let's only send the the
-			// finality target and it's parent block. Fishermen should detect any byzantine
-			// activity.
-			let len = finality_proof.unknown_headers.len();
-			if len > 2 {
-				finality_proof.unknown_headers =
-					finality_proof.unknown_headers[(len - 2)..].to_owned()
-			}
-		}
+		// if let Some((relay_hash, _)) = para_headers.into_iter().min_by_key(|(_, h)| *h.number())
```
Wizdave97, line 318: why was this commented out?

Wizdave97 (changes requested):

code/centauri/algorithms/grandpa/prover/src/lib.rs
vmarkushin: Yes, exactly

code/centauri/algorithms/grandpa/prover/src/lib.rs
vmarkushin: I think it's because we don't need to prune the headers anymore. But It's better to ask @seunlanlege, he suggested removing it.

code/centauri/algorithms/grandpa/prover/src/lib.rs
Wizdave97: I mean the former code ensured we don’t miss any blocks

Wizdave97: What you changed it to is what he had there previously.

code/centauri/algorithms/beefy/prover/src/lib.rs
```diff
@@ -84,7 +84,7 @@ impl HostFunctions for Crypto {
 /// This contains methods for fetching BEEFY proofs for parachain headers.
 pub struct Prover<T: Config> {
 	/// Subxt client for the relay chain
-	pub relay_client: OnlineClient<T>,
+	pub relay_client: OnlineClient<PolkadotConfig>,
```
seunlanlege, line 87:
```suggestion
	pub relay_client: OnlineClient<T>,
```

code/centauri/algorithms/beefy/prover/src/lib.rs
```diff
@@ -232,10 +232,11 @@ where
 	) -> Result<(Vec<ParachainHeader>, BatchProof<H256>), Error>
 	where
 		T::BlockNumber: Ord + sp_runtime::traits::Zero,
+		u32: From<T::BlockNumber>,
 	{
-		let header_numbers = header_numbers.into_iter().collect();
+		let header_numbers = header_numbers.into_iter().map(From::from).collect();
 		let FinalizedParaHeads { leaf_indices, raw_finalized_heads: finalized_blocks } =
-			fetch_finalized_parachain_heads(
+			fetch_finalized_parachain_heads::<PolkadotConfig>(
```
seunlanlege, line 239:
```suggestion
			fetch_finalized_parachain_heads::<T>(
```

code/centauri/algorithms/grandpa/primitives/src/justification.rs
```diff
@@ -101,21 +104,23 @@ where
...
+
 			check_message_signature::<Host, _, _>(
 				&message,
 				&signed.id,
 				&signed.signature,
 				self.round,
 				set_id,
-			)?;
+			)
+			.map_err(|x| anyhow!("{} {:?}", x, (message.encode(), self.round, set_id)))?;
```
seunlanlege, line 115: why the weird error message?

code/centauri/algorithms/grandpa/primitives/src/lib.rs
```diff
@@ -98,8 +99,15 @@ pub struct ParachainHeadersWithFinalityProof<H: Header> {
 
 /// Host functions that allow the light client perform cryptographic operations in native.
 pub trait HostFunctions: light_client_common::HostFunctions + 'static {
+	/// RelayChain header type.
+	type Header: Header;
+
 	/// Verify an ed25519 signature
 	fn ed25519_verify(sig: &ed25519::Signature, msg: &[u8], pub_key: &ed25519::Public) -> bool;
+	/// Stores the given list of RelayChain header hashes in the light client's storage.
+	fn add_relaychain_header_hashes(headers: &[<Self::Header as Header>::Hash]);
```
seunlanlege, line 108:
```suggestion
	fn insert_relay_header_hashes(headers: &[<Self::Header as Header>::Hash]);
```

code/centauri/algorithms/grandpa/primitives/src/lib.rs
```diff
@@ -98,8 +99,15 @@ pub struct ParachainHeadersWithFinalityProof<H: Header> {
...
+	/// Checks if a RelayChain header hash exists in the light client's storage.
+	fn exists_relaychain_header_hash(hash: <Self::Header as Header>::Hash) -> bool;
```
seunlanlege, line 110:
```suggestion
	fn contains_relay_header_hash(hash: <Self::Header as Header>::Hash) -> bool;
```

code/centauri/algorithms/grandpa/prover/src/lib.rs
```diff
@@ -79,15 +77,15 @@ where
...
+		let relay_client = OnlineClient::from_rpc_client(relay_ws_client.clone()).await?;
 		let para_ws_client = Arc::new(WsClientBuilder::default().build(para_ws_url).await?);
 		let para_client = OnlineClient::<T>::from_rpc_client(para_ws_client.clone()).await?;
 
 		Ok(Self { relay_ws_client, relay_client, para_ws_client, para_client, para_id })
 	}
 
 	/// Construct the inital client state.
-	pub async fn initialize_client_state(&self) -> Result<ClientState<T::Hash>, anyhow::Error> {
+	pub async fn initialize_client_state(&self) -> Result<ClientState<H256>, anyhow::Error> {
```
seunlanlege, line 88:
```suggestion
	pub async fn initialize_client_state(&self) -> Result<ClientState<T::Hash>, anyhow::Error> {
```

code/centauri/algorithms/grandpa/prover/src/lib.rs
seunlanlege: David is correct, this code is wrong. Should be `previous_finalized_height`.

code/centauri/algorithms/grandpa/verifier/src/lib.rs
```diff
@@ -69,18 +69,19 @@ where
...
+	let from = client_state.latest_relay_hash;
+	let mut finalized = headers.ancestry(from, target.hash()).map_err(|_| {
+		anyhow!("[verify_parachain_headers_with_grandpa_finality_proof] Invalid ancestry!")
+	})?;
 	finalized.sort();
 
 	// 2. verify justification.
 	let justification = GrandpaJustification::<H>::decode(&mut &finality_proof.justification[..])?;
+
+	if justification.commit.target_hash != finality_proof.block {
```
seunlanlege, line 81: rather than having to do this check everywhere, let's just remove `block` from `FinalityProof` and use the `justification.commit.target_hash` everywhere instead

code/centauri/hyperspace/bin/chain.rs
```diff
@@ -562,6 +589,22 @@ impl Chain for AnyChain {
 			_ => unreachable!(),
 		}
 	}
+
+	async fn query_client_message(
+		&self,
+		host_block_hash: [u8; 32],
+		transaction_index: usize,
+		event_index: usize,
```
seunlanlege, line 597: this is still not chain agnostic, this should instead only take the block hash and consensus height.

code/centauri/hyperspace/parachain/src/chain.rs
```diff
@@ -38,6 +56,86 @@ type BeefyJustification =
...
+		&self,
+		counterparty: &C,
+		client_message: AnyClientMessage,
+	) -> Result<(), anyhow::Error> {
+		match client_message {
+			AnyClientMessage::Grandpa(ClientMessage::Header(header)) => {
+				// todo: verify block num. This block number may not exist on relaychain
+				// (see `verify_parachain_headers_with_grandpa_finality_proof` implementation)
+				let base_block_number =
+					header.finality_proof.unknown_headers[0].number.saturating_sub(1);
```
seunlanlege, line 88: this isn't correct. should use the `justification.commit.target_number`

code/centauri/hyperspace/parachain/src/chain.rs
```diff
@@ -38,6 +56,86 @@ type BeefyJustification =
...
+							"No justification found for block: {:?}",
+							header.finality_proof.block
+						)
+					})?
+					.0;
+
+				let trusted_finality_proof =
+					FinalityProof::<RelayChainHeader>::decode(&mut &encoded[..])?;
+
+				if header.finality_proof.block != trusted_finality_proof.block {
```
seunlanlege, line 106: let's compare justification targets here instead

code/centauri/algorithms/grandpa/verifier/src/lib.rs
```diff
@@ -69,18 +69,19 @@ where
...
 
-	let from = finality_proof
-		.unknown_headers
-		.iter()
-		.min_by_key(|h| *h.number())
-		.ok_or_else(|| anyhow!("Unknown headers can't be empty!"))?;
-	let mut finalized = headers
-		.ancestry(from.hash(), target.hash())
-		.map_err(|_| anyhow!("Invalid ancestry!"))?;
+	let from = client_state.latest_relay_hash;
```
seunlanlege, line 72: seeing as there might be relay chain headers < `client_state.latest_relay_hash`, we should also check that there exists a route between the minimum header in unknown headers to `client_state.latest_relay_hash`.

code/centauri/algorithms/grandpa/verifier/src/lib.rs
vmarkushin: In some places (like [here](https://github.com/ComposableFi/composable/blob/centauri/code/centauri/light-clients/ics10-grandpa/src/client_def.rs#L148)) we don't decode justification and use this field. How about using the last item in `unknown_headers`?

code/centauri/algorithms/grandpa/prover/src/lib.rs
vmarkushin: Ok, I see what you mean

code/centauri/algorithms/grandpa/verifier/src/lib.rs
seunlanlege: Ok as long as we enforce in the verifier that 

```rust
unkown_headers.iter().max_by_key(Header::number) == finality_proof.block
```

code/centauri/hyperspace/parachain/src/chain.rs
seunlanlege: Lets sync on this one over a huddle

code/centauri/algorithms/grandpa/verifier/src/lib.rs
vmarkushin: Yes, I think we should enforce that since it's one of the [invariants](https://github.com/ComposableFi/composable/blob/centauri/code/centauri/algorithms/grandpa/primitives/src/lib.rs#L55)

code/centauri/hyperspace/bin/chain.rs
vmarkushin: How about making it generic, so the client can use chain-specific types? Because otherwise, we'd have to run through all transactions in the block and then through all the messages in the `deliver` arguments

vmarkushin:
Moved to: https://github.com/ComposableFi/centauri/pull/84

PR #1937, opened 2022-10-11
[CU-31pc6c3] Add a building of composable-bridge-devnet docker image

andor0:
## Issue
https://app.clickup.com/t/31pc6c3


.github/workflows/build-test-deploy.yml
```diff
@@ -856,51 +856,49 @@ jobs:
...
+        with:
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+      - run: |
+          # enable required features (see https://github.com/cachix/install-nix-action/issues/19)
+          mkdir -p ~/.config/nix
+          echo "system-features = kvm" >> ~/.config/nix/nix.conf
+      - run: |
+          cachix watch-exec --jobs 16 --compression-level $CACHIX_COMPRESSION_LEVEL composable-community nix -- build .#bridge-devnet-dali-container --no-update-lock-file --show-trace -L
+          docker load --input ./result
```
dzmitry-lahoda, line 899: i would prefer it builds commit hash too non on main. people may need branches to test. there is example on how to do it in this file (actaully hat 100% cp paste with just change of package and container name)

.nix/mmr-polkadot-bin.nix
```diff
@@ -0,0 +1,28 @@
...
+with pkgs;
+rustPlatform.buildRustPackage rec {
+  # HACK: break the nix sandbox so we can build the runtimes. This
+  # requires Nix to have `sandbox = relaxed` in its config.
+  # We don't really care because polkadot is only used for local devnet.
+  __noChroot = true;
+  name = "mmr-polkadot-v${version}";
+  version = "0.9.27";
+  src = fetchFromGitHub {
+    repo = "polkadot";
```
dzmitry-lahoda, line 11: please make reuse of this stuff. share it with usual polkadot and replace hashes via nix. version, rev, hash, cargoSha256, name could be inputs to template.

code/parachain/node/src/command.rs
```diff
@@ -41,7 +41,7 @@ fn load_spec(id: &str) -> std::result::Result<Box<dyn sc_service::ChainSpec>, St
 		// Must define the default chain here because `export-genesis-state` command
 		// does not support `--chain` and `--parachain-id` arguments simultaneously.
 		#[cfg(feature = "dali")]
-		"dali-dev" => Box::new(chain_spec::dali_dev()),
+		id if id.contains("dali") => Box::new(chain_spec::dali_dev(id)),
```
dzmitry-lahoda, line 44: not sure, what if we will have dali-rococo like others do tuned for rococo remote by parity?

dzmitry-lahoda: that is one of the reason why we use nix - we share and compose.

KaiserKarel (approved):

andor0:
@Mergifyio refresh

.github/workflows/build-test-deploy.yml
andor0: done.

.nix/mmr-polkadot-bin.nix
andor0: done.

code/parachain/node/src/command.rs
andor0: then we will update the code for this case.

dzmitry-lahoda (approved):

PR #1945, opened 2022-10-11
rm log4rs from hyperspace

blasrodri:
removing unused `log4rs` which is also giving us an error on unmaintained dependency `typemap`
https://github.com/ComposableFi/composable/actions/runs/3229723835/jobs/5287337940#step:3:5

### DIFF
--- a/code/Cargo.lock
+++ b/code/Cargo.lock
@@ -3220,6 +3220,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "destructure_traitobject"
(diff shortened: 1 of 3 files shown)


seunlanlege:
Do we even use this lib?

Wizdave97:
> Do we even use this lib?

Don't think so.

vimukthi-git (approved):

blasrodri:
@Wizdave97 @seunlanlege dependency removed

PR #1946, opened 2022-10-11
Ben/extend stake logic/#2v8m6fj

benluelo:
Updates to extend stake behaviour

### DIFF
--- a/code/parachain/frame/staking-rewards/src/lib.rs
+++ b/code/parachain/frame/staking-rewards/src/lib.rs
@@ -63,6 +63,7 @@ pub use pallet::*;
 #[frame_support::pallet]
 pub mod pallet {
 	pub use crate::weights::WeightInfo;
+
 	use composable_support::{
 		math::safe::{SafeAdd, SafeDiv, SafeMul, SafeSub},
 		validation::{validators::GeOne, TryIntoValidated, Validated},
@@ -77,6 +78,7 @@ pub mod pallet {
 		time::{DurationSeconds, ONE_MONTH, ONE_WEEK},
 	};
 	use frame_support::{
+		defensive,
 		traits::{
 			fungibles::{
 				Inspect as FungiblesInspect, InspectHold as FungiblesInspectHold,
@@ -90,7 +92,7 @@ pub mod pallet {
 				},
 				WithdrawConsequence,
 			},
-			DefensiveSaturating, TryCollect, UnixTime,
+			Defensive, DefensiveSaturating, TryCollect, UnixTime,
 		},
 		transactional, BoundedBTreeMap, PalletId,
 	};
@@ -261,8 +263,6 @@ pub mod pallet {
 		StakedAmountTooLow,
 		/// Staked amount after split is less than the minimum staking amount for the pool.
 		StakedAmountTooLowAfterSplit,
-		/// A staked position can only be extended to a duration >= the existing duration.
-		CannotExtendToShorterDuration,
 	}
 
 	pub(crate) type AssetIdOf<T> = <T as Config>::AssetId;
@@ -573,8 +573,11 @@ pub mod pallet {
 				&fnft_collection_id,
 				&fnft_instance_id,
 			)?;
+
+			// TODO(benluelo): This needs to be passed in through the extrinsic
 			let keep_alive = true;
-			let _position_id = <Self as Staking>::extend(
+
+			<Self as Staking>::extend(
 				&who,
 				(fnft_collection_id, fnft_instance_id),
 				amount,
@@ -850,7 +853,6 @@ pub mod pallet {
 				Self::compute_rewards_and_reductions(awarded_shares, &rewards_pool)?;
 
 			rewards_pool.rewards = rewards;
-			rewards_pool.total_shares = rewards_pool.total_shares.safe_add(&awarded_shares)?;
 
 			let fnft_collection_id = rewards_pool.financial_nft_asset_id;
 			let fnft_instance_id = T::FinancialNft::get_next_nft_id(&fnft_collection_id)?;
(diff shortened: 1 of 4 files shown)


ComposableFiB0t:
Task linked: [CU-2v8m6fj Staking Rewards - validate 'extend stake amount' logic](https://app.clickup.com/t/2v8m6fj)

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -565,18 +568,23 @@ pub mod pallet {
 			origin: OriginFor<T>,
 			fnft_collection_id: T::AssetId,
 			fnft_instance_id: T::FinancialNftInstanceId,
+			duration: DurationSeconds,
```
vimukthi-git, line 571: We are not going to allow setting a duration, just that original duration gets applied again.

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -904,78 +914,135 @@ pub mod pallet {
...
-					WithdrawConsequence::Success
-				),
-				Error::<T>::NotEnoughAssets
-			);
+				RewardPools::<T>::try_mutate(stake.reward_pool_id, |maybe_rewards_pool| {
+					let rewards_pool =
+						maybe_rewards_pool.as_mut().ok_or(Error::<T>::RewardsPoolNotFound)?;
 
-			let awarded_shares = Self::boosted_amount(reward_multiplier, amount)?;
+					let reward_multiplier = Self::reward_multiplier(rewards_pool, duration)
```
vimukthi-git, line 929: Reward multiplier must not be allowed to mutate from the originally set one.

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -904,78 +914,135 @@ pub mod pallet {
...
+				ensure!(duration >= stake.lock.duration, Error::<T>::CannotExtendToShorterDuration);
 
-			ensure!(
-				matches!(
-					T::Assets::can_withdraw(rewards_pool.asset_id, who, amount),
-					WithdrawConsequence::Success
-				),
-				Error::<T>::NotEnoughAssets
-			);
+				RewardPools::<T>::try_mutate(stake.reward_pool_id, |maybe_rewards_pool| {
```
vimukthi-git, line 925: Indentation looks off

vimukthi-git (changes requested):
Need to remove duration parameter and modification to reward multiplier that would be caused by it

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -904,78 +914,135 @@ pub mod pallet {
...
+				})
+			})
 		}
 
 		#[transactional]
 		fn unstake(
 			who: &Self::AccountId,
 			(fnft_collection_id, fnft_instance_id): &Self::PositionId,
 		) -> DispatchResult {
+			// TODO(benluelo): Use ::take here instead of try_get and then remove
```
PoisonPhang, line 1030: +1

PoisonPhang (approved):
LGTM! :rocket: 
(Assuming existing comments are addressed)

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: looks fine to me, I know github has issues rendering indents sometimes though

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -905,77 +912,122 @@ pub mod pallet {
...
+						matches!(
+							T::Assets::can_withdraw(rewards_pool.asset_id, who, amount),
+							WithdrawConsequence::Success
+						),
+						Error::<T>::NotEnoughAssets
+					);
 
-			let awarded_shares = Self::boosted_amount(reward_multiplier, amount)?;
+					let new_shares = Self::boosted_amount(
+						// SAFETY: The duration preset on an existing stake should be valid in the
```
vimukthi-git, line 932: I'm afraid we can not make this assumption as it risks throwing off the math for rewarding all users in the pool if we make a mistake in the future. So we are forced to store the `duration_preset` in the `Stake`.

benluelo: Not sure I follow - do we need to store the `reward_multiplier` in the stake as well?

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -905,77 +912,122 @@ pub mod pallet {
...
+						// pool since it's currently not possible to modify the presets after pool
+						// creation.
+						rewards_pool
```
vimukthi-git, line 935: In-lining this in the function decreases the readability at least in review because it's not clear what value is being passed.

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -905,77 +912,122 @@ pub mod pallet {
...
-			rewards_pool.total_shares = rewards_pool.total_shares.safe_add(&awarded_shares)?;
-			rewards_pool.rewards = rewards;
-			stake.stake = stake.stake.safe_add(&amount)?;
-			stake.share = stake.share.safe_add(&awarded_shares)?;
-			for (asset, additional_inflation) in reductions.iter() {
-				let inflation =
-					stake.reductions.get_mut(asset).ok_or(Error::<T>::ReductionConfigProblem)?;
-				*inflation = inflation.safe_add(additional_inflation)?;
-			}
+					for (reward_asset_id, reward) in &mut rewards_pool.rewards {
```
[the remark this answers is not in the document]
benluelo: I had the same thought, but I would want to make changes to `stake` as well and I figured it was out of scope for this PR. I can do a followup PR that does that change

code/parachain/frame/staking-rewards/src/lib.rs
vimukthi-git: Yeah as we now use that for the extension and the risk of any mutation to that is the math being wrong for the entire pool

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: https://github.com/ComposableFi/composable/commit/dda75e674d708b76807572b440ffc30306e49f41

code/parachain/frame/staking-rewards/src/test/mod.rs
```diff
@@ -625,73 +625,267 @@ fn stake_in_case_of_not_zero_inflation_should_work() {
...
+				Stake {
+					fnft_instance_id,
+					reward_pool_id: STAKED_ASSET::ID,
+					stake: staked_amount + extended_amount,
+					share: Pallet::<Test>::boosted_amount(
+						rewards_pool.lock.duration_presets[&ONE_MINUTE],
+						staked_amount + extended_amount
+					)
+					.expect("boosted amount calculation should not fail"),
+					// TODO(benluelo): Figure out a way to not have a magic number here (i.e. how to
```
vimukthi-git, line 763: Having a magic number is ok in a unit test as long as the calculation is explained in a comment. Calculating it just hard codes the test to the current implementation of the protocol

vimukthi-git (approved):
LGTM

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: The only way this would error is if the duration presets in the pool changed somehow - if that functionality were added in the future and this is no longer upheld, then the tests would catch it since `defensive*()` methods panic in debug mode when they're triggered. I can replace this with an `expect` call, I just figured a fallback was the best option

code/parachain/frame/staking-rewards/src/test/mod.rs
vimukthi-git: Think of a property test like rsTest for example

PR #1963, opened 2022-10-13
efficient grandpa prover

seunlanlege:
Experimenting with optimizing the amount of data that needs to be sent for GRANDPA proofs, not needed for launch

### DIFF
--- a/code/centauri/algorithms/grandpa/primitives/src/lib.rs
+++ b/code/centauri/algorithms/grandpa/primitives/src/lib.rs
@@ -57,6 +57,7 @@ pub struct FinalityProof<H: Header> {
 }
 
 /// Previous light client state.
+#[derive(Clone)]
 pub struct ClientState<H> {
 	/// Current authority set
 	pub current_authorities: AuthorityList,
--- a/code/centauri/algorithms/grandpa/primitives/src/lib.rs
+++ b/code/centauri/algorithms/grandpa/primitives/src/lib.rs
@@ -64,6 +64,8 @@ pub struct ClientState<H> {
 	pub current_set_id: u64,
 	/// latest finalized height on the relay chain.
 	pub latest_relay_height: u32,
+	/// latest finalized height on the parachain.
+	pub latest_para_height: u32,
 	/// latest finalized hash on the relay chain.
 	pub latest_relay_hash: H,
 	/// para_id of associated parachain
--- a/code/centauri/algorithms/grandpa/primitives/src/lib.rs
+++ b/code/centauri/algorithms/grandpa/primitives/src/lib.rs
@@ -85,6 +85,7 @@ pub struct ParachainHeaderProofs {
 }
 
 /// Parachain headers with a Grandpa finality proof.
+#[derive(Clone)]
 pub struct ParachainHeadersWithFinalityProof<H: Header> {
 	/// The grandpa finality proof: contains relay chain headers from the
 	/// last known finalized grandpa block.
--- a/code/centauri/algorithms/grandpa/verifier/src/lib.rs
+++ b/code/centauri/algorithms/grandpa/verifier/src/lib.rs
@@ -40,7 +40,10 @@ use sp_trie::{LayoutV0, StorageProof};
 #[cfg(test)]
 mod tests;
 
-/// Verify a new grandpa justification, given the old state.
+/// This function verifies the GRANDPA finality proof for relay chain headers.
+///
+/// Next, we prove the finality of parachain headers, by verifying patricia-merkle trie state proofs
+/// of these headers, stored at the recently finalized relay chain heights.
 pub fn verify_parachain_headers_with_grandpa_finality_proof<H, Host>(
 	mut client_state: ClientState<H::Hash>,
 	proof: ParachainHeadersWithFinalityProof<H>,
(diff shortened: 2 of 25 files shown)


code/centauri/algorithms/grandpa/prover/build.rs
```diff
@@ -15,6 +15,7 @@
 
 #[tokio::main]
 async fn main() -> color_eyre::Result<()> {
-	subxt_codegen::build_script("ws://127.0.0.1:9944", "runtime").await?;
+	subxt_codegen::build_script("ws://127.0.0.1:9944", "polkadot").await?;
+	subxt_codegen::build_script("ws://127.0.0.1:9188", "parachain").await?;
 	Ok(())
 }
```
KaiserKarel, line 21: build.rs will make deterministic builds a lot more difficult; so at some point we'll remove this and generate this stuff with nix.

dzmitry-lahoda: actually it is blocker for having deterministic(nix) hyperspace and docker compose builds [docker](https://app.clickup.com/t/3p6g7gr). so we cannot help bridge team until that is fixed. i would prefer CI nix and local dev would follow same flow. so in this case build.rs will be removed. also build.rs will fail to maintain several relays version (non local) and runtimes.  cc @seunlanlege

dzmitry-lahoda: cc @andor0 @cor

code/centauri/algorithms/grandpa/verifier/src/lib.rs
```diff
@@ -51,57 +54,78 @@ where
...
 
-	// 2. next check that there exists a route from client.latest_relay_hash to target.
-	let finalized = headers
-		.ancestry(client_state.latest_relay_hash, finality_proof.block)
+	// this is illegal
+	if target.hash() != finality_proof.block {
+		Err(anyhow!("Latest finalized block should be highest block in unknown_headers"))?;
+	}
+
+	let from = finality_proof
```
blasrodri, line 72: not sure if worth it:
but perhaps it's better to loop over all unknown headers once and return the min and max by key?

blasrodri (approved):

code/centauri/algorithms/grandpa/primitives/src/lib.rs
```diff
@@ -57,13 +57,16 @@ pub struct FinalityProof<H: Header> {
 }
 
 /// Previous light client state.
+#[derive(Clone)]
```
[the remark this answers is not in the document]
seunlanlege: > do we need to clone the entire struct?

Not really sure what you're asking sir

code/centauri/algorithms/grandpa/verifier/src/lib.rs
seunlanlege: > but perhaps it's better to loop over all unknown headers once and return the min and max by key?

Well given that we're pruning unknown headers, it's not worth it.

code/centauri/algorithms/grandpa/primitives/src/lib.rs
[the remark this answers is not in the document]
blasrodri: the `Clone` trait is derived. Usually is because you want to call `.clone()` over the entire struct. And given that it contains authority sets, and other attributes that can be expensive to clone, I was wondering if we needed to clone the entire struct. Or if there's a cheaper alternative where only a partial bit it's copied, or we could copy a pointer to the struct, etc...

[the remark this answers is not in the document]
seunlanlege: The clone is used in tests sir

code/centauri/algorithms/grandpa/verifier/src/tests.rs
```diff
@@ -74,44 +77,51 @@ async fn follow_grandpa_justifications() {
...
 			)
 			.await
 			.expect("Failed to fetch finalized parachain headers with proof");
 
-		client_state = verify_parachain_headers_with_grandpa_finality_proof::<
+		let new_client_state = verify_parachain_headers_with_grandpa_finality_proof::<
 			Header,
 			HostFunctionsProvider,
-		>(client_state, proof)
+		>(client_state.clone(), proof.clone())
```
seunlanlege, line 117: @blasrodri

code/centauri/light-clients/ics10-grandpa/src/tests.rs
```diff
@@ -207,27 +210,22 @@ async fn test_continuous_update_of_grandpa_client() {
...
+		let header_numbers = ((client_state.latest_para_height + 1)..=finalized_para_header.number)
 			.collect::<Vec<_>>();
 
+		if header_numbers.len() == 0 {
+			continue
+		}
+
 		let proof = prover
 			.query_finalized_parachain_headers_with_proof(
+				&(client_state.clone().into()),
```
seunlanlege, line 227: @blasrodri

code/centauri/algorithms/grandpa/verifier/src/tests.rs
blasrodri: oh, tests :)

PR #1980, opened 2022-10-14
feat(devnet): improve

hussein-aitlahcen:


.github/workflows/build-test-deploy.yml
```diff
@@ -755,6 +759,66 @@ jobs:
       - run: |
           cachix watch-exec --jobs 16 --compression-level $CACHIX_COMPRESSION_LEVEL composable-community nix -- build .#polkadot-node --no-update-lock-file --show-trace -L
 
+  package-statemine-node:
```
dzmitry-lahoda, line 762: you can use matrix for nodes

.github/workflows/build-test-deploy.yml
```diff
@@ -566,7 +566,11 @@ jobs:
   check-simnode:
     name: "Simnode"
     needs: package-simnode-tests
-    runs-on: ubuntu-latest
+    runs-on:
```
dzmitry-lahoda, line 569: speed for running on sre vs gh node for almost exactly same. and that fixed ownership issue. you may have returned back owership. but other stuff is useful

dzmitry-lahoda (approved):

.github/workflows/build-test-deploy.yml
hussein-aitlahcen: If you do that then you can't have dependency on only one node

.github/workflows/build-test-deploy.yml
hussein-aitlahcen: It's currently broken because GH runners have size limitation that our backup exceed

.github/workflows/persistent-devnet.yml
```diff
@@ -96,8 +97,8 @@ jobs:
     - name: Deploy
       run: |
         # For the love of God can anyone tell me why github runners are not stateless.
-        nixops check --include composable-persistent-devnet
-        nixops deploy --allow-recreate --confirm --deployment devnet-gce --debug --show-trace --option narinfo-cache-negative-ttl 0 --include composable-persistent-devnet
+        nixops check --include composable-persistent-devnet || true
```
cor, line 100: i think `|| true` should probably be removed?

cor (changes requested):

.github/workflows/persistent-devnet.yml
hussein-aitlahcen: No it shouldn't because the exit code of check is very opiniated :)

cor (approved):
LGTM 🚀

vimukthi-git (approved):
:rocket:

PR #1992, opened 2022-10-14
Ben/use all instead of reserve for crowdloan locks

benluelo:
changes ::RESERVE to ::all() when locking crowdloan rewards


vimukthi-git (approved):

vimukthi-git:
Checking code it's not entirely clear if this is the reason still, but it's worth the try

PoisonPhang (approved):

vimukthi-git:
@mergifyio refresh

PR #1998, opened 2022-10-17
Ics10 cw v2

blasrodri:
As part of the Cosmos Composable bridge, we're going to package our GRANDPA light client in CosmWasm. There is an ongoing [effort](https://github.com/strangelove-ventures/ibc-go/pull/3). Once our implementation of [ICS-08](https://github.com/cosmos/ibc/tree/main/spec/client/ics-008-wasm-client) is integrated in the `go-sdk` we will be able to bridge into Cosmos using our existing Rust implementations.

This PR is the first of many that will wrap GRANDPA into a CW contract. For now, this one only creates a simple scaffolding structure and upgrades the dependencies between both projects (i.e. `cosmswasm-crypto` and `substrate` -> `ecdsa` crate)

Ref: ComposableFi/centauri#61

to check:

```
cargo +nightly c -p ics10-grandpa-cw --target=wasm32-unknown-unknown --lib
```


code/centauri/light-clients/ics10-grandpa-cw/src/context.rs
```diff
@@ -1,132 +1,60 @@
...
+pub struct Context<'a, H> {
+	deps: DepsMut<'a>,
+	_phantom: PhantomData<H>,
 }
 
-// TODO: validate whether there is any risk in introducing this manually
-// I do not see another way of doing it, since [`DepsMut`] isn't Eq or PartialEq
-impl<'a> PartialEq for Context<'a> {
-	fn eq(&self, other: &Self) -> bool {
+impl<'a, H> PartialEq for Context<'a, H> {
```
blasrodri, line 28: I honestly wonder why do we need this constraint at all upstream

PR #2006, opened 2022-10-17
#nix #security proposing to have 2 signatures on #release branch

dzmitry-lahoda:

## Issue
- adding security to release

## Description
- asks people to sign git tag or note to ensure releasable branch approved

## Possible more security

- git tag or note to contain (nix) hash of wasm so that build proposed wasm for upgrade also can be verified


KaiserKarel (approved):

dzmitry-lahoda:
@Mergifyio refresh

dzmitry-lahoda:
@dzmitry-lahoda ergifyio merge

dzmitry-lahoda:
@mergifyio refresh

RELEASE.md
```diff
@@ -42,3 +42,28 @@ If bug fixes need to be backported to previous releases, a PR through main is al
     - `release-vx.y.1 <- main`
 3. Create a PR to cherry-pick the fix into `release-vx.y.2`.
     - `release-vx.y.2 <- main`
+
+## Security
+
+Tip of the branch from which runtime release is considered should be signed at least by 2 signatures.
+
+These to be owned by Technical committee or by Council members.
```
vimukthi-git, line 50: What does it mean to be owned?

RELEASE.md
```diff
@@ -42,3 +42,28 @@ If bug fixes need to be backported to previous releases, a PR through main is al
     - `release-vx.y.1 <- main`
 3. Create a PR to cherry-pick the fix into `release-vx.y.2`.
     - `release-vx.y.2 <- main`
+
+## Security
+
+Tip of the branch from which runtime release is considered should be signed at least by 2 signatures.
```
vimukthi-git, line 48: Is this some convention in other parachains or arbitrary ?

andyjsbell (approved):
LGTM 👍

RELEASE.md
dzmitry-lahoda: Other chains do not use nix, and cannot do deterministic builds. They are unsecure by design. So they likely talk to each other during release in Matrix chats. 

This proposal allows to make security which is on chain == Security off chain. At least nominally. Other just cannot do it without nix.

RELEASE.md
dzmitry-lahoda: I think in this sense https://www.babypips.com/crypto/learn/what-is-a-digital-signature   .  The only owner of private key can sign by key. So he kind of owns ability to sign. It is shortcut to avoid explaining that you own the key. And if you sign something then it is unique. owner => key => signature === owner => signature

RELEASE.md
dzmitry-lahoda: Look into mention of `crev` in our chats. Recent mentions will be with nix and git tags and dotsama indentity. 

Imho this PR is last missing step in possible security in dotsama.

dzmitry-lahoda: At least nominally :)

vimukthi-git: Without something verifying the signatures automatically for deployment, the security is just an assumption IMO

RELEASE.md
vimukthi-git: Eight the breakup of the paragraph threw me off lol

RELEASE.md
vimukthi-git, line 48:
```suggestion
Tip of the branch from which runtime release is considered should be signed at least by 2 keys owned by Technical committee or by Council members.
```

dzmitry-lahoda: also i proved this works in basic bash example.

RELEASE.md
vimukthi-git:
```suggestion
```

vimukthi-git (approved):

RELEASE.md
vimukthi-git: Just curious, What problem/requirement are we trying to solve anyway?

dzmitry-lahoda: Security as low as its lowest element in chain. We have ability to full multisignature security on chain and in vendoing code. 

But in git we have 0 signature security with ability of Mergifyio or GitHub teams to put anycode they want into our release. So our CI and hosting is not secure at all. 

SRE may like to push secure CI agenda, but that just shift lower security point from them to SRE. The only option to fix that link to multisig.

dzmitry-lahoda: In this case git/CI security is same as multisig on testnet at least.

dzmitry-lahoda: I think signatures is cheaper and safer alternative than building badly working non secured (because centralized) CI. @bostercf Bret do you agree?

dzmitry-lahoda:
@Mergifyio update

RELEASE.md
dzmitry-lahoda: closing until further comments

PR #2020, opened 2022-10-19
[cu-33e4tdu] Attempt#2 Change order of execution of lock

vimukthi-git:
## Description
Changing order of locking to be after the balance has been updated with the new lock amount.

### DIFF
--- a/code/parachain/frame/crowdloan-rewards/src/lib.rs
+++ b/code/parachain/frame/crowdloan-rewards/src/lib.rs
@@ -500,8 +500,8 @@ pub mod pallet {
 						false,
 					)?;
 
-					// IMPORTANT: Order of execution of this lock is important for proper locking of
-					// funds.Refer https://app.clickup.com/t/33e4tdu
+					// IMPORTANT: Order of execution of this lock matters for proper locking of
+					// funds. Refer https://app.clickup.com/t/33e4tdu
 					if T::LockByDefault::get() && !RemoveRewardLocks::<T>::exists() {
 						T::RewardAsset::set_lock(
 							T::LockId::get(),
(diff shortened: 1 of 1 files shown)


ComposableFiB0t:
Task linked: [CU-33e4tdu Crowdloan rewards pallet - config changes (lock PICA)](https://app.clickup.com/t/33e4tdu)

code/parachain/frame/crowdloan-rewards/src/lib.rs
```diff
@@ -490,14 +490,6 @@ pub mod pallet {
...
-						T::RewardAsset::set_lock(
-							T::LockId::get(),
-							reward_account,
-							reward.claimed,
-							WithdrawReasons::all(),
-						);
-					}
 
 					let funds_account = Self::account_id();
 					// No need to keep the pallet account alive.
```
vimukthi-git, line 495: Moving the lock to be after the actual transfer of funds

dzmitry-lahoda (approved):

PoisonPhang:
From ClickUp
> All the above scenarios were tested several time and I can confirm that the balance acting as locked now. 

A new issue has arise now: - after the fist claiming, the new ones require native token that is free to pay GAS in the user's balance. This will block users to claim forward so wo need to address this behavior. 

We either need this Tx to be `Pays::No` by default, or we can't be locking the entirety of their native token. (Unless one of the lock types allows for locked funds to pay for gas).

vimukthi-git:
@PoisonPhang As explained on slack,

Locks have scopes, initially we had the scope as `reserve` then `all` , now I changed it to `transfer`  to allow fee payments. It's well thought out actually

LyonSsS:
Re tested the behavior of this ticket: https://app.clickup.com/t/33e4tdu using the current branch locally. All the issues were fixed and this PR can now be merged.

LyonSsS (approved):

vimukthi-git:
@mergifyio refresh

PR #2021, opened 2022-10-19
#xcmp fixed some dotsama simulator. added rococo

dzmitry-lahoda:

## Issue
* we do not test rococo in simulator, but it hast bugs different from kusama https://substrate.stackexchange.com/questions/5393/is-teleport-of-roc-from-rockmine-to-my-parachain-should-be-used-instead-of-reser
* some tests were broken after some runtime upgrades, fixed these
* some tests failed after fee changed

## Description
* fixed and unified multiasset limit config (each asset increase weigth which not yet calculated until v3)
* fixed weird Balances issue is that free balance minting is not equal free balance after (ED balance is not free)
* added rococo just to be built as deps for now by cachix (it takes 10 mintues to run any test - so prefer nix will cache it :) )
* started more resonable bounding test strategy for fees
* while waiting for rustc added some XCMP links :)

### DIFF
--- a/code/Cargo.lock
+++ b/code/Cargo.lock
@@ -6407,6 +6407,7 @@ dependencies = [
  "polkadot-primitives",
  "polkadot-runtime-parachains",
  "primitives 0.1.0",
+ "rococo-runtime",
  "scale-info",
  "smallvec 1.9.0",
  "sp-api",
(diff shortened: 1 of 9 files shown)


dzmitry-lahoda:
@obsessed-cake @KaiserKarel please take a look

dzmitry-lahoda:
@Mergifyio refresh

obsessed-cake (approved):

dzmitry-lahoda:
@Mergifyio refresh

PR #2035, opened 2022-10-19
fixing #xcmp issue and making simulator tests on rococo and with some assets scenarios

dzmitry-lahoda:
Signed-off-by: Dzmitry Lahoda <dzmitry@lahoda.pro>

## Issue
XCM was not working well, needed more testing and fixes and restoration after previous polkadot crate updates

## Description
- https://drive.google.com/file/d/169Jb6bdJKIGtK0bVqYlQgDSvN0nSfNhq/view?usp=sharing
- made dali run with rococo in sumulator (not kusama)
- made testing of fee much more robust and correct and managable
- improve ratio calculation math
- added more XCM tests
- split some big xcm test, clarified others, made some tests of gradual complexity increase
- clean up from old obsolete comments
- stopped to use manual scale encoding of abis in simulator
- set harcoded ration according future release
- ensured that hardcoded things used everywhere they should be used
- added more tests for dynamic assets 
- split out value for harcoded chains and assets into separate file for reuse
- ensured sending assets to self mapped well
- moved all XCM in direction using native decimals (ensure that we test real price of USDT/6 vs PICA/12)
- dropped xTokens support to make encoding according Dr Gavin Wood guidlines (and consitent simpler asset encoding with Statemine)  https://github.com/open-web3-stack/open-runtime-module-library/discussions/820 
- KSM/AUSD/USDT are harcoded assets

PR #2043, opened 2022-10-20
[CU-3p0r4ry] rpc for fe assets

andresmechali:
## Issue
https://app.clickup.com/t/3p0r4ry

### DIFF
--- a/code/parachain/frame/assets-registry/src/lib.rs
+++ b/code/parachain/frame/assets-registry/src/lib.rs
@@ -30,6 +30,7 @@ pub mod pallet {
 	pub use crate::weights::WeightInfo;
 	use codec::FullCodec;
 	use composable_traits::{
+		assets::Asset,
 		currency::{BalanceLike, CurrencyFactory, Exponent, RangeId},
 		defi::Ratio,
 		xcm::assets::{
@@ -41,10 +42,9 @@ pub mod pallet {
 	use frame_support::{
 		dispatch::DispatchResultWithPostInfo, pallet_prelude::*, traits::EnsureOrigin,
 	};
-
 	use frame_system::pallet_prelude::*;
 	use scale_info::TypeInfo;
-	use sp_std::{fmt::Debug, str};
+	use sp_std::{fmt::Debug, str, vec::Vec};
 
 	/// The module configuration trait.
 	#[pallet::config]
@@ -62,6 +62,7 @@ pub mod pallet {
 			+ Into<u128>
 			+ Debug
 			+ Default
+			+ Ord
 			+ TypeInfo;
 
 		/// Identifier for the class of foreign asset.
@@ -308,6 +309,26 @@ pub mod pallet {
 			remote_asset_id: Self::AssetNativeLocation,
 		) -> Option<Self::Balance> {
 			<MinFeeAmounts<T>>::get(parachain_id, remote_asset_id)
+		}
+
+		fn get_foreign_assets_list() -> Vec<Asset<Self::AssetNativeLocation>> {
+			ForeignToLocal::<T>::iter()
+				.map(|(_, asset_id)| {
+					let foreign_metadata = LocalToForeign::<T>::get(asset_id)
+						.expect("Must exist, as it does in ForeignToLocal");
+					let decimals = match foreign_metadata.decimals {
+						Some(exponent) => exponent,
+						_ => 12_u32,
+					};
+
+					Asset {
+						name: None,
+						id: asset_id.into(),
+						decimals,
+						foreign_id: Some(foreign_metadata.location),
+					}
+				})
+				.collect::<Vec<_>>()
 		}
 	}
 
--- a/code/parachain/frame/composable-traits/src/assets.rs
+++ b/code/parachain/frame/composable-traits/src/assets.rs
@@ -4,6 +4,7 @@ use composable_support::collections::vec::bounded::BiBoundedVec;
 use scale_info::TypeInfo;
 use sp_std::vec::Vec;
 
+use crate::currency::Exponent;
 #[cfg(feature = "std")]
 use serde::{Deserialize, Serialize};
 
@@ -25,7 +26,9 @@ impl BasicAssetMetadata {
 
 #[derive(Decode, Encode, Debug, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
-pub struct Asset {
-	pub name: Vec<u8>,
-	pub id: u64,
+pub struct Asset<ForeignId> {
+	pub name: Option<Vec<u8>>,
+	pub id: u128,
+	pub decimals: Exponent,
+	pub foreign_id: Option<ForeignId>,
 }
--- a/code/parachain/runtime/composable/src/lib.rs
+++ b/code/parachain/runtime/composable/src/lib.rs
@@ -26,8 +26,8 @@ mod weights;
 mod xcmp;
 use common::{
 	impls::DealWithFees, AccountId, AccountIndex, Address, Amount, AuraId, Balance, BlockNumber,
-	Hash, Moment, Signature, AVERAGE_ON_INITIALIZE_RATIO, DAYS, HOURS, MAXIMUM_BLOCK_WEIGHT,
-	MILLISECS_PER_BLOCK, NORMAL_DISPATCH_RATIO, SLOT_DURATION,
+	ForeignAssetId, Hash, Moment, Signature, AVERAGE_ON_INITIALIZE_RATIO, DAYS, HOURS,
+	MAXIMUM_BLOCK_WEIGHT, MILLISECS_PER_BLOCK, NORMAL_DISPATCH_RATIO, SLOT_DURATION,
 };
 use composable_traits::assets::Asset;
 use orml_traits::parameter_type_with_key;
@@ -877,12 +877,12 @@ mod benches {
 }
 
 impl_runtime_apis! {
-	impl assets_runtime_api::AssetsRuntimeApi<Block, CurrencyId, AccountId, Balance> for Runtime {
+	impl assets_runtime_api::AssetsRuntimeApi<Block, CurrencyId, AccountId, Balance, ForeignAssetId> for Runtime {
 		fn balance_of(SafeRpcWrapper(asset_id): SafeRpcWrapper<CurrencyId>, account_id: AccountId) -> SafeRpcWrapper<Balance> /* Balance */ {
 			SafeRpcWrapper(<Assets as frame_support::traits::fungibles::Inspect::<AccountId>>::balance(asset_id, &account_id))
 		}
 
-		fn list_assets() -> Vec<Asset> {
+		fn list_assets() -> Vec<Asset<ForeignAssetId>> {
 			CurrencyId::list_assets()
 		}
 	}
--- a/code/parachain/runtime/dali/src/lib.rs
+++ b/code/parachain/runtime/dali/src/lib.rs
@@ -40,7 +40,7 @@ use common::{
 	},
 	impls::DealWithFees,
 	multi_existential_deposits, AccountId, AccountIndex, Address, Amount, AuraId, Balance,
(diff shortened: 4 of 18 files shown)


ComposableFiB0t:
Task linked: [CU-3p0r4ry RPC/API for Front End Assets](https://app.clickup.com/t/3p0r4ry)

code/parachain/frame/assets-registry/src/lib.rs
```diff
@@ -309,6 +315,24 @@ pub mod pallet {
...
+			LocalToForeign::<T>::iter()
+				.map(|(local_id, foreign_metadata)| {
+					let decimals = match foreign_metadata.decimals {
+						Some(exponent) => exponent,
+						_ => 12_u32,
+					};
+
+					Asset {
+						name: None,
+						id: local_id.into() as u64,
```
benluelo, line 329: .into() followed by `as` is a bit strange, using .into() only should be enough

code/parachain/frame/assets/src/lib.rs
```diff
@@ -116,6 +118,15 @@ pub mod pallet {
 		#[pallet::constant]
 		type NativeAssetId: Get<Self::AssetId>;
 		type GenerateCurrencyId: CurrencyFactory<AssetId = Self::AssetId, Balance = Self::Balance>;
+		/// Identifier for the class of foreign asset.
+		type ForeignAssetId: FullCodec
```
benluelo, line 122: Is this used anywhere in the pallet?

code/parachain/runtime/dali/src/lib.rs
```diff
@@ -1455,13 +1457,17 @@ impl_runtime_apis! {
...
 		}
 
-		fn list_assets() -> Vec<Asset> {
-			CurrencyId::list_assets()
+		fn list_assets() -> Vec<Asset<ForeignAssetId>> {
+			let mut assets = CurrencyId::list_assets();
+			let mut foreign_assets = assets_registry::Pallet::<Runtime>::get_foreign_assets_list();
+			assets.append(&mut foreign_assets);
+
+			assets
```
benluelo, line 1470: This needs to take into account assets that may be in both

benluelo: (otherwise there will be duplicated assets)

code/parachain/frame/assets-registry/src/lib.rs
andresmechali: That's what I though but it's not. `local_id` is a `LocalAssetId` and with `.into()` it turns into a `u128`

code/parachain/runtime/dali/src/lib.rs
andresmechali: Makes sense. But how can we know? Local ids would be different anyway no?

code/parachain/frame/assets-registry/src/lib.rs
benluelo: then you should use `.try_into()` to go from `u128` to `u64`, or change the `Asset.id` type to `u128` - `as` will silently truncate the top 64 bits of `u128` when converting from `u64`

code/parachain/frame/assets-registry/src/lib.rs
```diff
@@ -91,8 +92,13 @@ pub mod pallet {
 	/// Mapping local asset to foreign asset.
 	#[pallet::storage]
 	#[pallet::getter(fn from_local_asset)]
-	pub type LocalToForeign<T: Config> =
-		StorageMap<_, Twox128, T::LocalAssetId, ForeignMetadata<T::ForeignAssetId>, OptionQuery>;
+	pub type LocalToForeign<T: Config> = StorageMap<
+		_,
+		Blake2_128Concat,
```
benluelo, line 97: why was this changed from `Twox128`?

code/parachain/frame/composable-traits/src/assets.rs
```diff
@@ -25,7 +26,9 @@ impl BasicAssetMetadata {
 
 #[derive(Decode, Encode, Debug, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
-pub struct Asset {
-	pub name: Vec<u8>,
+pub struct Asset<ForeignId> {
+	pub name: Option<Vec<u8>>,
 	pub id: u64,
```
benluelo, line 31: This should probably be a generic field instead of a concrete `u64`

code/parachain/frame/assets-registry/src/lib.rs
andresmechali: changed to u128 👌🏼

code/parachain/frame/assets/src/lib.rs
andresmechali: Actually no, I added it at some point and then removed it. Well spotted.

code/parachain/frame/assets-registry/src/lib.rs
andresmechali: `Twox128` doesn't allow me to `iter()` as `ReversibleStorageHasher` is not satisfied.

code/parachain/frame/composable-traits/src/assets.rs
andresmechali: We should be ok with a `u128` no?

code/parachain/runtime/composable/src/lib.rs
```diff
@@ -813,6 +813,7 @@ construct_runtime!(
 		CrowdloanRewards: crowdloan_rewards = 56,
 		Assets: assets = 57,
 		GovernanceRegistry: governance_registry = 58,
+		// AssetsRegistry: assets_registry = 59,
```
dzmitry-lahoda, line 816: why commented out?

code/parachain/runtime/primitives/src/currency.rs
```diff
@@ -89,9 +89,9 @@ macro_rules! list_assets {
 			}
 		}
 
-		pub fn list_assets() -> Vec<Asset> {
+		pub fn list_assets() -> Vec<Asset<XcmAssetLocation>> {
 			[
-				$(Asset { id: CurrencyId::$NAME.0 as u64, name: stringify!($NAME).as_bytes().to_vec() },)*
+				$(Asset { id: CurrencyId::$NAME.0 as u128, name: Some(stringify!($NAME).as_bytes().to_vec()), decimals: 12_u8.into(), foreign_id: None },)*
```
dzmitry-lahoda, line 94: KSM and USD* have foreign asset id too or will have as soon as registered.

dzmitry-lahoda: so it  there should be join of registry with hard coded

code/parachain/frame/assets-registry/src/lib.rs
dzmitry-lahoda: alocal asset id is permissioed, so i  can be 64 bit hash with concat. i am sure there is such. in other examples.

dzmitry-lahoda: twox128 was overkill

dzmitry-lahoda: you could iterate values and get local if by reverse map. foreign to local and local to foreign

dzmitry-lahoda: we have some assets on rococo, may be on picasso. not big deal now. but recreating them is gated by multisig and democracy.

code/parachain/runtime/composable/src/lib.rs
andresmechali: Forgot to remove it

code/parachain/frame/assets-registry/src/lib.rs
andresmechali: `Twox64Concat` works as well, but still has the problem of changing storage. Should I use that, or go for the `ForeignToLocal-LocalToForeign` double iteration?

benluelo: RPCs have no on-chain cost, so it should be fine to do the double iteration

code/parachain/frame/composable-traits/src/assets.rs
benluelo: Yes, but it would probably be best to make it generic. This can be addressed in another ticket though, feel free to resolve :+1:

code/parachain/runtime/dali/src/lib.rs
benluelo: @dzmitry-lahoda left a comment relating to this as well, I think he may have more insight into how to properly handle this than I do

code/parachain/runtime/primitives/src/currency.rs
andresmechali: @dzmitry-lahoda  What can we use to identify already existing assets, given that when we register we are not giving any input on `local_id` or `name` or `symbol`?

code/parachain/frame/composable-traits/src/assets.rs
andresmechali: https://app.clickup.com/t/3mzpfw2

code/parachain/runtime/primitives/src/currency.rs
dzmitry-lahoda: Encoding works like that https://drive.google.com/file/d/169Jb6bdJKIGtK0bVqYlQgDSvN0nSfNhq/view  and can see some examples https://github.com/ComposableFi/composable/pull/2035#issuecomment-1288758296

dzmitry-lahoda: will update after check

andresmechali: Thanks @dzmitry-lahoda , it helped me understanding better xcm.
Can you provide me with a concrete example on how a `MultiLocation` would look like when we register on our chain for example `USDT` (which locally has id `130`) , which let's assume has id `111` on some parachain with id `222`?

code/parachain/runtime/dali/src/lib.rs
andresmechali: @benluelo as far as I understand, there will be no duplicated assets. Registering an asset requires sudo, and nobody will register one that is already hardcoded.

dzmitry-lahoda (approved):

code/parachain/runtime/dali/src/lib.rs
dzmitry-lahoda: these are hardcode in PR for XCM. I can fix this RPC if you will at least hardcode KSM = (1,Here).into()

vimukthi-git (approved):
LGTM

andresmechali:
@mergify refresh

andresmechali:
@mergify refresh

easteregg (approved):

saadjhk (approved):

andresmechali:
@mergify refresh

hussein-aitlahcen (approved):

code/parachain/runtime/primitives/src/currency.rs
```diff
@@ -89,9 +89,24 @@ macro_rules! list_assets {
 			}
 		}
 
-		pub fn list_assets() -> Vec<Asset> {
+		const RELAY_NATIVE_CURRENCIES: [CurrencyId; 1] = [CurrencyId::KSM];
```
benluelo, line 92: Does this (and known_location) need to be in the macro?

benluelo (approved):

code/parachain/runtime/primitives/src/currency.rs
andresmechali: Not exactly, it made sense for me as it was only used there but it can be extracted if it makes more sense

andresmechali:
@mergify refresh

code/parachain/runtime/primitives/src/currency.rs
andresmechali: I will resolve this to get it merged asap and can extract it later on some other PR

PR #2060, opened 2022-10-21
Rm build from hyperspace

blasrodri:
## Issue

## Description

## Checklist


blasrodri:
@seunlanlege can you please confirm that this is what you were referring to?
https://github.com/ComposableFi/composable/pull/2065#pullrequestreview-1150880249

seunlanlege:
Yes it's similar, but in my comment we don't need polkadot launch running in the background. You only need the built composable node. So the cli is on the Composable node.

blasrodri:
> Yes it's similar, but in my comment we don't need polkadot launch running in the background. You only need the built composable node. So the cli is on the Composable node.

For simplicity sake, can we just accept this solution for now and unblock us? I'm happy to discuss an improvement - but I don't even know where to start if it's adding a CLI for the composable node to be honest.

I've tested it using `docker run --platform linux/amd64 --rm -ti -u1000:1000 -p9944:9944 -p9188:9188 -p9988:9988 composablefi/composable-centauri:latest`

seunlanlege:
> For simplicity sake, can we just accept this solution for now and unblock us? I'm happy to discuss an improvement - but I don't even know where to start if it's adding a CLI for the composable node to be honest.

We can discuss this on monday.

PR #2065, opened 2022-10-21
hyperspace-parachain: trigger subxt_codegen conditionally

blasrodri:
Only trigger the subxt code generation in case it's requested through an environment variable (SUBXT_CODEGEN).

Tried getting rid of the build.rs file directly here: https://github.com/ComposableFi/composable/pull/2060

However, David pointed out that because `hyperspace`, `hyperspace-parachain` and the `parachain` all are
different binaries their `OUT_DIR` will differ. We want to ensure that the generated code isn't removed between steps.

By doing this _hack_, we are able to generate the code only when we need to (running the integration-tests for instance). But this isn't needed
when compiling or checking the entire project for other reasons (running unit tests for example).

### DIFF
--- a/code/centauri/hyperspace/parachain/build.rs
+++ b/code/centauri/hyperspace/parachain/build.rs
@@ -13,9 +13,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::env;
+
 #[tokio::main]
 async fn main() -> color_eyre::Result<()> {
-	subxt_codegen::build_script("ws://127.0.0.1:9944", "polkadot").await?;
-	subxt_codegen::build_script("ws://127.0.0.1:9188", "parachain").await?;
+	// only trigger the subxt code generation in case it's requested
+	// through an environment variable.
+	if env::var("SUBXT_CODEGEN").is_ok() {
+		subxt_codegen::build_script("ws://127.0.0.1:9944", "polkadot").await?;
+		subxt_codegen::build_script("ws://127.0.0.1:9188", "parachain").await?;
+	}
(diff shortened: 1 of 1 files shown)


code/centauri/hyperspace/parachain/build.rs
```diff
@@ -13,9 +13,15 @@
...
 
+use std::env;
+
 #[tokio::main]
 async fn main() -> color_eyre::Result<()> {
-	subxt_codegen::build_script("ws://127.0.0.1:9944", "polkadot").await?;
-	subxt_codegen::build_script("ws://127.0.0.1:9188", "parachain").await?;
+	// only trigger the subxt code generation in case it's requested
+	// through an environment variable.
+	if env::var("SUBXT_CODEGEN").is_ok() {
```
dzmitry-lahoda, line 22: we thought about `if !args.parse().NIX { codegen(); }

blasrodri: That's nix syntax?
And if so, you'd be calling the rust functions from there through a script?

seunlanlege (changes requested):
I'd prefer if we instead added a cli command that writes the node Metadata to a file, and use subxt to generate the file to the utils folder.

In CI we can assert that there's no diff between what exists there and the newly written file.

This solves this issue once and for all

code/centauri/hyperspace/parachain/build.rs
dzmitry-lahoda: not nix. it is that disabling generation from ws via NIX var is good.

PR #2073, opened 2022-10-24
Change misbehavior report mechanism for Grandpa IBC client

vmarkushin:
Instead of requiring `Equivocation`, we now require two finality proofs, one of which is for a forked chain

### DIFF
--- a/code/centauri/light-clients/ics10-grandpa/src/client_def.rs
+++ b/code/centauri/light-clients/ics10-grandpa/src/client_def.rs
@@ -107,10 +107,26 @@ where
 				let first_header = first_proof.unknown_headers.first().ok_or_else(|| {
 					Error::Custom("No headers in first finality proof".to_string())
 				})?;
-
 				let second_header = second_proof.unknown_headers.first().ok_or_else(|| {
 					Error::Custom("No headers in second finality proof".to_string())
 				})?;
+
+				let first_last_header = first_proof
+					.unknown_headers
+					.last()
+					.expect("first finality proof has at least one header; qed");
+				let second_last_header = second_proof
+					.unknown_headers
+					.last()
+					.expect("second finality proof has at least one header; qed");
+
+				if first_last_header != first_proof.block ||
+					second_last_header != second_proof.block
+				{
+					return Err(Error::Custom(
+						"Misbehaviour proofs are not for the same chain".into(),
+					))
+				}
 
 				let first_parent = first_header.parent_hash;
 				let second_parent = second_header.parent_hash;
@@ -144,7 +160,6 @@ where
 					&mut &first_proof.justification[..],
 				)
 				.map_err(|_| Error::Custom("Could not decode first justification".to_string()))?;
-
 				let second_justification = GrandpaJustification::<RelayChainHeader>::decode(
 					&mut &second_proof.justification[..],
 				)
@@ -165,6 +180,7 @@ where
 				let second_valid = second_justification
 					.verify::<H>(client_state.current_set_id, &client_state.current_authorities)
 					.is_ok();
+
 				if !first_valid || !second_valid {
 					Err(Error::Custom("Invalid justification".to_string()))?
 				}
--- a/code/centauri/light-clients/ics10-grandpa/src/client_def.rs
+++ b/code/centauri/light-clients/ics10-grandpa/src/client_def.rs
@@ -101,7 +101,9 @@ where
 				let second_proof = misbehavior.second_finality_proof;
 
 				if first_proof.block == second_proof.block {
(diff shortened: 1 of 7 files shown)


code/centauri/light-clients/ics10-grandpa/src/client_def.rs
```diff
@@ -94,24 +97,76 @@ where
...
+
+				let second_header = second_proof.unknown_headers.first().ok_or_else(|| {
+					Error::Custom("No headers in second finality proof".to_string())
+				})?;
+
+				if first_header.number != second_header.number {
+					Err(Error::Custom(
+						"First and second finality proofs are for different blocks".to_string(),
+					))?
+				}
```
seunlanlege, line 118: nahhh, they can be for diffferent blocks

seunlanlege: 1. Check that the header with the lowest height (`unknown_headers.iter().min_by_key(|h| h.number())`) for both finality proofs exist on-chain.
2. assert that they both describe two distinct chains
3. verify that both justifications are valid

vmarkushin: Yeah, forgot to remove

code/parachain/frame/ibc/src/light_clients.rs
```diff
@@ -76,19 +84,73 @@ impl tendermint_light_client_verifier::host_functions::HostFunctionsProvider
...
-	fn add_relaychain_headers(header: &Self::Header) {
-		todo!()
+	fn add_relaychain_headers(headers: &[Self::Header]) {
+		if headers.is_empty() {
+			return
+		}
+
+		for header in headers {
+			let hash = header.hash();
+			GrandpaBlockHashStorage::insert(header.number, hash);
```
seunlanlege, line 126: we don't need this

code/centauri/light-clients/ics10-grandpa/src/client_def.rs
```diff
@@ -94,24 +97,73 @@ where
...
+						if header.hash() != last_hash {
+							return Err(Error::Custom("Chain is not canonical".to_string()))
+						}
+						last_hash = header.parent_hash;
+					}
+					Ok(())
+				};
+
+				check_canonicity(&first_proof.unknown_headers, first_parent)?;
+				check_canonicity(&second_proof.unknown_headers, second_parent)?;
```
blasrodri, line 126: [nit] an alternative is to accept the two proofs in the `check_canonicity`

code/parachain/frame/ibc/src/light_clients.rs
```diff
@@ -76,19 +84,73 @@ impl tendermint_light_client_verifier::host_functions::HostFunctionsProvider
...
+
+		GrandpaHeaderHashesSetStorage::mutate(|hashes_set| {
+			GrandpaHeaderHashesStorage::mutate(|hashes| {
+				for hash in new_hashes {
+					match hashes.try_push(*hash) {
+						Ok(_) => {},
+						Err(_) => {
+							let old_hash = hashes.remove(0);
+							hashes_set.remove(&old_hash);
+							let _ = hashes.try_push(*hash);
```
blasrodri, line 138: why are you ignoring the error?

code/centauri/light-clients/ics10-grandpa/src/client_def.rs
blasrodri: `check_canonicity` could accept a slice of proofs for instance instead of calling it twice. But it's a minor thing, you choose :)

code/parachain/frame/ibc/src/light_clients.rs
```diff
@@ -76,19 +84,75 @@ impl tendermint_light_client_verifier::host_functions::HostFunctionsProvider
 
 impl ics07_tendermint::HostFunctionsProvider for HostFunctionsManager {}
 
+pub struct GrandpaHederHashesStorageInstance;
+impl StorageInstance for GrandpaHederHashesStorageInstance {
```
Wizdave97, line 88: why do we need to storage instances for this?

code/centauri/light-clients/ics10-grandpa/src/client_def.rs
Wizdave97: That might make the function less succint.

code/parachain/frame/ibc/src/light_clients.rs
vmarkushin: because I don't use substrate's storage macro here

blasrodri (approved):

PR #2076, opened 2022-10-24
adds cli for generating runtime metadata

seunlanlege:
Use it like so: 

```bash
 ./code/target/release/composable metadata --chain=dali-dev > ./dali.metadata
 subxt codegen --file=dali.metadata | rustfmt --edition=2018 --emit=stdout > dali.rs
```

### DIFF
--- a/code/centauri/hyperspace/parachain/src/lib.rs
+++ b/code/centauri/hyperspace/parachain/src/lib.rs
@@ -84,7 +84,7 @@ pub struct ParachainClient<T: subxt::Config> {
 	pub client_id: Option<ClientId>,
 	/// Connection Id
 	pub connection_id: Option<ConnectionId>,
-	/// Commitment prefix
+	/// ICS-23 provable store commitment prefix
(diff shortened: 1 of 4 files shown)


blasrodri (approved):

blasrodri:
@cor

blasrodri:
@seunlanlege how can we get `polkadot`? For that we need to use `ws`?

seunlanlege:
> @seunlanlege how can we get polkadot? For that we need to use ws?

Yes but that changes less often

blasrodri:
> > @seunlanlege how can we get polkadot? For that we need to use ws?
> 
> Yes but that changes less often

how do you suggest we run the polkadot part in our CI?

seunlanlege:
> @seunlanlege how can we get polkadot? For that we need to use ws?

It'll be generated when we upgrade our polkadot version manually

dzmitry-lahoda:
@blasrodri we can run it each time in nix. if dep did not changed, so it will use cache. so you do not write if changed then run. it just runs.

PR #2079, opened 2022-10-24
Subxt exports

blasrodri:
`subxt` code generated is used by multiple crates. This PR moves the logic into its own crate, so that it can be referred by others as a dependency.


dzmitry-lahoda:
It seems will break work of people if they regenerate dali api on build. i would prefer this crate retains include out for now. I will write for nix `if nix, use other stuff` . and only final step handle work final state of what people prefere in build.rs

so can we retain exactly same build.rs and exactly same include, but in new crate? only difference is that hyperspace referenes this crate instead of including code directly?

PR #2083, opened 2022-10-24
[CU-3q301nu] Get ED from CurrencyFactory

PoisonPhang:
## Issue
[[CU-3q301nu]](https://app.clickup.com/t/3q301nu)

## Description
Given the interest in eventually removing currency factory and to minimize refactoring, this is routed through asset registry before going through currency factory.

## Checklist
- [X] Get ED from Currency Factory / Asset Registry
- [X] Add new hard-coded ED values
- [x] Write dev guide/docs for creating/adding ED values
- [x] Resolve merge conflicts


ComposableFiB0t:
Task linked: [CU-3q301nu [TRIAGE] Picasso/Transfers - Registered foreign assets existential deposits doesn't affect account removal](https://app.clickup.com/t/3q301nu)

code/integration-tests/local-integration-tests/src/helpers.rs
```diff
@@ -20,13 +20,14 @@ pub fn para_account_id(id: u32) -> AccountId {
...
+	AssetsRegistry: AssetRatioInspect<AssetId = CurrencyId>
+		+ AssetExistentialDepositInspect<AssetId = CurrencyId, Balance = Balance>,
+>(
 	asset_id: LocalAssetId,
 	_instruction_count: usize,
 ) -> Balance {
-	PriceConverter::<AssetsRegistry>::get_price_inverse(asset_id, NativeExistentialDeposit::get())
-		.unwrap() /
-		Balance::from(2_u128)
+	multi_existential_deposits::<AssetsRegistry>(&asset_id) / 2_u128
```
vimukthi-git, line 30: What's the magic number 2 here?

code/parachain/runtime/common/src/lib.rs
```diff
@@ -217,21 +219,24 @@ pub fn multi_existential_deposits<AssetsRegistry>(_currency_id: &CurrencyId) ->
 	Balance::zero()
 }
 
+/// Given a `currency_id`, returns the existential deposit of a MultiAsset.
+/// Returns `1_000_000` as the existential deposit if unable to get an existential deposit for the
```
vimukthi-git, line 223: Does it not make sense to just error out for non-existing ED?

code/integration-tests/local-integration-tests/src/helpers.rs
```diff
@@ -50,23 +51,19 @@ pub fn sibling_account() -> AccountId {
...
+	assert!(multi_existential_deposits::<AssetsRegistry>(&asset_id) <= amount);
 	amount
 }
 
 /// weigh enough to handle any XCMP message
 pub fn enough_weight() -> u128 {
-	let this_liveness_native_amount = BaseXcmWeight::get() as u128 +
-		100 * UnitWeightCost::get() as Balance * MaxInstructions::get() as Balance;
-	this_liveness_native_amount
+	BaseXcmWeight::get() as u128 +
```
vimukthi-git, line 67: Can we explain where this formula comes from?

code/integration-tests/local-integration-tests/src/helpers.rs
dzmitry-lahoda: it ensures the name of method to be true 'under_existential_deposit'

code/integration-tests/local-integration-tests/src/helpers.rs
dzmitry-lahoda: from xcm configuration

code/parachain/runtime/common/src/lib.rs
dzmitry-lahoda: no. it should be fixed to tranfer ksm and usdt.

rfcs/0009-asset-registry-existential-deposit-workflow.md
```diff
@@ -0,0 +1,100 @@
...
+our native token (`0.1` or `100_000_000_000`) and convert that into the 
+equivilant value in the necessary token useing 
+`PriceConverter::get_price_inverse`.
+
+Instead of the current implimentation we want to resolve the ED of a token with 
+the following alogrithim:
+
+  Given some token ID, `currency_id`
+
+  1. Querey Currency Factory for the ED of `currency_id`
```
vimukthi-git:
```suggestion
  1. Query Currency Factory for the ED of `currency_id`
```

rfcs/0009-asset-registry-existential-deposit-workflow.md
```diff
@@ -0,0 +1,100 @@
...
+---
+
+This document is a review of the implementation and usage of Existential Deposit
+values within our runtimes. This will review this maintainmance of ED values
+from a code and SUDO/Governance prospective.
+
+## Background
+---
+
+Existential Deposits are the minimum value required of a balance to exists on
```
vimukthi-git, line 13: A link to what "Existential Deposits" mean would be nice here :)

vimukthi-git:
```suggestion
Existential Deposits are the minimum balance required of an account to exist on
```

rfcs/0009-asset-registry-existential-deposit-workflow.md
```diff
@@ -0,0 +1,100 @@
...
+## Requirements
+---
+
+* Be able to define different Existential Deposit values for different
+  tokens.
+* Have a predefined collection of well-known tokens that each have their
+  own Existential Deposit.
+* SUDO/Governance must be able to overwrite the predefined Existential Deposit
+  values without modifying the code.
+* Be able to add Existential Deposit for new tokens reconized by the chain.
```
vimukthi-git: I think this requirement is already covered by the first one.

rfcs/0009-asset-registry-existential-deposit-workflow.md
```diff
@@ -0,0 +1,100 @@
...
+Existential Deposits are the minimum value required of a balance to exists on
+our chains. Given that different tokens will have different values, its 
+necessary to have different Existential Deposit values for different tokens.
+
+## Requirements
+---
+
+* Be able to define different Existential Deposit values for different
+  tokens.
+* Have a predefined collection of well-known tokens that each have their
```
vimukthi-git, line 22: I think this is a technicality of the implementation to not have to define initial set, rather than a requirement, so we can possibly remove

rfcs/0009-asset-registry-existential-deposit-workflow.md
```diff
@@ -0,0 +1,100 @@
...
+
+Existential Deposits are the minimum value required of a balance to exists on
+our chains. Given that different tokens will have different values, its 
+necessary to have different Existential Deposit values for different tokens.
+
+## Requirements
+---
+
+* Be able to define different Existential Deposit values for different
+  tokens.
```
vimukthi-git, line 21:
```suggestion
  tokens in the native currency.
```

rfcs/0009-asset-registry-existential-deposit-workflow.md
```diff
@@ -0,0 +1,100 @@
...
+
+```rust
+pub fn multi_existential_deposits<
+	AssetsRegistry: AssetRatioInspect<AssetId = CurrencyId>
+		+ AssetExistentialDepositInspect<AssetId = CurrencyId, Balance = Balance>,
+>(
+	currency_id: &CurrencyId,
+) -> Balance {
+	AssetsRegistry::existential_deposit(*currency_id)
+		.and_then(|ed| PriceConverter::<AssetsRegistry>::get_price_inverse(*currency_id, ed))
```
vimukthi-git, line 86: Can we add more detail on the conversion aspect? What algos are used and how those are updated?

vimukthi-git (commented):
Great thinking to add an RFC to cover this very important aspect. Thanks :)

rfcs/0009-asset-registry-existential-deposit-workflow.md
```diff
@@ -0,0 +1,100 @@
...
+
+  Given some token ID, `currency_id`
+
+  1. Query Currency Factory for the ED of `currency_id`
+
+  2. If not found in 1, attempt to match `currency_id` to our hard-coded ED 
+     values
+    
+  3. If not found in 1 or 2, return the maximum value of the `Balance` type so
+     that unknown assets balances are not tracked
```
vimukthi-git, line 72:
```suggestion
     that unknown assets balances are not tracked as ED would be an impossible amount to have in an account.
```

rfcs/0009-asset-registry-existential-deposit-workflow.md
PoisonPhang: This function deserves its own RFC :sweat_smile: 
I'll try and give better details on it here though.

dzmitry-lahoda (approved):

.config/dictionaries/names.dic
```diff
@@ -68,6 +68,7 @@ Whelan
 privi
 muzmatch
 biconomy
+Bitfrost
```
dzmitry-lahoda, line 71: biFrost

dzmitry-lahoda (approved):

rfcs/0009-asset-registry-existential-deposit-workflow.md
```diff
@@ -0,0 +1,100 @@
...
+pub fn multi_existential_deposits<
+	AssetsRegistry: AssetRatioInspect<AssetId = CurrencyId>
+		+ AssetExistentialDepositInspect<AssetId = CurrencyId, Balance = Balance>,
+>(
+	currency_id: &CurrencyId,
+) -> Balance {
+	AssetsRegistry::existential_deposit(*currency_id)
+		.and_then(|ed| PriceConverter::<AssetsRegistry>::get_price_inverse(*currency_id, ed))
+		.unwrap_or(match *currency_id {
+			CurrencyId::USDT => 1492,
```
vimukthi-git, line 88: seems we are matching converted values here for each currency. I don't see the reason why any one of these would have a different ED after being converted to native currency. Could you explain?

dzmitry-lahoda: yeas, indeed. It is possible to use ratios defined in BYOG. and map amount of ED for that formulas. ED -> ration -> ED or FEE -> ratio -> FEE. Same ratio coverter

Martin-Kiss (approved):

rfcs/0009-asset-registry-existential-deposit-workflow.md
PoisonPhang: Additionally, these assets aren't in CF/AR and need to default to something without allowing for all assets (including unknown assets) to also default. Also a minor performance bump

vimukthi-git (approved):
LGTM, I would still revisit this topic to clarify
1. Why native existential deposit is different in value to other asset existential deposits?
2. Can we avoid hard coding any ED by just calculating ED for every asset based on the native ED and the asset ratio between an asset and the native currency?

PR #2099, opened 2022-10-25
[CU-32bkw0g] Move RFCs to docs

Martin-Kiss:
## Issue
https://app.clickup.com/t/32bkw0g

## Description
RFCs should be part of docs.


vimukthi-git:
Why?

ComposableFiB0t:
Task linked: [CU-32bkw0g move RFCs](https://app.clickup.com/t/32bkw0g)

Martin-Kiss:
> Why?

Because having 'docs' generalized as the home for most documentation is intuitive (to the point I've been asked why this was not the case for RFCs). 
There have also been complaints in the past about not being able to find specific pieces of documentation, which this approach solves for most cases. 
So anyone can be sure to find whatever they're looking for within the `docs` folder or in the respective components directory.

PR #2118, opened 2022-10-26
#CU-2ynbwg5 #CU-31gy38v #xcm  #xcmp security and testing hardening

dzmitry-lahoda:
## Description
image

- overview of xcmp runtime https://drive.google.com/file/d/1dnNDDhb2UxZ0HvkfH4GwvZud2Sk5iDbd/view?usp=sharing for audit on this PR
- PR walkthrough https://drive.google.com/file/d/1_VnZu8Boz5mmLg3tPgTi6dqlYHIPGkK3/view?usp=sharing (sorry for low res - who knows how to meet record big?)
- static assets stuff https://github.com/ComposableFi/composable/pull/2231

## Fixes

- prevent root execute from relay 😱
- prevent raw send from us (we should not spam)⛓
- make multi ED according ratios unified (made prices/fees easy to manage and tested) ☯️
- assets location added to RPC
- outgoing xcm fees set right (like in Acala)
- ED is now can be updated via trait

## Support for fixes

- more tests in simulator
- removed nonsense like this from our test codebase inherited from acala https://github.com/bifrost-finance/bifrost/blob/1e4ffdbf4ac4a89d70084ae0ec7d7c5ec270269e/integration-tests/bifrost-kusama/src/kusama_cross_chain_transfer.rs#L98
- start building composable and polkadot in simulator  
- made Exponent u8 as in Parity Assets (reasonable) 🦀
- used fractional as ratio instead of fixed 🔢 
- removed obsolete docs and other docs clean up 📄

## Next PR

- https://github.com/ComposableFi/composable/pull/2168

## Checklist

- on devnet, transfer KSM back and forth to relay
- on testnet, transfer USDT from Rockmine back and forth
- if Karura sends messages out of it (and only iff messages are moved from it to HRMP) - can try kUSD transfer
- try transfer from Dali to Dali on devnet with fresh registered tokens


ComposableFiB0t:
Task linked: [CU-2ynbwg5 Cross chain asset(s) registry](https://app.clickup.com/t/2ynbwg5)

docs/docs/internal/xcmp/xcm-examples.md
```diff
@@ -2,14 +2,7 @@
...
 
-General flow for XCM is:
-
-1. Open channels
-2. Register foreign assets or from genesis
-3. Obtain assets (mint or swap on DEX)
-4. Execute XCM transactions
-
-If any of these steps is missing, parachain XCM messages will not work. 
+For general flow on how to make XCM and transfer happen, look [here](./ping.plantuml)
```
Martin-Kiss, line 5:
```suggestion
 For details on the general workflow of XCM and transfers [look here](./ping.plantuml)
```
Nit expression

Martin-Kiss (approved):

code/parachain/runtime/common/src/fees.rs
```diff
@@ -0,0 +1,162 @@
...
+pub struct WellKnownPriceConverter;
+
+impl WellKnownPriceConverter {
+	pub fn get_ratio(asset_id: CurrencyId) -> Option<Rational64> {
+		match asset_id {
+			CurrencyId::KSM => Some(rational!(375 / 1_000_000)),
+			CurrencyId::ibcDOT => Some(rational!(2143 / 1_000_000)),
+			CurrencyId::USDT | CurrencyId::USDC => Some(rational!(15 / 1_000_000_000)),
+			CurrencyId::aUSD | CurrencyId::kUSD => Some(rational!(15 / 1_000)),
+			CurrencyId::PICA => Some(rational!(1 / 1)),
```
vimukthi-git, line 80: I don't see why this stuff is in the "common", it should be runtime specific.

vimukthi-git: Or configurable at least

vimukthi-git: Best place is probably the assets-registry it's self

code/parachain/runtime/common/src/fees.rs
```diff
@@ -0,0 +1,162 @@
...
+			CurrencyId::PBLO => Some(rational!(1 / 1)),
+			_ => None,
+		}
+	}
+
+	pub fn existential_deposit(asset_id: CurrencyId) -> Option<Balance> {
+		Self::to_asset_balance(NATIVE_EXISTENTIAL_DEPOSIT, asset_id)
+	}
+
+	pub fn to_asset_balance(fee: NativeBalance, asset_id: CurrencyId) -> Option<Balance> {
```
vimukthi-git, line 90:
```suggestion
	pub fn to_asset_balance(balance: NativeBalance, asset_id: CurrencyId) -> Option<Balance> {
```

vimukthi-git (changes requested):
Common code(`code/parachain/runtime/common`) assuming PICA is the native currency is not correct.

Please let's avoid big PRs like this in the future that mixes critical fixes with refactorings/documentation etc. 
Ideally this PR could have been just about either one of,
```
- prevent root execute from relay
- prevent raw send from us (we should not spam)chains
- make multi ED according ratios unified (made prices/fees easy to manage and tested) 
- more tests in simulator
```
For a reviewer it's super hard to see which thing is modified and tested properly as of now.

code/parachain/runtime/common/src/fees.rs
dzmitry-lahoda: it is common for dali and for picasso. next improvements will be done by using consts overrides and/or trait with impl overrides. also there is idea to have other runtime without relay - so imho it is good enough. 

this code hard to move to pallet as of now.

dzmitry-lahoda:
talked with @andresmechali about tests

code/parachain/runtime/common/src/xcmp.rs
```diff
@@ -1,16 +1,15 @@
 //! proposed shared XCM setup parameters and impl
```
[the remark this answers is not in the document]
dzmitry-lahoda: opinionated. these are rewards for block authors.

code/parachain/runtime/common/src/fees.rs
dzmitry-lahoda: is it warning or something? 

imho having fee here indicates better what this conversion is about.

dzmitry-lahoda:
@vimukthi-git I cannot develop this stuff according your suggestion. it neither technically possible nor by current timings of builds/releases/cis/testing procedures.  if you think we can improve let talk with @andyjsbell

dzmitry-lahoda:
@vimukthi-git what exact changes do you want to do?

code/parachain/frame/composable-traits/src/currency.rs
```diff
@@ -142,13 +154,93 @@ impl<
...
+		}
+	};
+}
+
+impl_const_get!(ConstRational64, Rational64);
+
+impl Rational64 {
+	pub fn new(numer: u64, denom: u64) -> Self {
+		Rational64 { numer, denom }
+	}
```
PoisonPhang, line 213: For this an all other creations of `Rational*` types, we should be checking that `denom` is never `0`.

code/parachain/runtime/common/src/fees.rs
PoisonPhang: I agree that these conversions will eventually need to be specific to the runtime/native-token, but we are already giving Asset Registry priority in providing our converted ED when we first call `	AssetsRegistry::existential_deposit(*currency_id)` in `fn multi_existential_deposits`

code/parachain/runtime/common/src/fees.rs
```diff
@@ -0,0 +1,162 @@
...
+			CurrencyId::aUSD | CurrencyId::kUSD => Some(rational!(15 / 1_000)),
+			CurrencyId::PICA => Some(rational!(1 / 1)),
+			CurrencyId::PBLO => Some(rational!(1 / 1)),
+			_ => None,
+		}
+	}
+
+	pub fn existential_deposit(asset_id: CurrencyId) -> Option<Balance> {
+		Self::to_asset_balance(NATIVE_EXISTENTIAL_DEPOSIT, asset_id)
+	}
```
PoisonPhang, line 88: Can we get some cargo docs for this function that explain when to use this over `multi_existential_deposits`

vimukthi-git:
@dzmitry-lahoda 

> it neither technically possible nor by current timings of builds/releases/cis/testing procedures

I don't agree that it's not possible to do technically (We are doing this regularly in our team). Timelines will be managed according to the need. The amount of time spent trying to redo things because of not having good design/review justifies breaking things up.

code/parachain/runtime/common/src/fees.rs
vimukthi-git: reusable function's parameters need not be named after a specific use case. Then you can rename the function,
```suggestion
	pub fn to_fee_in_asset(fee: NativeBalance, asset_id: CurrencyId) -> Option<Balance> {
```

code/parachain/runtime/common/src/xcmp.rs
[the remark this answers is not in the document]
dzmitry-lahoda: neither it is about fees. one may but would not do as nobody does. please use common sense.

code/parachain/runtime/common/src/fees.rs
dzmitry-lahoda: sure can do it later. right now dali and picasso use it as common.

code/parachain/runtime/common/src/fees.rs
dzmitry-lahoda: weird, it implements trait from substrate. i cannot do it.

code/parachain/runtime/common/src/fees.rs
dzmitry-lahoda: this should not be used directly. it should be wrapped into decision use this or runtime - as that is the case.

dzmitry-lahoda:
@vimukthi-git please elaborate `I don't agree that it's not possible to do technically (We are doing this regularly in our team)` Example of change you can do it separately?

dzmitry-lahoda:
@vimukthi-git please elaborate `I don't agree that it's not possible to do technically (We are doing this regularly in our team)` Example of change you can do it separately?

code/parachain/runtime/common/src/xcmp.rs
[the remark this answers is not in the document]
dzmitry-lahoda: on can make fees 0 - it is not about fees there. @vimukthi-git please stop doing https://en.wikipedia.org/wiki/List_of_fallacies#Red_herring_fallacies in review :)

code/parachain/runtime/picasso/src/xcmp.rs
```diff
@@ -240,28 +213,20 @@ impl orml_xtokens::Config for Runtime {
 	type CurrencyIdConvert = AssetsIdConverter;
 	type AccountIdToMultiLocation = AccountIdToMultiLocation;
 	type SelfLocation = topology::this::Local;
+	type MinXcmFee = ParachainMinFee;
 	type XcmExecutor = XcmExecutor<XcmConfig>;
+	type MultiLocationsFilter = Everything;
```
vimukthi-git, line 218: I think we need to limit this at the start to the chains we want, like statemine and relay

code/parachain/runtime/common/src/xcmp.rs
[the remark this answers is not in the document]
vimukthi-git: > please stop doing https://en.wikipedia.org/wiki/List_of_fallacies in review

Don't see the relevance. Naming things non-ambiguously is not a black art to argue about.

code/parachain/runtime/common/src/fees.rs
vimukthi-git: Which trait is that sir?

code/parachain/runtime/picasso/src/xcmp.rs
```diff
@@ -67,10 +63,8 @@ parameter_types! {
 }
 
 pub type Barrier = (
-	XcmpDebug,
-	//DebugAllowUnpaidExecutionFrom<WellKnownsChains>,
 	AllowKnownQueryResponses<RelayerXcm>,
-	AllowSubscriptionsFrom<Everything>,
+	AllowSubscriptionsFrom<ParentOrSiblings>,
 	AllowTopLevelPaidExecutionFrom<Everything>,
```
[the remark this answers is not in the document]
dzmitry-lahoda: AllowTopLevelPaidExecutionFrom is only Relay. Acala and Statemine allow messages from Relay to go through paywall barrier if it is payed. So it is good default. 

Making it more strict is possible, but it requires separate research item proven by test runs.

code/parachain/runtime/common/src/xcmp.rs
```diff
@@ -383,3 +340,35 @@ impl<X, Y, Treasury: TakeRevenue, Z> Drop for TransactionFeePoolTrader<X, Y, Tre
...
+		Registry: RemoteAssetRegistryInspect<
+			AssetId = CurrencyId,
+			AssetNativeLocation = XcmAssetLocation,
+			Balance = Balance,
+		>,
+	> OutgoingFee<Registry>
+{
+	pub fn outgoing_fee(location: &MultiLocation) -> Option<Balance> {
+		match (location.parents, location.first_interior()) {
+			(1, None) => Some(400_000_000_000),
```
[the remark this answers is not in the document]
dzmitry-lahoda: Will recheck and update. At some point it was from Acala to ensure user outgoing XCM TX work ok on other side. So magic constants are by design now https://composablefinance.slack.com/archives/C031G5NT0CA/p1667043903048109

code/parachain/runtime/common/src/xcmp.rs
```diff
@@ -383,3 +340,35 @@ impl<X, Y, Treasury: TakeRevenue, Z> Drop for TransactionFeePoolTrader<X, Y, Tre
...
+			(1, Some(Parachain(id))) => {
+				let location = XcmAssetLocation::new(location.clone());
+				Registry::min_xcm_fee(ParaId::from(*id), location).or(Some(u128::MAX))
```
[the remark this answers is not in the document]
dzmitry-lahoda: Will recheck this part and tell what Acala does with this. That was some stuff taken from Acala some time ago.

code/parachain/runtime/common/src/xcmp.rs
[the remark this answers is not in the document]
dzmitry-lahoda: why? it was named impls.rs . sure it was bad.

code/parachain/runtime/common/src/fees.rs
```diff
@@ -0,0 +1,162 @@
...
+		constants::ExtrinsicBaseWeight, WeightToFeeCoefficient, WeightToFeeCoefficients,
+		WeightToFeePolynomial,
+	},
+};
+use num_traits::One;
+use primitives::currency::CurrencyId;
+use sp_runtime::Perbill;
+use sp_std::marker::PhantomData;
+
+pub const NATIVE_EXISTENTIAL_DEPOSIT: NativeBalance = 100_000_000_000;
```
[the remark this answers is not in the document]
dzmitry-lahoda: this will be NATIVE. it may be then abstracted with easy to refactor to Composable native.

code/parachain/runtime/common/src/fees.rs
```diff
@@ -0,0 +1,162 @@
...
+pub type NativeExistentialDeposit = ConstU128<NATIVE_EXISTENTIAL_DEPOSIT>;
```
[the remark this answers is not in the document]
dzmitry-lahoda: same

code/parachain/frame/composable-traits/src/currency.rs
```diff
@@ -38,8 +38,20 @@ pub trait AssetExistentialDepositInspect {
...
+		Err(DispatchError::Other("unimplemented!"))
+	}
+}
+
+/// foreign_amount / native_amount
+pub type ForeignByNative = Rational64;
+
+pub trait AssetRatioInspect {
+	type AssetId;
+	fn get_ratio(_asset_id: Self::AssetId) -> Option<ForeignByNative> {
```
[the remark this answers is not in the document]
dzmitry-lahoda: it is not price. it is ratio of prices. ratio of prices, but not price. and there is not such idea as unit in ratios. see how it initialized. it does not use units.

code/parachain/frame/composable-traits/src/currency.rs
```diff
@@ -38,8 +38,20 @@ pub trait AssetExistentialDepositInspect {
...
-	/// Given an asset ID, returns the existential_deposit of an asset as a `Balance`
-	fn existential_deposit(asset_id: Self::AssetId) -> Result<Self::Balance, DispatchError>;
+	/// Given an `asset_id`, returns the existential deposit of an asset in asset currency.
+	fn existential_deposit(_asset_id: Self::AssetId) -> Result<Self::Balance, DispatchError> {
+		Err(DispatchError::Other("unimplemented!"))
+	}
+}
+
+/// foreign_amount / native_amount
+pub type ForeignByNative = Rational64;
```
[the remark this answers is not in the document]
dzmitry-lahoda: It is OtherToNativeRatio. Neither unit nor price. Other can be the word. Ratio does not operates unit. I can sate 3/17 or 22 / 13. Where is unit?

code/parachain/runtime/picasso/src/xcmp.rs
dzmitry-lahoda: Likely it is something about sending tokens.  pallet owned by Acala and security was taken from here.

code/parachain/runtime/common/src/xcmp.rs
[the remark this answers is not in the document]
dzmitry-lahoda: Max prevented sending assets away. will set it to resonable default

code/parachain/runtime/common/src/fees.rs
dzmitry-lahoda: missed trait impl. it is BalanceCoversion. So BalanceConversion = AssetsRegistry + Hardcoded. Will name them same.

dzmitry-lahoda: so it just uses same method, `to_asset_balance =  configured to_asset_balance or hardcoded to_asset_balance
```
fn to_asset_balance(
		native_amount: NativeBalance,
		asset_id: CurrencyId,
	) -> Result<Balance, Self::Error> {
		AssetsRegistry::get_ratio(asset_id)
			...
			.or(WellKnownPriceConverter::to_asset_balance(native_amount, asset_id))
                       ...
	}
```

dzmitry-lahoda: naming them differently will be overrefactoring imho and neither target increased security of PR nor making fees better.

code/parachain/runtime/common/src/xcmp.rs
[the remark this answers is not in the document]
vimukthi-git: It's ok to have aggregations of code(impl.rs or whatever) until we have a good understanding about the overall product/repo/concepts. i.e We will refactor runtime level code with a good design(RFC) when we can think about it better with production experience.

code/parachain/runtime/common/src/fees.rs
vimukthi-git: I see then you can rename parameter `fee` -> `balance` then, or else resolve

code/parachain/runtime/common/src/fees.rs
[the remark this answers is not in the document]
vimukthi-git: As long as this package is called "common" there is no "native" here.

code/parachain/runtime/common/src/fees.rs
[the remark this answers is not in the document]
vimukthi-git: same here

code/parachain/frame/composable-traits/src/currency.rs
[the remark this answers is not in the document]
vimukthi-git: Any rational number (`x/y`) can be defined as amount of `x` per unit of `y` (basic math sir https://www.mathsisfun.com/measure/unit-price.html).

code/parachain/runtime/common/src/xcmp.rs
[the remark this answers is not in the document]
vimukthi-git: We need at least a comment explaining what this is used for if left like this

code/parachain/runtime/common/src/xcmp.rs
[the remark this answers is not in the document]
vimukthi-git: If the registry has no min fee defined it might make sense to prevent the assets from being sent anyway. We just need to prevent a scenario of assets getting trapped I guess

code/parachain/runtime/picasso/src/xcmp.rs
vimukthi-git: It's limited anyway by the channels that have been opened ultimately. Just that defining this would give better errors to users hopefully. Anyway feel free to resolve.

code/parachain/runtime/common/src/fees.rs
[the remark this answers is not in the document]
dzmitry-lahoda: `runtime/common` is place to handle common parts of runtimes. each runtime has native. we have 2 runtimes active. both has native to be PICA. as soon as new native will be here, common still will have native, but each its trait will be parametrized by some runtime trait. so PICA is not for common, but for that generic trait.

or you have other proposal how to make generalized code without copy pastes in future? I do what I think works and fits substrate coding practices. Renaming to PICA does not fit.

code/parachain/runtime/common/src/fees.rs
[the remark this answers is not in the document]
dzmitry-lahoda: same as above

code/parachain/frame/composable-traits/src/currency.rs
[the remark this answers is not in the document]
dzmitry-lahoda: I would leave such refactoring to next PRs if that is needed. Now ratios are defined according to what is desired. And better then fixed. As soon as all ratios will be normalized to some """unit""" let him rename the field too. 

For now there is not units nor prices.

code/parachain/runtime/common/src/xcmp.rs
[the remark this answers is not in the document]
dzmitry-lahoda: Sure. Will pint to XCM v4.

code/parachain/runtime/common/src/xcmp.rs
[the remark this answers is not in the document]
dzmitry-lahoda: I will add test which proves that assets just failed to sent, not trapped.

code/parachain/runtime/picasso/src/xcmp.rs
dzmitry-lahoda: `MultiLocationsFilter` != channel. So that is need separate handling. So can improve later.

code/parachain/runtime/common/src/xcmp.rs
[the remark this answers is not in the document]
dzmitry-lahoda: so you think `impls.rs` is better file name then `rewards.rs`? what impls is? rewards is clear - fees to block authors. if you not allow fees to block authors (which is weird) and rename it to anything else, e.g. treasury.rs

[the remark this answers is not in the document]
dzmitry-lahoda: i named the way i assume better to audit. specifically fees.rs must be audited for XCM(and that way BYOG) setup, while rewards.rs is not touching that area.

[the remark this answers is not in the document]
dzmitry-lahoda: > Don't see the relevance. Naming things non-ambiguously is not a black art to argue about. tbh I think you tend to get into this type of scenarios because you are over-refactoring code, then more names to give :)

please point me to overrefactoring? specific line of code?

code/parachain/runtime/common/src/fees.rs
benluelo: Then that should be put in the docs for the function

code/parachain/runtime/common/src/xcmp.rs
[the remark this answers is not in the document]
dzmitry-lahoda: impls.rs bad no matter what for review and audit. you can set final name to what ever what in next pr or rfc as you wish.

code/parachain/runtime/common/src/fees.rs
dzmitry-lahoda: This is already in polkadotjs storage constants and separate file about fees and in likely in generated pallet crates rust docs.

dzmitry-lahoda: may be some parachains move it into separate file just this one value so it can have specific codeowners

dzmitry-lahoda:
@vimukthi-git may you pin point me the lines showing bad design so we may consider to handle this now if possible as per you comment about `bad design`?

I do not do redesign so now. I do correctness and possibility to audit and security. You may consider redesign in other PRs

code/parachain/runtime/common/src/xcmp.rs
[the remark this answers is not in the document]
vimukthi-git: > so you think impls.rs is better file name then rewards.rs? what impls is?

No, just saying don't refactor(leave things as is) or give a different name that matches the function.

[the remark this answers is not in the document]
dzmitry-lahoda: new name to matches functions. old name distracts (should one look into it when audit fees or XCM - no, but it impls something? - so it is just bad). fees protect from spam. where is spam protection? here. 

imagine gov, it has taxes === fees. and it has budges = rewards. budges is from fees but != fees.

code/parachain/frame/composable-traits/src/currency.rs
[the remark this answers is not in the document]
dzmitry-lahoda: you used childing def of ratio(nal), here is proper science https://en.wikipedia.org/wiki/Ratio 

`In [mathematics](https://en.wikipedia.org/wiki/Mathematics), a ratio shows how many times one number contains another.`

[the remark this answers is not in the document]
dzmitry-lahoda: `unit` stuff is specialization which is not here - if somebody would enforce it (e.g. error if denom is not 10^12 - then rename it accrodingly)

dzmitry-lahoda:
I am fixing remaining issues. Hope most comments were done.

code/parachain/runtime/common/src/fees.rs
dzmitry-lahoda: fixed

code/parachain/runtime/common/src/fees.rs
dzmitry-lahoda: retaining as is for now.

code/parachain/runtime/common/src/fees.rs
```diff
@@ -0,0 +1,162 @@
...
+
+pub struct PriceConverter<AssetsRegistry>(PhantomData<AssetsRegistry>);
+
+pub mod cross_chain_errors {
+	pub const ASSET_IS_NOT_PRICEABLE: &str = "Asset is not priceable";
+	pub const AMOUNT_OF_ASSET_IS_MORE_THAN_MAX_POSSIBLE: &str =
+		"Amount of asset is more than max possible";
+}
+
+pub struct WellKnownPriceConverter;
```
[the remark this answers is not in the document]
dzmitry-lahoda: fixed, but used Native to encompass future Composable token into code.

code/parachain/runtime/common/src/fees.rs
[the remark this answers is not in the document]
dzmitry-lahoda: as per parity, they suggested to think yourself as if you are inside consensus. So it encodes (0, Here) or even Null in TS as native token. So each time your are in consensus you have native. so Native was used in goveranane.rs and other place. I will not changed used patter in this PR. feel free to rename consistently all place in next PRs.

code/parachain/runtime/common/src/fees.rs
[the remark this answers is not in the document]
dzmitry-lahoda: feel free to rename in next pr. I used existing name and convention which imho is way better.

code/parachain/runtime/picasso/src/xcmp.rs
dzmitry-lahoda: acala does not sets this one.

dzmitry-lahoda: added test for case when location or asset does not exists. it is good enough.

code/parachain/runtime/picasso/src/xcmp.rs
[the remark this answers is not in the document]
dzmitry-lahoda: we handle assets from only chains channels opened too.

dzmitry-lahoda:
```
type OnChargeTransaction =
		transaction_payment::CurrencyAdapter<Balances, ToStakingPot<Runtime, NativeTreasury>>;
```

So it was wrongly named and overengneered if you will check. 


Read like that `on charge  transaction do pay transaction in native token (Balances) into staking pot and native treasury`. So it is not fee but where fee is going.

dzmitry-lahoda:
MultiLocationFilter on xTokens is useless while we use XCM only for XCMP/HRMP/UMP with registered assets. As soon as XCM will be used for IBC or  extended to smart contracts or else. It would start make sence.

dzmitry-lahoda:
```
			let mut non_fee_reserve: Option<MultiLocation> = None;
			let asset_len = assets.len();
			for i in 0..asset_len {
				let asset = assets.get(i).ok_or(Error::<T>::AssetIndexNonExistent)?;
				ensure!(
					matches!(asset.fun, Fungibility::Fungible(x) if !x.is_zero()),
					Error::<T>::InvalidAsset
				);
				// `assets` includes fee, the reserve location is decided by non fee asset
				if (fee != *asset && non_fee_reserve.is_none()) || asset_len == 1 {
					non_fee_reserve = T::ReserveProvider::reserve(asset);
				}
				// make sure all non fee assets share the same reserve
				if non_fee_reserve.is_some() {
					ensure!(
						non_fee_reserve == T::ReserveProvider::reserve(asset),
						Error::<T>::DistinctReserveForAssetAndFee
					);
				}
			}

			let fee_reserve = T::ReserveProvider::reserve(&fee);
			if fee_reserve != non_fee_reserve {
				// Current only support `ToReserve` with relay-chain asset as fee. other case
				// like `NonReserve` or `SelfReserve` with relay-chain fee is not support.
				ensure!(non_fee_reserve == dest.chain_part(), Error::<T>::InvalidAsset);

				let reserve_location = non_fee_reserve.clone().ok_or(Error::<T>::AssetHasNoReserve)?;
				let min_xcm_fee = T::MinXcmFee::get(&reserve_location).ok_or(Error::<T>::MinXcmFeeNotDefined)?;
````

We never hit condition `fee_reserve != non_fee_reserve` because we do not maintain bug to backward compatiblity

https://docs.moonbeam.network/builders/xcm/xc-integration/#re-anchoring-support

will prove tommorow we never hit that

File: code/parachain/frame/composable-traits/src/lib.rs [hunk 16]
```diff
@@ -34,6 +33,12 @@
 	trivial_numeric_casts,
 	unused_extern_crates
 )]
+#![allow(incomplete_features)]
+#![feature(associated_type_defaults)]
+#![feature(trait_alias)]
+#![feature(const_trait_impl)]
+#![feature(const_convert)]
+#![feature(adt_const_params)]
```
benluelo, line 41: enabling incomplete features in production code probably isn't the best idea

code/parachain/frame/composable-traits/src/currency.rs
```diff
@@ -142,13 +163,123 @@ impl<
...
+	Ord,
+	Encode,
+	Decode,
+	MaxEncodedLen,
+	TypeInfo,
+)]
+pub struct Rational64 {
+	pub n: u64,
+	pub d: u64,
+}
```
benluelo, line 184: This is almost entirely copied from https://paritytech.github.io/substrate/master/src/sp_arithmetic/rational.rs.html#92 - why not just use `Rational128`?

benluelo: (I'm assuming it's due to size requirements - since `Ratio` was previously used, which was a type alias for `FixedU128`?)

code/parachain/frame/composable-traits/src/currency.rs
```diff
@@ -142,13 +163,123 @@ impl<
...
+			}
+		}
+		impl<const T: $t> TypedGet for $name<T> {
+			type Type = $t;
+			fn get() -> $t {
+				T
+			}
+		}
+	};
+}
```
benluelo, line 218: copied from here: https://paritytech.github.io/substrate/master/src/sp_core/lib.rs.html#422

please add docs and link to the original when copying code so maintainers down the line know where it came from

code/parachain/frame/currency-factory/src/lib.rs
```diff
@@ -210,9 +210,20 @@ pub mod pallet {
...
+
+		fn update_existential_deposit(asset_id: Self::AssetId, ed: Option<Self::Balance>) {
+			AssetEd::<T>::set(asset_id, ed)
+		}
+	}
+
 	impl<T: Config> LocalAssets<T::AssetId> for Pallet<T> {
+		// NOTE: it is not true as of now, so we should not rely on chain for this.
+		// NOTE: XCM does not supports decimals and Statemine can have assets without decimals
+		// NOTE: hence we cannot default to 12 decimals robustly
```
benluelo, line 225: What do you recommend as an alternative?

code/parachain/frame/composable-traits/src/lib.rs
dzmitry-lahoda: in theory can make these test only and add more unwraps/if let into prod codebase

code/parachain/frame/composable-traits/src/currency.rs
```diff
@@ -142,13 +163,123 @@ impl<
...
+impl<const N: u64, const D: u64> RationalLike<N, D> for Rational64 {
+	fn new() -> Self {
+		Self::from_unchecked(N, D)
+	}
+}
+
+impl Rational64 {
+	pub const fn from(n: u64, d: u64) -> Self {
+		Self::from_unchecked(n, d.max(1))
+	}
```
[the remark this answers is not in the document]
dzmitry-lahoda: See Rational128 from Parity. I agree with you so, but imho if parity did that can do too.

code/parachain/frame/composable-traits/src/currency.rs
dzmitry-lahoda: partially yes. 
Rational128 is not SCALE.

code/parachain/frame/currency-factory/src/lib.rs
dzmitry-lahoda: remove all 12 decimals here and fix all places - not in this PR.

code/parachain/frame/composable-traits/src/lib.rs
dzmitry-lahoda: also it will work if to encode Rational as u128 high and low bits. so it will be ugly big impl. and will not be able to use as #[const] parameter.

code/parachain/frame/composable-traits/src/currency.rs
```diff
@@ -142,13 +163,123 @@ impl<
...
+	pub n: u64,
+	pub d: u64,
+}
+
+#[macro_export]
+macro_rules! rational {
+	($n:literal / 0) => {
+		panic!("denominator cannot be zero")
+	};
+	($n:literal / $d: literal) => {
```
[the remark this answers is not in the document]
dzmitry-lahoda: NonZero will lead to some incompatibilities in storage and in compile time imho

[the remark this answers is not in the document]
dzmitry-lahoda: so one may try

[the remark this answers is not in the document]
benluelo: NonZero* types encode as their inner type: https://docs.rs/parity-scale-codec/latest/src/parity_scale_codec/codec.rs.html#689

code/parachain/frame/composable-traits/src/currency.rs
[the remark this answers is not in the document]
benluelo: yeah, this comment was written before I realized it was copied from parity. I think some docs explaining the behaviour would still be good though :+1:

code/parachain/frame/composable-traits/src/currency.rs
benluelo: Could you use a newtype wrapper around Rational128 instead? Strange that it doesn't implement any of the scale traits.

code/parachain/frame/composable-traits/src/lib.rs
benluelo: Not sure what you mean by not being able to use it in a const parameter - this works: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4b669b68d44e5d039f4b034317534934

code/parachain/frame/composable-traits/src/currency.rs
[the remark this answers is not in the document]
dzmitry-lahoda: nice.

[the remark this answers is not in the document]
dzmitry-lahoda: hope it composes with other places too. risk it may require some feature too.

code/parachain/frame/composable-traits/src/lib.rs
dzmitry-lahoda: const type parameter and #[const] config parameter

benluelo: 'const type parameter' is referring to the `RationalLike` trait?

code/parachain/frame/composable-traits/src/currency.rs
dzmitry-lahoda: it will take 256 bit. it is not 128 as fixed was, nor we need such precision.  not sure why parity has not u64?

code/parachain/frame/composable-traits/src/currency.rs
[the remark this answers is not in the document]
dzmitry-lahoda: fixed macro anyway.

code/parachain/frame/composable-traits/src/lib.rs
dzmitry-lahoda: #[pallet::const]
type FooBar = Get<Rational64>

benluelo: Not sure I understand - `Get<Rational64>` will work with what I sent before, unstable features aren't required for that

code/parachain/frame/composable-traits/src/currency.rs
benluelo: They tend to half-way implement a lot of things, it would have been trivial to have Rational8/16/32/64 implemented with a macro. Perhaps we should raise an issue or submit a PR.

code/parachain/frame/composable-traits/src/currency.rs
```diff
@@ -142,13 +160,122 @@ impl<
...
+)]
+pub struct Rational64 {
+	pub n: u64,
+	pub d: u64,
+}
+
+pub trait RationalLike<const N: u64, const D: u64> {
+	fn new() -> Self;
+	const CHECK: bool = if D != 0 { true } else { panic!("denominator cannot be zero") };
+}
```
benluelo, line 186: I really don't think this trait is necessary - as far as I can tell it's never used also?

code/parachain/runtime/common/src/xcmp.rs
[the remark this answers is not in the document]
dzmitry-lahoda: As as I see Acala does not prevents sending. So set to None as they do. But improved us for a case.

code/parachain/frame/composable-traits/src/currency.rs
dzmitry-lahoda: In theory it can be fixed using NonZero, but had bad experience trying to use NonZero in Substrate. 
So const check always work well.

dzmitry-lahoda: but yeah, may be replacement with NonZero (assuming it encodes into usual number) and it can be const generic, is nicer

dzmitry-lahoda: improved. for non zero keeped other issue opened.

benluelo: I think using NonZero and having a few simple tests to check it's encoding would be good

code/parachain/frame/composable-traits/src/currency.rs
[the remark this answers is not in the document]
dzmitry-lahoda: imho this better to delete this method so. and use const generics or non zero const generic. and in other places unchecked.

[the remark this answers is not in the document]
benluelo: This is related to the NonZero discussion in another thread - I definitely agree that this method isn't ideal, but a fallible constructor would be more ideal IMO

code/parachain/frame/composable-traits/src/lib.rs
benluelo: You could also use https://docs.rs/static_assertions/latest/static_assertions/macro.const_assert_ne.html

code/parachain/frame/composable-traits/src/currency.rs
dzmitry-lahoda: const CHECK: () = assert!(D != 0, "denominator cannot be zero");

dzmitry-lahoda: Sure, NonZero may work if to try.

code/parachain/frame/composable-traits/src/currency.rs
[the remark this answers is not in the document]
dzmitry-lahoda: with const generics features on may invent any kind of rules onto type.

code/parachain/frame/composable-traits/src/lib.rs
dzmitry-lahoda: > Not sure I understand - Get<Rational64> will work with what I sent before, unstable features aren't required for that
My test (may be 3 month ago) shown that without const cannot have nice config. So may be did something wrong.

code/parachain/frame/composable-traits/src/currency.rs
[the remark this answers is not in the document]
dzmitry-lahoda: closing as fixed macro. remain open NonZero trait stuff in other comments.

code/parachain/frame/composable-traits/src/lib.rs
dzmitry-lahoda: other refactoring https://github.com/ComposableFi/composable/pull/2118#discussion_r1015873472

code/parachain/frame/composable-traits/src/currency.rs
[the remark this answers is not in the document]
dzmitry-lahoda: closing this and retain other one for Rational changes

code/parachain/frame/assets-registry/src/tests.rs
```diff
@@ -126,7 +125,7 @@ fn update_asset() {
 		assert_eq!(AssetsRegistry::asset_ratio(local_asset_id), Some(ratio));
 
 		let new_decimals = 12;
-		let new_ratio = Ratio::from_inner(456);
+		let new_ratio = rational!(42 / 123);
```
benluelo, line 128: you're updating to the same ratio here

obsessed-cake (approved):

code/parachain/frame/composable-traits/src/currency.rs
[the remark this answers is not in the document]
dzmitry-lahoda: PricePerNativeUnit is confusing regardless of foreign or not.  there is no unit, and that is not price.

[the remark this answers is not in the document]
benluelo: Perhaps something as simple as `Ratio` then?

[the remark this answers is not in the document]
benluelo: If there is no unit, and it's not necessarily foreign/native, all that's left is ratio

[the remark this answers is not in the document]
dzmitry-lahoda: Ratio does not tells you that.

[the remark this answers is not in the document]
dzmitry-lahoda: AnyCurrencyToNativeCurrencyRatio ≡ X in Other currency / Y in native.

PR #2122, opened 2022-10-27
[cu-3m8xu0g] Adding release PR checklist

vimukthi-git:
## Description
Release checklist to be enforced

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/release-checklist.md
+++ b/.github/PULL_REQUEST_TEMPLATE/release-checklist.md
@@ -14,7 +14,7 @@ In the PR description,
 - [ ] Make sure proper logic/storage migrations are included as necessary
 - [ ] Verify documentation has been updated.
 
-## Action
+## Act
 
 - [ ] Generate weights, i.e run `benchmark`
 - [ ] Runtime [versioning](https://docs.substrate.io/build/upgrade-the-runtime/) updates
--- a/.github/PULL_REQUEST_TEMPLATE/release-checklist.md
+++ b/.github/PULL_REQUEST_TEMPLATE/release-checklist.md
@@ -1,0 +1,28 @@
+# Release Checklist
+
+## Understand
+
+In the PR description,
+
+- [ ] List updates to each runtime together with their audit reports since the last runtime upgrade to each of them.
+- [ ] List updates to the node codebase.
+- [ ] List upgrades to main dependencies such as substrate, core substrate pallets, ORML etc.
+
+## Verify
+
+- [ ] Storage/logic migrations from the previous versions has been removed.
+- [ ] Make sure proper logic/storage migrations are included as necessary
+- [ ] Verify documentation has been updated.
+
+## Action
+
+- [ ] Generate weights, i.e run `benchmark`
+- [ ] Runtime [versioning](https://docs.substrate.io/build/upgrade-the-runtime/) updates
+  - [ ] Update `spec_version` (automate-able)
+  - [ ] Update `transaction_version` if existing extrinsics or their ordering has changed. Can be verified via [metadata comparison](https://github.com/paritytech/polkadot/blob/master/doc/release-checklist.md#extrinsic-ordering).
+- [ ] Update composable node version if the code has changed
+- [ ] Update composableJs version (if necessary to be released)
+- [ ] Update FE version (if necessary to be released)
+- [ ] Update Subsquid version (if necessary to be released)
+- [ ] Consider and list possible proxy filter updates for available calls.
+- [ ] Categorize (and give a title) the release according to the types of changes it does, eg: security patch, bugfix, feature etc.
--- a/.github/PULL_REQUEST_TEMPLATE/release-checklist.md
+++ b/.github/PULL_REQUEST_TEMPLATE/release-checklist.md
@@ -26,3 +26,4 @@ In the PR description,
 - [ ] Update Subsquid version (if necessary to be released)
 - [ ] Consider and list possible proxy filter updates for available calls.
 - [ ] Categorize (and give a title) the release according to the types of changes it does, eg: security patch, bugfix, feature etc.
+- [ ] Run `cargo build` once to update the Cargo.lock file.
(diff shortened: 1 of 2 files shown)


ComposableFiB0t:
Task linked: [CU-3m8xu0g Create PR Release checklist](https://app.clickup.com/t/3m8xu0g)

.github/PULL_REQUEST_TEMPLATE/release-checklist.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Verify
+
+- [ ] Storage/logic migrations from the previous versions has been removed.
+- [ ] Make sure proper logic/storage migrations are included as necessary
+- [ ] Verify documentation has been updated.
+
+## Act
+
+- [ ] Generate weights, i.e run `benchmark`
```
KaiserKarel, line 19:
```suggestion
- [ ] Regenerate weights for all pallets.
```

.github/PULL_REQUEST_TEMPLATE/release-checklist.md
```diff
@@ -0,0 +1,28 @@
...
+- [ ] Runtime [versioning](https://docs.substrate.io/build/upgrade-the-runtime/) updates
+  - [ ] Update `spec_version` (automate-able)
+  - [ ] Update `transaction_version` if existing extrinsics or their ordering has changed. Can be verified via [metadata comparison](https://github.com/paritytech/polkadot/blob/master/doc/release-checklist.md#extrinsic-ordering).
+- [ ] Update composable node version if the code has changed
+- [ ] Update composableJs version (if necessary to be released)
+- [ ] Update FE version (if necessary to be released)
+- [ ] Update Subsquid version (if necessary to be released)
+- [ ] Consider and list possible proxy filter updates for available calls.
+- [ ] Categorize (and give a title) the release according to the types of changes it does, eg: security patch, bugfix, feature etc.
```
KaiserKarel, line 28: I feel like individually versioning these is a little bit error prone; i.e changing extrinsic ordering is a breaking change according to substrate and requires incrementing the spec_version, but according to Rust semver it is non-breaking and only requires `PATCH` to be incremented. Since composableJS relies on metadata, that would need to be incremented by `MAJOR` as it is a breaking change. 

Personally think that everything that relies on the node version ought to be versioned the same to avoid confusion, else we will run into

Rust code -> patch increment
JS code -> major
runtime version -> major (spec),

So we can end up with the rust code being at version 0.0.8, but the js code being at 8.0.0, which will catch people by suprise. Alternatively, if we rerelease everything (even if it is unchanged), we don't need to think about versioning. If the node is on 3.3.4, use JS version 3.3.4.

Only the frontend should perhaps be versioned slightly differently, so that we can separate out UI changes from other changes.

KaiserKarel: Frontend versioning could be as simple as ${VERSION}-{label}, where label is something we use to indicate UI changes

vimukthi-git: This document intentionally avoids the release/versioning strategy as that topic/discussion is to be capture by an RFC, here it only captures the versions that needs to be updated as a checklist to keep it brief without getting into detail on how to do it. We can definitely link that here when we have that RFC (or other doc). 

Anyway just for discussion

> Alternatively, if we rerelease everything (even if it is unchanged), we don't need to think about versioning

Though it's convenient, the problem with incrementing versions without an actual update is that when non runtime/rust related components(JS) gets updated we are forced to make updates to runtime version and do a runtime upgrade(or node install) without a reason or risk having an out of sync runtime that can create confusion when trying to resolve issues. It's best to be explicit in this case IMO and do proper versioning of each component and have the release page list the latest versions of each component tested together as being compatible to avoid confusion for users. 

Because of the directed dependency between runtime version and the other components, we can and must keep other dependent rust components(node, pallets, libraries) versions following the same version prefix to match runtime. Polkadot in this case [seems to use](https://github.com/paritytech/polkadot/pull/5982/files#diff-e5e76e02c0d16e79c70b024cbe3c6ea56f3249382a0f987ba203c34fcb40ed66) something like `x.y.z` for the rust components and `xyz0` (concatenated numbers with 0 postfix) for the runtime. 

> Frontend versioning could be as simple as ${VERSION}-{label} ..

Agree and I think we can do the same for all JS component versions as they can evolve separately.

andyjsbell: > Though it's convenient, the problem with incrementing versions without an actual update is that when non runtime/rust related components(JS) gets updated we are forced to make updates to runtime version and do a runtime upgrade(or node install) without a reason or risk having an out of sync runtime that can create confusion when trying to resolve issues. It's best to be explicit in this case IMO and do proper versioning of each component and have the release page list the latest versions of each component tested together as being compatible to avoid confusion for users.

I am not quite sure we would need to make runtime upgrades if there is a change in a front end component.  This would be incremented accordingly with the suggested versioning ${VERSION}-${label} above and deployed independently from the runtime.

vimukthi-git: I think there is consensus on how FE is versioned. My point is more about things like the ComposableJS library which contains runtime types, if both those are forced to have the same version scheme then make a major interface change in ComposableJS then we would have to upgrade the runtime as well. Therefore I suggest just having different versions for those. For rust components we can definitely go for the same version scheme.

dzmitry-lahoda:
Release Rejection routine:
1. collect all commits
2. collect all release notes
3. each release note has one or more commits
4. if any commit without release note - reject release

can have allow to have release note reference CU/GH item which references commit. 
this way can compress commits into on CU/GH reference

.github/PULL_REQUEST_TEMPLATE/release-checklist.md
vimukthi-git: I'm resolving this comment as per discussions in https://github.com/ComposableFi/composable/pull/2211

vimukthi-git:
>there is no signed tag by gpg identity registered on Kusama owners who are either council or tech or PICA holder which verifies approves git commit and hash for deterministic artifacts.

Let me add this @dzmitry-lahoda

andyjsbell (approved):
lgtm 👍

PR #2126, opened 2022-10-27
[CU-2vxvfbd] Remove max_rewards from Rewards and add 'pause' logic for accumulation

andresmechali:
## Issue
https://app.clickup.com/t/2vxvfbd

### DIFF
--- a/code/parachain/frame/staking-rewards/src/lib.rs
+++ b/code/parachain/frame/staking-rewards/src/lib.rs
@@ -63,7 +63,10 @@ use frame_support::{
 };
 pub use pallet::*;
 use sp_runtime::SaturatedConversion;
-use sp_std::{cmp, ops::Div};
+use sp_std::{
+	cmp,
+	ops::{Div, Mul},
+};
 
 #[frame_support::pallet]
 pub mod pallet {
@@ -1699,8 +1702,14 @@ pub(crate) fn do_reward_accumulation<T: Config>(
 		let releasable_periods_surpassed =
 			cmp::min(maximum_releasable_periods, periods_surpassed.into());
 
+		// SAFETY: Usage of mul is safe here because newly_accumulated_rewards =
+		//      (    total_locked_rewards            elapsed_time        )
+		//  min ( ------------------------- , -------------------------- )
+		//      ( reward.reward_rate.amount   reward_rate_period_seconds )
+		// If LHS is smaller, this will result in total_locked_rewards, which we know fits in a
+		// u128. If RHS is smaller, this has to be < total_locked_rewards, so it will also fit.
 		let newly_accumulated_rewards =
-			releasable_periods_surpassed.saturating_mul(reward.reward_rate.amount.into());
+			releasable_periods_surpassed.mul(reward.reward_rate.amount.into());
 		let new_total_rewards = newly_accumulated_rewards
 			.safe_add(&reward.total_rewards.into())
 			.map_err(|_| RewardAccumulationCalculationError::ArithmeticError)?;
--- a/code/parachain/frame/staking-rewards/src/lib.rs
+++ b/code/parachain/frame/staking-rewards/src/lib.rs
@@ -1705,7 +1705,7 @@ pub(crate) fn do_reward_accumulation<T: Config>(
 				.saturating_mul(releasable_periods_surpassed.saturated_into::<u64>()),
 		);
 
-		if !total_locked_rewards.is_zero() {
+		if !total_locked_rewards.is_zero() && !newly_accumulated_rewards.is_zero() {
 			T::Assets::release(
 				asset_id,
 				pool_account,
--- a/code/parachain/frame/staking-rewards/src/lib.rs
+++ b/code/parachain/frame/staking-rewards/src/lib.rs
@@ -81,7 +81,6 @@ pub mod pallet {
 		fnft::{FinancialNft, FinancialNftProtocol},
 		staking::{
 			lock::LockConfig, RewardPoolConfiguration::RewardRateBasedIncentive, RewardRatePeriod,
-			DEFAULT_MAX_REWARDS,
 		},
 		time::{DurationSeconds, ONE_MONTH, ONE_WEEK},
 	};
@@ -114,7 +113,7 @@ pub mod pallet {
 		traits::{AccountIdConversion, BlockNumberProvider, One},
 		ArithmeticError, PerThing,
 	};
-	use sp_std::{cmp::max, fmt::Debug, ops::Mul, vec, vec::Vec};
+	use sp_std::{fmt::Debug, ops::Mul, vec, vec::Vec};
 
 	use crate::{
 		add_to_rewards_pot, claim_of_stake, do_reward_accumulation, prelude::*,
@@ -193,10 +192,6 @@ pub mod pallet {
 			pool_id: T::AssetId,
 			asset_id: T::AssetId,
 			error: RewardAccumulationHookError,
-		},
-		MaxRewardsAccumulated {
-			pool_id: T::AssetId,
-			asset_id: T::AssetId,
 		},
 		RewardPoolUpdated {
 			pool_id: T::AssetId,
@@ -1489,16 +1484,6 @@ pub mod pallet {
 						});
 					}
 				},
-				Err(MaxRewardsAccumulated) => {
-					Self::deposit_event(Event::<T>::MaxRewardsAccumulated {
-						pool_id,
-						asset_id: reward_asset_id,
-					});
-				},
-				Err(MaxRewardsAccumulatedPreviously) => {
-					// max rewards were accumulated previously, silently continue since the
-					// MaxRewardsAccumulated event has already been emitted for this pool
-				},
 			}
 		}
 
@@ -1569,7 +1554,6 @@ pub mod pallet {
 							total_rewards: amount,
 							claimed_rewards: Zero::zero(),
 							total_dilution_adjustment: T::Balance::zero(),
-							max_rewards: max(amount, DEFAULT_MAX_REWARDS.into()),
 							reward_rate: RewardRate {
 								amount: T::Balance::zero(),
 								period: RewardRatePeriod::PerSecond,
@@ -1605,22 +1589,25 @@ fn add_to_rewards_pot<T: Config>(
 	amount: T::Balance,
 	keep_alive: bool,
 ) -> DispatchResult {
-	RewardPools::<T>::get(pool_id)
-		.ok_or(Error::<T>::RewardsPoolNotFound)?
-		.rewards
-		.get(&asset_id)
-		.ok_or(Error::<T>::RewardAssetNotFound)?;
+	RewardPools::<T>::try_mutate(pool_id, |pool| {
+		let pool = pool.as_mut().ok_or(Error::<T>::RewardsPoolNotFound)?;
 
-	let pool_account = Pallet::<T>::pool_account_id(&pool_id);
+		let reward = pool.rewards.get_mut(&asset_id).ok_or(Error::<T>::RewardAssetNotFound)?;
 
-	T::Assets::transfer(asset_id, &who, &pool_account, amount, keep_alive)?;
-	T::Assets::hold(asset_id, &pool_account, amount)?;
+		if RewardsPotIsEmpty::<T>::contains_key(pool_id, asset_id) {
+			reward.last_updated_timestamp = T::UnixTime::now().as_secs();
+			RewardsPotIsEmpty::<T>::remove(pool_id, asset_id);
+		}
 
-	RewardsPotIsEmpty::<T>::remove(pool_id, asset_id);
+		let pool_account = Pallet::<T>::pool_account_id(&pool_id);
 
-	Pallet::<T>::deposit_event(Event::<T>::RewardsPotIncreased { pool_id, asset_id, amount });
+		T::Assets::transfer(asset_id, &who, &pool_account, amount, keep_alive)?;
+		T::Assets::hold(asset_id, &pool_account, amount)?;
 
-	Ok(())
+		Pallet::<T>::deposit_event(Event::<T>::RewardsPotIncreased { pool_id, asset_id, amount });
+
+		Ok(())
+	})
 }
 
 #[transactional]
@@ -1645,15 +1632,6 @@ fn update_rewards_pool<T: Config>(
 				Ok(()) => {},
 				Err(RewardAccumulationCalculationError::BackToTheFuture) =>
 					return Err(Error::<T>::BackToTheFuture.into()),
-				Err(RewardAccumulationCalculationError::MaxRewardsAccumulated) => {
-					Pallet::<T>::deposit_event(Event::<T>::MaxRewardsAccumulated {
-						pool_id,
-						asset_id,
-					});
-					continue
-				},
-				Err(RewardAccumulationCalculationError::MaxRewardsAccumulatedPreviously) =>
-					continue,
 				Err(RewardAccumulationCalculationError::RewardsPotEmpty) =>
 					return Err(Error::<T>::RewardsPotEmpty.into()),
 			}
@@ -1691,7 +1669,8 @@ pub(crate) fn do_reward_accumulation<T: Config>(
 	if periods_surpassed.is_zero() {
 		Ok(())
 	} else {
-		let total_locked_rewards: u128 = T::Assets::balance_on_hold(asset_id, pool_account).into();
+		let balance_on_hold = T::Assets::balance_on_hold(asset_id, pool_account);
+		let total_locked_rewards: u128 = balance_on_hold.into();
 
 		// the maximum amount repayable given the reward rate.
 		// i.e. if total locked is 50, and the reward rate is 15, then this would be 3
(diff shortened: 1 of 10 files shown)


ComposableFiB0t:
Task linked: [CU-2vxvfbd Review usage of max_rewards for pools](https://app.clickup.com/t/2vxvfbd)

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1710,76 +1686,37 @@ pub(crate) fn do_reward_accumulation<T: Config>(
...
-				.expect("funds should be available to release based on previous check; qed;");
-
-				// return an error, but update the reward first
-				reward.total_rewards = reward.max_rewards;
-				reward.last_updated_timestamp = last_updated_timestamp;
-				Err(RewardAccumulationCalculationError::MaxRewardsAccumulated)
-			} else {
-				Err(RewardAccumulationCalculationError::RewardsPotEmpty)
+		if !total_locked_rewards.is_zero() {
+			let amount_to_unlock = new_total_rewards.saturating_sub(reward.total_rewards.into());
```
vimukthi-git, line 1690: This is just `newly_accumulated_rewards` no?

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1710,76 +1686,37 @@ pub(crate) fn do_reward_accumulation<T: Config>(
...
+			// Note: will this always be Ok or need to handle error?
+			.expect("funds should be available to release based on previous check; qed;");
+
+			reward.total_rewards = new_total_rewards.into();
+			reward.last_updated_timestamp = last_updated_timestamp;
+
+			let new_balance_on_hold = T::Assets::balance_on_hold(asset_id, pool_account);
+
+			if new_balance_on_hold.into().is_zero() {
+				return Err(RewardAccumulationCalculationError::RewardsPotEmpty)
```
vimukthi-git, line 1707: Without max_reward , this error does not serve any purpose as the intended use of the pool of disbursing whatever the amount in the pot is done. So we can get rid of it.

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1710,76 +1686,37 @@ pub(crate) fn do_reward_accumulation<T: Config>(
...
+				asset_id,
+				pool_account,
+				amount_to_unlock.into(),
+				false, // not best effort, entire amount must be released
+			)
+			// Note: will this always be Ok or need to handle error?
+			.expect("funds should be available to release based on previous check; qed;");
+
+			reward.total_rewards = new_total_rewards.into();
+			reward.last_updated_timestamp = last_updated_timestamp;
```
vimukthi-git, line 1702: Setting this timestamp has to be done each round regardless of the fact that there are no rewards to disburse. This is to ensure that no rewards are disbursed retroactively given there is no max_reward agreed in the pool.

code/parachain/frame/staking-rewards/src/lib.rs
andresmechali: Alright. Should we still have some event stating that the pot is empty or it's not relevant?

code/parachain/frame/staking-rewards/src/lib.rs
andresmechali: You' actually right sir

code/parachain/frame/staking-rewards/src/lib.rs
andresmechali: Rewards will not be disbursed retroactively, as the timestamp is updated when an empty pot is refilled. This happens on the `add_to_rewards_pot` function. I don't think we need to update it regardless, as this value should represent the last time the rewards were updated, and this only happen if there are rewards to disburse.

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1597,22 +1584,25 @@ fn add_to_rewards_pot<T: Config>(
...
+	RewardPools::<T>::try_mutate(pool_id, |pool| {
+		let pool = pool.as_mut().ok_or(Error::<T>::RewardsPoolNotFound)?;
 
-	let pool_account = Pallet::<T>::pool_account_id(&pool_id);
+		let reward = pool.rewards.get_mut(&asset_id).ok_or(Error::<T>::RewardAssetNotFound)?;
 
-	T::Assets::transfer(asset_id, &who, &pool_account, amount, keep_alive)?;
-	T::Assets::hold(asset_id, &pool_account, amount)?;
+		if RewardsPotIsEmpty::<T>::contains_key(pool_id, asset_id) {
+			reward.last_updated_timestamp = T::UnixTime::now().as_secs();
```
benluelo, line 1593: why are you updating last_updated_timestsamp here, when you're not accumulating any rewards?

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1710,76 +1694,35 @@ pub(crate) fn do_reward_accumulation<T: Config>(
...
-				)
-				.expect("funds should be available to release based on previous check; qed;");
-
-				// return an error, but update the reward first
-				reward.total_rewards = reward.max_rewards;
-				reward.last_updated_timestamp = last_updated_timestamp;
-				Err(RewardAccumulationCalculationError::MaxRewardsAccumulated)
-			} else {
-				Err(RewardAccumulationCalculationError::RewardsPotEmpty)
+		if !total_locked_rewards.is_zero() {
```
benluelo, line 1697:
```suggestion
		if !newly_accumulated_rewards.is_zero() {
```

benluelo: you need to check that the newly accumulated rewards isn't zero, because total_locked_balance can be less than one period's worth of rewards but not zero

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1710,76 +1694,35 @@ pub(crate) fn do_reward_accumulation<T: Config>(
...
+			T::Assets::release(
+				asset_id,
+				pool_account,
+				newly_accumulated_rewards.into(),
+				false, // not best effort, entire amount must be released
+			)
+			// Note: will this always be Ok or need to handle error?
```
benluelo, line 1704: newly_accumulated_rewards is defined as:

```rust
let newly_accumulated_rewards = reward.reward_rate.amount * min(
    // maximum periods available
    total_locked_rewards / reward.reward_rate.amount,
    // periods elapsed
    elapsed_time / reward_rate_period_seconds,
);
```

if `periods elapsed` is smaller, that means that there are more rewards to release than have been surpassed; if `maximum periods available` is smaller, that means that only the available amount of rewards will be released - so this should always succeed as it never attempts to release more than is available.

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1710,76 +1694,35 @@ pub(crate) fn do_reward_accumulation<T: Config>(
...
+			.expect("funds should be available to release based on previous check; qed;");
+
+			reward.total_rewards = new_total_rewards.into();
+			reward.last_updated_timestamp = last_updated_timestamp;
+
+			let new_balance_on_hold = T::Assets::balance_on_hold(asset_id, pool_account);
+
+			if new_balance_on_hold.into().is_zero() {
```
benluelo, line 1712: you should check that it's below the reward rate for one period, otherwise it could end up in a situation where the amount is less than required to accumulate one period but it still attempts to accumulate every block (which would be wasteful)

code/parachain/frame/staking-rewards/src/test/test_reward_accumulation_hook.rs
```diff
@@ -148,6 +139,7 @@ fn test_accumulate_rewards_pool_empty_refill() {
 		// cap will be hit after 5_000 seconds (834 blocks)
 		const A_A_REWARD_RATE: u128 = A::units(2) / 100_000;
 		const A_A_MAX_REWARDS: u128 = A::units(1) / 10;
```
benluelo, line 141: as far as I can tell, this is unused now? you should also remove the comments about the cap being hit

code/parachain/frame/staking-rewards/src/test/test_reward_accumulation_hook.rs
```diff
@@ -148,6 +139,7 @@ fn test_accumulate_rewards_pool_empty_refill() {
...
+		const A_A_AMOUNT_TO_ADD: u128 = A_A_REWARD_RATE * block_seconds(4);
```
benluelo, line 142: I would rename this to something slightly more descriptive - `A_A_AMOUNT_TO_ADD_TO_REWARDS_POT`; I was confused as to what 'amount to add' meant until I saw it used

code/parachain/frame/staking-rewards/src/test/test_reward_accumulation_hook.rs
```diff
@@ -244,11 +224,8 @@ fn test_accumulate_rewards_pool_empty_refill() {
 
 		progress_to_block(current_block + 1, &mut current_block);
 
-		check_events([crate::Event::<Test>::RewardAccumulationHookError {
-			pool_id: A::ID,
-			asset_id: A::ID,
-			error: RewardAccumulationHookError::RewardsPotEmpty,
-		}]);
+		// Note: reward pot for A_B should become empty, once it adds the remaining rewards.
+		// Should we emit some event on the pallet? If yes, it should be asserted here.
```
benluelo, line 228: an event is emitted, see `reward_accumulation_hook_reward_update_calculation`. We could consider a new event for this though; thoughts @vimukthi-git?

Something like `RewardsPoolPaused` and `RewardPoolUnpaused`?

benluelo (changes requested):

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: updating the reward pot empty storage is now being done in two places - here and in `reward_accumulation_hook_reward_update_calculation`. It should be removed from one of them.

since `do_reward_accumulation` doesn't currently read the `RewardPotIsEmpty` storage, I don't think it should write to it either. It just accumulates if it can, and reports the relevant errors which are then handled at the callsite.

code/parachain/frame/staking-rewards/src/lib.rs
andresmechali: This is to avoid retroactively accumulating rewards. When it accumulates the next time, it will start from this timestamp instead of the last time it accumulated. Originally I added a `last_time_pot_funded: Option<u64>` field which served this purpose. Do you think that's a better approach?

code/parachain/frame/staking-rewards/src/test/test_reward_accumulation_hook.rs
andresmechali: You are right. I rephrased those comments as `initial amount will be fully rewarded after 25_000 seconds (4167 blocks)`, so that there's some explanation within the code about why we later use `progress_to_block(STARTING_BLOCK + 4167, &mut current_block);`

code/parachain/frame/staking-rewards/src/test/test_reward_accumulation_hook.rs
andresmechali: Yes sorry, that's what I meant. A specific event for when we pause/unpause. I think we should.

benluelo: I think that would be a good change, but I'm OK if we get this merged first and then update the event(s) in a followup PR so this one doesn't get too big

code/parachain/frame/staking-rewards/src/test/test_reward_accumulation_hook.rs
benluelo: Sounds good :+1:

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: Ah I see now, I think that makes more sense than updating every block. We could even change the `RewardPotIsEmpty` storage to contain the last updated timestamp as an `Option<u64>` (since we have to read it anyways), with `None` meaning the pot is empty.

We could also consider a more descriptive enum, something like:

```rust
enum RewardsPotState {
    Paused,
    Running {
        last_updated_timestamp: u64
    },
}
```

The `last_updated_timestamp` field could then be removed from the pool storage, which would make more sense.

andresmechali: Yeah that could work as well, but that would mean reading the `RewardPotIsEmpty` storage every block to calculate the accumulated rewards. I'm not sure how often will pools be paused (in which case we avoid getting the whole pool from storage) and how often now (which will mean having an extra read compared to what we have now)

code/parachain/frame/staking-rewards/src/test/test_reward_accumulation_hook.rs
andresmechali: Sure let's do that. I created a task for this, which I'll do once this PR is merged: https://app.clickup.com/t/3qftw8g

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: this pallet isn't live, so we can change the storage

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1691,7 +1686,8 @@ pub(crate) fn do_reward_accumulation<T: Config>(
 	if periods_surpassed.is_zero() {
 		Ok(())
 	} else {
-		let total_locked_rewards: u128 = T::Assets::balance_on_hold(asset_id, pool_account).into();
+		let balance_on_hold = T::Assets::balance_on_hold(asset_id, pool_account);
+		let total_locked_rewards: u128 = balance_on_hold.into();
```
benluelo, line 1690: this could be just one variable, no need for the temp variable balance_on_hold

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1718,79 +1714,39 @@ pub(crate) fn do_reward_accumulation<T: Config>(
...
-				)
-				.expect("funds should be available to release based on previous check; qed;");
-
-				// return an error, but update the reward first
-				reward.total_rewards = reward.max_rewards;
-				reward.last_updated_timestamp = last_updated_timestamp;
-				Err(RewardAccumulationCalculationError::MaxRewardsAccumulated)
-			} else {
-				Err(RewardAccumulationCalculationError::RewardsPotEmpty)
+		if !total_locked_rewards.is_zero() && !newly_accumulated_rewards.is_zero() {
```
benluelo, line 1717:
```suggestion
		if !newly_accumulated_rewards.is_zero() {
```
Only `newly_accumulated_rewards` needs to be checked to be non-zero

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1704,12 +1700,12 @@ pub(crate) fn do_reward_accumulation<T: Config>(
...
 			cmp::min(maximum_releasable_periods, periods_surpassed.into());
 
-		// saturating is safe here since these values are checked against max_rewards anyways, which
-		// is <= u128::MAX
-		let newly_accumulated_rewards =
-			u128::saturating_mul(releasable_periods_surpassed, reward.reward_rate.amount.into());
-		let new_total_rewards =
-			newly_accumulated_rewards.saturating_add(reward.total_rewards.into());
+		let newly_accumulated_rewards = releasable_periods_surpassed
+			.safe_mul(&reward.reward_rate.amount.into())
```
benluelo, line 1704: There's no need for fallible arithmetic here (see this comment: https://github.com/ComposableFi/composable/pull/2126#discussion_r1008393598)

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1700,9 +1699,8 @@ pub(crate) fn do_reward_accumulation<T: Config>(
 		let releasable_periods_surpassed =
 			cmp::min(maximum_releasable_periods, periods_surpassed.into());
 
-		let newly_accumulated_rewards = releasable_periods_surpassed
-			.safe_mul(&reward.reward_rate.amount.into())
-			.map_err(|_| RewardAccumulationCalculationError::ArithmeticError)?;
+		let newly_accumulated_rewards =
+			releasable_periods_surpassed.saturating_mul(reward.reward_rate.amount.into());
```
benluelo, line 1703: If you're going to use saturating, leave a comment about why saturating is safe and doesn't break logic.

Here, just `mul` should be used, and a comment explaining why it's safe

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1641,21 +1641,15 @@ fn update_rewards_pool<T: Config>(
...
-						asset_id,
-					});
-					continue
-				},
-				Err(RewardAccumulationCalculationError::MaxRewardsAccumulatedPreviously) =>
-					continue,
 				Err(RewardAccumulationCalculationError::RewardsPotEmpty) =>
 					return Err(Error::<T>::RewardsPotEmpty.into()),
+				Err(RewardAccumulationCalculationError::ArithmeticError) =>
+					return Err(Error::<T>::ArithmeticError.into()),
```
benluelo, line 1652: is this error still required?

andresmechali: It is, given the current use of
 ```
 let new_total_rewards = newly_accumulated_rewards
  .safe_add(&reward.total_rewards.into())
  .map_err(|_| RewardAccumulationCalculationError::ArithmeticError)?;
```
I think this will never result in an error, as it should be accumulating `u128::MAX` for this to happen. However, I don't know what values can `reward.reward_rate.amount` take in order to establish some practical boundaries and be sure that `newly_accumulated_rewards + reward.total_rewards` will never overflow.

benluelo (approved):

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: makes sense, we can review the logic there at a later date :+1:

PR #2144, opened 2022-10-28
[CU-35d6w57] Store decimals in subsquid

andresmechali:
## Issue
https://app.clickup.com/t/35d6w57

### DIFF
--- a/subsquid/src/dbHelper.ts
+++ b/subsquid/src/dbHelper.ts
@@ -4,7 +4,7 @@ import { randomUUID } from "crypto";
 import { ApiPromise, WsProvider } from "@polkadot/api";
 import { hexToU8a } from "@polkadot/util";
 import { chain } from "./config";
-import { encodeAccount, getNormalizedAmount } from "./utils";
+import { encodeAccount, getAmountWithoutDecimals } from "./utils";
 import {
   Account,
   Activity,
@@ -235,7 +235,7 @@ export async function storeHistoricalLockedValue(
     }
     const lockedValue =
       oraclePrices[assetId] *
-      getNormalizedAmount(amountsLocked[assetId], assetDecimals[assetId]);
+      getAmountWithoutDecimals(amountsLocked[assetId], assetDecimals[assetId]);
     return BigInt(agg) + lockedValue;
   }, BigInt(0));
 
--- a/subsquid/src/dbHelper.ts
+++ b/subsquid/src/dbHelper.ts
@@ -4,10 +4,11 @@ import { randomUUID } from "crypto";
 import { ApiPromise, WsProvider } from "@polkadot/api";
 import { hexToU8a } from "@polkadot/util";
 import { chain } from "./config";
-import { encodeAccount } from "./utils";
+import { encodeAccount, getNormalizedAmount } from "./utils";
 import {
   Account,
   Activity,
+  Asset,
   Currency,
   Event,
   EventType,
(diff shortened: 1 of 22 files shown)


dzmitry-lahoda:
is title still corect? xcm does not propagates decimals nor allows to hook to convert,  or any other known bridge. also unkown assets do. ot have decimals and statemine assets can be without. so decimals only ui side concern only. and usdt is 6 decimals.

andresmechali:
> is title still corect? xcm does not propagates decimals nor allows to hook to convert, or any other known bridge. also unkown assets do. ot have decimals and statemine assets can be without. so decimals only ui side concern only. and usdt is 6 decimals.

@dzmitry-lahoda This is, as you said, for UI purpose. It's only FE who will use the Subsquid queries that are affected by this, in order to correctly show accumulated values that involve multiple currencies.

subsquid/src/dbHelper.ts
```diff
@@ -200,6 +201,17 @@ export async function storeHistoricalLockedValue(
...
+  const assetDecimals: Record<string, number> = {};
+
+  for (const assetId of Object.keys(amountsLocked)) {
+    const asset = await ctx.store.get(Asset, {
+      where: {
+        id: assetId,
+      },
+    });
+
+    assetDecimals[assetId] = asset?.decimals || 12;
```
dzmitry-lahoda, line 213: please remove 12 from this line.

dzmitry-lahoda: retain asset to undefined if it is such in this code.

subsquid/src/model/generated/asset.model.ts
```diff
@@ -27,6 +27,12 @@ export class Asset {
   @Column_("numeric", {transformer: marshal.bigintTransformer, nullable: false})
   price!: bigint
 
+  /**
+   * Defaults to 12 when not specified
```
dzmitry-lahoda, line 31: as was dicussed in slack with other people either no default or (that is mine) make 12 only last step decision with TODO item to remove that. do not provide decimals from data layer if it is not aware of decimals.

dzmitry-lahoda: OR add gh/cu item to fix data layer with no hardcode.

subsquid/src/dbHelper.ts
saadjhk: @andresmechali we should remove this

subsquid/src/model/generated/asset.model.ts
saadjhk: > is title still corect? xcm does not propagates decimals nor allows to hook to convert, or any other known bridge. also unkown assets do. ot have decimals and statemine assets can be without. so decimals only ui side concern only. and usdt is 6 decimals.

@andresmechali yeah i believe you dont need to store normalized values, but one thing keep in mind are pablo transactions, when there are two assets involved (they might have a different decimals) we need to be sure in that case whether we will calculate correctly or not (e.g spot price)

vimukthi-git (approved):

subsquid/src/dbHelper.ts
andresmechali: If I remove this we cannot properly normalize accumulations. Meaning, if for some reason we don't know the number of decimals, how can we sum it with the previous value in a way that makes sense?

subsquid/src/model/generated/asset.model.ts
andresmechali: The issue is that if I don't normalize values, I cannot properly provide a value that makes sense for queries like `totalValueLocked`.

subsquid/src/dbHelper.ts
andresmechali: Yes I understand, this is meant to be exclusively for the FE to use, to provide total locked values without FE needing to get the decimals and do it themselves. As discussed today, ignoring any token that doesn't have decimals. If this is still not convenient I can send them the values of each token and they can add them up as pleased.

subsquid/src/model/generated/asset.model.ts
andresmechali: @saadjhk regarding your comment about Pablo, we do treat assets independently so that's not an issue

easteregg (approved):

andresmechali:
@mergify refresh

PR #2176, opened 2022-11-02
cor/flake-parts

cor:
## Issue
This migrates our flake to use `flake-parts`, which is `NixOS`'s module system applied to their flake system. It should make our `.nix` configs a lot more mainatianable.

This is by no means a perfect refactor, there is still a lot of cleanup to do, but this is a good checkpoint.

Please review carefully as it affects _all of our `.nix` configurations_.

https://app.clickup.com/t/346hfek

## Checklist

- [x] Updated the rust/typescript docs.
- [x] Updated the `docs/`.


dzmitry-lahoda (approved):

hussein-aitlahcen (approved):

flake.nix
```diff
@@ -37,1605 +27,114 @@
...
-          frontend-pablo-server = let PORT = 8002;
-          in pkgs.writeShellApplication {
-            name = "frontend-pablo-server";
-            runtimeInputs = [ pkgs.miniserve ];
-            text = ''
-              miniserve -p ${
-                builtins.toString PORT
-              } --spa --index index.html ${frontend-static}/pablo
-            '';
+          # NOTE: crane can't be used because of how it vendors deps, which is incompatible with some packages in polkadot, an issue must be raised to the repo
```
Dom-Roth, line 87: Do we have a ClickUp ticket to track this?

Dom-Roth, line 87:
```
          # NOTE: crane can't be used because of how it vendors deps, which is incompatible with some packages in polkadot, an issue must be raised to the repo
```
Do we have a ClickUp ticket to track this?

dzmitry-lahoda: it is polkadot issue afaik, not us.

Dom-Roth (approved):

flake.nix
Dom-Roth: My question was more about raising the issue on their side and keeping track of their progress. 😅

dzmitry-lahoda: that is hard one. its like ask them to use nix afaik. or change their build process with some expanation about nix when they do not use it. thar is hars imho.

easteregg (approved):

flake.nix
cor: Yeah I agree with @dzmitry-lahoda, we can inform them about it, but they will most likely note care

Martin-Kiss (approved):

KaiserKarel (approved):

dzmitry-lahoda:
@mergifyio refresh

seunlanlege (approved):

vimukthi-git (approved):
Can we have a comment in each file to explain what it does for non-nix experts like my self?

PR #2181, opened 2022-11-02
[CU-34hy71m] Added RFC 0010 - Calculation of all On-Chain Existential Deposit Values

PoisonPhang:
## Issue

[[CU-34hy71m]](https://app.clickup.com/t/34hy71m)

## Additional Changes

Very much a **Request For Comments**. Not to be merged until conversation is concluded.


ComposableFiB0t:
Task linked: [CU-34hy71m [RFC] Define New Standards for ED Storage and Retrieval](https://app.clickup.com/t/34hy71m)

code/0010-define-calculations-of-existential-deposit.md
```diff
@@ -0,0 +1,149 @@
...
+
+* Will inflation/deflation be a concern as assets enter and leave the chain if
+  the ED of a foreign asset is lower/higher on our chain than it is on the
+  assets native chain?
+  
+* Can an RPC expose a function that only exists in the runtime? If not, we may
+  need to move `multi_existential_deposits` (or at least its funtionality) into
+  Asset Registry.
+  
+* Do we already have an in-use storage for AssetRegistry in Picaso?
```
PoisonPhang, line 149: >  Do we already have an in-use storage for AssetRegistry in Picaso?

code/0010-define-calculations-of-existential-deposit.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Questions
+
+* Will inflation/deflation be a concern as assets enter and leave the chain if
+  the ED of a foreign asset is lower/higher on our chain than it is on the
+  assets native chain?
+  
+* Can an RPC expose a function that only exists in the runtime? If not, we may
+  need to move `multi_existential_deposits` (or at least its functionality) into
+  Asset Registry.
```
PoisonPhang, line 147: > Can an RPC expose a function that only exists in the runtime? If not, we may need to move `multi_existential_deposits` (or at least its functionality) into Asset Registry.

code/0010-define-calculations-of-existential-deposit.md
```diff
@@ -0,0 +1,149 @@
...
+  
+* Do we already have an in-use storage for AssetRegistry in Picasso?
```
PoisonPhang, line 149: > Do we already have an in-use storage for AssetRegistry in Picasso?

code/0010-define-calculations-of-existential-deposit.md
```diff
@@ -0,0 +1,149 @@
...
+  2. **Otherwise, pull a hard-coded ED from a `match` statement**.
+  
+  3. If no matches are found, return `Balance::MAX` which will prune unknown
+     assets.
+    
+It should be updated as follows:
+
+  1. If an ED is defined for an asset in Asset Registry, return the ED as is.
+  
+  2. **If a asset ratio is defined for an asset in Asset Registry, calculate the
```
dzmitry-lahoda, line 104: what would be the way initial values initialized on devnets, testnet and prod?

dzmitry-lahoda, line 104: how values would be initialized and released to devnet, testnet, kusama, and later to polkadot?

code/0010-define-calculations-of-existential-deposit.md
dzmitry-lahoda: how would variability be handled with Composable runtime?

code/0010-define-calculations-of-existential-deposit.md
```diff
@@ -0,0 +1,149 @@
...
+The function `AssetRegistry::register_asset` should be updated with the 
+following signature:
+
+```rust
+pub fn register_asset(
+  origin: OriginFor<T>,
+  location: T::ForeignAssetId,
+  ed: Option<T::Balance>, // Used to be required
+  ratio: Ratio, // Used to be optional
+  decimals: Option<Exponent>,
```
andresmechali, line 130: For what we discussed yesterday, we should also make `decimals` required no?

dzmitry-lahoda: decimals no - because `assets` pallet allows assets without decimals.

dzmitry-lahoda: i think we should not prevent us receiving such tokens

code/0010-define-calculations-of-existential-deposit.md
vimukthi-git: I think it's good to have the list of asset exchange rates defined initially here for sure.

code/0010-define-calculations-of-existential-deposit.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Initial Standard ED Value
+
+With the following definitions: 
+
+* `local_ed` - The functional ED of a asset on our chain.
+
+* `native_token_ed` - The ED of our chains native asset.
+
+* `foreign_asset_ratio` - The ratio of 'one in the foreign asset' over 'one in
```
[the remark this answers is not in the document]
dzmitry-lahoda: that is not exchange. nobody exchanges nothing here. it may be even illegal for some users. it just pays in relevant amount. there is no exchange.

KaiserKarel (commented):
Alternatively, we could define ED as follows.

"If the amount held in an account is too low to work within our DeFi protocols because of `precision`, we consider it unusable dust.

We then do not rely on asset registry state at all, nor do we concern ourselves with USD value.

PoisonPhang:
> Alternatively, we could define ED as follows.
> 
> "If the amount held in an account is too low to work within our DeFi protocols because of `precision`, we consider it unusable dust. This threshold is the ED.
> 
> We then do not rely on asset registry state at all, nor do we concern ourselves with USD value.
@KaiserKarel 

In the context of our chain, this would mean only considering values below `0.000_000_000_001` as dust, correct?
I think that might be too low of a threshold for dust if we're considering ED to be the minimum value to take up space in our storage.

code/0010-define-calculations-of-existential-deposit.md
```diff
@@ -0,0 +1,149 @@
...
+  1. If an ED is defined for an asset in Asset Registry, return the ED as is.
+  
+  2. **Otherwise, pull a hard-coded ED from a `match` statement**.
+  
+  3. If no matches are found, return `Balance::MAX` which will prune unknown
+     assets.
+    
+It should be updated as follows:
+
+  1. If an ED is defined for an asset in Asset Registry, return the ED as is.
```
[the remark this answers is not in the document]
PoisonPhang: In this case, it would be less straight forward to set ED to a desired value. However, I would still be achievable.
I don't have a strong opinion either way.

code/0010-define-calculations-of-existential-deposit.md
PoisonPhang: Ideally, we could provide Asset Registry with a genesis config that would contain our initial assets and values. Otherwise, we'll need an approved origin to do this.

PoisonPhang: Also, we may need another extrinsic in Asset Registry for creating assets with a defined asset ID rather than just getting the next value from Currency Factory.

code/0010-define-calculations-of-existential-deposit.md
PoisonPhang: > how would variability be handled with Composable runtime?
I think this provides enough reason to move this functionality into Asset Registry and have the runtime call into it.

code/0010-define-calculations-of-existential-deposit.md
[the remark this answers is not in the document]
vimukthi-git: Just because someone does not exchange it does not mean it's not a exchange rate. Eg: Your bank may show your account balance in USD even though you are actually saving in LKR.

code/0010-define-calculations-of-existential-deposit.md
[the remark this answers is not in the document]
vimukthi-git: Any place where we could cut down on runtime code/storage, we should. So let's get rid of it.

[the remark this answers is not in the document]
dzmitry-lahoda: I do not know what is better. Acala and Statemine do store ED. Acala does ED, ratio(in very complicated manner so) and DEX.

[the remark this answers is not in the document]
dzmitry-lahoda: ED protects from data (storage spam) and ratio from CPU spam.  Weights in substrate are moddelled by having sums of inpute data, storage calls and CPU.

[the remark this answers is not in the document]
dzmitry-lahoda: Statemine assets has callback which allows using ED as way to calculate fee but they do not force to do so (so if somebody has ratio - it can override)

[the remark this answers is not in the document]
dzmitry-lahoda: Also ED may be more or less static - to avoid accounts cleanup often. While ratios can vary to allow bias toward earning some currency (so you want more statble coint? make ratio small).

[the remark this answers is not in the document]
vimukthi-git: Seems statemine is using a deterministic function to calculate ED https://github.com/paritytech/cumulus/blob/master/parachains/runtimes/assets/statemine/src/lib.rs#L217 
Better double check though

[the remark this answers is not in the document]
vimukthi-git: > callback which allows using ED as way to calculate fee but they do not force to do so

Not sure how this is done, But I think I'm also ok with resolving this for now as we can change it anytime in the future without trouble.

vimukthi-git (approved):
LGTM, as my comments are not blockers. Conceptually everything looks good.

code/0010-define-calculations-of-existential-deposit.md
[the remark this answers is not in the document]
PoisonPhang: > Seems statemine is using a deterministic function to calculate ED https://github.com/paritytech/cumulus/blob/master/parachains/runtimes/assets/statemine/src/lib.rs#L217
@vimukthi-git can you update this link? I don't think it still points to where you were wanting it to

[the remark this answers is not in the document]
vimukthi-git: It is here https://github.com/paritytech/cumulus/blob/3fb9c8a7beb8fb0fc31ea3335f8483d8086f60c8/parachains/runtimes/assets/statemine/src/constants.rs#L28 , seems someone just changed stuff around

rfcs/0010-define-calculations-of-existential-deposit.md
```diff
@@ -0,0 +1,163 @@
...
+
+  ```rust
+  pub fn register_asset_with_id(
+    origin: OriginFor<T>,
+    asset_id: T::AssetId,
+    location: T::ForeignAssetId,
+    ratio: Ratio, // Used to be optional
+    decimals: Option<Exponent>,
+  ) -> DispatchResultWithPostInfo;
+  ```
```
PoisonPhang, line 137: Looking back at this, I don't think this will be needed for Albula.

We no longer wish to directly overwrite the ED of well-known assets and with @andresmechali 's work, we don't need to worry about exposing every well-known asset through Asset Registry.

The only other use case I could see for this would be overwriting the ratios of well-known assets with Asset Registry.

Thoughts? @vimukthi-git @dzmitry-lahoda

vimukthi-git: Yes the idea is to hard code assets anyway.

rfcs/0010-define-calculations-of-existential-deposit.md
```diff
@@ -0,0 +1,150 @@
...
+  substantial for ED. In other words, a user cannot "delete" their balance when 
+  they convert it into another asset.
+
+## Method
+
+### Initial Standard ED Value
+
+With the following definitions: 
+
+* `local_ed` - The functional ED of a asset on our chain.
```
benluelo, line 34:
```suggestion
* `local_ed` - The functional ED of an asset on our chain.
```

KaiserKarel (approved):

rfcs/0010-define-calculations-of-existential-deposit.md
```diff
@@ -0,0 +1,150 @@
...
+
+assert(asset_a_ed_as_native == native_asset_ed)
+assert(asset_b_ed_as_native == native_asset_ed)
+assert(asset_a_ed_as_native == asset_b_ed_as_native)
+assert(native_as_asset_b == asset_a_as_asset_b)
+```
+
+### Well-Known Assets & Asset Registry Storage
+
+For a asset to have its ED be correctly calculated, and therefore to exists on
```
benluelo, line 77:
```suggestion
For an asset to have its ED correctly calculated, and therefore to exist on 
```

rfcs/0010-define-calculations-of-existential-deposit.md
```diff
@@ -0,0 +1,150 @@
...
+     is.**
+  
+  2. **Otherwise, pull a hard-coded ED from a `match` statement**.
+  
+  3. If no matches are found, return `Balance::MAX` which will prune unknown
+     assets.
+    
+It should be updated as follows:
+
+  1. **If a asset ratio is defined for an asset in Asset Registry or as a
```
benluelo, line 97:
```suggestion
  1. **If an asset ratio is defined for an asset in Asset Registry or as a 
```

rfcs/0010-define-calculations-of-existential-deposit.md
```diff
@@ -0,0 +1,150 @@
...
+to be possible.
+
+* The function `AssetRegistry::register_asset` should be updated with the 
+  following signature:
+
+  ```rust
+  pub fn register_asset(
+    origin: OriginFor<T>,
+    location: T::ForeignAssetId,
+    ratio: Ratio, // Used to be optional
```
benluelo, line 121:
```suggestion
    ratio: Rational64, // Used to be optional
```
Was recently updated to new type

rfcs/0010-define-calculations-of-existential-deposit.md
```diff
@@ -0,0 +1,150 @@
...
+  
+  > This is not really a solvable problem AFAIK given that we don't control
+    validations done on other chains when transferring to Picasso.
+  
+* Can an RPC expose a function that only exists in the runtime? If not, we may
+  need to move `multi_existential_deposits` (or at least its functionality) into
+  Asset Registry.
+  
+  Moving this functionality into Asset Registry will be necessary for 
+  different runtimes to have different asset configurations.
```
benluelo, line 145: Should this be `>`'d?

benluelo (approved):
LGTM, just some minor nits

andyjsbell (approved):

rfcs/0010-define-calculations-of-existential-deposit.md
PoisonPhang: It's not a quote, this was my own summarized of an answer

PoisonPhang: I get where your coming from in terms of formatting Q/A. Will update

benluelo (approved):

PR #2205, opened 2022-11-07
deleted old #ibc config

dzmitry-lahoda:

## Issue

In main repo we have old IBC pallet infrastructure in runtime which is not compatible with new version in its API and config (router is not fully configurable in runtime as I see). As pers discussion with bridge team we should use centauri repo.

Also while adding new IBC, we should ask if extensions of runtime functions deleted is needed, because afaik these may not work on parity relay.

Some useful code (specifically mapping Substrate account and IBC account can be restored from this commit when adding new IBC.

## Checklist

- CI green


Wizdave97 (approved):

dzmitry-lahoda:
https://composablefinance.slack.com/archives/D049FC12Z61/p1667813737952259

dzmitry-lahoda:
image

PR #2211, opened 2022-11-07
[cu-31gytxh][RFC-0011] Release from tags

vimukthi-git:
- Only allow release from a tag formatted `release-v[0-9]+\.[0-9]+` 
- RFC


ComposableFiB0t:
Task linked: [CU-31gytxh Update release artifacts workflow](https://app.clickup.com/t/31gytxh)

rfcs/0011-releases.md
```diff
@@ -0,0 +1,73 @@
...
+
+Eg: 
+
+For Picasso FE : `fe-v2.4-picasso-abcd`
+
+For Subsquid : `subsquid-v2.5-1.0.1`
+
+etc.
+
+#### 3.1.1. Expected Typical Release Artefact List
```
vimukthi-git, line 46:
```suggestion
#### 3.1.1. Expected Typical Release Artifact List 
```

.github/workflows/release-artifacts.yml
```diff
@@ -33,6 +33,10 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Exit if not on a release-* branch
+        if: startsWith(github.ref, 'release-') == false
+        run: exit -1
```
KaiserKarel, line 38: Is this needed, since we have the `tags` key in the `on`?

rfcs/0011-releases.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+### 3.2. Release Process
+
+Typical Composable releases involve multiple rounds of QA and external audits/testing that may cause multiple patch(rc) versions to be released based on feedback/issues. This means that a release branch may have a longer maintenance life cycle independent of the main branch where most of the bleeding edge development happens. In order to execute this expected workflow, following release process steps are proposed.
+
+As the work starts for a `major.minor` (eg: 2.5) release,
+
+1. Create a branch `release-v2.5`.
+2. in order to make/deploy (in staging) a release create a tag `release-v2.5.0` which should trigger a workflow.
```
KaiserKarel, line 70: So you create a branch with release-xyz format, and then additionally tag that branch with a tag that is the same to the name? Shall we just omit 2?

rfcs/0011-releases.md
```diff
@@ -0,0 +1,73 @@
...
+2. QA/Audit happens on these released tag.
+3. Any reported issues must be fixed on `main` and merged/cherry picked to the `release-v2.5` branch. Then a tag should be created for the next round and so on until "release-able" version is found.
+3. Node and runtimes are release together from the same tag while other components(eg: fe) must have their own tag/workflows to release.
```
KaiserKarel, line 73:
```suggestion
3. QA/Audit happens on these released tag.
4. Any reported issues must be fixed on `main` and merged/cherry picked to the `release-v2.5` branch. Then a tag should be created for the next round and so on until "release-able" version is found.
5. Node and runtimes are release together from the same tag while other components(eg: fe) must have their own tag/workflows to release.
```

KaiserKarel, line 73: Might be nice to still include these in the release, even if their versions haven't been bumped? Then a single release in the overview always contains all components so you don't need to look and match compatible versions yourself.

.github/workflows/release-artifacts.yml
vimukthi-git: Right this check is supposed to be about the current branch, though it may not do that as it stands, let me check more

rfcs/0011-releases.md
vimukthi-git: Branch format is different from tag. Branch is just major.minor version.

rfcs/0011-releases.md
```diff
@@ -0,0 +1,73 @@
...
+### 3.2. Release Process
+
+Typical Composable releases involve multiple rounds of QA and external audits/testing that may cause multiple patch(rc) versions to be released based on feedback/issues. This means that a release branch may have a longer maintenance life cycle independent of the main branch where most of the bleeding edge development happens. In order to execute this expected workflow, following release process steps are proposed.
+
+As the work starts for a `major.minor` (eg: 2.5) release,
+
+1. Create a branch `release-v2.5`.
+2. in order to make/deploy (in staging) a release create a tag `release-v2.5.0` which should trigger a workflow.
+2. QA/Audit happens on these released tag.
+3. Any reported issues must be fixed on `main` and merged/cherry picked to the `release-v2.5` branch. Then a tag should be created for the next round and so on until "release-able" version is found.
```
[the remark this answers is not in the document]
vimukthi-git: We do not want automated next versions from commits as that would not allow us to organize releases with suitable grouping.

rfcs/0011-releases.md
vimukthi-git: Appologies!

.github/workflows/release-artifacts.yml
vimukthi-git: Removed, as it is meaningless with a tag

rfcs/0011-releases.md
vimukthi-git: Yes we should be able to automate this further to include all components together in a single release page. cc @easteregg what do you think?

rfcs/0011-releases.md
KaiserKarel: So we're planning to skip minor version during QA? So we start of attempting to release v2.5, and end up with v2.5.6?

KaiserKarel (approved):

rfcs/0011-releases.md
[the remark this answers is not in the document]
KaiserKarel: Ah I am derping. So we create a branch for each MAJOR-MINOR release, and then PATCH is incremented using tags?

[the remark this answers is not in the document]
KaiserKarel: I read this as: 1 branch called release, with tags, and to create a new tag on that release, first create a `release-xy` branch, test that, and if that gets merged, the release branch is tagged with the last tag of the `release-xy` branch. (Which seemed convoluted to me)

rfcs/0011-releases.md
```diff
@@ -0,0 +1,73 @@
...
+
+Each composable node release involves a release of (at-least) the following components,
+
+1. The Composable Node - main node executable
+
+    Node follows [semver](https://semver.org/) versioning with an additional `v` prefix. Eg: `v2.3.1`. 
+
+2. Runtime wasm - Dali, Picasso and Composable at the time of this writing.
+
+    In order to allow clear identification with the native runtime version of each node release the runtime version for each runtime is an integer derived by concatenating all numbers in the node version and adding to 10000. Eg: `v2.5.1` -> `251` -> 251 + 10000 -> (`dali-10251`, `picasso-10251`, `composable-10251`). The 10000 addition is to prevent collision of version numbers of this scheme with earlier versions before the scheme was introduced.
```
obsessed-cake, line 24: Why did you choose an addition? Perhaps, concatenation is better, e.g.
```
λ> "10" <> "251"
"10251"

λ> "10" <> "10251"
"1010251"
```
So, we can avoid collisions with future releases.

rfcs/0011-releases.md
KaiserKarel: Might it be better then to reduce it to a single number, like postgres does? Since the Major/Minor seem to not carry much information?

KaiserKarel: Albula being release 1, but fixes to albula being 1.1, 1.14 etc?

vimukthi-git: Sorry my wording was not correct before, the major version is always 2 above for v2.5, 5 being the minor. 

> Albula being release 1, but fixes to albula being 1.1, 1.14 etc?

I don't have a particular issue with this other than it breaking the previous scheme we were following for the node.

rfcs/0011-releases.md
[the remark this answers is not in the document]
vimukthi-git: Right let me try to reword this to be clear.

dzmitry-lahoda:
Not sure about non Rust. But for Rust.

```rust
pub struct RuntimeVersion {
	pub authoring_version: u32,
	pub spec_version: u32,
	pub impl_version: u32,
	pub transaction_version: u32,
	pub state_version: u8,
}
```
There are several numbers for runtime.

Our `Cargo.toml`   has `version`. Also our dependencies has versions. (cumulus/orml/polkadot/substrate). 

Can these somehow to be used to version things naturally?


Could also obtain 2 numbers for versioning:
 ```
<current_picasso_block_height>.<git rev-parse --short=16>
```

**Comment by vimukthi-git** [comment]
@dzmitry-lahoda This document is only concerned about the components we directly release. Any other library/component versions included within released components are out of scope as per the requirement.

Regarding the various versions included in the `RuntimeVersion`, guidance is being published at https://github.com/ComposableFi/composable/pull/2122, which I can link to this document once it's merged.

File: rfcs/0011-releases.md (hunk 6 above)
Comment (vimukthi-git) [reply rfcs/0011-releases.md:24]: The function specified here is still a "one to one" function(y = x + 10_000) for any node release version(for vA.B.C, x = ABC). Could you qualify why your suggestion is "better"? The function you propose has a downside of producing larger numbers unnecessarily, eg: "10" + "10251" = 1010251 instead of 10000 + 10251 = 20251.

**Comment by dzmitry-lahoda** [comment]
@vimukthi-git semantic versioning in nonsense

File: rfcs/0011-releases.md (hunk 3 above)
Comment (KaiserKarel) [reply rfcs/0011-releases.md:70]: We would then have that each component has an indepedent semver, compatible with the language standards of that ecosystem, and a monotonic increasing counter for the full release.

**Comment by vimukthi-git** [comment]
@dzmitry-lahoda software project maintainer decides if semantic versioning becomes non-sense or not by their actions. If we care about semantic versioning then we need to understand the changes to dependencies and how they are categorized in terms of effect on the users of our software (a dependency making major version update does not necessarily translate to a major version update to users of our software).

File: rfcs/0011-releases.md (hunk 3 above)
Comment (andyjsbell) [reply rfcs/0011-releases.md:70]: If we are happy to break the previous scheme, which I am, then it could be nice to have this encapsulating version `Albula` with its subcomponents versioned independently.  @vimukthi-git I suppose this would mean a change or two right?

**Review by dzmitry-lahoda — APPROVED** [review]

**Comment by dzmitry-lahoda** [comment]
@vimukthi-git feature slices could alternative way(alternative to artifacts semantic versioning) to track user facing functional. so what ever allows to release faster is good.

File: rfcs/0011-releases.md [hunk 7]
```diff
@@ -0,0 +1,73 @@
+# Releasing Composable Node and Related Components
+
+## 1. Abstract
+
+Given the complexity of the number of components and teams involved a standard process is needed to release Composable node and related components.
+
+## 2. Goals
+
+1. Define release numbering convention.
```
vimukthi-git, line 9:
```suggestion
1. Define release numbering convention.
2. Define branching strategy for releases.
```

rfcs/0011-releases.md
vimukthi-git: Had a call with @KaiserKarel on this, so the discussion points were,
1. What are the benefits of having a `major.minor.patch` version, when we could probably manage with just `major.patch`. 
2. Then another idea was that we could use `major=spec_version` and then node patches without runtime change could be `spec_version.patch`.

After discussion I feel 2) may complicate the branching proposed here with each release having a branch. So I propose going with `major.minor` where `major=branch name eg: 5` and `minor=spec_version eg: 2400` for the node main version. Where 
1. Each node only patch release would require a new branch, with `spec_version` remaining the same. 
2. Each runtime only patch requires keeping updates in the same branch as before.

vimukthi-git: Discussed with @andyjsbell and we will do following
```
Do major.minor where major=branch name eg: 5 and minor=spec_version eg: 2400 for the node main version
```

LyonSsS (approved):
Everything seems fine, clear from the QA scope.

rfcs/0011-releases.md
vimukthi-git: Resolving as this scheme is modified now

vimukthi-git:
@mergifyio refresh

PR #2228, opened 2022-11-09
[CU-3383bzz] fix list_assets rpc

andresmechali:
https://app.clickup.com/t/3383bzz

### DIFF
--- a/code/parachain/runtime/primitives/src/currency.rs
+++ b/code/parachain/runtime/primitives/src/currency.rs
@@ -129,7 +129,7 @@ macro_rules! list_assets {
 				$(Asset {
 					id: CurrencyId::$NAME.0 as u128,
 					name: Some(stringify!($NAME).as_bytes().to_vec()),
-					decimals: 12_u8.into(),
+					decimals: Self::decimals(),
 					foreign_id: Self::local_to_xcm_reserve(CurrencyId::$NAME).map(XcmAssetLocation::new)
 				},)*
 			]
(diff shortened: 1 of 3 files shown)


ComposableFiB0t:
Task linked: [CU-3383bzz Statemine USDT - Updating decimals for RPC assets](https://app.clickup.com/t/3383bzz)

dzmitry-lahoda (commented):
@benluelo please sync up on this. As you will see in your PR currency.rs was fixed and will have conflict.

code/parachain/runtime/primitives/src/currency.rs
```diff
@@ -89,22 +89,12 @@ macro_rules! list_assets {
...
-			None
-		}
-
 		pub fn list_assets() -> Vec<Asset<XcmAssetLocation>> {
 			[
 				$(Asset {
 					id: CurrencyId::$NAME.0 as u128,
 					name: Some(stringify!($NAME).as_bytes().to_vec()),
-					decimals: 12_u8.into(),
+					decimals: Self::decimals(),
```
dzmitry-lahoda, line 97: not sure if we can break UI now, but better assumption could be if decimals are made Option

dzmitry-lahoda: but that just wish, very low prio and not blocker.

andresmechali: As far as I understand, we are not planning to include assets that don't have decimals. At least not for now.

PoisonPhang (approved):

Dom-Roth (approved):

PR #2231, opened 2022-11-09
#rfc hardcoded static assets

dzmitry-lahoda:

## Issue

How we can release assets.

### DIFF
--- a/rfcs/0012-hardcoded-static-assets.md
+++ b/rfcs/0012-hardcoded-static-assets.md
@@ -1,0 +1,58 @@
+# Overview
+
+This RFC argues usage of hardcoded assets routes, identifiers, and metadata are generally reasonable.  
+
+It describes how one would initialize, update and override static assets.
+
+Hopefully, that description will be good enough to favor hardcoded assets as a viable approach for asset engineering.
+
+## Prereqisites
+
+Reader clear on general consequences of code as data, data as code, and code vs. data approaches for configuration and other small data parameters.
+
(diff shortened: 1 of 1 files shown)


rfcs/0012-hardcoded-static-assets.md
```diff
@@ -0,0 +1,72 @@
...
+RPC and extrinsic check runtime configured asset state before using 
+hardcode.
+
+Foreign locations and symbols will not change.
+
+## Who can update hardcoded assets?
+
+The root can update hardcoded assets. Such assets are permissioned.
+
+## Who uses hardcoded assets?
```
vimukthi-git, line 42: Could you link to the place in code they hard code? Just want to see an example of what this means in the runtime.

KaiserKarel (approved):
Agree, @andyjsbell and @vimukthi-git have the final call from me on this, but discussed it in DMs too and think it is very reasonable and reduces operational load during launch.

rfcs/0012-hardcoded-static-assets.md
dzmitry-lahoda: currency.rs in https://github.com/ComposableFi/composable/pull/2118

PoisonPhang: code/parachain/runtime/primitives/src/currency.rs

vimukthi-git (approved):
Prefer if we follow some standard format for these docs where we layout the implementation better. Anyway LGTM for now.

andyjsbell (commented):
lgtm

PR #2239, opened 2022-11-09
Cosmwasm Stargate SG-1

hussein-aitlahcen:
- upgrade VM to the latest, stargate capable
- upgrade host functions
- implement pallet stubs

### DIFF
--- a/code/parachain/frame/cosmwasm/src/runtimes/wasmi.rs
+++ b/code/parachain/frame/cosmwasm/src/runtimes/wasmi.rs
@@ -344,7 +344,7 @@ impl<'a, T: Config> VMBase for CosmwasmVM<'a, T> {
 		&mut self,
 		address: Self::Address,
 		message: &[u8],
-	) -> Result<cosmwasm_minimal_std::QueryResult, Self::Error> {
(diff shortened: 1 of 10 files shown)


code/parachain/frame/cosmwasm/src/runtimes/wasmi.rs
```diff
@@ -473,25 +426,33 @@ impl<'a, T: Config> VMBase for CosmwasmVM<'a, T> {
 		}
 	}
 
-	fn db_scan(
+	fn addr_validate(&mut self, input: &str) -> Result<Result<(), Self::Error>, Self::Error> {
```
blasrodri, line 429: why Result of Result?

hussein-aitlahcen: This change is a reorder by rust analyzer and unrelated to the pr. Other than that, it's a double result because the inner one does not halt execution and is returned to the contract. The outer one would halt exec

dzmitry-lahoda (approved):

aeryz (approved):

dzmitry-lahoda (approved):

PR #2266, opened 2022-11-11
[CU-3rxr068] remove unnecessary .expect and add change clippy config

andresmechali:
## Issue

https://app.clickup.com/t/3rxr068

### DIFF
--- a/code/parachain/frame/vesting/src/lib.rs
+++ b/code/parachain/frame/vesting/src/lib.rs
@@ -34,7 +34,6 @@
 		clippy::disallowed_types,
 		clippy::indexing_slicing,
 		clippy::todo,
-		clippy::expect_used
 	),
 	deny(clippy::unwrap_used, clippy::panic)
 )] // allow in tests
--- a/code/parachain/frame/vesting/src/lib.rs
+++ b/code/parachain/frame/vesting/src/lib.rs
@@ -34,9 +34,9 @@
 		clippy::disallowed_types,
 		clippy::indexing_slicing,
 		clippy::todo,
-		clippy::unwrap_used,
-		clippy::panic
-	)
+		clippy::expect_used
+	),
+	deny(clippy::unwrap_used, clippy::panic)
 )] // allow in tests
 #![warn(clippy::unseparated_literal_suffix)]
 #![cfg_attr(not(feature = "std"), no_std)]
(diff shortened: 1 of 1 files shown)


ComposableFiB0t:
Task linked: [CU-3rxr068 Replace every .expect on the Vesting pallet by proper errors](https://app.clickup.com/t/3rxr068)

code/parachain/frame/vesting/src/lib.rs
```diff
@@ -606,7 +610,7 @@ impl<T: Config> Pallet<T> {
 
 					claims_per_schedule
 						.try_insert(*id_to_claim, available_amount)
```
vimukthi-git, line 612: This `expect` is ok IMO as there is no chance(invariant) of it having a `MaxVestingSchedulesExceeded` given it is reinserting a value that was read in this same code.

code/parachain/frame/vesting/src/lib.rs
```diff
@@ -34,9 +34,9 @@
 		clippy::disallowed_types,
 		clippy::indexing_slicing,
 		clippy::todo,
-		clippy::unwrap_used,
-		clippy::panic
-	)
+		clippy::expect_used
+	),
+	deny(clippy::unwrap_used, clippy::panic)
```
vimukthi-git, line 39: What's the expected effect here?

andresmechali: To forbid `unwrap` and `panic` and warn `expect`s.

code/parachain/frame/vesting/src/lib.rs
andresmechali: Yeah, but I was suggested not to have any `expect` besides the `GenesisBuild`, and I guess it doesn't hurt to remove it right?

benluelo:
Why not make GenesisConfig.vesting a BoundedVec with the same size limit as the max vesting schedules?

code/parachain/frame/vesting/src/lib.rs
vimukthi-git: In the build we don't allow "warn" level stuff also, @benluelo is this line needed?

andresmechali: Well I'll better remove those warnings as things like `#[pallet::pallet]` and the `expect`s from the `GenesisBuild` will trigger them.

benluelo: `expect` should be allowed, and all uses of it should be audited

andresmechali:
> Why not make GenesisConfig.vesting a BoundedVec with the same size limit as the max vesting schedules?

@benluelo that makes sense

vimukthi-git:
> Why not make GenesisConfig.vesting a BoundedVec with the same size limit as the max vesting schedules?

This is ok, but I don't see this as a critical fix for release as the genesis is already done. So let's keep that separate.

andresmechali:
> > Why not make GenesisConfig.vesting a BoundedVec with the same size limit as the max vesting schedules?
> 
> This is ok, but I don't see this as a critical fix for release as the genesis is already done. So let's keep that separate.

Alright. Do we even need to handle schedules in the `GenesisBuild`? (see https://app.clickup.com/t/3rxr068?comment=1795517562)

PR #2275, opened 2022-11-11
xcm-review: reviewing xcm config

blasrodri:
## Issue
ref: https://app.clickup.com/t/35mag8u

### DIFF
--- a/code/integration-tests/local-integration-tests/src/cross_chain_transfer.rs
+++ b/code/integration-tests/local-integration-tests/src/cross_chain_transfer.rs
@@ -801,6 +801,7 @@ fn unspent_xcm_fee_is_returned_correctly() {
 	});
 }
 
+// what is ED?
 #[test]
 fn trap_assets_larger_than_ed_works() {
 	simtest();
@@ -824,11 +825,13 @@ fn trap_assets_larger_than_ed_works() {
 			);
 		// TODO: if we do not top up account initially, than any deposit_creating do not create
 		// anything may be something with zero block or like - fix it better way
+		// -> I don't understand what's the action here
+
 		let _ =
 			<balances::Pallet<Runtime> as frame_support::traits::Currency<AccountId>>::deposit_creating(
-				&this_runtime::TreasuryAccount::get(),
-				7 * CurrencyId::unit::<Balance>(),
-			);
+			&this_runtime::TreasuryAccount::get(),
+			7 * CurrencyId::unit::<Balance>(),
+		);
 
 		native_treasury_amount =
 			Assets::free_balance(CurrencyId::PICA, &this_runtime::TreasuryAccount::get());
--- a/code/parachain/frame/assets-registry/src/lib.rs
+++ b/code/parachain/frame/assets-registry/src/lib.rs
@@ -307,7 +307,7 @@ pub mod pallet {
 						.expect("Must exist, as it does in ForeignToLocal");
 					let decimals = match foreign_metadata.decimals {
 						Some(exponent) => exponent,
-						_ => 12_u8,
+						_ => 12_u8, // add doc that 12 is the default exponent?
 					};
 
 					Asset {
@@ -317,7 +317,7 @@ pub mod pallet {
 						foreign_id: Some(foreign_metadata.location),
 					}
 				})
-				.collect::<Vec<_>>()
+				.collect()
 		}
 	}
 
--- a/code/parachain/runtime/common/src/xcmp.rs
+++ b/code/parachain/runtime/common/src/xcmp.rs
@@ -132,16 +132,19 @@ impl<
 		Err(XcmError::TooExpensive)
 	}
 
+	// how does this refunding works? - add docs
 	fn refund_weight(&mut self, weight: Weight) -> Option<MultiAsset> {
 		if let Some(ref asset_location) = self.asset_location {
 			let fee = WeightToFeeConverter::weight_to_fee(&weight);
 			let fee = self.fee.min(fee);
 			let price = fee.saturating_mul(self.price) / self.fee;
+			if price <= 0 {
+				// optimization: check for price before and return None
+				return None
+			}
 			self.price = self.price.saturating_sub(price);
 			self.fee = self.fee.saturating_sub(fee);
-			if price > 0 {
-				return Some((asset_location.clone(), price).into())
-			}
+			return Some((asset_location.clone(), price).into())
 		}
 
 		None
@@ -243,6 +246,8 @@ impl<
 			MultiLocation { parents: 0, interior: X1(GeneralIndex(index)) } =>
 				Some(CurrencyId(index)),
 			_ =>
+			// there is a risk with this `_` that if a new Junction gets added, it will come
+			// directly to this path are we sure that this is futureproof?
 				if let Some(currency_id) = WellKnownXcmpAssets::remote_to_local(location.clone()) {
 					Some(currency_id)
 				} else {
--- a/code/parachain/runtime/composable/src/xcmp.rs
+++ b/code/parachain/runtime/composable/src/xcmp.rs
@@ -195,7 +195,7 @@ impl pallet_xcm::Config for Runtime {
 	type Origin = Origin;
 	type Call = Call;
 
-	const VERSION_DISCOVERY_QUEUE_SIZE: u32 = 100;
+	const VERSION_DISCOVERY_QUEUE_SIZE: u32 = 100; // any reasoning behind this number?
 	type AdvertisedXcmVersion = pallet_xcm::CurrentXcmVersion;
 }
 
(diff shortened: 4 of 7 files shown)


code/parachain/runtime/common/src/xcmp.rs
```diff
@@ -243,6 +246,8 @@ impl<
 			MultiLocation { parents: 0, interior: X1(GeneralIndex(index)) } =>
 				Some(CurrencyId(index)),
 			_ =>
+			// there is a risk with this `_` that if a new Junction gets added, it will come
```
dzmitry-lahoda, line 249: if Root will not configured bad path it will be ok

code/parachain/runtime/composable/src/xcmp.rs
```diff
@@ -195,7 +195,7 @@ impl pallet_xcm::Config for Runtime {
 	type Origin = Origin;
 	type Call = Call;
 
-	const VERSION_DISCOVERY_QUEUE_SIZE: u32 = 100;
+	const VERSION_DISCOVERY_QUEUE_SIZE: u32 = 100; // any reasoning behind this number?
```
dzmitry-lahoda, line 198: copy pasted from Acala and Statemine. That is how much QueryVersions we handles. Will reject if too many. In other place we accept request only from UMP and HRMP from Relay and from Parachains.

code/parachain/runtime/dali/src/xcmp.rs
```diff
@@ -253,7 +253,7 @@ pub fn xcm_asset_fee_estimator(instructions: u8, asset_id: CurrencyId) -> Balanc
 	assert!((instructions as u32) <= MaxInstructions::get());
 	let total_weight = UnitWeightCost::get() * instructions as u64;
 	Trader::weight_to_asset(total_weight, asset_id)
-		.expect("use only in simulator")
+		.expect("use only in simulator") // Isn't it better to have a default value than potentially panic?
```
dzmitry-lahoda, line 256: this is for tests, may be move it somewhere. so it can be used later in RPC (sure uwnrap to be removed)

code/parachain/frame/assets-registry/src/lib.rs
```diff
@@ -307,7 +307,7 @@ pub mod pallet {
 						.expect("Must exist, as it does in ForeignToLocal");
 					let decimals = match foreign_metadata.decimals {
 						Some(exponent) => exponent,
-						_ => 12_u8,
+						_ => 12_u8, // add doc that 12 is the default exponent?
```
dzmitry-lahoda, line 310: this should be removed and return None if no exponent.

code/integration-tests/local-integration-tests/src/cross_chain_transfer.rs
```diff
@@ -824,11 +825,13 @@ fn trap_assets_larger_than_ed_works() {
 			);
 		// TODO: if we do not top up account initially, than any deposit_creating do not create
```
dzmitry-lahoda, line 826: can remove comment

code/integration-tests/local-integration-tests/src/cross_chain_transfer.rs
```diff
@@ -824,11 +825,13 @@ fn trap_assets_larger_than_ed_works() {
...
 		// anything may be something with zero block or like - fix it better way
+		// -> I don't understand what's the action here
```
dzmitry-lahoda, line 828: need to recheck test. 
but idea is that non deposited assets trapped.
but if account of trap non existing for asset - it will not trap (because amount is lower than ED and cannot create account at all). 
better to clean up and split test

dzmitry-lahoda: So it is question if need to init account with some assets on first ever transfer (these are non user account, parachains/relaychain/treasury accounts) or somehow allow well known account to ignore ED requirement.

code/parachain/runtime/common/src/xcmp.rs
```diff
@@ -132,16 +132,19 @@ impl<
 		Err(XcmError::TooExpensive)
 	}
 
+	// how does this refunding works? - add docs
 	fn refund_weight(&mut self, weight: Weight) -> Option<MultiAsset> {
 		if let Some(ref asset_location) = self.asset_location {
 			let fee = WeightToFeeConverter::weight_to_fee(&weight);
 			let fee = self.fee.min(fee);
 			let price = fee.saturating_mul(self.price) / self.fee;
+			if price <= 0 {
```
dzmitry-lahoda, line 141: price should never be 0. should recheck that and do 0.max(1)

code/parachain/runtime/common/src/xcmp.rs
```diff
@@ -132,16 +132,19 @@ impl<
 		Err(XcmError::TooExpensive)
 	}
 
+	// how does this refunding works? - add docs
```
dzmitry-lahoda, line 135: it seems overcomplicated.  should check statemine.
but.
idea is that on start of TX we collect price.
after tx PICA fee could change. BUT we should not change PRICE, we use new FEE ratio.
so it avoids manipulation DEX price to lower fee I guess (from Acala) or just optimization.
1. save PICA fee == USD price 
2. used PICA fee, find new USD price based on old/new PICA * old USD price
3. tell XCM actual price (before we told oiringal expected PRICE)
4. on drop it will fund treasury

code/parachain/runtime/common/src/xcmp.rs
```diff
@@ -132,16 +132,19 @@ impl<
...
 	fn refund_weight(&mut self, weight: Weight) -> Option<MultiAsset> {
 		if let Some(ref asset_location) = self.asset_location {
 			let fee = WeightToFeeConverter::weight_to_fee(&weight);
 			let fee = self.fee.min(fee);
 			let price = fee.saturating_mul(self.price) / self.fee;
```
dzmitry-lahoda, line 140: multiply bar ration to be used here

code/parachain/runtime/common/src/xcmp.rs
dzmitry-lahoda: so yeah - better to test it.

code/parachain/runtime/common/src/xcmp.rs
dzmitry-lahoda: sure. this is correct. that is price_reduction, and fee is fee_reduction. better to rename for clarity. == 0

code/integration-tests/local-integration-tests/src/cross_chain_transfer.rs
dzmitry-lahoda: TLRD; if Treasury or other blockchain proxy account does not exist, and we tries to deposit (on transfer) or tip with fee (into treasury) it will fail. so this is bug. at least treasury must be ED 1 for all currencies IMHO. this is kind of bug/issue - can look how Acala/Statemine handle it (but test originally was from Acala)

dzmitry-lahoda: cc @vimukthi-git - not urgent, but may be an issue

code/parachain/runtime/common/src/xcmp.rs
dzmitry-lahoda: @vimukthi-git not an issue, but better to document this or write 1 unit test. logic seems right

code/parachain/runtime/common/src/xcmp.rs
dzmitry-lahoda: > // how does this refunding works? - add docs

See `drop` impl in the end of file. better group end to end fee/price/drop together if not.

code/parachain/runtime/common/src/xcmp.rs
dzmitry-lahoda: fee and price are needed because price is frozen on start of TX. that why it records price and fee. instead of refunding based on new price (may help against some exploits of unfairness I guess)

code/parachain/runtime/composable/src/xcmp.rs
dzmitry-lahoda: we need our monitor, sre, bots or bridge team to start metric in prometheus size of our queues. likely other teams do kind of this. non urgent. but should be planned. IBC will be similar. cc

code/parachain/runtime/common/src/xcmp.rs
dzmitry-lahoda: https://composablefinance.slack.com/archives/C031G5NT0CA/p1668168585791579

cc @vimukthi-git

dzmitry-lahoda: not urgent not critical, but can be issue later.

code/parachain/runtime/dali/src/xcmp.rs
dzmitry-lahoda: best is to have None in code. and wrapper for tests with expect (tests expect fee asked for only sufficient codes)

blasrodri:
I have no further comments. None of my findings are critical. Thanks @dzmitry-lahoda for reviewing them, and responding to each of these. I believe we can close this PR for now and reference it on the PRs that tackle what I've pointed out?

dzmitry-lahoda:
@vimukthi-git may you help us to decide when we should fix issue?

dzmitry-lahoda:
creating maintainance task on Picasso.

PR #2281, opened 2022-11-11
[CU-35txqkb] Removed AR Overwrite of ED

PoisonPhang:
## Issue

[[CU-35txqkb]](https://app.clickup.com/t/35txqkb)
[[CU-35txtwh]](https://app.clickup.com/t/35txtwh)

## Additional Changes

* `WellKnownForeignToNativePriceConverter::to_asset_balance()` now returns `None` instead of `Balance::one()` when encountering an `ArithmeticError` 
* Currency Factory no longer accepts or tracks ED when creating or storing assets. Given that the storage map for ED was independent of any other storage items, it should not need any storage migrations when getting removed.

### DIFF
--- a/code/parachain/frame/assets-registry/src/lib.rs
+++ b/code/parachain/frame/assets-registry/src/lib.rs
@@ -31,7 +31,12 @@ pub mod weights;
 pub mod pallet {
 	use crate::prelude::*;
 	pub use crate::weights::WeightInfo;
-	use composable_traits::currency::{CurrencyFactory, ForeignByNative, RangeId};
+	use codec::FullCodec;
+	use composable_traits::{
+		assets::Asset,
+		currency::{BalanceLike, CurrencyFactory, Exponent, ForeignByNative, RangeId},
+		xcm::assets::{ForeignMetadata, RemoteAssetRegistryInspect, RemoteAssetRegistryMutate},
+	};
 	use cumulus_primitives_core::ParaId;
 	use frame_support::{
 		dispatch::DispatchResultWithPostInfo, pallet_prelude::*, traits::EnsureOrigin,
@@ -75,8 +80,7 @@ pub mod pallet {
 		type ParachainOrGovernanceOrigin: EnsureOrigin<Self::Origin>;
 		type WeightInfo: WeightInfo;
 		type Balance: BalanceLike;
-		type CurrencyFactory: CurrencyFactory<AssetId = Self::LocalAssetId, Balance = Self::Balance>
-			+ AssetExistentialDepositInspect<AssetId = Self::LocalAssetId, Balance = Self::Balance>;
+		type CurrencyFactory: CurrencyFactory<AssetId = Self::LocalAssetId, Balance = Self::Balance>;
 	}
 
 	#[pallet::pallet]
@@ -186,8 +190,7 @@ pub mod pallet {
 		pub fn register_asset(
 			origin: OriginFor<T>,
 			location: T::ForeignAssetId,
-			ed: T::Balance,
-			ratio: Option<Rational>,
+			ratio: Rational,
 			decimals: Option<Exponent>,
 		) -> DispatchResultWithPostInfo {
 			T::UpdateAssetRegistryOrigin::ensure_origin(origin)?;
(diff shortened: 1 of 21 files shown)


ComposableFiB0t:
Task linked: [CU-35txqkb Update multi_existential_deposits](https://app.clickup.com/t/35txqkb)

PoisonPhang:
Anticipating needing to resolve merge conflicts and retest after #2313

code/parachain/frame/assets-registry/src/lib.rs
```diff
@@ -258,12 +261,12 @@ pub mod pallet {
...
 			asset_id: Self::AssetId,
 			location: Self::AssetNativeLocation,
-			ratio: Option<Rational>,
+			ratio: Rational,
 			decimals: Option<Exponent>,
 		) -> DispatchResult {
 			ForeignToLocal::<T>::insert(&location, asset_id);
 			LocalToForeign::<T>::insert(asset_id, ForeignMetadata { decimals, location });
-			AssetRatio::<T>::mutate_exists(asset_id, |x| *x = ratio);
+			AssetRatio::<T>::mutate_exists(asset_id, |x| *x = Some(ratio));
```
andresmechali, line 269: Given that `ratio` is no longer an `Option` shouldn't we also change it in the storage?

vimukthi-git (commented):
Let's remove/modify storage modifications. Let's create tickets to do that with a later release and a data migration which drains storage.

code/parachain/frame/assets-registry/src/lib.rs
andresmechali: Resolving as it's been decided not to modify storage for now

code/parachain/frame/currency-factory/src/lib.rs
```diff
@@ -114,10 +111,6 @@ pub mod pallet {
 	pub type AssetIdRanges<T: Config> =
 		StorageValue<_, Ranges<T::AssetId>, ValueQuery, RangesOnEmpty<T>>;
 
-	#[pallet::storage]
-	#[pallet::getter(fn get_assets_ed)]
-	pub type AssetEd<T: Config> = StorageMap<_, Twox128, T::AssetId, T::Balance, OptionQuery>;
```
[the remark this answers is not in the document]
PoisonPhang: Shouldn't it get drained from storage as soon as blocks are created that don't reference the item in their Merkle root?

code/parachain/runtime/common/src/fees.rs
```diff
@@ -48,15 +47,10 @@ pub fn multi_existential_deposits<AssetsRegistry>(_currency_id: &CurrencyId) ->
...
-	AssetsRegistry: AssetRatioInspect<AssetId = CurrencyId>
-		+ AssetExistentialDepositInspect<AssetId = CurrencyId, Balance = Balance>,
->(
+pub fn multi_existential_deposits<AssetsRegistry: AssetRatioInspect<AssetId = CurrencyId>>(
 	currency_id: &CurrencyId,
 ) -> Balance {
-	AssetsRegistry::existential_deposit(*currency_id)
-		.ok()
-		.or_else(|| WellKnownForeignToNativePriceConverter::existential_deposit(*currency_id))
+	PriceConverter::<AssetsRegistry>::to_asset_balance(NATIVE_EXISTENTIAL_DEPOSIT, *currency_id)
```
PoisonPhang, line 53: Instead of using the conversion of 1/10 PICA * ratio, would it be wiser to use 1/10 of whatever the asset as the ED for said asset?
@KaiserKarel @vimukthi-git @dzmitry-lahoda 
Reasoning/Context:
> Ratio is pre Dex predictor of price. Later dex will be used. So not sure if ED to be as volatile as DEX.

> Meaning you're concerned that only using ration to calculate ED will be too volatile once the market is active?

> Yes. We use ratio for fees because there is no dex to tell the ratio (or we do not want market price).  For ED it may be default we do not want market price ever. I do not know.

> Yeah, I'd be a bit annoyed if my balance that wasn't previously dust got turned into dust and removed due to a spike in value

> It might make senses to define ED relative to whatever 1 is considered to be in any given asset

vimukthi-git: If you had such a small amount that got turned into dust, I don't think this would matter tbh. But having a more predictable value like TWAP for asset ratio might make sense in the future.

dzmitry-lahoda: dust accounts are in wallet.

andresmechali (approved):

PR #2263, opened 2022-11-11
Release v1 - Albula

vimukthi-git:
## Release v1 

- [x] If it is the first release of the `v<Major>` (eg: v5) line then create a branch `release-v<Major>`. Execute the following steps on that branch.
- [x] Generate weights, i.e run `benchmark`
- [x] Runtime [versioning](https://docs.substrate.io/build/upgrade-the-runtime/) updates
   - [x] Update `spec_version` (automate-able)
- [x] Update composable node version if the code has changed
- [x] Update composableJs version (if necessary to be released)
- [x] Update FE version (if necessary to be released)
- [x] Update Subsquid version (if necessary to be released)
- [x] Update relevant frame pallets being released in runtimes to the latest node version
- [ ] Consider and list possible proxy filter updates for available calls.
- [x] Categorize (and give a title) the release according to the types of changes it does, eg: security patch, bugfix, feature etc.
- [x] Run `cargo build` once to update the Cargo.lock file.
- [ ] Update the `v<Major>` branch on Github and make PR to master. Get it merged to main before next step. Note that `v<Major>` branch must not be deleted.
- [ ] Get the tip of the branch signed according to 3.2.4 section.
- [x] Finally, create a tag `v<Branch.Spec_version>` (eg: `v5.4201`) to trigger the release artifact build.

PR #2282, opened 2022-11-12
[Bugfix] Weights traits for Pablo

vimukthi-git:
## Issue

There was a trait method mismatch for weights because of the Pablo refactor, caught while running benchmarks on #2263


dzmitry-lahoda (approved):

vimukthi-git:
Merging with one approval after locally verifying that this compiles and works.

PR #2283, opened 2022-11-12
start replacing github rules with mergify to scale monorepo

dzmitry-lahoda:

## Issue

Basic to expand monorepo(rfc?) for many teams which want to develop faster.

Monorepo does not mean all must use same tooling, same nix, same merge rules, same security level.

```
Configuration changed. This pull request must be merged manually — 1 potential rule
⚠️ The configuration has been changed, queue and merge actions are ignored. ⚠️
```

## Notes

why hash tags instead of tags? sure people can use both - just encoded one variant of tagging (twitter tags are right, gh tags are clumsy)

## Checklist

- all PR merged as before with semantics for rules
- PRs with #wip title are not queued

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -12,6 +12,8 @@ pull_request_rules:
       - and:
         - "#approved-reviews-by>=2"
         - base=main
+        - not:
+          - title~=#wip
     actions:
       queue:
         name: default
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,12 +1,17 @@
 queue_rules:
  - name: default
    allow_inplace_checks: false
-   conditions: []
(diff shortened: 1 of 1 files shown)


KaiserKarel (approved):

KaiserKarel (approved):

dzmitry-lahoda:
@fdr-abag @bostercf @andyjsbell are you ok to use mergify rule for gate(>=2 review for merge) instead of gh rules? goal is to allow more flexible rules (e.g. internal dev docs can be 1 approve rule) or some poc/prototype/wip(like centauri)

dzmitry-lahoda:
image

PR #2285, opened 2022-11-12
#cosmwasm dev tools to play with contracts

dzmitry-lahoda:
## Issue

Started to move cosmwasm tools into flake-part.

Remove subxt from wasm tools (as it is not for sure such).

Add cosmwasm check.


KaiserKarel (approved):

code/xcvm/cosmos/flake-parts.nix
```diff
@@ -0,0 +1,34 @@
+{ self, ... }: {
```
cor, line 1: Please rename this file to `cosmwasm.nix` instead of `flake-parts.nix`

cor (changes requested):
Looks good! Glad you're getting started with `flake-parts`. Only one nitpick, please rename the flake-parts file

code/xcvm/cosmos/flake-parts.nix
dzmitry-lahoda: https://github.com/hercules-ci/flake-parts/tree/main/template/multi-module/hello

dzmitry-lahoda: renamed, hopefully according dicussion

cor (approved):
LGTM

PR #2294, opened 2022-11-14
[CU-35ed5ra] Obsessed cake/nixify subxt exports

obsessed-cake:
## Issue
https://app.clickup.com/t/35ed5ra

## Description
This PR has changes to nixify and configure subxt.
I'll move the changes and subxt from centauri to main branch in the next PR.


dzmitry-lahoda:
the goal is to enable run metadata export in ci so.

dzmitry-lahoda (approved):

PR #2331, opened 2022-11-16
Added zlib to dev-shell to fix cargo nightly

PoisonPhang:
## Issue

Can't run things with cargo +nightly (i.e `SKIP_WASM_BUILD=1 bacon clippy -- --all --tests`)

## Additional Changes

Added zlib to dev-shell. Fix inspired by [this](https://github.com/oxalica/rust-overlay/issues/54#issuecomment-1223758506)

Concerns: bacon ran with `SKIP_WASM_BUILD=1 bacon clippy -- --all --tests` is reporting many things on main as errors with this fix, which does not match with CI (20 errors, 1918 warnings :face_with_spiral_eyes: )

Note, this is probably a naive fix - feel free to correct/roast

### DIFF
--- a/dev-shells.nix
+++ b/dev-shells.nix
@@ -11,8 +11,11 @@
           buildInputs = base.buildInputs
             ++ (with pkgs; [ clang nodejs python3 yarn ])
             ++ (with self'.packages; [ rust-nightly ]);
-          LD_LIBRARY_PATH = pkgs.lib.strings.makeLibraryPath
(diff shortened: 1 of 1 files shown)


hussein-aitlahcen (approved):

cor (approved):
LGTM! Regarding the difference in error count: we can fix that in a later PR, actually good that we're able to run `bacon` now

dzmitry-lahoda (approved):

PR #2349, opened 2022-11-17
[CU-3vdqb2b] Improve locked value storing and lazily calculate total locked value

andresmechali:
## Issue

https://app.clickup.com/t/3vdqb2b

This requires reprocessing.


ComposableFiB0t:
Task linked: [CU-3vdqb2b Fix daily active users chart (Subsquid)](https://app.clickup.com/t/3vdqb2b)

subsquid/src/server-extension/resolvers/picassoOverviewStats.ts
```diff
@@ -36,26 +44,45 @@ export class PicassoOverviewStatsResolver
...
 
-    if (!lockedValue?.[0]) {
-      return Promise.resolve(0n);
+    for (const assetId of Object.keys(assetPrices)) {
+      try {
+        const oraclePrice = await api.query.oracle.prices(assetId);
+        if (oraclePrice?.price) {
+          assetPrices[assetId] = BigInt(oraclePrice?.price?.toString() || 0n);
+        }
+        // TODO: what to do if no there's no Oracle price?
```
dzmitry-lahoda, line 72: return "N/A"

dzmitry-lahoda (approved):

subsquid/src/server-extension/resolvers/picassoOverviewStats.ts
andresmechali: In the case of this resolver it's accumulating over all the locked assets, so if one locked asset is missing its price, I can only think on ignoring it, __unless__ there's some other source I can get prices from.

saadjhk (approved):

andresmechali:
@mergify refresh

PR #2354, opened 2022-11-18
Fix transfers bugs

easteregg:
## Issue
Properly remove karura related dependencies from asset fetching

Use multiplier for both short and long version

update fee token if byog

Improve validations and recover properly from validation errors

Update destChainFee for picasso=>statemine


PoisonPhang (approved):
LGTM
NIT: Maybe remove commented out code and rely on git history if it needs to return eventually.

PR #2358, opened 2022-11-18
chore(cosmwasm): add tests for crypto host functions

aeryz:
* Add tests for all crypto related host functions.

* Fix crypto related host functions.

### DIFF
--- a/code/parachain/frame/cosmwasm/src/tests.rs
+++ b/code/parachain/frame/cosmwasm/src/tests.rs
@@ -198,7 +198,7 @@ fn ed25519_batch_verify_fails_if_one_fail() {
 }
 
 #[test]
-fn ed25519_batch_verify_fails_if_input_lengts_are_incorrect() {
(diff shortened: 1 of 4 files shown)


dzmitry-lahoda (approved):

code/parachain/frame/cosmwasm/src/tests.rs
```diff
@@ -1,6 +1,221 @@
...
+			&ref_messages,
+			&ref_signatures,
+			&ref_public_keys
+		));
+	})
+}
+
+#[test]
+fn ed25519_batch_verify_fails_if_input_lengths_are_incorrect() {
+	new_test_ext().execute_with(|| {
```
hussein-aitlahcen, line 202: The functions can be extracted from the pallet for the mock context not to be required. But it is not a big deal.

hussein-aitlahcen (approved):

PR #2312, opened 2022-11-15
[CU-3t3ay05] Add ratio and ED to assets RPC

andresmechali:
## Issue
https://app.clickup.com/t/3t3ay05

I will keep this open until https://github.com/ComposableFi/composable/pull/2281 has been merged, as some minor changes will need to be made

PR #2392, opened 2022-11-22
[CU-3vz1bh3] update assets RPC on Picasso

andresmechali:
## Issue

https://app.clickup.com/t/3vz1bh3

### DIFF
--- a/code/parachain/runtime/picasso/src/lib.rs
+++ b/code/parachain/runtime/picasso/src/lib.rs
@@ -873,14 +873,11 @@ impl_runtime_apis! {
 			// Override asset data for hardcoded assets that have been manually updated, and append
 			// new assets without duplication
 			foreign_assets.into_iter().fold(assets, |mut acc, mut foreign_asset| {
-				if let Some(i) = acc.iter().position(|asset_i| asset_i.id == foreign_asset.id) {
-					// Assets that have been updated
-					if let Some(asset) = acc.get_mut(i) {
-						// Update asset with data from assets-registry
-						asset.decimals = foreign_asset.decimals;
-						asset.foreign_id = foreign_asset.foreign_id.clone();
-						asset.ratio = foreign_asset.ratio;
-					}
+				if let Some(asset) = acc.iter_mut().find(|asset_i| asset_i.id == foreign_asset.id) {
+					// Update asset with data from assets-registry
+					asset.decimals = foreign_asset.decimals;
+					asset.foreign_id = foreign_asset.foreign_id.clone();
+					asset.ratio = foreign_asset.ratio;
 				} else {
 					foreign_asset.existential_deposit = multi_existential_deposits::<AssetsRegistry>(&foreign_asset.id.into());
 					acc.push(foreign_asset.clone())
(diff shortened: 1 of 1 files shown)


ComposableFiB0t:
Task linked: [CU-3vz1bh3 Update assets RPC on Picasso runtime](https://app.clickup.com/t/3vz1bh3)

code/parachain/runtime/picasso/src/lib.rs
```diff
@@ -858,7 +859,34 @@ impl_runtime_apis! {
...
+			// Override asset data for hardcoded assets that have been manually updated, and append
+			// new assets without duplication
+			foreign_assets.into_iter().fold(assets, |mut acc, mut foreign_asset| {
+				if let Some(i) = acc.iter().position(|asset_i| asset_i.id == foreign_asset.id) {
+					// Assets that have been updated
+					if let Some(asset) = acc.get_mut(i) {
+						// Update asset with data from assets-registry
+						asset.decimals = foreign_asset.decimals;
+						asset.foreign_id = foreign_asset.foreign_id.clone();
+						asset.ratio = foreign_asset.ratio;
```
PoisonPhang, line 882: I thought we agreed to not overwrite well-known assets with what's inside Asset Registry?
Regardless, I think this assignment could be simplified. See:
https://doc.rust-lang.org/core/option/enum.Option.html#method.replace
https://github.com/ComposableFi/composable/blob/main/code/parachain/frame/crowdloan-rewards/src/lib.rs#L458-L462

code/parachain/runtime/picasso/src/lib.rs
```diff
@@ -858,7 +859,34 @@ impl_runtime_apis! {
...
+				asset
+			}).collect::<Vec<_>>();
+
+			// Assets from the assets-registry pallet
+			let foreign_assets = assets_registry::Pallet::<Runtime>::get_foreign_assets_list();
+
+			// Override asset data for hardcoded assets that have been manually updated, and append
+			// new assets without duplication
+			foreign_assets.into_iter().fold(assets, |mut acc, mut foreign_asset| {
+				if let Some(i) = acc.iter().position(|asset_i| asset_i.id == foreign_asset.id) {
```
benluelo, line 876: Use .iter_mut().find() instead of .position()

code/parachain/runtime/picasso/src/lib.rs
andresmechali: yes, Vim had the same question when I added this to Dali (this PR is only copying the RPC from Dali) https://github.com/ComposableFi/composable/pull/2312#discussion_r1025806521

As we are not going to override it, this part won't have any effect. But if at some point we want to override one of the well known assets, it will use the updated information.

Regarding the `.replace` method, looks neat! I'll change it to use that.

code/parachain/runtime/picasso/src/lib.rs
andresmechali: Yes sir

PoisonPhang (approved):

dzmitry-lahoda (approved):

andyjsbell (approved):

vimukthi-git:
@andresmechali Did you get the chance to dev-test that this works?

vimukthi-git (approved):
LGTM

andresmechali:
@mergify refresh

PR #2453, opened 2022-11-26
#done removed old democracy

dzmitry-lahoda:

## Issue

no used and likely will not be used and obsolete copy paste


andyjsbell:
Thanks @dzmitry-lahoda - do we have a task linked to this?

KaiserKarel (approved):

dzmitry-lahoda:
@andyjsbell this was an exclusive case where it can be done under one hour :) really it took 15 minutes (ok - with rebase review 15+ more minutes). it was discussed 3 month ago afaik to remove it. so it was not clear before that if Gov v2 will support multi Origin (it appeared yes) and easy to make it multi currency (it seems yes) - so deleting is easy.

andyjsbell:
@dzmitry-lahoda No problem. Sounds like it would have take longer to create the task than do it.

andyjsbell (approved):
Thanks @dzmitry-lahoda

vimukthi-git (approved):

dzmitry-lahoda:
oh, sorry. was mistake.

PR #2454, opened 2022-11-28
Integration Tests: Re- enabling KSM XCM Tests & minor doc, clean code fixes

Dom-Roth:
## Issue

Would be great if i could create a ClickUp ticket, right?
XCM Integration Tests for Kusama test transfers were disabled last week due to an issue, that couldn't be reproduced.
Therefore re- enabling them here & hoping for the best.

## Additional Changes

The code comments were partially wrong / misleading, and there was a double `before` block. Which by current integration test standards we're trying to avoid.

## Checklist

~~- [ ] Updated the rust/typescript docs.~~
~~- [ ] Updated the `docs/`.~~

### DIFF
--- a/code/integration-tests/runtime-tests/test/tests/xcmp/test_configuration.json
+++ b/code/integration-tests/runtime-tests/test/tests/xcmp/test_configuration.json
@@ -1,5 +1,5 @@
 {
   "enabledTests": {
-    "enabled": false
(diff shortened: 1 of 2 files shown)


kollegian (approved):

LyonSsS (approved):
Ready to be merged

PR #2465, opened 2022-11-28
fixed version for simnode

dzmitry-lahoda:

## Checklist

- [ ] Updated the rust/typescript docs.
- [ ] Updated the `docs/`.

### DIFF
--- a/code/parachain/runtime/composable/src/lib.rs
+++ b/code/parachain/runtime/composable/src/lib.rs
@@ -119,7 +119,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// The version of the runtime specification. A full node will not attempt to use its native
 	//   runtime in substitute for the on-chain Wasm runtime unless all of `spec_name`,
 	//   `spec_version`, and `authoring_version` are the same between Wasm and native.
-	spec_version: 10_002,
+	spec_version: 10_003,
 	impl_version: 3,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,
--- a/code/parachain/runtime/dali/src/lib.rs
+++ b/code/parachain/runtime/dali/src/lib.rs
@@ -62,7 +62,7 @@ use primitives::currency::{CurrencyId, ValidateCurrencyId};
 use sp_api::impl_runtime_apis;
 use sp_core::{crypto::KeyTypeId, OpaqueMetadata};
 use sp_runtime::{
-	create_runtime_str, generic, impl_opaque_keys,
+	generic, impl_opaque_keys,
 	traits::{
 		AccountIdConversion, AccountIdLookup, BlakeTwo256, Block as BlockT, Bounded, Convert,
 		ConvertInto, Zero,
@@ -1490,7 +1490,7 @@ impl_runtime_apis! {
 	}
 
 	impl sp_api::Core<Block> for Runtime {
-		fn version() -> RuntimeVersion {
+		fn version() -> sp_version::RuntimeVersion {
 			VERSION
 		}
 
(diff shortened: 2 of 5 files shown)


code/parachain/runtime/composable/src/lib.rs
```diff
@@ -119,7 +119,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	// The version of the runtime specification. A full node will not attempt to use its native
 	//   runtime in substitute for the on-chain Wasm runtime unless all of `spec_name`,
 	//   `spec_version`, and `authoring_version` are the same between Wasm and native.
-	spec_version: 10_002,
+	spec_version: 10_003,
```
KaiserKarel, line 122: Are we supposed to be incrementing this right now?

dzmitry-lahoda: It blocks CI as per thread (simnode requires new runtime new version).

```
    /// Version of the runtime specification. A full-node will not attempt to use its native
    /// runtime in substitute for the on-chain Wasm runtime unless all of `spec_name`,
    /// `spec_version` and `authoring_version` are the same between Wasm and native.
    pub spec_version: u32,
```

Imho we should have incremented when did wasm upgrade.  cc @vimukthi-git 

Now is too late (but we are lucky we are not running native).  So it can be ticket be sure and update our release guide. 

It means if we are going to release new wasm, we must release new node and new native version ups.

dzmitry-lahoda: If any node will try to run native now, it will try to produce blocks native, and will be punished as I read this.

My changes do not relevant to anything happening, just update to fix simnode.

Imho simnode is wrong too. It should be proactive and connected to our release (release should run simnode upgrade with release and current main to avoid issues) or something.

dzmitry-lahoda: But this is not relevant to this PR and PR cannot fix what already happened. It just unblocks work :)

dzmitry-lahoda: https://stackoverflow.com/questions/56873179/substrate-has-runtime-versioning-what-is-the-purpose-and-use-case-for-these <- so we need task for better guide and test and sync for release. not in this PR so.

obsessed-cake (approved):

hussein-aitlahcen (approved):

dzmitry-lahoda:
@Mergifyio refresh

code/parachain/runtime/composable/src/lib.rs
dzmitry-lahoda: resolving for now. @vimukthi-git @andyjsbell 
1. seems need bump version right after release (assuming next release needed to change logic)
2. factor out version into separate files like in example (so can easy to diff/change detection, parse and paset, sed)
3. document that DO not need to update version in sync (i.e. composable version will lag far behind)
4. document that native runs are NOT supported for next year (premature optimization).

dzmitry-lahoda:
https://github.com/Mergifyio refresh

code/parachain/runtime/composable/src/lib.rs
vimukthi-git: Just found out about this change

> Now is too late (but we are lucky we are not running native). So it can be ticket to check to be sure and update our release guide.

I don't get why we are lucky here, could you explain? We already released new node version at the time of wasm upgrade.

>It means if we are going to release new wasm, we must release new node and new native version ups.

We are doing that already. as per the clearly defined scheme in updating all the relevant versions for a node release (https://github.com/ComposableFi/composable/blob/main/rfcs/0011-releases.md#31-release-numbers), this update is against that.

PR #2426, opened 2022-11-24
hyperspace in devnet

cor:

PR #2474, opened 2022-11-28
Include `pallet-ibc` in dali runtime #NX-37

vmarkushin:
Most of the code was copied from the `centauri` repo. Some parts are incompatible with the current repo, so I tried to rewrite them, leaving the old code version for comparison.

### DIFF
--- a/code/Cargo.lock
+++ b/code/Cargo.lock
@@ -718,14 +718,37 @@ dependencies = [
 [[package]]
 name = "beefy-light-client"
 version = "0.1.0"
-source = "git+https://github.com/ComposableFi/centauri/?rev=c80e2b203ecfb1925b0330aafba0f736c99fa5f6#c80e2b203ecfb1925b0330aafba0f736c99fa5f6"
+source = "git+https://github.com/ComposableFi/centauri?rev=c7d2f2d812074850c3b5ab7e42fa582f72582355#c7d2f2d812074850c3b5ab7e42fa582f72582355"
 dependencies = [
- "beefy-light-client-primitives",
+ "beefy-light-client-primitives 0.1.0 (git+https://github.com/ComposableFi/centauri?rev=c7d2f2d812074850c3b5ab7e42fa582f72582355)",
  "beefy-primitives",
  "ckb-merkle-mountain-range",
  "derive_more",
  "frame-support",
- "light-client-common",
+ "light-client-common 0.1.0 (git+https://github.com/ComposableFi/centauri?rev=c7d2f2d812074850c3b5ab7e42fa582f72582355)",
+ "pallet-beefy-mmr",
+ "pallet-mmr",
+ "parity-scale-codec",
+ "rs_merkle",
+ "sp-core",
+ "sp-core-hashing",
+ "sp-mmr-primitives",
+ "sp-runtime",
+ "sp-std 4.0.0 (git+https://github.com/ComposableFi/substrate?rev=9bdc51acad943bd07fae159f59564a1b464d33b5)",
+ "sp-trie",
+]
+
+[[package]]
+name = "beefy-light-client"
+version = "0.1.0"
+source = "git+https://github.com/ComposableFi/centauri/?rev=c80e2b203ecfb1925b0330aafba0f736c99fa5f6#c80e2b203ecfb1925b0330aafba0f736c99fa5f6"
+dependencies = [
+ "beefy-light-client-primitives 0.1.0 (git+https://github.com/ComposableFi/centauri/?rev=c80e2b203ecfb1925b0330aafba0f736c99fa5f6)",
+ "beefy-primitives",
+ "ckb-merkle-mountain-range",
+ "derive_more",
+ "frame-support",
+ "light-client-common 0.1.0 (git+https://github.com/ComposableFi/centauri/?rev=c80e2b203ecfb1925b0330aafba0f736c99fa5f6)",
  "pallet-beefy-mmr",
  "pallet-mmr",
  "parity-scale-codec",
@@ -741,12 +764,28 @@ dependencies = [
 [[package]]
 name = "beefy-light-client-primitives"
 version = "0.1.0"
+source = "git+https://github.com/ComposableFi/centauri?rev=c7d2f2d812074850c3b5ab7e42fa582f72582355#c7d2f2d812074850c3b5ab7e42fa582f72582355"
+dependencies = [
+ "beefy-primitives",
+ "ckb-merkle-mountain-range",
+ "derive_more",
+ "light-client-common 0.1.0 (git+https://github.com/ComposableFi/centauri?rev=c7d2f2d812074850c3b5ab7e42fa582f72582355)",
+ "parity-scale-codec",
+ "rs_merkle",
+ "sp-core",
+ "sp-mmr-primitives",
+ "sp-std 4.0.0 (git+https://github.com/ComposableFi/substrate?rev=9bdc51acad943bd07fae159f59564a1b464d33b5)",
+]
+
+[[package]]
+name = "beefy-light-client-primitives"
+version = "0.1.0"
 source = "git+https://github.com/ComposableFi/centauri/?rev=c80e2b203ecfb1925b0330aafba0f736c99fa5f6#c80e2b203ecfb1925b0330aafba0f736c99fa5f6"
 dependencies = [
  "beefy-primitives",
  "ckb-merkle-mountain-range",
  "derive_more",
- "light-client-common",
+ "light-client-common 0.1.0 (git+https://github.com/ComposableFi/centauri/?rev=c80e2b203ecfb1925b0330aafba0f736c99fa5f6)",
  "parity-scale-codec",
  "rs_merkle",
  "sp-core",
@@ -1607,6 +1646,7 @@ dependencies = [
  "pablo-runtime-api",
  "pallet-assets 1.0.0",
  "pallet-crowdloan-rewards",
+ "pallet-ibc 0.0.1 (git+https://github.com/ComposableFi/centauri?rev=c7d2f2d812074850c3b5ab7e42fa582f72582355)",
  "pallet-transaction-payment-rpc",
  "pallet-transaction-payment-rpc-runtime-api",
  "parity-scale-codec",
@@ -3187,6 +3227,9 @@ dependencies = [
  "frame-system-rpc-runtime-api",
  "hex",
  "hex-literal",
+ "ibc 0.15.0 (git+https://github.com/ComposableFi/centauri?rev=c7d2f2d812074850c3b5ab7e42fa582f72582355)",
+ "ibc-primitives 0.1.0 (git+https://github.com/ComposableFi/centauri?rev=c7d2f2d812074850c3b5ab7e42fa582f72582355)",
+ "ibc-runtime-api",
  "lending-runtime-api",
  "orml-tokens",
  "orml-traits",
@@ -3212,6 +3255,8 @@ dependencies = [
  "pallet-dutch-auction",
  "pallet-fnft",
  "pallet-governance-registry",
+ "pallet-ibc 0.0.1 (git+https://github.com/ComposableFi/centauri?rev=c7d2f2d812074850c3b5ab7e42fa582f72582355)",
+ "pallet-ibc-ping",
  "pallet-identity",
  "pallet-indices",
  "pallet-lending",
@@ -3862,9 +3907,9 @@ dependencies = [
 
 [[package]]
 name = "filetime"
-version = "0.2.18"
+version = "0.2.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4b9663d381d07ae25dc88dbdf27df458faa83a9b25336bcac83d5e452b5fc9d3"
+checksum = "4e884668cd0c7480504233e951174ddc3b382f7c2666e3b7310b5c4e7b0c37f9"
 dependencies = [
  "cfg-if 1.0.0",
  "libc",
@@ -4607,6 +4652,27 @@ dependencies = [
 [[package]]
 name = "grandpa-light-client-primitives"
 version = "0.1.0"
+source = "git+https://github.com/ComposableFi/centauri?rev=c7d2f2d812074850c3b5ab7e42fa582f72582355#c7d2f2d812074850c3b5ab7e42fa582f72582355"
+dependencies = [
+ "anyhow",
+ "derive_more",
+ "finality-grandpa",
+ "frame-support",
+ "hash-db",
+ "light-client-common 0.1.0 (git+https://github.com/ComposableFi/centauri?rev=c7d2f2d812074850c3b5ab7e42fa582f72582355)",
+ "parity-scale-codec",
+ "sp-core",
+ "sp-finality-grandpa",
+ "sp-io",
+ "sp-runtime",
+ "sp-std 4.0.0 (git+https://github.com/ComposableFi/substrate?rev=9bdc51acad943bd07fae159f59564a1b464d33b5)",
+ "sp-storage 6.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "sp-trie",
+]
+
+[[package]]
+name = "grandpa-light-client-primitives"
+version = "0.1.0"
 source = "git+https://github.com/ComposableFi/centauri/?rev=c80e2b203ecfb1925b0330aafba0f736c99fa5f6#c80e2b203ecfb1925b0330aafba0f736c99fa5f6"
 dependencies = [
  "anyhow",
@@ -4614,7 +4680,7 @@ dependencies = [
  "finality-grandpa",
  "frame-support",
  "hash-db",
- "light-client-common",
+ "light-client-common 0.1.0 (git+https://github.com/ComposableFi/centauri/?rev=c80e2b203ecfb1925b0330aafba0f736c99fa5f6)",
  "parity-scale-codec",
  "sp-core",
  "sp-finality-grandpa",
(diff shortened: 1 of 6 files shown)


code/parachain/runtime/dali/src/lib.rs
```diff
@@ -1196,6 +1211,157 @@ impl cosmwasm::Config for Runtime {
...
+impl DenomToAssetId<Runtime> for IbcDenomToAssetIdConversion {
+	type Error = DispatchError;
+
+	fn from_denom_to_asset_id(denom: &String) -> Result<CurrencyId, Self::Error> {
+		let denom_bytes = denom.as_bytes().to_vec();
+		if let Some(id) = IbcDenoms::<Runtime>::get(&denom_bytes) {
+			return Ok(id)
+		}
+
+		// let pallet_id: AccountId = PalletId(*b"pall-ibc").into_account_truncating();
```
dzmitry-lahoda, line 1256: what is this?

code/parachain/runtime/dali/src/lib.rs
```diff
@@ -1196,6 +1211,157 @@ impl cosmwasm::Config for Runtime {
 	type WeightInfo = cosmwasm::weights::SubstrateWeight<Runtime>;
 }
 
+parameter_types! {
+	pub const RelayChainId: RelayChain = RelayChain::Rococo;
+	pub const SpamProtectionDeposit: Balance = 1_000_000_000_000;
+	pub const MinimumConnectionDelay: u64 = 0; // well we don't need the security tbh.
```
dzmitry-lahoda, line 1217: why we do not need it?

code/parachain/runtime/dali/src/lib.rs
```diff
@@ -1196,6 +1211,157 @@ impl cosmwasm::Config for Runtime {
...
+			location,
+			Rational64::one(),
+			Some(12),
+		)?;
+
+		Ok(asset_id)
+	}
+
+	fn from_asset_id_to_denom(id: CurrencyId) -> Option<String> {
+		// let name = <assets::Pallet<Runtime> as InspectMetadata<AccountId>>::name(&id);
```
dzmitry-lahoda, line 1301: what is it?

code/parachain/runtime/dali/src/lib.rs
```diff
@@ -1196,6 +1211,157 @@ impl cosmwasm::Config for Runtime {
...
+		// String::from_utf8(name).ok()
+		// TODO: the code above is outdated
```
dzmitry-lahoda, line 1303: why you add outdated code?

code/parachain/runtime/dali/src/lib.rs
```diff
@@ -1196,6 +1211,157 @@ impl cosmwasm::Config for Runtime {
...
+			Rational64::one(),
+			Some(12),
+		)?;
+
+		Ok(asset_id)
+	}
+
+	fn from_asset_id_to_denom(id: CurrencyId) -> Option<String> {
+		// let name = <assets::Pallet<Runtime> as InspectMetadata<AccountId>>::name(&id);
+		// String::from_utf8(name).ok()
```
dzmitry-lahoda, line 1302: what is it?

code/parachain/runtime/dali/src/lib.rs
```diff
@@ -1270,8 +1436,13 @@ construct_runtime!(
...
 		CallFilter: call_filter = 140,
 
-	  // Cosmwasm support
-	  Cosmwasm: cosmwasm = 180
+		// Cosmwasm support
+		Cosmwasm: cosmwasm = 180,
+
+		// IBC support
+		IbcPing: pallet_ibc_ping = 190,
+		// pallet-ibc, should be the last module in your runtime
```
dzmitry-lahoda, line 1444: why? what if there is other module which will want to be last? from what this requirements steams?

dzmitry-lahoda: may we can make it 200 and tell which modules can be after and before? please guide what should be here and why.

code/parachain/runtime/dali/src/lib.rs
```diff
@@ -1196,6 +1211,157 @@ impl cosmwasm::Config for Runtime {
...
+			if limit == 0 {
+				break
+			}
+		}
+
+		(denoms, IbcAssetIds::<Runtime>::count() as u64, iterator.next().map(|(id, ..)| id))
+	}
+}
+
+impl pallet_ibc::Config for Runtime {
```
dzmitry-lahoda, line 1336: pallet is single instance, how we going to migrate to beffy?

code/parachain/runtime/dali/src/lib.rs
```diff
@@ -1196,6 +1211,157 @@ impl cosmwasm::Config for Runtime {
...
+	type TimeProvider = Timestamp;
+	type Event = Event;
+	type NativeCurrency = Balances;
+	type Balance = Balance;
+	type AssetId = CurrencyId;
+	type NativeAssetId = NativeAssetId;
+	type IbcDenomToAssetIdConversion = IbcDenomToAssetIdConversion;
+	const PALLET_PREFIX: &'static [u8] = b"ibc/";
```
dzmitry-lahoda, line 1344: why not using pallet id?

code/parachain/runtime/dali/src/lib.rs
dzmitry-lahoda: also, would beffy migration happen in place?

dzmitry-lahoda: so sure 2 modules cannot be both last

code/parachain/runtime/dali/src/lib.rs
```diff
@@ -95,6 +97,18 @@ use frame_support::{
 	weights::ConstantMultiplier,
 };
 use frame_system as system;
+use ibc::core::{
+	ics24_host::identifier::PortId,
+	ics26_routing::context::{Module, ModuleId},
+};
+// use orml_asset_registry::{AssetMetadata, AssetProcessor, DefaultAssetMetadata};
```
dzmitry-lahoda, line 104: what is it?

code/parachain/runtime/dali/src/lib.rs
```diff
@@ -23,12 +23,14 @@ pub const WASM_BINARY_V2: Option<&[u8]> = Some(include_bytes!(env!("DALI_RUNTIME
 #[cfg(not(feature = "builtin-wasm"))]
 pub const WASM_BINARY_V2: Option<&[u8]> = None;
 
+#[macro_use]
```
dzmitry-lahoda, line 26: why it is here?

code/Cargo.lock
```diff
@@ -653,6 +653,45 @@ dependencies = [
  "thiserror",
 ]
 
+[[package]]
+name = "beefy-light-client"
```
dzmitry-lahoda, line 657: why we depend on beefy client if we use grandpa?

code/parachain/runtime/dali/src/lib.rs
vmarkushin: This is the code from `centauri` repo. It's not working with the current `composable` code, so I had to rewrite it. I left the old code here so you can compare it with the new realization.

code/parachain/runtime/dali/src/lib.rs
vmarkushin: With a runtime upgrade?

blasrodri: I believe so. It won't happen anytime soon. Target is Q2

code/parachain/runtime/dali/src/lib.rs
dzmitry-lahoda: i see. thank you. 
so please remove and retain just link with comment.

ComposableFiB0t:
Task linked: [CU-441wwc9 Include pallet-ibc into Dali devnet](https://app.clickup.com/t/441wwc9)

code/parachain/runtime/dali/src/lib.rs
vmarkushin: Removed

code/parachain/runtime/dali/src/lib.rs
vmarkushin: Removed

code/parachain/node/src/chain_spec/dali.rs
```diff
@@ -77,5 +77,6 @@ pub fn genesis_config(
 		vesting: Default::default(),
 		lending: Default::default(),
 		liquidations: Default::default(),
+		ibc: dali_runtime::IbcConfig { asset_ids: vec![(1u128.into(), b"UNIT".to_vec())] },
```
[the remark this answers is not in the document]
seunlanlege: this is a mapping of asset-id to denoms

code/parachain/runtime/dali/src/lib.rs
seunlanlege: Oh we'll definitely need it in production

code/parachain/runtime/dali/src/lib.rs
seunlanlege: This code should be uncommented, it registers the native token as an asset on pallet assets

code/parachain/runtime/dali/src/lib.rs
seunlanlege: This fetches the asset id from pallet assets, uncomment as well

code/parachain/runtime/dali/src/lib.rs
seunlanlege: This todo shouldn't be trusted, the code checks out

code/parachain/runtime/dali/src/lib.rs
seunlanlege: actually, you can put it anywhere

code/parachain/node/src/chain_spec/dali.rs
KaiserKarel, line 80: What's the meaning of 1u128.into() and b"UNIT".to_vec()? Can we perhaps refactor the asset_ids type to take in a struct with named fields, instead a tuple?

code/parachain/runtime/dali/src/lib.rs
KaiserKarel: @vmarkushin Can this be commented better? (Capitalized, punctuated, explaining why 0 is acceptable for Dali but not acceptable for Production). As well as, is there a reason to use unsafe configurations in Dali, instead of just having correct configurations everywhere?

code/parachain/runtime/dali/src/lib.rs
```diff
@@ -1196,6 +1211,157 @@ impl cosmwasm::Config for Runtime {
...
+		// 	.to_vec();
+		// let asset_id = generate_asset_id()?;
+
+		// <assets::Pallet<Runtime> as Create<AccountId>>::create(
+		// 	asset_id,
+		// 	pallet_id.clone(),
+		// 	true,
+		// 	1,
+		// )?;
+		//
```
KaiserKarel, line 1272: Please remove outdated/commented code.

code/parachain/runtime/dali/src/lib.rs
```diff
@@ -1196,6 +1211,157 @@ impl cosmwasm::Config for Runtime {
...
+
+		let location = XcmAssetLocation::new(
+			MultiLocation::new(1, X1(Junction::GeneralIndex(asset_id.into()))).into(),
+		);
+		assets_registry::Pallet::<Runtime>::set_reserve_location(
+			asset_id,
+			location,
+			Rational64::one(),
+			Some(12),
+		)?;
```
KaiserKarel, line 1295: I think this will work for now; but @PoisonPhang is working on a proposal to change currency-factory and such.

code/parachain/runtime/dali/src/lib.rs
KaiserKarel: Then the todo needs to be removed @vmarkushin :)

KaiserKarel: And bridging team needs to figure out what is outdated and what is not, before requesting a merge into main.

code/parachain/runtime/dali/src/lib.rs
```diff
@@ -1196,6 +1211,157 @@ impl cosmwasm::Config for Runtime {
...
+		mut limit: u64,
+	) -> (Vec<Vec<u8>>, u64, Option<CurrencyId>) {
+		let mut iterator = if let Some(asset_id) = start_key {
+			let raw_key = asset_id.encode();
+			IbcAssetIds::<Runtime>::iter_from(raw_key).skip(0)
+		} else if let Some(offset) = offset {
+			IbcAssetIds::<Runtime>::iter().skip(offset as usize)
+		} else {
+			IbcAssetIds::<Runtime>::iter().skip(0)
+		};
```
KaiserKarel, line 1321: This code will fail when called as `ibc_assets(Some(key), Some(offset)` no? As it will then ignore the Some(offset).

code/parachain/runtime/dali/src/lib.rs
```diff
@@ -1196,6 +1211,157 @@ impl cosmwasm::Config for Runtime {
...
+
+		let mut denoms = vec![];
+		for (_, denom) in iterator.by_ref() {
+			denoms.push(denom);
+			limit -= 1;
+			if limit == 0 {
+				break
+			}
+		}
```
KaiserKarel, line 1330: Can be something like

```
iterator.by_ref().take(limit).collect().
```

Technically you are also returning both an iterator and denoms, which contain the same value? Bit of a codesmell but don't think it needs to be fixed in this PR.

code/parachain/runtime/dali/src/lib.rs
KaiserKarel: Then @vmarkushin let's remove the comment on it needing to be last.

code/parachain/runtime/dali/src/lib.rs
```diff
@@ -1196,6 +1211,157 @@ impl cosmwasm::Config for Runtime {
...
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Default)]
+pub struct Router {
+	pallet_ibc_ping: pallet_ibc_ping::IbcModule<Runtime>,
+}
+
+impl ModuleRouter for Router {
+	fn get_route_mut(&mut self, module_id: &impl Borrow<ModuleId>) -> Option<&mut dyn Module> {
+		match module_id.borrow().to_string().as_str() {
```
KaiserKarel, line 1227: Do we really need to convert a ModuleId to a String to take an &str?

code/parachain/runtime/dali/src/lib.rs
```diff
@@ -1196,6 +1211,157 @@ impl cosmwasm::Config for Runtime {
...
+	pub const MinimumConnectionDelay: u64 = 0; // well we don't need the security tbh.
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Default)]
+pub struct Router {
+	pallet_ibc_ping: pallet_ibc_ping::IbcModule<Runtime>,
+}
+
+impl ModuleRouter for Router {
+	fn get_route_mut(&mut self, module_id: &impl Borrow<ModuleId>) -> Option<&mut dyn Module> {
```
KaiserKarel, line 1226: A reference to something which implements Borrow seems like a codesmell to me.

code/parachain/runtime/dali/src/lib.rs
KaiserKarel: Can just implement PartialEq and Eq for ModuleId and &str no?

code/parachain/runtime/dali/src/lib.rs
vmarkushin: @seunlanlege it doesn't work with the composable pallets, that's why I commented it and wrote new code below

code/parachain/runtime/dali/src/lib.rs
vmarkushin: @KaiserKarel the commented code (above the TODO) is copied from `centauri` repo so reviewers can compare it with the one I wrote for this repo (below the TODO)

dzmitry-lahoda: just add link to code in repo so we can compare.

code/parachain/runtime/dali/src/lib.rs
vmarkushin: @seunlanlege could you answer @KaiserKarel's comment?

code/parachain/runtime/dali/src/lib.rs
vmarkushin: It's the semantics of the method. But I agree that it'd be better if it were accepting something like `Option<Either<T::AssetId, u32>>`. Will create a PR to change the signature

code/parachain/runtime/dali/src/lib.rs
vmarkushin: Thanks. No, the iterator is not returned, there is a call to `.next`

code/parachain/runtime/dali/src/lib.rs
vmarkushin: Will change to `impl Into<ModuleId>`

code/Cargo.lock
vmarkushin: Because it's one of the clients in pallet-ibc

code/parachain/runtime/dali/src/lib.rs
vmarkushin: Link to the old code: https://github.com/ComposableFi/centauri/blob/c80e2b203ecfb1925b0330aafba0f736c99fa5f6/utils/parachain-node/runtime/src/lib.rs#L647

code/parachain/runtime/dali/src/lib.rs
vmarkushin: @KaiserKarel @seunlanlege and @Wizdave97 are unavailable at the moment. I can set it to 5 minutes for now as it's done in the [docs](https://github.com/ComposableFi/centauri/blob/b97264308da51e67e0002e6706fab741a6a996a6/contracts/pallet-ibc/README.md?plain=1#L31) and leave a TODO to refine the parameter later, how about that?

dzmitry-lahoda: imho any NON zero is good to check setting it at all works.

blasrodri:
blocking https://github.com/ComposableFi/composable/pull/2426 to do a proper test

ComposableFiB0t:
Task linked: [CU-4737ne1 Integrate pallet-ibc into Dali runtime](https://app.clickup.com/t/4737ne1)

dzmitry-lahoda (approved):
ok, task to run e2e on this will be created. so it is not clear if current setup works.

code/parachain/runtime/dali/src/lib.rs
andyjsbell: @vmarkushin will resolve this as I believe this will be handled in a subsequent PR

andyjsbell (approved):

PR #2489, opened 2022-11-29
[CU-37t8q47] Added RFC 0013 - Depracate Currency Factory

PoisonPhang:
## Issue

[[CU-37t8q47]](https://app.clickup.com/t/37t8q47)

## Additional Changes

This is an RFC, please comment on any concerns.

### DIFF
--- a/rfcs/0013-deprecate-currency-factory.md
+++ b/rfcs/0013-deprecate-currency-factory.md
@@ -121,6 +121,25 @@ pub enum CurrencyId {
 }
 ```
 
+### Moonbeam
+
+Moonbeam has two asset classifications: 
+
+* Local
+* Foreign
+
+Moonbeam maintains separate instances of Parity's pallet-assets to track each of
+these asset classifications. Another pallet, pallet-asset-manager, is used to 
+route between the two.
+
+Local asset IDs are determined by the result of hashing a counter that tracks
+the number of local assets. While Foreign asset IDs are determined by the assets
+XCM Multilocation.
+
+#### Foreign
+
+Local assets are registered with a `ForeignAssetType`, `AssetRegisterarMetadata`, a `Balance` representing the ED, and a `bool` `is_sufficent`.
+
 ## Requirements
 
 If we are to deprecate Currency Factory, we will need to ensure the requirements
@@ -145,18 +164,20 @@ These solutions are detailed below:
 
 One way to remove the need for pallet-currency-factory is to move its 
 functionality to pallet-asset-registry. If we also simplify the asset ID 
-reservation system to no longer use predefined ranges, this will be a minimal change.
+reservation system to no longer use predefined ranges, this will be a minimal 
+change.
 
-This retains the functionality of Currency Factory, and does not require a new pallet. However, this does not enable permissionless asset creation without the
+This retains the functionality of Currency Factory, and does not require a new 
+pallet. However, this does not enable permissionless asset creation without the
 need for more design.
 
 #### Technical Implementation
 
 Instead of reserving asset IDs via our current range system, we could simply 
-have one nonce for reserving new asset IDs, which would (by the definition of a nonce) ensure no collisions. 
-This nonce can start at an arbitrary but high number so that our hard-coded 
-asset IDs are still safe. This would reduce the complexity of our current 
-reservation system while still avoiding collision.
+have one nonce for reserving new asset IDs, which would (by the definition of a 
+nonce) ensure no collisions. This nonce can start at an arbitrary but high 
+number so that our hard-coded asset IDs are still safe. This would reduce the 
+complexity of our current reservation system while still avoiding collision.
 
 #### Consequences
 
@@ -251,6 +272,9 @@ After analyzing possible solutions, the best course of action to deprecate
 Currency Factory is to maintain a fork of Parity's pallet-assets, embrace the 
 design Moonbeam uses to maintain two instances of their pallet, and to remove 
 our asset ID reservation system (for release two).
+
+Additionally, for foreign asset IDs, we can move to the standard of using XCM 
+Multi Locations. We can move to using a simple nonce for local asset IDs.
 
 This can be accomplished in two main phases:
 
--- a/rfcs/0013-deprecate-currency-factory.md
+++ b/rfcs/0013-deprecate-currency-factory.md
@@ -35,6 +35,92 @@ As for asset metadata, we don't have standards for formatting this nor do we
 utilize it. Assuming we replace the reservation system, Currency Factory will be 
 left without value.
 
+## Other Solutions
+
+The goal of this section is to overview the solutions developed by other networks
+in the ecosystem.
+
+### Acala
+
+Acala has four asset classifications:
+
+* Foreign
+* Stable
+* Erc20
+* Native
+
+Acala has separate registration processes for each of these within their Asset 
+Registry pallet. Of interest to is are the foreign and native asset registration
+process.
+
+#### Foreign Assets
+
+Foreign assets are registered with a `MultiLocation` and `AssetMetadata`. The 
+`MultiLocation` then mapped to a `ForeignAssetId` that is determined by a nonce. 
+A mapping of `ForeignAssetId` to `MultiLocation` is also created so that either 
+can be used to look up the other. The provided metadata provides the assets 
+name, symbol, decimals, and minimum balance. The Metadata is indexed by the 
+`ForeignAssetId`.
+
+Foreign assets can then be updated by their `ForeignAssetId`, in which case
+both the assets `MultiLocation` and `AssetMetadata` can be updated.
+
+#### Native Assets
+
+Native assets are registered with a `CurrencyId` and `AssetMetadata`. The 
+`CurrencyId` maps to the `AssetMetadata` and this mapping is used to ensure that 
+the asset does not already exist.
+
+Native assets can then be updated by their `CurrencyId`, in which case
+the assets `AssetMetadata` can be updated.
+
+#### General Notes
+
+* The `AssetMetadatas` storage map is used for all types and is quarried by the 
+`AssetIds` `enum`.
+
+* The `AssetIds` type is also used for routing regular currency functions (i.e. 
+transfer, withdraw, deposit).
+
+* For non-erc20 assets, Acala uses 'orml-tokens' for currency functions
+
+#### Data Structures & Types
+
+Note: `MultiLocation` is supplied by XCM.
+
+```rust
+pub type ForeignAssetId = u16;
+```
+
+```rust
+pub struct AssetMetadata<Balance> {
+  pub name: Vec<u8>,
+	pub symbol: Vec<u8>,
+	pub decimals: u8,
+	pub minimal_balance: Balance,
+}
+```
+
+```rust
+pub enum AssetIds {
+	Erc20(EvmAddress),
+	StableAssetId(StableAssetPoolId),
+	ForeignAssetId(ForeignAssetId),
+	NativeAssetId(CurrencyId),
+}
+```
+
+```rust
+pub enum CurrencyId {
+	Token(TokenSymbol),
+	DexShare(DexShare, DexShare),
+	Erc20(EvmAddress),
+	StableAssetPoolToken(StableAssetPoolId),
+	LiquidCrowdloan(Lease),
+	ForeignAsset(ForeignAssetId),
+}
+```
+
 ## Requirements
 
 If we are to deprecate Currency Factory, we will need to ensure the requirements
@@ -206,5 +292,7 @@ our new pallet-assets
 * [Slack thread detailing issues with Currency Factory Ranges](https://composablefinance.slack.com/archives/C031G5NT0CA/p1667492928188269)
 
 * [Parity's pallet-assets](https://github.com/paritytech/substrate/tree/master/frame/assets)
+
+* [Acala's Asset Registry](https://github.com/AcalaNetwork/Acala/tree/master/modules/asset-registry)
 
 * [Moonbeam's asset-manager](https://github.com/PureStake/moonbeam/tree/master/pallets/asset-manager)
--- a/rfcs/0013-deprecate-currency-factory.md
+++ b/rfcs/0013-deprecate-currency-factory.md
@@ -130,6 +130,8 @@ idea but would require much more substantial changes.
 
   * Supported by Polkadot wallets
   
+  * Supports asset metadata
+  
   #### Cons
   
   * No asset ID reservation system, only prevents duplicates
@@ -137,6 +139,9 @@ idea but would require much more substantial changes.
   * Hold functionality is not present (We depend on this in several cases)
 
 * **Moonbeam's pallet-asset-manager**
+
+Note: Not looking at using or forking this directly, but instead how they use 
+different instances of the same pallet for mintable and external assets.
 
   #### Pros
   
@@ -154,6 +159,40 @@ idea but would require much more substantial changes.
 existing storage for pallet-assets/pallet-asset-registry/pallet-currency-factory 
 to an alternative.
 
+## Conclusion
+
+After analyzing possible solutions, the best course of action to deprecate 
+Currency Factory is to maintain a fork of Parity's pallet-assets, embrace the 
+design Moonbeam uses to maintain two instances of their pallet, and to remove 
+our asset ID reservation system (for release two).
+
+This can be accomplished in two main phases:
+
+1. Create a fork of Parity's pallet-assets and add the features we need
+
+2. Replace our pallet-assets and pallet-currency-factory with two instances of
+Parity's pallet-assets.
+
+Parts of either of these phases can be scaled down for release two.
+
+### Fork of Parity's pallet-assets
+
+1. Create a non-traditional fork of Parity's pallet-assets
+
+2. Implement `frame_support::traits::tokens::fungibles::MutateHold` for 
+pallet-assets
+
+3. Translate calls into our old pallet-assets and pallet-currency-factory to 
+our new pallet-assets
+
+### Manage Two Instances of pallet-assets
+
+1. Configure instance for mintable assets
+
+2. Configure instance for external assets
+
+3. Route calls between the two instances
+
 ## Questions
 
 * > As for asset metadata, we don't have standards for formatting this nor do we 
@@ -165,3 +204,7 @@ to an alternative.
 # References
 
 * [Slack thread detailing issues with Currency Factory Ranges](https://composablefinance.slack.com/archives/C031G5NT0CA/p1667492928188269)
+
+* [Parity's pallet-assets](https://github.com/paritytech/substrate/tree/master/frame/assets)
+
+* [Moonbeam's asset-manager](https://github.com/PureStake/moonbeam/tree/master/pallets/asset-manager)
--- a/rfcs/0013-deprecate-currency-factory.md
+++ b/rfcs/0013-deprecate-currency-factory.md
@@ -15,7 +15,8 @@ stand to gain from. We really only have two types of currencies native
 (mintable) and non-native (external). However, we currently have many more 
 ranges that enforce artificial limitations on various abstractions of these two 
 currency types. This complicates the currency creation process and leads to 
-confusion while providing no benefit to us.
+confusion while providing no benefit to us. 
+Since asset IDs aren't intended for human consumption, there is no requirement for them to be "easily consumable" - the end user should be concerned with the asset's _symbol_, not it's id.
 
 ### No Permissionless Asset Creation
 
@@ -51,23 +52,22 @@ enforce in a future solution.
 ## Solution Proposals
 
 There are multiple ways we could go about replacing Currency Factory. Possible 
-solutions are not necessarily exclusive (i.e. we could implement multiple). 
+solutions are not necessarily mutually exclusive (i.e. we could implement multiple). 
 These solutions are detailed below:
 
 ### Simplify & Move the Asset ID Reservation System
 
 One way to remove the need for pallet-currency-factory is to move its 
 functionality to pallet-asset-registry. If we also simplify the asset ID 
-reservation system, this will be a minimal change.
+reservation system to no longer use predefined ranges, this will be a minimal change.
 
-This keeps the functionality of Currency Factory around but removes the need for
-another pallet. This does not enable premissionless asset creation without the
+This retains the functionality of Currency Factory, and does not require a new pallet. However, this does not enable permissionless asset creation without the
 need for more design.
 
 #### Technical Implementation
 
 Instead of reserving asset IDs via our current range system, we could simply 
-have a nonce for reserving new asset IDs while ensuring a lack of collisions. 
+have one nonce for reserving new asset IDs, which would (by the definition of a nonce) ensure no collisions. 
 This nonce can start at an arbitrary but high number so that our hard-coded 
 asset IDs are still safe. This would reduce the complexity of our current 
 reservation system while still avoiding collision.
--- a/rfcs/0013-deprecate-currency-factory.md
+++ b/rfcs/0013-deprecate-currency-factory.md
@@ -34,6 +34,20 @@ As for asset metadata, we don't have standards for formatting this nor do we
 utilize it. Assuming we replace the reservation system, Currency Factory will be 
 left without value.
 
+## Requirements
+
+If we are to deprecate Currency Factory, we will need to ensure the requirements
+it fulfills are still met by some solution.
+
+* Permissioned asset creation can be done without collision in the asset ID
+
+* Asset metadata is available
+
+Additionally, Currency Factory failed to meet some requirements that we should
+enforce in a future solution.
+
+* Permissionless asset creation can be done without collision in the asset ID
+
 ## Solution Proposals
 
 There are multiple ways we could go about replacing Currency Factory. Possible 
@@ -45,6 +59,10 @@ These solutions are detailed below:
 One way to remove the need for pallet-currency-factory is to move its 
 functionality to pallet-asset-registry. If we also simplify the asset ID 
 reservation system, this will be a minimal change.
+
+This keeps the functionality of Currency Factory around but removes the need for
+another pallet. This does not enable premissionless asset creation without the
+need for more design.
 
 #### Technical Implementation
 
@@ -68,6 +86,10 @@ reservation (any way making fork with reservation is easy later).
 Asset reservation helps us automate new asset IDs. However, for the scope of 
 release two - this may not be necessary. We could instead hard-code asset IDs 
 for LP tokens and other instances.
+
+Given we can currently manage without automated asset creation, as we can forgo
+the asset creation requirements for release two and ensure that all asset data
+is still made available.
 
 #### Consequences
 
--- a/rfcs/0013-deprecate-currency-factory.md
+++ b/rfcs/0013-deprecate-currency-factory.md
@@ -145,14 +145,13 @@ Local assets are registered with a `ForeignAssetType`, `AssetRegisterarMetadata`
 If we are to deprecate Currency Factory, we will need to ensure the requirements
 it fulfills are still met by some solution.
 
-* Permissioned asset creation can be done without collision in the asset ID
+* Permissioned asset creation can be done without collision
 
 * Asset metadata is available
 
-Additionally, Currency Factory failed to meet some requirements that we should
-enforce in a future solution.
+* Can registrar both local and foreign asset types
 
-* Permissionless asset creation can be done without collision in the asset ID
+* Support XCM Multi Locations for foreign assets
 
 ## Solution Proposals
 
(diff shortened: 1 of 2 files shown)


ComposableFiB0t:
Task linked: [CU-37t8q47 [Assessment] Final confirmation on deprecating Currency-factory pallet](https://app.clickup.com/t/37t8q47)

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,92 @@
...
+  * Partially specialized
+  
+  * No asset ID reservation system, only prevents duplicates
+
+## Questions
+
+* > As for asset metadata, we don't have standards for formatting this nor do we 
+  utilize it.
+  Is this true? A breif investigation revealed that outside of `scripts/polkadot-launch/initialization/src/interfaces/basilisk/definitions.ts` 
+  there is no mention of the Asset Metadata structure.
```
PoisonPhang, line 92: > As for asset metadata, we don't have standards for formatting this nor do we 
  utilize it.
  
 Is this true? A breif investigation revealed that outside of `scripts/polkadot-launch/initialization/src/interfaces/basilisk/definitions.ts` there is no mention of the Asset Metadata structure.

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,93 @@
...
+depreciating Currency Factory and its replacements.
+
+## Reasoning
+
+There are several reasons we stand to benefit from replacing Currency Factory.
+These reasons are detailed below:
+
+### Currency Factory Ranges are Confusing
+  
+As detailed in [this](https://composablefinance.slack.com/archives/C031G5NT0CA/p1667492928188269)
```
dzmitry-lahoda, line 13: please summarize details out of thread here

dzmitry-lahoda: (or you have done it already, just remove it or put into some no so distracting place - so we are not forced to read - i  know I do same way sometimes, but IMHO rfc could be cleaner)

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,93 @@
...
+enforce artificial limitations on various abstractions of these two currency 
+types. This complicates the currency creation process and leads to confusion 
+while providing no benefit to us.
+
+### Currency Factory is Already Losing Functionality
+
+Recently, we moved away from storing ED in Currency Factory. This leaves two 
+uses for currency factory: reserving asset IDs and storing asset metadata. 
+
+As we have already discussed, currency factory over complicates the asset ID
```
dzmitry-lahoda, line 26: not sure if this section and paragraph clear (clear for me in general, not sure it is clear for others)

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,93 @@
...
+### Simplify & Move the Asset ID Reservation System
+
+One way to remove the need for pallet-currency-factory is to move its 
+functionality to pallet-asset-registry. If we also simplify the asset ID 
+reservation system, this will be a minimal change.
+
+Instead of reserving asset IDs via our current range system, we could simply 
+have a nonce for reserving new asset IDs while ensuring a lack of collisions. 
+This nonce can start at an arbitrary but high number so that our hard-coded 
+asset IDs are still safe (i.e `u32::MAX`). This would reduce the complexity of
```
dzmitry-lahoda, line 47: Hardcoded could be u16::MAX, u32::MAX could be reserved for permissioned assets. 
So may be we can avoid to be too detailed in this RFC? I mean what is abstractions on top of registry? Just reservation of some subset of ids - we can specify max later.

Imho paragraphs which jump from biz value to tech details are hard to read. May be outline tech proposal later.

Imho using assets is more of biz and ecosystem and general development proposal, then tech.

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,93 @@
...
+Neither option presented seems to handle the reserving of asset IDs. Therefore, 
+neither of these will fully replace currency factory and would more correctly 
+replace our own pallet-assets. A replacement for pallet-assets may be a good 
+idea but would require much more substantial changes A replacement for 
+pallet-assets may be a good idea but would require much more substantial 
+changes.
+
+* **Parity's pallet-assets**
+  
+  #### Pros
```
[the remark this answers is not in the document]
dzmitry-lahoda: - supported by polkadot wallets

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,93 @@
...
+  
+  * Simple Interface
+  
+  #### Cons
+  
+  * No asset ID reservation system, only prevents duplicates
+  
+  * Hold functionality is not present (We depend on this in several cases)
+
+* **Moonbeam's pallet-asset-manager**
```
dzmitry-lahoda, line 75: we should not use assets-manager imho. i used moonbeam just as example of good use of pallet-assets (it is proves assets are good for ETH bridges )

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,93 @@
...
+
+There are several reasons we stand to benefit from replacing Currency Factory.
+These reasons are detailed below:
+
+### Currency Factory Ranges are Confusing
+  
+As detailed in [this](https://composablefinance.slack.com/archives/C031G5NT0CA/p1667492928188269) 
+thread, the Currency Factory ranges are a confusing abstraction that we don't 
+stand to gain from. We really only have two types of currencies native 
+(mintable) and non-native (external). However, we have many more ranges that
```
dzmitry-lahoda, line 16: second issue - we do not have permissionless asset (non payable/no sufficitent) - it is impossible to create no marketable assets (which is needed for governance/airdrops)

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,93 @@
...
+  
+  * Very nice UI compatibility with Polkadot Dashboard
+  
+  * Maintained by Parity
+  
+  * Simple Interface
+  
+  #### Cons
+  
+  * No asset ID reservation system, only prevents duplicates
```
dzmitry-lahoda, line 71: we do not need. we just runtime config assets to be premissioned. when we will have permisionless assets they may already have reservation (any way making fork with reservation is easy later)

dzmitry-lahoda: so really we do not care lack of reservation system

dzmitry-lahoda: now

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,93 @@
...
+  
+  * Hold functionality is not present (We depend on this in several cases)
```
dzmitry-lahoda, line 73: we can use assets-registry here. it can do next
1. implement orml traits
2. implement substrate traits
3. implement free/reserved/hold/lock in consistent/property tested/fuzzied way
4. route calls to assets

rfcs/0013-deprecate-currency-factory.md
dzmitry-lahoda: pallet-asset-manager - idea we will have our own `assets-registy` which will be tune for our needs.

it is important to check that `pallet-assets` can be wrapped

dzmitry-lahoda:
we have now orml-tokens - what other teams do with these?

dzmitry-lahoda:
also, if assets are wrappet with our assets-registry, we should be clear what Call to assets would be forbidden (governance?)

dzmitry-lahoda:
about locks and reservations. how wallets/polkadotjs will work with that in assets? One way is if assets are locked or else, just move them to derived account. So you can easy transfer and send, up to ED on main account. But all locks, reserves, holds will never be touched. Other options - is to surface 100% same exstrincs/storage in assets-registy OR in our fork.

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Solution Proposals
+
+There are multiple ways we could go about replacing Currency Factory. These 
+solutions are detailed below:
+
+### Simplify & Move the Asset ID Reservation System
+
+One way to remove the need for pallet-currency-factory is to move its 
+functionality to pallet-asset-registry. If we also simplify the asset ID
```
dzmitry-lahoda, line 41: 1. we are 4 month old orml-tokens. what state they have now? may be they did some stuff for non payable assets and substrate assets compatibility?
2. could we move functionally to our assets pallet? can we make our assets pallet to be as compatible as possible with substrate assets?

so in theory we can ditch currency factory and move all to our assets in a way compable on binary level with substrate assets.

also we can throw away our governance-registry imho

dzmitry-lahoda:
It is not clear from RFC why substrate assets are good. Imho it should be clarified - year of man work to make UI/QA/FE/users happy is what we lack. Imho better to have fork of assets and collaborate with ecosystem to tune wallets. Do not write own wallets.

rfcs/0013-deprecate-currency-factory.md
PoisonPhang: +1
I've included this in the solution section

rfcs/0013-deprecate-currency-factory.md
PoisonPhang: > we are 4 month old orml-tokens. what state they have now? may be they did some stuff for non payable assets and substrate assets compatibility?

We'll either need to request another back-port or upgrade our runtime to enjoy their newer additions

> could we move functionally to our assets pallet? can we make our assets pallet to be as compatible as possible with substrate assets?

Like you concluded later, I think maintaining a fork would be the easiest way to get all the benefits of Parity's pallet-assets

rfcs/0013-deprecate-currency-factory.md
PoisonPhang: > implement free/reserved/hold/lock in consistent/property tested/fuzzied way

Is there any particular reason that hold and lock/freeze are implemented differently?

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,145 @@
...
+Through RFCs 0009 and 0010, we decided to move away from storing ED in Currency 
+Factory. After the removal of ED storage, Currency Factory was left with two 
+uses: reserving asset IDs and storing asset metadata. 
+
+As discussed, Currency Factory overcomplicates the asset ID reservation system. 
+As for asset metadata, we don't have standards for formatting this nor do we 
+utilize it. Assuming we replace the reservation system, Currency Factory will be 
+left without value.
+
+## Solution Proposals
```
vimukthi-git, line 37: Going directly to solutions without giving any concrete requirements fulfilled by this solution based on the provided reasoning seems odd to me. That means the readers can not actually assess whether a solution is both necessary and sufficient.

vimukthi-git (changes requested):
RFC is incomplete without concrete goals or requirements it's trying to fulfill. Hence can not review as of now.

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,210 @@
...
+3. Translate calls into our old pallet-assets and pallet-currency-factory to 
+our new pallet-assets
+
+### Manage Two Instances of pallet-assets
+
+1. Configure instance for mintable assets
+
+2. Configure instance for external assets
+
+3. Route calls between the two instances
```
PoisonPhang, line 194: Looking for comments on this.
I was considering using Assets Registry as a routing layer, but that may be out of scope for Assets Registry.

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,210 @@
...
+### Currency Factory is Already Losing Functionality
+
+Through RFCs 0009 and 0010, we decided to move away from storing ED in Currency 
+Factory. After the removal of ED storage, Currency Factory was left with two 
+uses: reserving asset IDs and storing asset metadata. 
+
+As discussed, Currency Factory overcomplicates the asset ID reservation system. 
+As for asset metadata, we don't have standards for formatting this nor do we 
+utilize it. Assuming we replace the reservation system, Currency Factory will be 
+left without value.
```
benluelo, line 35: are there plans to replace the reservation system? what are possible alternatives?

rfcs/0013-deprecate-currency-factory.md
benluelo: Is our old substrate version holding us back from using newer features? And if so, what is keeping us on an older substrate version?

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,210 @@
...
+
+Additionally, Currency Factory failed to meet some requirements that we should
+enforce in a future solution.
+
+* Permissionless asset creation can be done without collision in the asset ID
+
+## Solution Proposals
+
+There are multiple ways we could go about replacing Currency Factory. Possible 
+solutions are not necessarily exclusive (i.e. we could implement multiple).
```
benluelo, line 54:
```suggestion
solutions are not necessarily mutually exclusive (i.e. we could implement multiple). 
```

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,210 @@
...
+
+There are multiple ways we could go about replacing Currency Factory. Possible 
+solutions are not necessarily exclusive (i.e. we could implement multiple). 
+These solutions are detailed below:
+
+### Simplify & Move the Asset ID Reservation System
+
+One way to remove the need for pallet-currency-factory is to move its 
+functionality to pallet-asset-registry. If we also simplify the asset ID 
+reservation system, this will be a minimal change.
```
benluelo, line 61:
```suggestion
reservation system to no longer use predefined ranges, this will be a minimal change.
```

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,210 @@
...
+
+This keeps the functionality of Currency Factory around but removes the need for
+another pallet. This does not enable premissionless asset creation without the
+need for more design.
```
benluelo, line 65:
```suggestion
This retains the functionality of Currency Factory, and does not require a new pallet. However, this does not enable permissionless asset creation without the
need for more design.
```

benluelo: Not sure what's meant by "more design" here, perhaps it could be rephrased?

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,210 @@
...
+
+#### Technical Implementation
+
+Instead of reserving asset IDs via our current range system, we could simply 
+have a nonce for reserving new asset IDs while ensuring a lack of collisions. 
+This nonce can start at an arbitrary but high number so that our hard-coded 
+asset IDs are still safe. This would reduce the complexity of our current 
+reservation system while still avoiding collision.
```
benluelo, line 73:
```suggestion
Instead of reserving asset IDs via our current range system, we could simply 
have one nonce for reserving new asset IDs, which would (by the definition of a nonce) ensure no collisions. 
This nonce can start at an arbitrary but high number so that our hard-coded 
asset IDs are still safe. This would reduce the complexity of our current 
reservation system while still avoiding collision.
```

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,210 @@
...
+These reasons are detailed below:
+
+### Currency Factory Ranges are Confusing
+  
+The Currency Factory ranges are a confusing abstraction that we don't 
+stand to gain from. We really only have two types of currencies native 
+(mintable) and non-native (external). However, we currently have many more 
+ranges that enforce artificial limitations on various abstractions of these two 
+currency types. This complicates the currency creation process and leads to 
+confusion while providing no benefit to us.
```
benluelo, line 18:
```suggestion
The Currency Factory ranges are a confusing abstraction that we don't 
stand to gain from. We really only have two types of currencies native 
(mintable) and non-native (external). However, we currently have many more 
ranges that enforce artificial limitations on various abstractions of these two 
currency types. This complicates the currency creation process and leads to 
confusion while providing no benefit to us. 
Since asset IDs aren't intended for human consumption, there is no requirement for them to be "easily consumable" - the end user should be concerned with the asset's _symbol_, not it's id.
```

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,210 @@
...
+reservation system while still avoiding collision.
+
+#### Consequences
+
+* If we already have assets in various ranges, the runtime migration process
+will become more complicated as we translate them to existing in one range.
+  
+### Remove Asset ID Reservation
+
+> We do not need [asset ID reservation]. We just runtime config assets to be
```
benluelo, line 82: what is meant by "We just runtime config assets to be permissioned"?

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,210 @@
...
+This nonce can start at an arbitrary but high number so that our hard-coded 
+asset IDs are still safe. This would reduce the complexity of our current 
+reservation system while still avoiding collision.
+
+#### Consequences
+
+* If we already have assets in various ranges, the runtime migration process
+will become more complicated as we translate them to existing in one range.
+  
+### Remove Asset ID Reservation
```
benluelo, line 80: What is the purpose of asset id reservation? (this is a gap in my knowledge, but a short explanation/ refresher would be nice to have in this document regardless)

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,210 @@
...
+* Any of these options will require a difficult runtime migration as we move 
+existing storage for pallet-assets/pallet-asset-registry/pallet-currency-factory 
+to an alternative.
+
+## Conclusion
+
+After analyzing possible solutions, the best course of action to deprecate 
+Currency Factory is to maintain a fork of Parity's pallet-assets, embrace the 
+design Moonbeam uses to maintain two instances of their pallet, and to remove 
+our asset ID reservation system (for release two).
```
benluelo, line 167: what benefits does forking bring over simply wrapping the pallet?

benluelo: Could we upstream the changes we need? This may be a more long-term task, so possibly out of scope for what we want to do with this RFC; perhaps a "future possibilities" section would be good?

rfcs/0013-deprecate-currency-factory.md
PoisonPhang: We lack a system for creating assets without root/governance
Which may become something we need in the future. Currency Factory is permissioned

rfcs/0013-deprecate-currency-factory.md
PoisonPhang: Meaning that, as long as a permissioned source is creating an asset (root or governance) we are safe to assume that we don't need to automate the ID acquisition process

rfcs/0013-deprecate-currency-factory.md
PoisonPhang: Ensuring new asset IDs don't collide with existing ones

rfcs/0013-deprecate-currency-factory.md
PoisonPhang: Up-streaming is an option, probably not one for release two tho.

rfcs/0013-deprecate-currency-factory.md
benluelo: got it, so if we end up having a requirement for permissionless asset creation we would have to design it

rfcs/0013-deprecate-currency-factory.md
benluelo: that would be solved by using a single nonce right?

rfcs/0013-deprecate-currency-factory.md
benluelo: makes sense, i think that should be mentioned as explicitly out of scope then

rfcs/0013-deprecate-currency-factory.md
PoisonPhang: It could be, yes

benluelo: This is related to https://github.com/ComposableFi/composable/pull/2489/files#r1037535250, I think that section could be updated with this

benluelo: Maybe not in that section specifically, but somewhere (lol)

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,210 @@
...
+As for asset metadata, we don't have standards for formatting this nor do we 
+utilize it. Assuming we replace the reservation system, Currency Factory will be 
+left without value.
+
+## Requirements
+
+If we are to deprecate Currency Factory, we will need to ensure the requirements
+it fulfills are still met by some solution.
+
+* Permissioned asset creation can be done without collision in the asset ID
```
vimukthi-git, line 43: The requirement is not clear here. Why do we need permissioned asset creation? and through what interface? through an extrinsic or pallet interface trait?

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,210 @@
...
+stand to gain from. We really only have two types of currencies native 
+(mintable) and non-native (external). However, we currently have many more 
+ranges that enforce artificial limitations on various abstractions of these two 
+currency types. This complicates the currency creation process and leads to 
+confusion while providing no benefit to us. 
+Since asset IDs aren't intended for human consumption, there is no requirement for them to be "easily consumable" - the end user should be concerned with the asset's _symbol_, not it's id.
+
+### No Permissionless Asset Creation
+
+> We do not have permissionless asset (non-payable/no sufficient) - it is
```
vimukthi-git, line 23: I'm don't really get this statement, why do we need to create new assets for governance and airdrops?
And what is meant by "no marketable assets"?

rfcs/0013-deprecate-currency-factory.md
vimukthi-git: Good question, this background section is not detailed enough on what other projects are doing for example. Is there a best practice?

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,210 @@
...
+it fulfills are still met by some solution.
+
+* Permissioned asset creation can be done without collision in the asset ID
+
+* Asset metadata is available
+
+Additionally, Currency Factory failed to meet some requirements that we should
+enforce in a future solution.
+
+* Permissionless asset creation can be done without collision in the asset ID
```
vimukthi-git, line 50: The requirement is not clear here. Why do we need permissionless asset creation? and through what interface? through an extrinsic or pallet interface trait?

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,210 @@
...
+left without value.
+
+## Requirements
+
+If we are to deprecate Currency Factory, we will need to ensure the requirements
+it fulfills are still met by some solution.
+
+* Permissioned asset creation can be done without collision in the asset ID
+
+* Asset metadata is available
```
vimukthi-git, line 45: We need to be more specific here what type of meta data are we talking about?

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,210 @@
...
+Through RFCs 0009 and 0010, we decided to move away from storing ED in Currency 
+Factory. After the removal of ED storage, Currency Factory was left with two 
+uses: reserving asset IDs and storing asset metadata. 
+
+As discussed, Currency Factory overcomplicates the asset ID reservation system. 
+As for asset metadata, we don't have standards for formatting this nor do we 
+utilize it. Assuming we replace the reservation system, Currency Factory will be 
+left without value.
+
+## Requirements
```
vimukthi-git, line 38: I recommend following [MosCow prioritization](https://www.productplan.com/glossary/moscow-prioritization) to understand if a requirement is actually a requirement while also understanding priority.

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,210 @@
...
+
+This assessment aims to gather all relevant details and consequences of 
+depreciating Currency Factory and its replacements.
+
+## Reasoning
+
+There are several reasons we stand to benefit from replacing Currency Factory.
+These reasons are detailed below:
+
+### Currency Factory Ranges are Confusing
```
vimukthi-git, line 11: :+1:

rfcs/0013-deprecate-currency-factory.md
vimukthi-git: I'm not even sure we need general collision avoidance tbh, We may only need some namespace mechanism such as [MultiLocation](https://polkadot.network/blog/xcm-the-cross-consensus-message-format/#-assets-in-xcm)

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,210 @@
...
+
+* Permissioned asset creation can be done without collision in the asset ID
+
+* Asset metadata is available
+
+Additionally, Currency Factory failed to meet some requirements that we should
+enforce in a future solution.
+
+* Permissionless asset creation can be done without collision in the asset ID
+
```
vimukthi-git, line 51: Are we missing any migration requirement from the existing solution to any new solution?

vimukthi-git (commented):
I would like this to cover 
- Background: the diff between what we are doing as of now assets/currency-factory and what other well known projects are doing to understand the best way to move forward
- Requirement: As commented in the section

rfcs/0013-deprecate-currency-factory.md
PoisonPhang: If @dzmitry-lahoda could elaborate on this, that would be appreciated

rfcs/0013-deprecate-currency-factory.md
PoisonPhang: I've started a section on what other chains have historically done. Acala section has been pushed with one on Moonbeam being a WIP

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,345 @@
...
+is to use Assets Registry to register both local and foreign assets. Assets
+Registry will act as a routing layer between our two instances of pallet-assets.
+
+This will leave us with two independently stored asset ranges (local and 
+foreign) without the overhead of Currency Factory. 
+
+### Technical
+
+1. Asset IDs & Balances
+
```
PoisonPhang, line 290: Instead of the complexity of the enum, routing can be handled internally by the pallet by confirm if some storage items exist or not - at the cost of storage access

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,345 @@
...
+```
+
+### Moonbeam
+
+Moonbeam has two asset classifications: 
+
+* Local
+* Foreign
+
+Moonbeam maintains separate instances of Parity's pallet-assets to track each of
```
vimukthi-git, line 131: This approach sounds neat for our asset-registry to eventually evolve in to

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,345 @@
...
+## Requirements
+
+If we are to deprecate Currency Factory, we will need to ensure the requirements
+it fulfills are still met by some solution.
+
+* Permissioned asset creation can be done without collision
+
+* Asset metadata is available
+
+* Can registrar both local and foreign asset types
```
vimukthi-git, line 152: Can we try to define what assets are considered foreign(assets created by other sovereign chains than the current executing chain) vs local(assets created by the chain that executes the logic). This will be helpful for anyone trying to understand how we handle assets.

dzmitry-lahoda: did not looked, but likely moonbeam uses Call filters to prevent any calls to extrinsic on foreign assets registry to prevent any minting or burning even via sudo.

dzmitry-lahoda:
one thing, all assets in Statemine have an implicit location. As it has only locally issued assets and only one pallet. It just harcoded translation layer for any new asset to to (0, 50, asset_id). So they do not store default values.

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,345 @@
...
+Given we can currently manage without automated asset creation, as we can forgo
+the asset creation requirements for release two and ensure that all asset data
+is still made available.
+
+#### Consequences
+
+* Minor runtime migration will be needed for deleting Currency Factory's 
+storage.
+
+### Use an Externally Maintained Asset ID Reservation System
```
vimukthi-git, line 205: This is a separate concern as proposed before, we can focus on concern one for release 2. But not a problem to document the solution. I agree in  concept though we need to assess how well it supports our requirements such as asset locking at the time of implementation(release 3 or after)

rfcs/0013-deprecate-currency-factory.md
vimukthi-git: Agree for not using Moonbeam stuff but we can learn from their impl :)

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,345 @@
...
+  
+  #### Cons
+  
+  * Partially specialized
+
+  * No asset ID reservation system, only prevents duplicates
+
+#### Consequences
+
+* Any of these options will require a difficult runtime migration as we move
```
vimukthi-git, line 264: I don't think runtime migration is difficult if we design and plan the steps correctly.

rfcs/0013-deprecate-currency-factory.md
```diff
@@ -0,0 +1,345 @@
...
+#### Consequences
+
+* Any of these options will require a difficult runtime migration as we move 
+existing storage for pallet-assets/pallet-asset-registry/pallet-currency-factory 
+to an alternative.
+
+## Conclusion
+
+After analyzing possible solutions, the best course of action to deprecate 
+Currency Factory is to maintain a fork of Parity's pallet-assets, embrace the
```
vimukthi-git, line 271: Note that pallet-assets is still a work in progress https://github.com/paritytech/substrate/issues/8453 , we need `fungibles::*Hold` traits for staking in release 3, which pallet-assets does not support. Any solution suggesting adoption of that pallet need to take into account orml-tokens features that we need for release 3. I would rather deal with a migration into pallet-assets(it supports that for genesis https://github.com/paritytech/substrate/blob/master/frame/assets/src/lib.rs#L362) later than delay our feature roadmap. Note that the number of tokens(LPT * 3, PBLO, pool shares * 3) would not be that large even after release 3 and staking. We can even hard code a migration for that. 

I propose the following steps to impl in brief

#### Release 2: 
1. Make sure that we don't use currency-factory and create and track(1) the assets in assets-registry with `MultiLocation` even for local assets.
2. Allow Pablo to create assets(LPT) through assets-registry (avoid use of currency-factory)
3. Make sure one can not use local assets to pay gas fees in XCM transfers (Which seems to be the case AFAIK)
4. For ED for LPT: Asset ratios gets defined in a suitable way at the time of creation for LP (1 LP = value of provided liquidity in PICA / number of tokens)

#### Release 3: 
1. Introduce staking with pool share tokens (Just for context, not related to RFC)
2. Remove currency-factory all together (Just to minimize the changes for release 2)
3. Possibly upgrade substrate to latest production version (assess later).

#### Release 4 or later
1. Assess/Introduce frame/pallet-assets with a migration for existing set of assets if `fungibles::*Hold` trait impls are available.
2. Remove use of our own pallet-assets together with orml-tokens use.

PoisonPhang:
Closing in favor of #2877

PR #2490, opened 2022-11-29
#wip #draft process items RFC

dzmitry-lahoda:

## Checklist

- [ ] Updated the rust/typescript docs.
- [ ] Updated the `docs/`.

### DIFF
--- a/rfcs/0013-development-process-and-project-items.md
+++ b/rfcs/0013-development-process-and-project-items.md
@@ -35,22 +35,45 @@ Next is set of guidlines and constrains to operate CU items which are considered
 
 ### CU guidlines
 
-People tagged or mentioned in CI or watching relevant tasks are expected to respond within 24 working hours. 
-GitHub (GH) issues, GH metions or review requests notifications are good help, but not required. 
+`ETA` - is 24 working hours
+`Item` - CU item 
+
+#### Commnication
+
+People tagged or mentioned in item or anybody from watching an item are expected to respond within ETA.
+GitHub (GH) issues, GH metions or review requests notifications are good help, but not required.
+Slack notifications are also good will, but optional.
+
+#### Depedencies
+
 If there is dependency on exteranal our of companu CU work item work or dependcy, it should be accompanied by in CU referencing work item
+
+#### Item existence
+
+Any work which is not considered to be done or planned should not be in CU. 
+Repeating, work which is not considered to be done should not be click up item.
+
+CU creation can be dicussed. If dicussion is long and inteream, task to align/decide should be created.
+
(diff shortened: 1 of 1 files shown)


andyjsbell:
@dzmitry-lahoda thanks for this.  We are in the process of aligning CU/GH with our work practices.  Let's use this RFC as the base for this.  I will probably review/commit to this PR during this process so please keep this open and in draft for the time being 👍

dzmitry-lahoda:
(

> @dzmitry-lahoda thanks for this. We are in the process of aligning CU/GH with our work practices. Let's use this RFC as the base for this. I will probably review/commit to this PR during this process so please keep this open and in draft for the time being 👍

:)) In the end, there is todo to described on how I align with SAFE/Scrum (so it is not require any). Core CU item flow is more or less cleaned (so no Grammarly yet on it), but why and consideration and Ghant change through experiment and alignment with SAFE/Scrum is not clarified.

And importantly, it does not specifies any hierarchy or task structure (milestones, features) - actually it consider all items to be just items (so need to clarify how `long` term running milestone/feature is same as small feature).

Also need to clarify how PO/PMs can act as `manual labour for making our Ghant Chart` because all tools sucks to automate this (specifically LibreOffice only tool I used which does Ghant Chart right).

And also how to compose items flow with tags is not clarified too (so as with hierarchy - there is no need to use tags, but would be good to use)

andyjsbell:
@dzmitry-lahoda I thought we agreed to leave this open, why did you close it?

dzmitry-lahoda:
It was opened for a while so. I closed it because
1. Needs updates regarding ownership definition (so task is always owned directly or indirectly and we can always sure progresses by hierarchy of ownership).
2. Updates to handling ETA, mentioned and comments.
3. Other updates I recalled to make all rules consistent, non self contradictory and allow to be run (so the goal is clear to move forward the tasks chart and able resolve any state case) 

And that no solves then education people about. And lack of interest in topic. And there is no incentive to handle that. So nothing grounds investment into it more then I did (more then some time to draft).

dzmitry-lahoda:
@andyjsbell and the document will be 3x size. 1. rules 2. reasoning 3. examples. 
i am pretty sure it is possible to make it consistent and teach people. so there should be acceptance we want to work like that - specifically, people should care to review or comment on it.

dzmitry-lahoda:
if we 100% defined work to be done. we create placeholder task for sure in proper hierarchy and with proper tags/fields because we would do it 100% and then fill details.

dzmitry-lahoda:
export day to day report of activity of subordinate to manager

PR #2503, opened 2022-11-30
[CU-30xvvdv] Improve weight checks for normalization

PoisonPhang:
## Issue

[[CU-30xvvdv]](https://app.clickup.com/t/30xvvdv)

## Additional Changes

Added `WeightMath` trait to dex math.

Also improved check for non-zero weights.

### DIFF
--- a/code/parachain/frame/composable-maths/src/dex/mod.rs
+++ b/code/parachain/frame/composable-maths/src/dex/mod.rs
@@ -5,23 +5,31 @@ pub mod price;
 #[cfg(test)]
 pub mod tests;
 
-pub trait PoolWeightMathExt<'t>: PerThing + 't {
+pub trait PoolWeightMathExt<'t>: 't {
 	/// Sums all weights in an array. If weights total greater than `Perthing::one()`, returns
 	/// `None`.
-	fn sum_weights(weights: impl IntoIterator<Item = &'t Self>) -> Option<Self>;
+	fn sum_weights<Weight: 't + PerThing + CheckedAdd>(
+		weights: impl IntoIterator<Item = &'t Weight>,
+	) -> Option<Weight>;
 
 	/// Returns `true` if all weights in array are non-zero.
-	fn non_zero_weights(weights: impl IntoIterator<Item = &'t Self>) -> bool;
+	fn non_zero_weights<Weight: 't + PerThing>(
+		weights: impl IntoIterator<Item = &'t Weight>,
+	) -> bool;
 }
 
 impl<'t, T: PerThing + CheckedAdd + 't> PoolWeightMathExt<'t> for T {
-	fn sum_weights(weights: impl IntoIterator<Item = &'t Self>) -> Option<Self> {
+	fn sum_weights<Weight: 't + PerThing + CheckedAdd>(
+		weights: impl IntoIterator<Item = &'t Weight>,
+	) -> Option<Weight> {
 		weights
 			.into_iter()
-			.try_fold(Self::zero(), |total_weight, weight_n| total_weight.checked_add(weight_n))
+			.try_fold(Weight::zero(), |total_weight, weight_n| total_weight.checked_add(weight_n))
 	}
 
-	fn non_zero_weights(weights: impl IntoIterator<Item = &'t Self>) -> bool {
+	fn non_zero_weights<Weight: 't + PerThing>(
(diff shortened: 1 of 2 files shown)


ComposableFiB0t:
Task linked: [CU-30xvvdv 5.3) 8. Ensure normalized weights are used everywhere.](https://app.clickup.com/t/30xvvdv)

code/parachain/frame/composable-maths/src/dex/mod.rs
```diff
@@ -1,4 +1,33 @@
...
+
+impl<T: PerThing + CheckedAdd> WeightMath for T {
+	fn sum_weights(weights: &[Self]) -> Option<Self> {
+		weights
+			.iter()
+			.try_fold(Self::zero(), |total_weight, weight_n| total_weight.checked_add(weight_n))
+	}
+
+	fn non_zero_weights(weights: &[Self]) -> bool {
+		for weight in weights {
```
vimukthi-git, line 25: We can probably combine the two functions into one with an error return if a weight is zero as we don't expect any weights to be zero in any case

PoisonPhang: While I agree it would be a slight performance optimization, I wanted to keep these functions with single responsibility. While the `non_zero_weights` function is likely only to be used in one context, the `sum_weights` function has some more generic utility.

vimukthi-git (approved):
LGTM

code/parachain/frame/composable-maths/src/dex/mod.rs
```diff
@@ -1,4 +1,33 @@
...
+			if weight.is_zero() {
+				return false
+			}
+		}
```
benluelo, line 29: this could just be `weights.iter().all(|w| !weight.is_zero())`

code/parachain/frame/composable-maths/src/dex/mod.rs
```diff
@@ -1,4 +1,33 @@
...
 #[cfg(test)]
 pub mod tests;
+
+pub trait WeightMath: PerThing {
+	/// Sums all weights in an array. If weights total greater than `Perthing::one()`, returns
+	/// `None`.
+	fn sum_weights(weights: &[Self]) -> Option<Self>;
+
+	/// Returns `true` if all weights in array are non-zero.
+	fn non_zero_weights(weights: &[Self]) -> bool;
```
benluelo, line 14: these fns could take `impl IntoIterator<Item = &Self>` to avoid unnecessary `.collect::<Vec>()` (and avoiding `.copied()`/`.cloned()` by not needing to take ownership of `Self`)

code/parachain/frame/composable-maths/src/dex/mod.rs
benluelo: it might also be possible to directly pass `Zero::is_zero`, not 100% sure though

code/parachain/frame/composable-maths/src/dex/mod.rs
```diff
@@ -1,4 +1,33 @@
+use sp_runtime::{traits::CheckedAdd, PerThing};
+
 pub mod constant_product;
 pub mod price;
 #[cfg(test)]
 pub mod tests;
+
+pub trait WeightMath: PerThing {
```
benluelo, line 8: NIT: I would rename this to `WeightMathExt`, since it's an extension trait

code/parachain/frame/pablo/src/dual_asset_constant_product.rs
```diff
@@ -27,14 +28,14 @@ impl<T: Config> DualAssetConstantProduct<T> {
...
-		);
-		ensure!(
-			weights[0].deconstruct() + weights[1].deconstruct() ==
-				Permill::from_percent(100).deconstruct(),
+			if let Some(total_weight) = assets_weights.values().sum_weights() {
+				total_weight.is_one()
+			} else {
+				// If `sum_weights` returns `None`, it overflowed and weights are not normalized
+				false
+			},
```
benluelo, line 38:
```suggestion
            assets_weights
                .values()
                .sum_weights()
                .map(|total_weight| total_weight.is_one())
				// If `sum_weights` returns `None`, it overflowed and weights are not normalized
                .unwrap_or(false)
```

benluelo (approved):
lgtm :rocket:

PR #2557, opened 2022-12-05
Fix implementation of compute_first_deposit_lp_

benluelo:
## Issue

The current implementation of `compute_first_deposit_lp_` was erroneously using the balance instead of the deposit. This PR removes the balance parameter from the function and refactors the input into a struct for clarity.

### DIFF
--- a/code/parachain/frame/composable-maths/src/dex/constant_product.rs
+++ b/code/parachain/frame/composable-maths/src/dex/constant_product.rs
@@ -465,29 +465,40 @@ pub fn compute_deposit_lp(
 ///
 /// https://github.com/ComposableFi/composable/blob/main/rfcs/0008-pablo-lbp-cpp-restructure.md#42-liquidity-provider-token-lpt-math-updates
 /// Equation 6
-pub fn compute_first_deposit_lp_<T: PerThing>(
-	pool_assets: &[(u128, u128, T)],
+pub fn compute_first_deposit_lp_<T, I>(
+	pool_assets: I,
 	_f: T,
-) -> ConstantProductAmmResult<ConstantProductAmmValueFeePair> {
+) -> ConstantProductAmmResult<ConstantProductAmmValueFeePair>
+where
+	T: PerThing,
+	I: IntoIterator<Item = ConstantProductFirstDepositInput<T>>,
+	I::IntoIter: ExactSizeIterator,
+{
+	let mut pool_assets = pool_assets.into_iter();
+
 	let k: u128 = pool_assets.len().try_into().map_err(|_| ArithmeticError::Overflow)?;
+
 	ensure!(!k.is_zero(), ConstantProductAmmError::InvalidTokensList);
 
-	let product = pool_assets.iter().try_fold::<_, _, Result<_, ArithmeticError>>(
-		Decimal::from(1),
-		|product, (_d_i, b_i, w_i)| {
-			let b_i = Decimal::safe_from_fixed_point(*b_i)?;
-			let w_i = Decimal::safe_from_per_thing(*w_i)?;
-			let pow = b_i.checked_powd(w_i).ok_or(ArithmeticError::Overflow)?;
+	let product =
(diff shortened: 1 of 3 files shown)


code/parachain/frame/composable-maths/src/dex/constant_product.rs
```diff
@@ -465,23 +465,29 @@ pub fn compute_deposit_lp(
 ///
 /// https://github.com/ComposableFi/composable/blob/main/rfcs/0008-pablo-lbp-cpp-restructure.md#42-liquidity-provider-token-lpt-math-updates
 /// Equation 6
-pub fn compute_first_deposit_lp_<T: PerThing>(
-	pool_assets: &[(u128, u128, T)],
+pub fn compute_first_deposit_lp_<T, I>(
+	pool_assets: I,
 	_f: T,
```
benluelo, line 470: Do we need this field if it's always 0? cc @vimukthi-git @PoisonPhang

PoisonPhang: I was under the impression that we planned to apply fees here in the future

code/parachain/frame/composable-maths/src/dex/constant_product.rs
```diff
@@ -465,23 +465,29 @@ pub fn compute_deposit_lp(
...
-) -> ConstantProductAmmResult<ConstantProductAmmValueFeePair> {
+) -> ConstantProductAmmResult<ConstantProductAmmValueFeePair>
+where
+	T: PerThing,
+	I: IntoIterator<Item = ConstantProductFirstDepositInput<T>>,
+	I::IntoIter: ExactSizeIterator,
```
benluelo, line 475: I also refactored this fn to be a bit more general, IMO this is a much cleaner interface.

PoisonPhang (approved):
LGTM
Good catch

code/parachain/frame/composable-maths/src/dex/constant_product.rs
```diff
@@ -490,6 +496,11 @@ pub fn compute_first_deposit_lp_<T: PerThing>(
 	Ok(ConstantProductAmmValueFeePair { value: product.safe_to_fixed_point()?, fee: 0 })
 }
 
+pub struct ConstantProductFirstDepositInput<T> {
```
vimukthi-git, line 499: Let's not introduce new types to keep consistency for the moment with all interfaces/functions in this file. For now simply removing the `balance` parameter works for first deposit use as the balance is always 0 for first deposit.

code/parachain/frame/composable-maths/src/dex/constant_product.rs
```diff
@@ -465,23 +465,29 @@ pub fn compute_deposit_lp(
 ///
 /// https://github.com/ComposableFi/composable/blob/main/rfcs/0008-pablo-lbp-cpp-restructure.md#42-liquidity-provider-token-lpt-math-updates
 /// Equation 6
-pub fn compute_first_deposit_lp_<T: PerThing>(
-	pool_assets: &[(u128, u128, T)],
```
vimukthi-git, line 469: Just removing the first `u128` param here would just solve the issue without a refactor as explained elsewhere.

code/parachain/frame/composable-maths/src/dex/constant_product.rs
benluelo: What consistency is being kept? Newtypes are used throughout this file (`ConstantProductAmmResult<_>`, `ConstantProductAmmValueFeePair`, `AssetAmount`, etc). I think using a newtype is actually _more_ consistent with what we currently have

code/parachain/frame/composable-maths/src/dex/constant_product.rs
benluelo: See https://github.com/ComposableFi/composable/pull/2572. I would still like to get these changes merged afterwards

vimukthi-git: Yes, we can consider better interface/structure for this file later after we finish with Pablo overall restructure.

PR #2577, opened 2022-12-06
[No Ticket] Protect Picasso runtime from out of consensus modification

vimukthi-git:
This is to protect Picasso runtime from being modified without consensus. Need a process to verify runtime changes so that we don't run into blockers at the time of release.


benluelo (approved):

KaiserKarel (approved):

andyjsbell (approved):
Thanks @vimukthi-git

vimukthi-git:
@mergifyio refresh

PR #2586, opened 2022-12-07
Chores for general CI hygiene.

KaiserKarel:


ComposableFiB0t:
Task linked: [CU-32qxxxb Github actions hygiene](https://app.clickup.com/t/32qxxxb)

KaiserKarel:
I'll need to rename some required checks in the settings tab to get this to merge; so once approved, will do so.

.github/workflows/benchmark.yml
```diff
@@ -1,68 +1,69 @@
-name: Run Benchmarks
+name: Update benchmarks weights
```
dzmitry-lahoda, line 1: imho we could simplify and make more consistent our ci if we would have rule of file = name = label

Update benchmarks weights -> snake -> update-benchmarks-weights.yaml -> kebab -> update-benchmarks-weights-${{ matrix.runtime }}-${{ github.ref }} -> label == update-benchmarks-weights

.github/workflows/build-test-deploy.yml
```diff
@@ -621,6 +621,35 @@ jobs:
           command: nix -- build .#devnet
           token: "${{ secrets.CACHIX_AUTH_TOKEN }}"
 
+  lint-test-frontend:
+    # Legacy because these should be nixified.
```
dzmitry-lahoda, line 625: imho may remove comment. we kind of know this. as soon as it will be non stable (because of vendor is not locked nor shared by nix) we would nixify.

dzmitry-lahoda (approved):

.github/workflows/benchmark.yml
KaiserKarel: Let's save that for a later PR, but agree

.github/workflows/build-test-deploy.yml
KaiserKarel: Just keeping that so FE knows why it is legacy.

hussein-aitlahcen (approved):

obsessed-cake (approved):

vimukthi-git (approved):

PR #2606, opened 2022-12-07
Routed Pablo LP Token Creation Through Assets Registry [No-CU]

PoisonPhang:
## Issue

LP Tokens were getting created without a ratio, leading to an ED of `Balance::MAX`


ComposableFiB0t:
Task linked: [CU-30xvvdg 5.3) 6. Implement Algorithm: Amm::add_liquidity.](https://app.clickup.com/t/30xvvdg)

code/parachain/frame/pablo/src/dual_asset_constant_product.rs
```diff
@@ -43,8 +44,11 @@ impl<T: Config> DualAssetConstantProduct<T> {
 			Error::<T>::WeightsMustSumToOne
 		);
 		ensure!(fee_config.fee_rate < Permill::one(), Error::<T>::InvalidFees);
-
-		let lp_token = T::CurrencyFactory::create(RangeId::LP_TOKENS)?;
+		let lp_token = T::AssetsRegistry::register_asset(
+			XcmAssetLocation::LOCAL_NATIVE,
```
vimukthi-git, line 48: This would not work AFAIK as we need to use a unique ID here

vimukthi-git (commented):
LGTM, looks pretty minimal

code/parachain/frame/pablo/src/dual_asset_constant_product.rs
vimukthi-git: Tracking an ever incrementing next LP token ID in Pablo would be needed

benluelo: It calls into currency factory

vimukthi-git: No I mean in asset-registry this value is used as a key for `ForeignToLocal` so this would not work after the first LP token if it's not unique

benluelo: so this should perhaps be called something like "create_and_register"?

PoisonPhang: Two+ counters

PoisonPhang: > Tracking an ever incrementing next LP token ID in Pablo would be needed

Two+ counters/nonce for one type of token seems like a terrible misuse of storage

vimukthi-git: hmm not really that bad when you think about the way `Multilocation` supposed to work https://github.com/paritytech/xcm-format#71-description. So Pablo is at the protocol/smart contract layer. Eventually we can migrate the assets-registry to remove tracking of another local identifier all together and just use `Multilocation` for mappings. This way we decentralize all asset "unique" ID tracking to each protocol. Anyway more stuff to add to your RFC later. For now just generating and storing a nonce is ok for Pablo IMO.

PoisonPhang: This asset should also never be considered foreign. We'll eventually want AR (or something else) to be able to route between foreign and local assets. Maybe it's worth taking the time to at least prepare the interface of AR for that. Even if we don't go all the way with a full asset revamp just yet.

This would allow us to dynamically create local assets and, with a Multi Location, create foreign assets without unnecessary complexity and keeping storage abuse at a minimum.

I think something like Moonbeam's Asset Registrar can be implemented here but with more elegance.

vimukthi-git: >This asset should also never be considered foreign.

Yes, as we agreed it's a temp solution. We can use the multi location to understand if the given asset is foreign or not later.

> We'll eventually want AR (or something else) to be able to route between foreign and local assets

yes for BYOG I guess.

>This would allow us to dynamically create local assets and, with a Multi Location, create foreign assets without unnecessary complexity and keeping storage abuse at a minimum.

Yes!

>I think something like Moonbeam's Asset Registrar can be implemented here but with more elegance.

That would be cool to evolve into :)

PoisonPhang: I made updates to #2489 with my thoughts on this system

PoisonPhang:
Closing as this will instead be handled by #2624

PR #2634, opened 2022-12-09
#XCVM-244 making two workspaces for rust compilation of incompatible coswasm and substrate deps

dzmitry-lahoda:
## Issue

need to build contracts from separate cargo to allow several version of deps to live (mostly substrate vs comswasm conflict)

## Additional Changes

- some hack to build multi folder src
- nix fmt
- all rs code is same as before (minus formatting)
- some bridge.rs parts are behind cosmwasm fg did (not sure why they did not fail compile before)

## Checklist

- native cargo builds ok
- crane builds ok
- ci passes (waiting)


ComposableFiB0t:
Task linked: [XCVM-244 split cargo workspace to solve compilation issues](https://app.clickup.com/t/20465559/XCVM-244)

dzmitry-lahoda:
seems all sketeched. need to fix nix rust deps and targets wasm builds - and it is done

PR #2643, opened 2022-12-09
[CU-34qhnj1] add lend and unlend functions to Lending pallet

kkast:
add lend and unlend functions to Lending pallet, fixing tests

### DIFF
--- a/code/parachain/frame/composable-traits/src/lending/mod.rs
+++ b/code/parachain/frame/composable-traits/src/lending/mod.rs
@@ -27,6 +27,8 @@ pub enum CollateralRatio<T> {
 
 pub type CollateralLpAmountOf<T> = <T as DeFiEngine>::Balance;
 
+pub type LendAssetAmountOf<T> = <T as DeFiEngine>::Balance;
+
 pub type BorrowAmountOf<T> = <T as DeFiEngine>::Balance;
 
 #[derive(Encode, Decode, Default, TypeInfo, RuntimeDebug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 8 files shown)


benluelo:
`lend` and `unlend` are pretty vague - perhaps something like `deposit` and `withdraw` (which is the same as the underlying vault functions called in these extrinsics)

Also, what's the purpose of these extrinsics? The relevant clickup ticket is empty: https://app.clickup.com/t/34qhnj1

benluelo:
Also please run `nix run ".#fmt"`

kkast:
> `lend` and `unlend` are pretty vague - perhaps something like `deposit` and `withdraw` (which is the same as the underlying vault functions called in these extrinsics)
> 
> Also, what's the purpose of these extrinsics? The relevant clickup ticket is empty: https://app.clickup.com/t/34qhnj1

the idea to use lend and unlend instead of deposit and withdraw i think is because there is already deposit_collateral and withdraw_collateral. rn, Lending pallet doesnt have extrinsic to work with vault. it needs to use vault extrinsics to deposit or withdraw. we want to have these functions here in order to pause them if necessary.

benluelo:
> > `lend` and `unlend` are pretty vague - perhaps something like `deposit` and `withdraw` (which is the same as the underlying vault functions called in these extrinsics)
> > Also, what's the purpose of these extrinsics? The relevant clickup ticket is empty: https://app.clickup.com/t/34qhnj1
> 
> the idea to use lend and unlend instead of deposit and withdraw i think is because there is already deposit_collateral and withdraw_collateral. rn, Lending pallet doesnt have extrinsic to work with vault. it needs to use vault extrinsics to deposit or withdraw. we want to have these functions here in order to pause them if necessary.

makes sense, but I see no reason why they couldn't be called `vault_deposit` and `vault_withdraw` though. Perhaps @kevin-composable can comment here

> we want to have these functions here in order to pause them if necessary.
Maybe adding pausing functionality would be good?

dzmitry-lahoda:
imho it is good to have pallet methods for that. so i never seen such naming before. can we have diffeerent names which are more common?

dzmitry-lahoda (approved):

kevin-aftermath (commented):
This looks perfect; `vault_deposit` and `vault_withdraw` sound good to me

kevin-composable (commented):
This looks perfect; `vault_deposit` and `vault_withdraw` sound good to me

kevin-composable (approved):

vimukthi-git (approved):
LGTM :) Though I'm unable to provide much in-depth review or insight without a design/spec or details in the task linked.

ComposableFiB0t:
Task linked: [CU-34qhnj1 Add vault_deposit + vault_withdraw extrinsics](https://app.clickup.com/t/34qhnj1)

PR #2659, opened 2022-12-12
Add add and delete extrinsics to pallet-crowdloan

KaiserKarel:


ComposableFiB0t:
Task linked: [CU-39qvt6b Add 'add' and 'delete' extrinsics to pallet-crowdloan](https://app.clickup.com/t/39qvt6b)

code/parachain/frame/crowdloan-rewards/src/lib.rs
```diff
@@ -101,25 +101,42 @@ pub mod pallet {
...
-		OverFunded { excess_funds: T::Balance },
+		OverFunded {
+			excess_funds: T::Balance,
+		},
 		/// A portion of rewards have been unlocked and future claims will not have locks
-		RewardsUnlocked { at: MomentOf<T> },
+		RewardsUnlocked {
+			at: MomentOf<T>,
+		},
+		///
```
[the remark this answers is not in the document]
KaiserKarel: woops, will add

code/parachain/frame/crowdloan-rewards/src/lib.rs
```diff
@@ -339,6 +359,32 @@ pub mod pallet {
 			Self::do_unlock(reward_accounts);
 			Ok(())
 		}
+
+		/// Deletes all accounts provided in the `deletions` vector. If the provided `Balance` does
+		/// not match the current balance of the account, the deletion will fail.
+		#[pallet::weight(<T as Config>::WeightInfo::populate(deletions.len() as _))]
+		#[transactional]
+		pub fn delete(
```
vimukthi-git, line 367: Not sure if we need two extrinsics here given that all we need is one transaction to correct the contributions as per original total

code/parachain/frame/crowdloan-rewards/src/lib.rs
```diff
@@ -350,6 +396,49 @@ pub mod pallet {
...
+			}
+
+			TotalRewards::<T>::set(total_rewards);
+			TotalContributors::<T>::set(total_contributors);
+			Ok(())
+		}
+
+		/// Add new rewards to the pallet. Updates total_rewards and contributors accordingly.
+		pub fn do_add(
+			additions: Vec<(RemoteAccountOf<T>, RewardAmountOf<T>, VestingPeriodOf<T>)>,
```
vimukthi-git, line 428: I think we can avoid adding a delete by using the new reward amounts as 0 for removed ones. Saves much of the changes needed.

vimukthi-git (commented):
IMO we may

code/parachain/frame/crowdloan-rewards/src/lib.rs
```diff
@@ -350,6 +396,49 @@ pub mod pallet {
...
+						ensure!(rewards.total == expected, Error::<T>::UnexpectedRewardAmount);
+						TotalRewards::<T>::mutate(|total| *total -= rewards.total);
+						TotalContributors::<T>::mutate(|total| *total -= 1);
+					}
+					*rewards = None;
+					Ok::<(), Error<T>>(())
+				})?;
+			}
+
+			TotalRewards::<T>::set(total_rewards);
```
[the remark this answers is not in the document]
KaiserKarel: Oof that's a mistake.

code/parachain/frame/crowdloan-rewards/src/lib.rs
KaiserKarel: So two separate extrinsics, one doing do_delete // set rewards to 0.
Or confounding them in 1 extrinsic? I initially had the do_replacements extrinsic perform both operations, but that was a bit messy. `delete` and `add` will be called within utility.batch anyway.

code/parachain/frame/crowdloan-rewards/src/lib.rs
KaiserKarel: Well we call them using batch, so it'll be 1 transaction.

code/parachain/frame/crowdloan-rewards/src/lib.rs
```diff
@@ -331,6 +350,7 @@ pub mod pallet {
 		}
 
 		#[pallet::weight(<T as Config>::WeightInfo::unlock_rewards_for(reward_accounts.len() as _))]
+		#[transactional]
```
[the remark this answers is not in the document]
KaiserKarel: <insert boomer meme>

code/parachain/frame/crowdloan-rewards/src/lib.rs
```diff
@@ -350,6 +381,54 @@ pub mod pallet {
...
 	impl<T: Config> Pallet<T> {
+		/// Patch existing rewards by deleting them.
+		/// This will cause inconsistencies if a claim has already been made. Errors can be avoided
+		/// by passing the expected amount of rewards claimable.
+		///
+		/// Security:
+		/// * Should be called within a transactional context, as this may return an error but still
+		///   change
+		/// storage.
+		pub(crate) fn do_delete(
```
vimukthi-git, line 389: NIT: We may comment this out as well not to have this affect the audits

code/parachain/frame/crowdloan-rewards/src/lib.rs
```diff
@@ -350,6 +378,54 @@ pub mod pallet {
...
+			additions: Vec<(RemoteAccountOf<T>, RewardAmountOf<T>, VestingPeriodOf<T>)>,
+		) -> DispatchResult {
+			Self::in_uninitialized(|| Self::do_populate(additions))
+		}
+
+		/// Runs the provided function while setting the crate to uninitialized. Useful during
+		/// patching.
+		pub(crate) fn in_uninitialized<F: FnOnce() -> DispatchResult>(f: F) -> DispatchResult {
+			let start = VestingTimeStart::<T>::take();
+			let result = f();
```
vimukthi-git, line 424: We may need this validation from `do_initialize`
```
let excess_funds = available_funds
				.checked_sub(&total_rewards)
				.ok_or(Error::<T>::RewardsNotFunded)?;

```
To avoid a case of not having enough funds to support the rewards

vimukthi-git (approved):
LGTM, I think the validation may be needed to avoid a case of users claims erroring out. But I think that is manageable. We may do some manual checks on the balance of the pallet account in Picasso chain to verify that would not happen.

code/parachain/frame/crowdloan-rewards/src/lib.rs
KaiserKarel: That plays poorly with 0ed updated; will add a check to do_populate.

code/parachain/frame/crowdloan-rewards/src/lib.rs
KaiserKarel: Agreed, we also want to keep the CTO at the bottom of contributors.

code/parachain/frame/crowdloan-rewards/src/lib.rs
```diff
@@ -339,6 +341,36 @@ pub mod pallet {
...
+		// 	Self::deposit_event(Event::RewardsDeleted { deletions });
+		// 	Ok(())
+		// }
+
+		/// Adds all accounts in the `additions` vector. Add may be called even if the pallet has
+		/// been initialized.
+		#[pallet::weight(<T as Config>::WeightInfo::populate(additions.len() as _))]
+		pub fn add(
+			origin: OriginFor<T>,
+			additions: Vec<(RemoteAccountOf<T>, RewardAmountOf<T>, VestingPeriodOf<T>)>,
```
andyjsbell, line 367: Could it be better to use a `BoundedVec` here?  Looks like we use `Vec` all the way through the pallet so maybe a subsequent thing to refactor.

code/parachain/frame/crowdloan-rewards/src/lib.rs
```diff
@@ -350,6 +382,63 @@ pub mod pallet {
...
+		// 			}
+		// 			*rewards = None;
+		// 			Ok::<(), Error<T>>(())
+		// 		})?;
+		// 	}
+
+		// 	TotalRewards::<T>::set(total_rewards);
+		// 	TotalContributors::<T>::set(total_contributors);
+		// 	Ok(())
+		// }
```
andyjsbell, line 415: Maybe we should clarify why this is also commented out with the line used above in the other commented block: 
`// README: Disabled to reduce the amount of auditing work. When enabling, also enable the
		// tests.`
		
Not a biggy but why do we need to leave the commented code in?

code/parachain/frame/crowdloan-rewards/src/tests.rs
```diff
@@ -188,6 +188,45 @@ fn test_populate_ok() {
...
+// 		let rewards = gen(100, DEFAULT_REWARD);
+// 		Balances::make_free_balance_be(&CrowdloanRewards::account_id(), expected_total_rewards);
+// 		assert_ok!(CrowdloanRewards::populate(Origin::root(), rewards));
+// 		assert_eq!(CrowdloanRewards::total_rewards(), expected_total_rewards);
+// 		assert_eq!(CrowdloanRewards::claimed_rewards(), 0);
+// 		CrowdloanRewards::delete(Origin::root(), rewards).unwrap();
+// 		assert_eq!(CrowdloanRewards::total_rewards(), 0);
+// 	});
+// }
+
```
andyjsbell, line 210: If we don't need it then let's just remove it

code/parachain/frame/crowdloan-rewards/src/lib.rs
KaiserKarel: Technically it is bounded by the maximum size of a transaction.

code/parachain/frame/crowdloan-rewards/src/lib.rs
KaiserKarel: Agree, will just remove.

PR #2666, opened 2022-12-13
Add filtering for rustSrc

KaiserKarel:


dzmitry-lahoda (commented):
rust may embeed anything. json for example. build.rs can co sume configs.

KaiserKarel:
> rust may embeed anything. json for example. build.rs can co sume configs.

Then the build will fail, and we can include it in the filter?

dzmitry-lahoda:
yes. build will fail. so imho it is premature optimization.

dzmitry-lahoda (approved):

cor (approved):

KaiserKarel:
@mergifyio requeue

PR #2686, opened 2022-12-13
#xcvm #cosmwasm build contracts in #ci

dzmitry-lahoda:


dzmitry-lahoda:
ok, lfs issue. will do new PR

PR #2691, opened 2022-12-13
[No Ticket] Require broader agreement for RFCs

vimukthi-git:


.mergify.yml
```diff
@@ -27,4 +27,17 @@ pull_request_rules:
...
       queue:
         priority: high
+        name: default
+
+  # Any RFCs must be reviewed and approved by more than 3 reviewers plus core group
+  - name: Automatic merge on approval (RFC)
+    conditions:
+      - and:
+          - "#approved-reviews-by>=3"
+          - approved-reviews-by=@ComposableFi/core
```
vimukthi-git, line 37: Could be
```suggestion
          - approved-reviews-by=@ComposableFi/parachain-leads
```

dzmitry-lahoda: imho 
1. this file should be same security because it can override rfc easy then
2. some documents, specifically release documents could be as important as rfc

dzmitry-lahoda: like there could be rfc suggesting way to do releases, but release document could suggest counter rfc without stating explicitly.

dzmitry-lahoda: cc @andyjsbell

dzmitry-lahoda: also, not sure, but last time i checked leads did not covered all from core. so this pr will disable some of core to sign (as it was - not sure - may be changed)

andyjsbell: I think this would help with some description for this PR on why this change is necessary. @vimukthi-git would you be able to help finish this please?

dzmitry-lahoda (approved):

.mergify.yml
dzmitry-lahoda: build-test-deploy.yml <- i see somebody disabled udeps, coverage and simnode
1. simnode - we do not know how changes in code  behave well on upgrades (at least picasso) let not do firefighting on release
2. we do not consider coverage report as part of risk reducing factor (example XCM)
3. udeps - XCVM team bloated our runtime with fork of everything without alignment with other teams

so this file is also important for higher review
1. release files
2. RFC process files
3. CI gates
4. merfigy

dzmitry-lahoda: >  Decrease lead time until after release 3

Could be done removing it from gate and continue on error. It still would work but faster lead time. Or disable behind label like #long

dzmitry-lahoda: So if if harden RFC files, we should harden other """process""" files (CI, merfgiy, release)

PR #2698, opened 2022-12-14
fix(cosmwasm): use migrate from cosmwasm vm

aeryz:
* Fix migrate to check if the admin check is done properly before execution.
* Make sure that gas is properly refund if the execution fails.

### DIFF
--- a/code/parachain/frame/cosmwasm/src/lib.rs
+++ b/code/parachain/frame/cosmwasm/src/lib.rs
@@ -536,11 +536,9 @@ pub mod pallet {
 				CodeIdentifier::CodeHash(code_hash) =>
 					CodeHashToId::<T>::try_get(code_hash).map_err(|_| Error::<T>::CodeNotFound)?,
 			};
-			let outcome = EntryPointCaller::<MigrateCall>::setup(who, contract, new_code_id)?.call(
-				&mut shared,
-				Default::default(),
-				message,
-			);
+			let outcome =
+				EntryPointCaller::<MigrateCall>::setup(&mut shared, who, contract, new_code_id)?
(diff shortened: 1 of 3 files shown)


ComposableFiB0t:
Task linked: [XCVM-257 Fix `migrate` entry point call](https://app.clickup.com/t/20465559/XCVM-257)

code/parachain/frame/cosmwasm/src/lib.rs
```diff
@@ -536,11 +536,9 @@ pub mod pallet {
 				CodeIdentifier::CodeHash(code_hash) =>
 					CodeHashToId::<T>::try_get(code_hash).map_err(|_| Error::<T>::CodeNotFound)?,
```
hussein-aitlahcen, line 537: this must not fail either

code/parachain/frame/cosmwasm/src/lib.rs
```diff
@@ -609,24 +548,8 @@ pub mod pallet {
...
-						vm,
-						&Addr::unchecked(Self::account_to_cosmwasm_addr(who)),
-						CosmwasmAccount::new(contract.clone()),
-						None,
-					)
-					.map_err(Into::into)
-				},
-			);
-			Self::deposit_event(Event::<T>::AdminCleared { contract, old_admin: info.admin });
+			let outcome = Self::do_update_admin(&mut shared, who, contract.clone(), None);
```
hussein-aitlahcen, line 551: Can we remove this extrinsic and make `update_admin` accept an option?

hussein-aitlahcen (approved):

KaiserKarel (approved):

PR #2714, opened 2022-12-14
Ben/pablo tests suite refine rewrite (#390y0vn)

benluelo:
For a description of the build script, see here: https://github.com/benluelo/substrate-pallet-testing-framework/blob/main/test-builder/README.md

Down the line we would be able to just use that package directly, but for now I have just copy-pasted the build script.

If opening in an editor with an LSP configuration that enables all features by default, I recommend setting `SKIP_TESTS_BUILD_SCRIPT=1` as the build script is not yet well optimized.

### DIFF
--- a/code/parachain/frame/composable-tests-helpers/src/lib.rs
+++ b/code/parachain/frame/composable-tests-helpers/src/lib.rs
@@ -1,2 +1,15 @@
 #![cfg_attr(not(feature = "std"), no_std)]
+
+use sp_runtime::AccountId32;
+
 pub mod test;
+
+pub const ALICE: AccountId32 = AccountId32::new([
+	0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1,
+]);
+pub const BOB: AccountId32 = AccountId32::new([
+	0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2,
+]);
+pub const CHARLIE: AccountId32 = AccountId32::new([
+	0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 3,
+]);
--- a/code/parachain/frame/composable-tests-helpers/src/lib.rs
+++ b/code/parachain/frame/composable-tests-helpers/src/lib.rs
@@ -1,18 +1,38 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
+use sp_keyring::sr25519::Keyring;
 use sp_runtime::AccountId32;
 
 pub mod test;
 
-pub const ALICE: AccountId32 = AccountId32::new([
-	0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1,
-]);
-pub const BOB: AccountId32 = AccountId32::new([
-	0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2,
-]);
-pub const CHARLIE: AccountId32 = AccountId32::new([
-	0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 3,
-]);
-pub const DAVE: AccountId32 = AccountId32::new([
-	0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 4,
-]);
+// pub const ALICE: AccountId32 = AccountId32::new([
+// 	0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1,
+// ]);
+// pub const BOB: AccountId32 = AccountId32::new([
+// 	0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2,
+// ]);
+// pub const CHARLIE: AccountId32 = AccountId32::new([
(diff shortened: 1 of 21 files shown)


ComposableFiB0t:
Task linked: [CU-390y0vn [Pablo/Dex] Refine & Rewrite Pallet Tests Suite](https://app.clickup.com/t/390y0vn)

code/parachain/frame/composable-tests-helpers/src/lib.rs
```diff
@@ -1,2 +1,15 @@
 #![cfg_attr(not(feature = "std"), no_std)]
+
+use sp_runtime::AccountId32;
+
 pub mod test;
+
+pub const ALICE: AccountId32 = AccountId32::new([
```
dzmitry-lahoda, line 7: in xcm tests i used real keyring. would be easier to write simnode tests.

code/parachain/frame/pablo/build.rs
```diff
@@ -0,0 +1,197 @@
+#![allow(clippy::disallowed_methods)] // temporarilly
+
+use std::{env, fmt::Write, fs, iter, path::PathBuf};
+
+use syn::{Generics, Item, ItemFn, ItemMod, ReturnType, Signature, Visibility};
+
```
dzmitry-lahoda, line 6: some comment what this script does would be nice

code/parachain/frame/pablo/src/testing.rs
```diff
@@ -0,0 +1,297 @@
...
+	type __SystemBlockNumber: From<u32> + Into<u64> + SafeAdd;
+	type __SystemAccountId: From<AccountId32> + Clone;
+	type __TimestampMoment: From<u64>;
+	type __PabloBalance: BalanceLike + From<u128> + Zero;
+	// type __AssetId: Clone;
+}
+
+impl<T> PabloRuntimeConstraints for T
+where
+	T: RuntimeTrait<Event<Self>>,
```
dzmitry-lahoda, line 55: hm. seems it may work for similar case in production code in cosmwasm pallet

code/parachain/frame/pablo/build.rs
benluelo: yeah, working on those right now. once it's ready i'll move the PR out of draft

code/parachain/frame/pablo/src/testing.rs
benluelo: it's a nifty trick, can save a lot of boilerplate if you use the same constraints a lot. there are some limitations with it though that i'm documenting as well

code/parachain/frame/composable-tests-helpers/src/lib.rs
benluelo: not sure what you're referring to - what's keyring?

code/parachain/frame/pablo/build.rs
benluelo: updated the PR description with a link to some documentation on what the build script does and how it works :+1:

code/parachain/frame/pablo/src/test/dual_asset_constant_product_tests.rs
```diff
@@ -629,16 +637,16 @@ fn staking_pool_test() {
...
-			matches!(e.event,
-	            mock::Event::StakingRewards(pallet_staking_rewards::Event::RewardPoolCreated { owner, .. })
-	            if owner == Pablo::account_id(&pool_id) )
-		});
+		// make sure a Staking pool is created.
+		// assert_has_event::<Test, _>(|e| {
+		// 	matches!(e.event,
+		//            mock::Event::StakingRewards(pallet_staking_rewards::Event::RewardPoolCreated {
+		// owner, .. })            if owner == Pablo::account_id(&pool_id) )
+		// });
```
PoisonPhang, line 645: NIT: Remove

code/parachain/frame/pablo/src/test/dual_asset_constant_product_tests.rs
```diff
@@ -629,16 +637,16 @@ fn staking_pool_test() {
...
 
 		let bob_usdt = 45_000_u128 * unit;
-        let trading_fee = Perbill::from_float(0.05).mul_floor(bob_usdt);
-        let protocol_fee = Perbill::from_float(0.2).mul_floor(trading_fee);
+		let trading_fee = Perbill::from_float(0.05).mul_floor(bob_usdt);
+		let protocol_fee = Perbill::from_float(0.2).mul_floor(trading_fee);
```
PoisonPhang, line 649: NIT: While here, perhaps we can use something other than `from_float`

code/parachain/frame/pablo/src/testing.rs
```diff
@@ -0,0 +1,299 @@
...
+				pool_id,
+				assets.clone(),
+				Zero::zero(),
+				true,
+			),
+			Event::<Runtime>::LiquidityAdded {
+				who: BOB.into(),
+				pool_id,
+				asset_amounts: assets,
+				minted_lp: 19_999_999_993_470_955_u128.into(),
```
PoisonPhang, line 191: Lets depend on the functions from `composable-math` here instead of a hard-coded value. If we want more testing for the math functions correctness, we can add those to the `composable-math` crate. This also applies to similar numbers throughout this test.

PoisonPhang (approved):
Mostly NIT comments. Functionally correct
:rocket:

code/parachain/frame/pablo/src/test/dual_asset_constant_product_tests.rs
benluelo: agreed, will switch to from_rational

code/parachain/frame/pablo/src/testing.rs
benluelo: good point, updated

code/parachain/frame/composable-tests-helpers/src/lib.rs
KaiserKarel: sp_keyring

KaiserKarel: It has well known accounts hardcoded, such as Alice and Bob

benluelo: yes, @PoisonPhang filled me in on that, cool stuff. I'm working on using that right now

code/parachain/frame/pablo/src/testing.rs
KaiserKarel: Why not just flip the bound around?

```
where Runtime: pablo::Config?
```

code/parachain/frame/pablo/build.rs
```diff
@@ -0,0 +1,197 @@
...
+fn main() {
+	aggregate_tests("testing");
+}
```
KaiserKarel, line 9: Please sanity check that nix does run this and that these test work under nix :)

KaiserKarel (changes requested):
My one nit is that RuntimeConstraints is verbose. Is there no way around that?

vimukthi-git (commented):
Looks like a good idea and I don't have a particular technical argument against it other than it looks way over-engineered at pallet level compared to the benefit we get out of it. I would like to see some alternative on running Pablo tests with Picasso configuration at Picasso runtime level without much modification to Pablo pallet if possible.

benluelo:
> alternative on running Pablo tests with Picasso configuration at Picasso runtime level without much modification to Pablo pallet

This isn't possible since the current Pablo tests are (very!) tightly coupled to the mock runtime; either we update the mock configuration to match Picasso, and refactor the tests to suit (which will then inevitably go out of sync), or we write the tests in a way that allows us to inject any runtime.

benluelo:
I worked on this branch over the weekend to evaluate the DX of the build script, and I've come to the conclusion that it's not acceptable. rust-analyzer is unable to provide any form of hints/types/etc, since any change to the tests requires re-checking both dali and picasso runtimes. I am currently looking into alternatives, with an external tool/ cargo plugin being the most promising (something like `cargo subtest ..` to generate and run the runtime tests, basically the exact same as the current build script but external to the code)

PR #2715, opened 2022-12-14
[CU-865bad17j] Add pool types to Pablo events

andresmechali:
https://app.clickup.com/t/865bad17j

### DIFF
--- a/code/parachain/frame/pablo/src/lib.rs
+++ b/code/parachain/frame/pablo/src/lib.rs
@@ -105,18 +105,24 @@ pub mod pallet {
 		},
 	}
 
+	#[derive(
+		RuntimeDebug, Encode, Decode, MaxEncodedLen, CloneNoBound, PartialEq, Eq, TypeInfo,
+	)]
+	pub enum PoolConfigurationType {
+		DualAssetConstantProduct,
+	}
+
 	impl<AccountId: Clone + PartialEq + Debug, AssetId: Clone + Ord + Debug>
 		PoolInitConfiguration<AccountId, AssetId>
 	{
-		fn get_type(&self) -> Vec<u8> {
-			let pool_type = match self {
+		fn get_type(&self) -> PoolConfigurationType {
+			match self {
 				PoolInitConfiguration::DualAssetConstantProduct {
 					owner: _,
 					fee: _,
 					assets_weights: _,
-				} => "DualAssetConstantProduct",
-			};
-			pool_type.as_bytes().to_vec()
+				} => PoolConfigurationType::DualAssetConstantProduct,
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


ComposableFiB0t:
Task linked: [CU-865bad17j Add pool type to Pablo events](https://app.clickup.com/t/865bad17j)

code/parachain/frame/pablo/src/lib.rs
```diff
@@ -105,13 +105,45 @@ pub mod pallet {
 		},
 	}
 
+	#[derive(
+		RuntimeDebug, Encode, Decode, MaxEncodedLen, CloneNoBound, PartialEq, Eq, TypeInfo,
+	)]
+	pub enum PoolConfigurationType {
```
vimukthi-git, line 111: Is there no way to avoid this?

andresmechali: The extra enum? probably yes, but it seemed the most clear/straightforward to me, and that makes it typed on Subsquid as well. We could use a simple string or `Vec<u8>` with the name but that might be more prone to errors. Any suggestion? Ben mentioned `core::mem::Discriminant`

andresmechali: The extra enum? probably yes, but it seemed the most clear/straightforward to me, and that makes it typed on Subsquid as well. We could use a simple string or `Vec<u8>` with the name but that might be more prone to errors. Any suggestion? Ben mentioned `core::mem::Discriminant`

benluelo (changes requested):
The pool type information required by subsquid shouldn't leak into the Pablo api - this information is available in storage and should be retrieved from there

andresmechali:
> The pool type information required by subsquid shouldn't leak into the Pablo api - this information is available in storage and should be retrieved from there

@benluelo  Is it preferable to query the chain on every event? I could also keep track of the `poolId -> type` on Subsquid if needed.

benluelo:
Either way works, but if you want to reduce the chain queries then I'd say go for the pool id -> pool type mapping (and you can use the discriminant for that, or perhaps something like [strum](https://docs.rs/strum/latest/strum/)?)

code/parachain/frame/pablo/src/lib.rs
KaiserKarel: I prefer an enum too. discrimants might change if we add enum variants to the top of an enum:

```
enum Foo {
    One // discriminant == 0
    Two // discriminant == 1
}

// adding a new type:
enum Foo {
    Three // discriminant == 0
    One // discriminant == 1
    Two // discriminant == 2
}

```

KaiserKarel: @andresmechali you can use https://docs.rs/enum-kinds-macros/latest/enum_kinds_macros/ to generate the enum.

benluelo: Adding variants at the beginning of the enum will also break the scale encoding (unless we explicitly add the `#[codec(index = n)]` ofc, but still bad practice; any new variants should be added to the end of the enum). enum-kinds looks good though, and I'm not opposed to it, as long as it's only used internally in subsquid.

andresmechali: Thanks @KaiserKarel  and @benluelo .
I think that to keep it simple and not having to maintain an extra enum I can live with just querying the type from storage when the pool is created and then query it internally on Subsquid when needed

andresmechali: Also we avoid bloating the events with data that's only relevant for Subsquid, as Ben mentioned in other comment

vimukthi-git: Given that pool creation is a low frequency event it's ok to query storage (though not a generally good practice for event sourcing systems because of the possibility of overloading upstream systems).

PR #2708, opened 2022-12-14
#legacy #CU-3v0yb6h #CU-3vhy8q9 #CU-3u4b2mw migrated staging to zombienet

dzmitry-lahoda:
Signed-off-by: Dzmitry Lahoda <dzmitry@lahoda.pro>

waiting for merge of other PRs

# Testing

- waits for PRs about preventing reducing risks to switch to be merged

- build and run devnets via nix (native, docker, docker-compose work - composable only and complete)
- ??? staging testing ?

PR #2724, opened 2022-12-15
#CU-31pqtuu checking nix rot and nixops (we finally """compile""" nix)

dzmitry-lahoda:

👁 https://github.com/NixOS/nix/issues/7464 <- vote for nix to be come """compiled"""  👁

- dry run nixops staging 🛡
- evaluates all nix branches to prevent any rot (cc @radupopa2010 )
- fixes some rot :)  ⛑
- sets these as gate to merge to main
- is required for https://github.com/ComposableFi/composable/pull/2708
- in modified commands started to use nix profile instead of env as per guidelines (with cachix which support deploy :) )


.github/workflows/build-test-deploy.yml
```diff
@@ -592,41 +620,22 @@ jobs:
...
-      image: niteo/nixpkgs-nixos-22.05:316b762afdb9e142a803f29c49a88b4a47db80ee
-    steps:
-      - uses: actions/checkout@v3
-      - uses: "./.github/templates/watch-exec"
-        with:
-          command: nix -- build .#kusama-picasso-karura-devnet
-          token: "${{ secrets.CACHIX_AUTH_TOKEN }}"
-
   cache-devnet-all-dev-local:
-    name: "Package Full Devnet"
```
KaiserKarel, line 618: We'll still need a name key here, to ensure we can add a branch protection rule

dzmitry-lahoda: imho should use display name = name so we can be sure that any refactoring does not breaks hidden GH setting.

dzmitry-lahoda:
template failed with 

```
    DOCKER_USER_OPTION: $UID:$GID
Watching /nix/store for new store paths ...
Error: Process completed with exit code 130.
```

ComposableFiB0t:
Task linked: [CU-31pqtuu Migrate to zombienet](https://app.clickup.com/t/31pqtuu)

.github/workflows/build-test-deploy.yml
```diff
@@ -592,41 +620,23 @@ jobs:
...
+    name: "Cache devnet all dev local"
```
KaiserKarel, line 624:
```suggestion
    name: "Build all devnets"
```

KaiserKarel (approved):

.github/workflows/build-test-deploy.yml
```diff
@@ -136,6 +137,33 @@ jobs:
...
+      - uses: cachix/cachix-action@f5f67badd061acb62b5c6e25e763572ca8317004
+        with:
+          skipPush: true
+          installCommand: |
+            nix-channel --add ${{ env.NIX_NIXPKGS_CHANNEL }} nixpkgs
+            nix-channel --update
+            nix-env -iA nixpkgs.cachix
+          name: composable-community
+      - run: |
+          chmod +x ./.github/workflows/check-nix-long.sh
```
obsessed-cake, line 164: this is extra
```
$ cat <<EOF > foo.sh
> export BAR=123
> EOF

$ echo $BAR
# empty string

$ source foo.sh 
$ echo $BAR
123
```

dzmitry-lahoda: you mean you propose to store sh in yaml where it is not listed neither testable without GH?

dzmitry-lahoda: or you propose to use `./` instead of `source ./`?

.github/workflows/build-test-deploy.yml
dzmitry-lahoda: this name seems simpler and better. 
But imho name should be same as display name.  `Build all devnets` = `build-all-devnets`.  
Why would somebody want to have name != display name?

.github/workflows/check-nix-long.sh
```diff
@@ -0,0 +1,11 @@
+NIX_DEBUG_COMMAND="" && [[ $ACTIONS_RUNNER_DEBUG = "true" ]] && NIX_DEBUG_COMMAND='--print-build-logs --debug --show-trace --verbose'
+set -o pipefail -o errexit
+NIXPKGS_ALLOW_BROKEN=1 nix flake check --keep-going --no-build --allow-import-from-derivation  --no-update-lock-file --fallback -L ${NIX_DEBUG_COMMAND} --impure --option sandbox relaxed --impure 2>&1 | tee "nix.check.log"  || true
+set +o pipefail +o errexit
+echo "exited with(https://github.com/NixOS/nix/issues/7464) ${$?}" 
+cat "nix.check.log" | grep --invert-match  "error: path [']/nix/store/[a-zA-Z0-9]\+-[a-zA-Z0-9\.-]\+['] is not valid" \
+| grep --invert-match  "error: cannot substitute path [']/nix/store/[a-zA-Z0-9]\+-[a-zA-Z0-9\.-]\+['] \- no write access to the Nix store" \
+| grep --invert-match '^error: some errors were encountered during the evaluation' > "filtered.nix.check.log"
+RESULT=$(cat "filtered.nix.check.log" | grep -c 'error:')
```
obsessed-cake, line 9:
```suggestion
RESULT=$(grep -c 'error:' "filtered.nix.check.log")
```

.github/workflows/check-nix-long.sh
```diff
@@ -0,0 +1,11 @@
...
+echo "Got errors $RESULT"
```
obsessed-cake, line 10:
```suggestion
echo "Got $RESULT errors"
```

obsessed-cake (approved):

.github/workflows/check-nix-long.sh
dzmitry-lahoda: fixed in next PR

.github/workflows/check-nix-long.sh
dzmitry-lahoda: haha :) fixed in next PR

.github/workflows/build-test-deploy.yml
obsessed-cake: > 

You can remove `chmod +x ./.github/workflows/check-nix-long.sh`. `source` doesn't require an executable flag.

.github/workflows/check-nix-long.sh
dzmitry-lahoda: so it is like `buildin.any`

grep is not responsible for reading stream, just consume (F#/C#/Ocaml style). 

you may compose cat with anything, while direct grep repl is not composable.

dzmitry-lahoda: please comment. 

so change in next pr for zombienet. so IMHO better write piped script so these compose better.

.github/workflows/build-test-deploy.yml
dzmitry-lahoda: changed `Cache devnet all dev local` -> `Build all devnets` and `cache-devnet-all-dev-local` -> `build-all-devnets` in next PR.

.github/workflows/build-test-deploy.yml
dzmitry-lahoda: Assuming suggestion was to remove source. removed in next pr for zombient. thank you.

dzmitry-lahoda:
Fixed comments in next PR for zombienet.

dzmitry-lahoda:
@mergify refresh

dzmitry-lahoda:
@cor @hussein-aitlahcen anything to say https://github.com/ComposableFi/composable/pull/2724 ? I fixed existing comments in next PR.

cor (approved):
LGTM, will be good to add ARM to this later on

.github/workflows/check-nix-long.sh
obsessed-cake: > grep is not responsible for reading stream, just consume (F#/C#/Ocaml style).

hmm... but `cat` also consumes a source wholly.

> while direct grep repl is not composable.

really?

```
grep foo bar.txt | grep ERROR | grep -v timeout | ...
```

dzmitry-lahoda: your first grep should match all lines.  other then that sure can grep 2 times.

i did not know that cat is not streaming. i thought it is. 
 
also would perfer timeout to be separate command, so can timeout regardless of grep or not. like separate timeout worki with any stream.

PR #2726, opened 2022-12-15
fix simulate add/remove liquidity

benluelo:
Updates the simulate add/remove liquidity functionality to be inline with the latest dex refactor updates. Also changes the input to `remove_liquidity` to require `min_amounts` length to be between 1 and 2 with a `BiBoundedVec`.

### DIFF
--- a/code/parachain/frame/pablo/src/lib.rs
+++ b/code/parachain/frame/pablo/src/lib.rs
@@ -58,18 +58,19 @@ pub mod pallet {
 		WeightInfo,
 	};
 	use codec::FullCodec;
-	use composable_support::math::safe::{SafeArithmetic, SafeSub};
+	use composable_support::{
+		collections::vec::bounded::{bi_bounded_vec::BiBoundedVecOutOfBounds, BiBoundedVec},
+		math::safe::{SafeArithmetic, SafeSub},
+	};
 	use composable_traits::{
 		currency::{CurrencyFactory, LocalAssets},
 		defi::{CurrencyPair, Rate},
-		dex::{
-			Amm, BasicPoolInfo, Fee, PriceAggregate, RedeemableAssets,
-			RemoveLiquiditySimulationResult,
-		},
+		dex::{Amm, BasicPoolInfo, Fee, PriceAggregate, RedeemableAssets},
 	};
 	use core::fmt::Debug;
 	use frame_support::{
 		pallet_prelude::*,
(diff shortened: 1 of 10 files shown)


PoisonPhang (approved):
LGTM! Great work! :rocket:

code/parachain/frame/pablo/src/lib.rs
```diff
@@ -696,24 +669,17 @@ pub mod pallet {
...
+				// since we're simulating this, we want to immediately roll it back no matter what
+				// the outcome is
+				TransactionOutcome::Rollback(<Self as Amm>::add_liquidity(
+					who,
+					pool_id,
+					amounts,
+					Zero::zero(),
+					false,
+				))
+			})
```
hussein-aitlahcen, line 682: I'm not sure it's possible to mutate the storage in RPCs, I think that you can use the base API and it won't interfere with the ledger. Must be checked but I believe already discussed this with @KaiserKarel

hussein-aitlahcen (approved):

code/parachain/frame/pablo/src/lib.rs
KaiserKarel: afaik yeah, but now I wonder too.

KaiserKarel (approved):

code/parachain/frame/pablo/src/lib.rs
benluelo: it's not possible in RPCs, no, but the simulate_*_liquidity fns are exposed outside of rpcs. IMO they shouldn't exist at all, and we should just be calling the extrinsics directly in the rpc impl, but this is fine for now

benluelo: to clarify what i mean by exposed - they aren't public facing API, but they are part of the `Amm` trait; so they can technically be called from anywhere.

obsessed-cake (approved):

code/parachain/frame/pablo/src/lib.rs
hussein-aitlahcen: Makes sense, thanks for clarifying :)

vimukthi-git (approved):
LGTM, please test locally before merging

cor (changes requested):
not toally sure yet

benluelo:
closing in favor of #2740

PR #2743, opened 2022-12-16
Disable flaky FE tests

KaiserKarel:


KaiserKarel:
Task linked: [CU-8676y2c83 Disable flaky UI tests](https://app.clickup.com/t/8676y2c83)

dzmitry-lahoda:
@KaiserKarel may be better https://jestjs.io/docs/cli#--testpathignorepatternsregexarray and no modification of target code a lot? 

Via CLI can easy wire #flaky or `flaky` in CI to try run (by using cli params).  With config file modification approach will need to patch file. Seem CLI easier for temporal off changes.

KaiserKarel:
> @KaiserKarel may be better https://jestjs.io/docs/cli#--testpathignorepatternsregexarray and no modification of target code a lot?
> 
> Via CLI can easy wire #flaky or `flaky` in CI to try run (by using cli params). With config file modification approach will need to patch file. Seem CLI easier for temporal off changes.
> 
> CI """owner""" should be able to disable without interfering a lot with coders imho.

I can move the ignore to CI too; but would prefer to enable strict testing of FE asap, because right now more bad tests are getting into the codebase.

dzmitry-lahoda (changes requested):
please add tests to effects gate. because it is very annoying to see failed red CI jobs and wait until all jobs need to retry one failed. so at least there will be urge to fix tests and we can ask for that. 

because also you disabled some flaky, but who knows if all. so we would be in same situation after pr merged, unless it is effects gate. 

so either we disable and allow to enable using label/tag or if enabled by default - put into gate. 
no gated red fails are 100% ignored.

KaiserKarel:
> please add tests to effects gate. because it is very annoying to see failed red CI jobs and wait until all jobs need to retry one failed. so at least there will be urge to fix tests and we can ask for that.
> 
> because also you disabled some flaky, but who knows if all. so we would be in same situation after pr merged, unless it is effects gate.
> 
> so either we disable and allow to enable using label/tag or if enabled by default - put into gate. no gated red fails are 100% ignored.

👍

scglenn (approved):

obsessed-cake (approved):

dzmitry-lahoda (approved):

PR #2744, opened 2022-12-16
Remove setup folder

KaiserKarel:


dzmitry-lahoda (approved):
So we do not use our oracle anymore as I got it.

KaiserKarel:
> So we do not use our oracle anymore as I got it.

No, but this is an unused script for interacting with the oracle right?

dzmitry-lahoda:
i do not know.

aeryz (approved):

obsessed-cake (approved):

KaiserKarel:
@mergifyio refresh

PR #2746, opened 2022-12-16
#draft #ibc #cosmwasm correct timeout  #CU-3uft9wy

dzmitry-lahoda:
- time contract returns as i see on gh codebases and docs are from this chain
- ibc expects as i see by specs, whitepaper and hyperspace  that chan

so this pr does best effort convert this to that

also seems ics20 module in centauri has bug and may lead to wrongly timeouts (chains start at same time - so timeouts are not so visible and tests are flaky)


KaiserKarel:
Task linked: [XCVM-94 fix what is lacking](https://app.clickup.com/t/20465559/XCVM-94)

PR #2758, opened 2022-12-19
NX-44 - fix: avoid infinite recursion with flake-parts

cor:


KaiserKarel:
Task linked: [NX-44 Update nix flake deps](https://app.clickup.com/t/20465559/NX-44)

flake.nix
```diff
@@ -25,9 +25,9 @@
     };
   };
 
-  outputs = { self, nixpkgs, flake-parts, ... }:
+  outputs = inputs@{ self, nixpkgs, flake-parts, ... }:
```
blasrodri, line 28: Can you add a comment referencing the issue? Might be useful for future nix people

obsessed-cake (approved):

flake.nix
cor: you automatically get a message in your terminal when you run `nix flake show` when you do it the old way, it's an api change for [flake-parts](hercules-ci/flake-parts) and all new flake parts projects do it correclty by default, so I don't see this being neccesary

KaiserKarel (approved):

PR #2760, opened 2022-12-20
review for pablo

blasrodri:


ComposableFiB0t:
Task linked: [CU-865bbb5y0 Pallet Review since last audit](https://app.clickup.com/t/865bbb5y0)

code/parachain/frame/pablo/src/lib.rs
```diff
@@ -490,6 +490,7 @@ pub mod pallet {
 		/// Note this function does not validate,
 		/// 1. if the pool is created by a valid origin.
 		/// 2. if a pool exists with the same pair already.
+		// blas: can't this (2) mean that someone can override the owner of an existing pool?
```
vimukthi-git, line 493: Owner may not be overridden as per current logic. But someone can create a pool with same assets that has them as the owner. But this function can only be called by trusted code, so the possibility of error is limited at this phase.

blasrodri: Thanks

PR #2761, opened 2022-12-20
Integration Tests: Pablo rework

Dom-Roth:


code/integration-tests/runtime-tests/test/tests/pablo/constantProductDex/constantProductTests.ts
```diff
@@ -1,285 +1,1888 @@
...
 
   before("Minting assets", async function () {
-    await mintAssetsToWallet(api, walletId1, sudoKey, [1, baseAssetId, quoteAssetId, baseAsset2, falseQuoteAsset]);
-    await mintAssetsToWallet(api, walletId2, sudoKey, [1, baseAssetId, quoteAssetId, baseAsset2, falseQuoteAsset]);
-    await mintAssetsToWallet(api, walletId3, sudoKey, [1]);
+    // ToDo: Enable following when pool creation works!
+    // await mintAssetsToWallet(api, poolOwnerWallet, sudoKey, [1]);
+    await mintAssetsToWallet(
+      api,
+      walletLpProvider1,
```
kollegian, line 187: An idea: we can pass the wallets as array and send this minting assets step in one go.

code/integration-tests/runtime-tests/test/tests/pablo/constantProductDex/constantProductTests.ts
```diff
@@ -1,285 +1,1888 @@
...
+            assetId: 4,
+            amount: baseAssetFundsCurrentlyInPoolsBeforeTx.free
+          },
+          false
+        )
+      ).catch(exc => exc);
+      expect(exc.toString()).to.contain("Currently buggy! Throws wrong error message.");
+    });
+
+    it("#4.8  I can not buy with the base asset being the same as the quote asset.", async function () {
```
kollegian, line 908: Here is our expectancy is this tx to fail right? But it seems it worked.

code/integration-tests/runtime-tests/test/tests/pablo/constantProductDex/constantProductTests.ts
```diff
@@ -1,285 +1,1888 @@
...
+        api.events.pablo.Swapped.is,
+        api.tx.pablo.swap(1337, { assetId: 131, amount: 1_000 }, { assetId: 1, amount: 10_000 }, false)
+      ).catch(exc => exc);
+      expect(err2.toString()).to.contain("pablo.PoolNotFound");
+    });
 
     it(
-      "Given that the pool has liquidity and the users have LPTokens, " +
-        "users can remove liquidity from the pool by using LP Tokens",
+      "#4.12 I can buy or swap with the minimum amount requested greater than the trade would give, " +
```
kollegian, line 1001: Wording seems a bit off, as we are expecting this tx to fail right?

code/integration-tests/runtime-tests/test/tests/pablo/constantProductDex/constantProductTests.ts
```diff
@@ -1,285 +1,1888 @@
...
+            false
+          )
+        ).catch(exc => exc);
+        expect(exc.toString()).to.contain("pablo.CannotRespectMinimumRequested");
       }
     );
 
-    it("Given that LPTokens are transferred to another user, other user can removeLiquidity", async function () {
-      const result = await removeLiquidityFromPool(
+    it("#4.13 I can buy a huge amount with very high slippage.", async function () {
```
kollegian, line 1019: Just to have a standard outlook throughout the tests, a simple assertion would be goo here as well.

code/integration-tests/runtime-tests/test/tests/pablo/constantProductDex/constantProductTests.ts
```diff
@@ -1,285 +1,1888 @@
...
+      const exc = await sendAndWaitForSuccess(
+        api,
+        walletLpProvider1,
+        api.events.pablo.LiquidityRemoved.is,
+        api.tx.pablo.removeLiquidity(hardCodedPool1.poolId, amountToRemove, assets)
+      ).catch(exc => exc);
+      expect(exc.toString()).to.contain("pablo.CannotRespectMinimumRequested");
+    });
+
+    it("#3.2  I can not remove liquidity amounts of 0.", async function () {
```
kollegian, line 1107: So shall we raise this as a bug as it allows people to remove liq with amounts of 0. Seems not of a big deal but still something to consider I would say.

code/integration-tests/runtime-tests/test/tests/pablo/constantProductDex/constantProductTests.ts
```diff
@@ -1,285 +1,1888 @@
...
+        "if there's been trades in the meantime, both wallets can withdraw their liquidity," +
+        "and will receive equal amounts of funds back.",
+      async function () {
+        this.timeout(5 * 60 * 1000);
+        /*
+          1. Provide liquidity
+          2. Move 50% of liquidity to another wallet.
+          3. Trade
+          4. Remove liquidity of first wallet (expect fees)
+          5. Remove liquidity of second wallet (expect no fees)
```
kollegian, line 1328: Shouldn't we expect fees for all lp holders, regardless of how they receive the lp tokens.

code/integration-tests/runtime-tests/test/tests/pablo/constantProductDex/constantProductTests.ts
```diff
@@ -1,285 +1,1888 @@
...
+          new BN(amountEarnedFeesQuoteAssetWallet1.toString()).add(new BN(priceFirstTrade.toFixed(0))),
+          new BN(1)
+        );
+      }
+    );
+
+    it(
+      "#3.8 After adding funds, and after a trade, transferring part of them to another wallet " +
+        "the initial wallet can withdraw the provided liquidity with received fees, " +
+        "while the second wallet won't receive fees",
```
kollegian, line 1480: The same goes for this one as well. ı don't think we are separately holding a struct for fee distribution. I couldn't fully understand the assertions but I would say if we try to match the amounts received they would be the same for both wallets.

code/integration-tests/runtime-tests/test/tests/pablo/constantProductDex/constantProductTests.ts
```diff
@@ -1,285 +1,1888 @@
...
+          new BN(amountEarnedFeesQuoteAssetWallet1.toString()).add(new BN(priceFirstTrade.toFixed(0))),
+          new BN(1)
         );
       }
     );
 
     it(
-      "Given that the users have sufficient balance," +
-        " users can't swap illegal token pairs(Non existing assetId in the pool)",
+      "#3.9 After adding funds, and after a trade, another wallet can add liquidity, withdraw it again without " +
```
kollegian, line 1627: Perhaps a wording issue, but I think in this case second lp provider will receive a slightly less lp token based on the amounts in the pool, so the only difference would be that I guess.

kollegian (approved):
Amazing work.

LyonSsS (approved):
This can be merged

code/integration-tests/runtime-tests/test/tests/pablo/constantProductDex/constantProductTests.ts
Dom-Roth: Yes but we'd either need to update this method incl. all tests that call this method.
Or make the method more complex by differentiating the input.
Or create a second method that can handle multiple wallets.

And I'm not sure what would be the best option here.

code/integration-tests/runtime-tests/test/tests/pablo/constantProductDex/constantProductTests.ts
Dom-Roth: Yes it works but a user looses the trading fee. To me this isn't a bug since it'd require more on-chain calculations to avoid which could increase the tx fee.
The wording here is definitely outdated, thanks for pointing out. Will update accordingly.

code/integration-tests/runtime-tests/test/tests/pablo/constantProductDex/constantProductTests.ts
Dom-Roth: You're correct. Will need to re- check and might need to bring this up for the engineering team. 
Ty sir 🙏.

code/integration-tests/runtime-tests/test/tests/pablo/constantProductDex/constantProductTests.ts
Dom-Roth: Thanks for seeing this. I forgot to finalize this test. 👀

code/integration-tests/runtime-tests/test/tests/pablo/constantProductDex/constantProductTests.ts
Dom-Roth: Don't think so. It will remove asset amounts of `0` anyway. And a user would loose the tx fee.

code/integration-tests/runtime-tests/test/tests/pablo/constantProductDex/constantProductTests.ts
Dom-Roth: Yes correct, the comment here is wrong. Thanks for pointing out.

code/integration-tests/runtime-tests/test/tests/pablo/constantProductDex/constantProductTests.ts
Dom-Roth: True and the assertions are actually checking this. Need to fix test title & comment.
Thanks sir.

code/integration-tests/runtime-tests/test/tests/pablo/constantProductDex/constantProductTests.ts
Dom-Roth: Good point, I'll update the test title be more precise.

Martin-Kiss (approved):

PR #2782, opened 2022-12-23
Fix release flow build

KaiserKarel:


vimukthi-git (approved):

.github/workflows/release-artifacts.yml
```diff
@@ -24,20 +24,12 @@ jobs:
...
     steps:
-      - run: |
-          apt-get update
-          apt-get install -y curl git
-          curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | bash
-          apt-get install git-lfs
       - uses: actions/checkout@v3
-        with:
-          lfs: true
-          fetch-depth: 0
```
vimukthi-git, line 40: This line may need to be kept as per original

KaiserKarel: I'm 99% sure not, but I can re-add the fetch-depth:0

vimukthi-git: Yeah was checking the docs, just trying to be safe

PR #2806, opened 2022-12-27
[Pbl-123] fix privacy policy and terms of use

fl-y:
https://app.clickup.com/t/20465559/PBL-123


andyjsbell (approved):
thanks @fl-y

easteregg (approved):

easteregg:
It would be better to use native MUI components to respect the typography of the app. But I think we can refactor this in the future.

PR #2809, opened 2022-12-27
[CU-86770765z] Add daily pool stats to Subsquid

andresmechali:


andyjsbell (approved):
thanks @andresmechali

vimukthi-git (approved):

PR #2829, opened 2022-12-29
[PBL-162, PBL-163, PBL-151] Subsquid - Remove Oracle and do not store historical prices in USD

andresmechali:
https://app.clickup.com/t/20465559/PBL-162
https://app.clickup.com/t/20465559/PBL-163
https://app.clickup.com/t/20465559/PBL-151


subsquid/src/processor.ts
```diff
@@ -105,14 +100,14 @@ processor.addEventHandler("Vesting.Claimed", processVestingClaimedEvent);
...
-);
+// processor.addEventHandler(
+//   "AssetsRegistry.AssetRegistered",
+//   processAssetRegisteredEvent
+// );
+//
+// processor.addEventHandler(
+//   "AssetsRegistry.AssetUpdated",
+//   processAssetUpdatedEvent
+// );
```
PoisonPhang, line 111: NIT: Remove commented out code

subsquid/src/server-extension/resolvers/pabloOverviewStats.ts
```diff
@@ -35,27 +44,34 @@ export class PabloOverviewStatsResolver
...
+    const lockedValue = await manager.find(HistoricalLockedValue, {
+      select: ["amount", "assetId"],
+      where: {
+        source: LockedSource.Pablo,
+      },
+    });
+
+    const totalValueLocked = lockedValue.reduce<Record<string, bigint>>(
+      (acc, value) => {
+        acc[value.assetId] = (acc[value.assetId] || 0n) + value.amount;
```
PoisonPhang, line 59: Should this instead be?
```suggestion
        acc[value.assetId] = (acc[value.assetId] ?? 0n) + value.amount;
```

PoisonPhang (approved):

subsquid/src/server-extension/resolvers/pabloOverviewStats.ts
andresmechali: It's the same in this case, as it is defaulting to 0 anyways. If this was `|| 1n`, it would matter as `0n || 1n = 1n`, but `0n ?? 1n = 0n`. Well noticed detail though 👍🏼

subsquid/src/model/generated/pabloPool.model.ts
```diff
@@ -47,4 +47,16 @@ export class PabloPool {
...
+     * Last updated block id
+     */
+    @Column_("text", {nullable: false})
+    blockId!: string
+
+    /**
+     * Asset to be used as reference when calculating the pool's TVL
+     */
+    @Column_("text", {nullable: false})
+    baseAssetId!: string
```
vimukthi-git, line 61:
```suggestion
    quoteAssetId!: string
```

vimukthi-git (approved):

subsquid/src/processors/pablo.ts
```diff
@@ -410,12 +441,38 @@ export async function processSwappedEvent(
...
   const weightRatio = baseAssetWeight.weight / quoteAssetWeight.weight;
 
+  const spotPrice = divideBigInts(quoteAmount, baseAmount) * weightRatio;
+
+  const feeSpotPrice = BigNumber(
+    fee.assetId.toString() === pool.baseAssetId ? 1 : spotPrice
+  );
+
+  const pabloFee = new PabloFee({
+    id: randomUUID(),
```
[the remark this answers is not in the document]
andresmechali: No reason. Just that as I see it, the ID is supposed to be a unique identifier and the event ID is neither unique (if as some point we want to create multiple `PabloFee` on a single event, we can't), nor is related to fees object

subsquid/src/model/generated/pabloPool.model.ts
andresmechali: This represents the asset to be used when normalizing values for a given pool. Naming it `quote` would be more accurate?

vimukthi-git: yes, `quote` is the correct naming for an asset when it is used for quoting amounts

subsquid/src/processors/pablo.ts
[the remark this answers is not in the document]
vimukthi-git: I don't see a reason to have multiple PabloFee objects for a single event in product, requirements so we can use that. Generally creating randomIDs should be limited to where it is a must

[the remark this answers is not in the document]
andresmechali: Alright, no problem, I'll change that

PR #2835, opened 2022-12-29
PBL-157: Discover PICA price off-of USDT pool

easteregg:


andyjsbell (approved):

vimukthi-git (approved):
LGTM, may best to indicate the source of the price so that there is not complaint of manipulation

PR #2846, opened 2022-12-31
feat: add working meta tags for images

fl-y:
Not the securest method to store images but works.


fl-y:
Closing since doesn't seem to be priority

PR #2847, opened 2023-01-02
PBL-171: Fix NaN in RemoveLiquidity page

easteregg:
Update return type for SimulateRemoveRPC

### DIFF
--- a/frontend/packages/defi-interfaces/pablo/definitions.ts
+++ b/frontend/packages/defi-interfaces/pablo/definitions.ts
@@ -103,8 +103,6 @@ export default {
       assetId: "u128",
     },
     ComposableTraitsDexStakingRewardPool: "Null",
-    RemoveLiquiditySimulationResult: {
-      assets: "BTreeMap<AssetId, Balance>",
-    },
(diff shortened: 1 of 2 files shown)


PoisonPhang (approved):

frontend/packages/defi-interfaces/pablo/definitions.ts
```diff
@@ -103,8 +103,6 @@ export default {
       assetId: "u128",
     },
     ComposableTraitsDexStakingRewardPool: "Null",
-    RemoveLiquiditySimulationResult: {
-      assets: "BTreeMap<AssetId, Balance>",
-    },
+    RemoveLiquiditySimulationResult: "BTreeMap<AssetId, Balance>",
```
benluelo, line 106: There is no longer a `RemoveLiquiditySimulationResult` type, the rpc just returns the BTreeMap directly; not sure if this typedef is required still

benluelo (approved):

frontend/packages/defi-interfaces/pablo/definitions.ts
easteregg: This is just an alias now.

PR #2849, opened 2023-01-02
Use `SS58` addresses in CosmWasm contracts

aeryz:
- `AccountId` can be generated from both hex and ss58 representation.
- `ss58` addresses are used as string representation export.


code/parachain/frame/cosmwasm/src/lib.rs
```diff
@@ -298,7 +298,8 @@ pub mod pallet {
 
 		/// A way to convert from our native account to cosmwasm `Addr`.
 		type AccountToAddr: Convert<AccountIdOf<Self>, String>
-			+ Convert<String, Result<AccountIdOf<Self>, ()>>;
+			+ Convert<String, Result<AccountIdOf<Self>, ()>>
+			+ Convert<Vec<u8>, Result<AccountIdOf<Self>, ()>>;
```
KaiserKarel, line 303: If something implements Convert<String>, doesn't it always implement Convert<Vec<u8>>, since String is a Vec<u8>?

aeryz: But the thing is string representation and the binary representation is different in this case:
eg. String representation of an address can be both `SS58` address or hex address. But the Vec is the canonical binary representation which means `String::as_bytes() != Vec`. So the implementor should provide how we can get an account id from the canonical binary representation.
But I think providing some comments is useful for people to see the rationale behind this, because you normally think "oh isn't string a vec at the end". Good catch.

code/parachain/runtime/dali/src/lib.rs
```diff
@@ -1085,21 +1086,29 @@ parameter_types! {
...
-/// TODO(hussein-aitlahcen): Probably nicer to have SS58 representation here.
 pub struct AccountToAddr;
 impl Convert<alloc::string::String, Result<AccountId, ()>> for AccountToAddr {
 	fn convert(a: alloc::string::String) -> Result<AccountId, ()> {
-		match a.strip_prefix("0x") {
-			Some(account_id) => Ok(<[u8; 32]>::try_from(hex::decode(account_id).map_err(|_| ())?)
-				.map_err(|_| ())?
-				.into()),
-			_ => Err(()),
+		if let Some(hex_account) = a.strip_prefix("0x") {
```
hussein-aitlahcen, line 1092: Do we really want to allow both hex and SS58?

aeryz: Actually, I don't see a use case where users/developers want to use hex over SS58. I provided that feature just in case but let me remove it and if we ever see a need or use case, then we add.

hussein-aitlahcen: Great!

hussein-aitlahcen (approved):

PR #2851, opened 2023-01-02
Ben/rfc 0015 pablo internals restructure

benluelo:


PoisonPhang (commented):
Goal 1's implementation is solid and clear. I think it would offer many benefits to our pallet development workflow.

Sadly, goal 1 is not truly complete until goal 2 is also complete. Given the proposed solutions for goal 2, I think the first solution is a clear winner - in most situations. 

If there is ever a case where we need to conditionally read/write storage at scale, solution one is quite the performance foot-gun. Instead of using the second solution in these cases, we can instead allow for direct storage read/write in these limited cases. I don't think developing hooks will be worth the effort here over a more traditional storage access.

In general, we should be able to design with solution one in mind to keep our functions pure.

Adding these to our pallet design philosophy will greatly improve the maintainability of our pallets. Good RFC!

benluelo:
> If there is ever a case where we need to conditionally read/write storage at scale, solution one is quite the performance foot-gun.

If there is an instance where there is a significant amount of conditional reads/writes, it's likely that the function in question is doing too much and should be broken down into smaller, more focused functions; I'll mention this in the relevant section.

I also agree that hooks are most likely not worth the time, but I thought I would mention it as an alternative either way. I think that realistically, explicit inputs and outputs is the best option, as it has much less boilerplate and indirection.

rfcs/0015-pablo-internals-restructure.md
```diff
@@ -0,0 +1,321 @@
...
+The [current implementation of `DualAssetConstantProduct`][dacp] is tightly
+coupled with the pablo pallet, depending on pablo's `Config` trait directly for
+all of it's functionality and directly modifying pablo's storage. This makes it
+very difficult to test, since a full runtime is required to provide the required
+mocks for it to work properly.
+
+## Goals
+
+1. Remove `DualAssetConstantProduct`'s direct dependency on pablo
+2. Make all of the `DualAssetConstantProduct` functionality pure
```
vimukthi-git, line 12: We don't have a requirement to make things "pure" in our code by it self, unless it is to handle an actual product/reuse requirement

rfcs/0015-pablo-internals-restructure.md
```diff
@@ -0,0 +1,321 @@
...
+
+The [current implementation of `DualAssetConstantProduct`][dacp] is tightly
+coupled with the pablo pallet, depending on pablo's `Config` trait directly for
+all of it's functionality and directly modifying pablo's storage. This makes it
+very difficult to test, since a full runtime is required to provide the required
+mocks for it to work properly.
+
+## Goals
+
+1. Remove `DualAssetConstantProduct`'s direct dependency on pablo
```
vimukthi-git, line 11: These are not goals, they are a solution to a problem. Problem should come first before the solution.
refer https://en.wikipedia.org/wiki/Circular_reasoning

vimukthi-git (changes requested):

rfcs/0015-pablo-internals-restructure.md
benluelo: Is the opening statement insufficient as a description of the problem? I can elaborate if so

PR #2877, opened 2023-01-06
Added RFC 0013 - Refactor assets-registry, assets and remove hard coded assets and currency-factory [INIT-13]

PoisonPhang:
[[INIT-13]](https://app.clickup.com/t/20465559/INIT-13)


rfcs/0013-redesign-assets-id-system.md
```diff
@@ -0,0 +1,136 @@
...
+
+* Support the traits from `frame_support::traits::tokens::fungibles`
+  * Specifically `MutateHold`
+
+* Ability to mint mintable assets and the inability to mint non-mintable assets
+
+* Support the existing local asset IDs we have hard-coded
+
+* Utilize both pallet-balances and pallet-assets from Parity for maintaining 
+balances within our consensus system
```
benluelo, line 41: I think we should add our own abstraction over the proposed assets manager to the requirements. pallets (such as pablo) may require information such as asset decimals, which AFAIK isn't exposed by any of parity's traits

rfcs/0013-redesign-assets-id-system.md
```diff
@@ -0,0 +1,246 @@
...
+
+## Local Asset ID Generation
+
+Each asset within our consensus system should have its own local asset ID. 
+Currently, we have hard-coded these local IDs rather than dynamically generating 
+them as needed. While existing assets should maintain their current ID, we can 
+define a new way to derive asset IDs automatically for assets created in the 
+future.
+
+A simple approach to this would be processing a nonce using a fixed-length
```
vimukthi-git, line 61: I think it's better to avoid global nonces for all protocols, we can use a namespace for each protocol that needs to create assets to avoid collisions

rfcs/0013-redesign-assets-id-system.md
```diff
@@ -0,0 +1,246 @@
...
+asset metadata (ticker-symbol, decimal precision, and ratio)
+
+* Support the traits from `frame_support::traits::tokens::fungibles`
+  * Specifically `MutateHold`
+
+* Ability to mint mintable assets and the inability to mint non-mintable assets
+
+* Support the existing local asset IDs we have hard-coded
+
+* Utilize both pallet-balances and pallet-assets from Parity for maintaining
```
vimukthi-git, line 40: This by it self is not a requirement, just that it's possibly the best approach given another constraint. Long terms compatibility between our chain and parity tools(Eg: UI) and conventions. I would avoid adding specific solutions in requirements.

rfcs/0013-redesign-assets-id-system.md
```diff
@@ -0,0 +1,246 @@
...
+
+When also considering pallet-balances for our native asset, it becomes clear 
+that some amount of assisted routing will be necessary. While routing between 
+pallet-assets and pallet-balances is trivial, routing between pallet-balances 
+and two instances of pallet-assets becomes more difficult.
+
+Ideally, this routing would not be needed, but both our local and foreign asset 
+will, in some instances, interact within the same scope (i.e. Pablo).
+
+Our current asset router (also called pallet-assets, to be renamed to assets
```
vimukthi-git, line 81: I think this routing possibly is not needed if we correctly follow a namespacing approach such as multi-asset to identify the pallet and/or instance the asset belongs to. Then the routing becomes just a library of code rather than a fully fledged pallet. Then the purpose of asset-registry would just be exposing any additional globabl meta data such as fees, symbol, ratios etc.

rfcs/0013-redesign-assets-id-system.md
```diff
@@ -0,0 +1,246 @@
...
+bottleneck would be much appreciated.
+
+One way to minimize this issue is to stick with orml-tokens but maintain 
+multiple instances. This would allow us to only provide a singe instance of 
+orml-tokens to the instances that need it without the context of the other 
+instance (scopes that only require local xor foreign assets).
+
+## Adapted Asset Registry
+
+Asset Registry must be updated with distinct paths for both local (mintable) and
```
vimukthi-git, line 101: The path mechanism is possibly not needed if we namespace assets

vimukthi-git (commented):
Looks neat. Few concerns along name spacing that we have discussed before.

rfcs/0013-redesign-assets-id-system.md
PoisonPhang: So, instead of the asset ID being `hash(global-nonce)`, we would have `hash(pablo-pablosNonce)`?

vimukthi-git: yeah that should work in theory. May be a good prototype is what we need to confirm

rfcs/0013-redesign-assets-id-system.md
vimukthi-git: Let me be more precise. Suggestion is not to remove the routing but to avoid the assets-registry in making the routing decisions. Given the multiaddress of the multiasset already carries the pallet instance information, whether it is mintable asset or not is a straightforward logic decision with static runtime information.

PoisonPhang: I see, that's more clear. Thank you.
I'll revise the asset ID generation section with this in mind and update the docs accordingly.

rfcs/0013-redesign-assets-id-system.md
```diff
@@ -0,0 +1,246 @@
...
+
+Instead of depending on these hard-coded asset IDs, we should instead develop
+a dynamic ID system that makes it easy for developers to both interact with 
+assets within our consensus system and to retrieve information relevant to a 
+given asset.
+
+## Requirements
+
+* There exists a double mapping between multi-locations and local asset IDs
+  * Zero or One multi-locations ↔ One asset ID
```
benluelo, line 28: can zero location map to an id?

rfcs/0013-redesign-assets-id-system.md
benluelo: We could have each pallet that wants to have it's own namespace registered with our assets pallet using their pallet id, and then have a `pallet id -> protocol nonce` storage. the question then is what to do with assets that aren't specific to one protocol

rfcs/0013-redesign-assets-id-system.md
PoisonPhang: This may have not been described clearly.
The idea is that each asset will have an local asset ID and will have up to one multi-location. A local asset, such as an LP token, may not have a multi-location.

rfcs/0013-redesign-assets-id-system.md
```diff
@@ -0,0 +1,246 @@
...
+* Support the traits from `frame_support::traits::tokens::fungibles`
+  * Specifically `MutateHold`
+
+* Ability to mint mintable assets and the inability to mint non-mintable assets
+
+* Support the existing local asset IDs we have hard-coded
+
+* Utilize both pallet-balances and pallet-assets from Parity for maintaining 
+balances within our consensus system
+
```
vimukthi-git, line 42: I think it would be nice to dig into xcm + byog(min-fee calculation, conversion) requirements and specify them here.

rfcs/0013-redesign-assets-id-system.md
benluelo: ah, so `id -> multilocation?` and `multilocation -> id`. makes sense :+1:

code/parachain/frame/assets-registry/src/lib.rs
```diff
@@ -187,6 +212,10 @@ pub mod pallet {
 		/// `ed` - same meaning as in for foreign asset account (if None, then asset is not
 		/// sufficient)
 		#[pallet::weight(<T as Config>::WeightInfo::register_asset())]
+		// TODO(RFC-0013): Update Assets Registry - Pallet Functions
+		// | Update `register_*_asset` routes
```
vimukthi-git, line 216: Local and foreign need not be separate, they both have a universal asset identifier in MultiAsset/MultiLocation. You register once with full path you can derive if they are local or not with code.

PoisonPhang: I was not originally planning to derive an XCM Multi-Location for each asset, only those who need it.
If I'm not mistaken, we'll need to decide on a way to derive locations for assets in our consensus system.

rfcs/0013-redesign-assets-id-system.md
PoisonPhang: Add trait definitions to assets registry

code/parachain/frame/assets-registry/src/lib.rs
```diff
@@ -72,27 +77,39 @@ pub mod pallet {
...
+
 		type CurrencyFactory: CurrencyFactory<AssetId = Self::LocalAssetId, Balance = Self::Balance>;
+
+		/// Maximum number of characters allowed in an asset symbol
+		#[pallet::constant]
+		type AssetSymbolMaxChars: Get<u32>;
+
+		/// Maximum number of characters allowed in an asset name
+		#[pallet::constant]
+		type AssetNameMaxChars: Get<u32>;
```
PoisonPhang, line 101: REVIEW POINT:
Two new config items

code/parachain/frame/assets-registry/src/lib.rs
```diff
@@ -117,6 +134,17 @@ pub mod pallet {
...
+	#[pallet::storage]
+	#[pallet::getter(fn decimals)]
+	pub type Metadata<T: Config> = StorageMap<
+		_,
+		Twox128,
+		T::LocalAssetId,
+		AssetMetadata<BoundedVec<u8, T::AssetNameMaxChars>, BoundedVec<u8, T::AssetSymbolMaxChars>>,
+		OptionQuery,
+	>;
+
```
PoisonPhang, line 147: REVIEW POINT:
Removal of the `ForeignMetadata` storage in favor of a `Metadata` type used by both local and foreign assets

code/parachain/frame/assets-registry/src/lib.rs
```diff
@@ -182,45 +221,103 @@ pub mod pallet {
...
-			location: T::ForeignAssetId,
-			ratio: Rational,
-			decimals: Option<Exponent>,
-		) -> DispatchResultWithPostInfo {
+			local_or_foreign: LocalOrForeignAssetId<T::LocalAssetId, T::ForeignAssetId>,
+			ratio: Option<Rational>,
+			name: Vec<u8>,
+			symbol: Vec<u8>,
+			decimals: Exponent,
+		) -> DispatchResult {
```
PoisonPhang, line 240: REVIEW POINT:
Updated interface supporting both local and foreign assets.

code/parachain/frame/assets-registry/src/lib.rs
```diff
@@ -182,45 +221,103 @@ pub mod pallet {
...
-		/// Given well existing asset, update its remote information.
-		/// Use with caution as it allow reroute assets location.
-		/// See `register_asset` for parameters meaning.
+		/// Update the location of a foreign asset.
+		///
+		/// Emmits:
+		/// * `AssetLocationUpdated`
 		#[pallet::weight(<T as Config>::WeightInfo::update_asset())]
-		pub fn update_asset(
+		pub fn update_asset_location(
```
PoisonPhang, line 271: REVIEW POINT:
Using less monolithic `update_assets` functions

code/parachain/frame/assets-registry/src/lib.rs
```diff
@@ -253,30 +350,52 @@ pub mod pallet {
...
 		type Balance = T::Balance;
 
+		fn register_asset(
+			asset_id: Self::AssetId,
+			location: Option<Self::AssetNativeLocation>,
+			ratio: Option<Rational>,
+			name: Vec<u8>,
+			symbol: Vec<u8>,
+			decimals: u8,
+		) -> DispatchResult {
```
PoisonPhang, line 363: REVIEW POINT:
New trait function allowing for easier asset registration from other protocols.

code/parachain/frame/assets-transactor-router/src/lib.rs
```diff
@@ -0,0 +1,918 @@
+//! # Assets Pallet
+//!
+//! The Assets pallet provides implementation of common currency traits
+//! (e.g. from [`orml`](https://docs.rs/orml-traits) or `frame_support`)
+//! and functionality for handling transfers and minting.
+//!
+//! - [`Config`]
+//! - [`Call`]
+//! - [`Pallet`]
+//!
```
PoisonPhang, line 10: All rust docs need updating

code/parachain/frame/assets-transactor-router/src/lib.rs
```diff
@@ -0,0 +1,918 @@
...
+mod orml;
+
+// #[cfg(test)]
+// mod mocks;
+
+// #[cfg(test)]
+// mod tests;
+
+// #[cfg(any(feature = "runtime-benchmarks", test))]
+// mod benchmarking;
```
PoisonPhang, line 84: Need to re-introduce these

code/parachain/frame/assets-transactor-router/src/lib.rs
```diff
@@ -0,0 +1,918 @@
...
+				ratio,
+				name,
+				symbol,
+				decimals,
+			)?;
+
+			Ok(asset_id)
+		}
+	}
+}
```
PoisonPhang, line 524: REVIEW POINT:
New asset creation interface

PoisonPhang (commented):
Highlighted more notable parts of these changes that deserve a focus during reviews.
Also noted some lingering To-Dos

Changes since last meeting:
* `MultiLocation` cannot be used directly as the asset ID by any of Parity's asset interfaces, it does not implement `Copy`
  * Instead, each asset must have its own asset ID, see RFC
* Parity's pallet-assets does not implement *any* hold/lock functionality. Adding this via a wrapper is more risk than its worth.
  * The main thing we were missing from this pallet were asset creation traits and metadata traits
  * I've added the metadata traits here - more research is needed to be sure this will enable the assets dashboard

code/parachain/frame/assets-transactor-router/src/lib.rs
```diff
@@ -0,0 +1,918 @@
...
+		/// origin of admin of this pallet
+		type AdminOrigin: EnsureOrigin<Self::Origin>;
+
+		// will be assets-registry
+		type AssetLookup: AssetTypeInspect<AssetId = Self::AssetId>
+			+ RemoteAssetRegistryMutate<
+				AssetId = Self::AssetId,
+				AssetNativeLocation = Self::AssetLocation,
+			> + InspectRegistryMetadata<AssetId = Self::AssetId>
+			+ MutateRegistryMetadata<AssetId = Self::AssetId>;
```
PoisonPhang, line 218: Rename to `AssetsRegistry`

code/parachain/frame/composable-traits/src/assets.rs
```diff
@@ -34,3 +35,105 @@ pub struct Asset<Balance, ForeignId> {
...
+	) -> DispatchResult;
+}
+
+/// Structure to represent basic asset metadata such as: name, symbol, decimals.
+#[derive(Clone, Encode, Decode, Eq, PartialEq, Default, RuntimeDebug, MaxEncodedLen, TypeInfo)]
+pub struct AssetMetadata<BoundedName, BoundedSymbol> {
+	/// Name of the asset.
+	pub name: BoundedName,
+	/// Symbol of the asset.
+	pub symbol: BoundedSymbol,
```
benluelo, line 92: These could just be bounded vecs directly with the generic parameter being a `Get` type

PoisonPhang: So, `(AssetId) -> Vec<???>`

benluelo: `BoundedVec<u8, MaxNameLen>` where `MaxNameLen: Get<u32>`

code/parachain/frame/assets-transactor-router/src/lib.rs
```diff
@@ -0,0 +1,918 @@
...
+		}
+
+		fn symbol(asset: &Self::AssetId) -> Vec<u8> {
+			<Self as fungibles::metadata::Inspect<T::AccountId>>::symbol(*asset)
+		}
+
+		fn decimals(asset: &Self::AssetId) -> u8 {
+			<Self as fungibles::metadata::Inspect<T::AccountId>>::decimals(*asset)
+		}
+	}
```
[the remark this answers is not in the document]
vimukthi-git: It would be awesome if this works

rfcs/0013-redesign-assets-id-system.md
```diff
@@ -0,0 +1,255 @@
...
+# Other Solutions
+
+* [Acala](./0013/acala-analysis.md)
+* [Moonbeam](./0013/moonbeam-analysis.md)
+* [Parallel](./0013/parallel-finance-analysis.md)
+
+<!-- TODO: These are mostly done, but I will introduce them to the repo in 
+their own files -->
+
+# Design
```
vimukthi-git, line 51: Would be nice to have some overview diagram of how the assets system components fit together

code/parachain/frame/composable-traits/src/assets.rs
```diff
@@ -34,3 +35,105 @@ pub struct Asset<Balance, ForeignId> {
...
+	/// The number of decimals this asset uses to represent one unit.
+	pub decimals: u8,
```
vimukthi-git, line 94: Probably a future extension is to allow more custom fields with a map or so

code/parachain/frame/assets-transactor-router/src/lib.rs
```diff
@@ -0,0 +1,918 @@
...
+		}
+
+		fn create_foreign_asset(
+			foreign_asset_id: Self::ForeignAssetId,
+			name: Vec<u8>,
+			symbol: Vec<u8>,
+			decimals: u8,
+			ratio: Option<Rational64>,
+		) -> Result<Self::LocalAssetId, DispatchError> {
+			let asset_id = Self::LocalAssetId::from(u128::from_be_bytes(sp_core::blake2_128(
```
vimukthi-git, line 508: Good to have a comment on the implementation of unique id how it's supposed to work with multiasset/multilocation. Specially since this looks non-unique without detail

vimukthi-git (commented):
LGTM

code/parachain/frame/assets-transactor-router/src/lib.rs
```diff
@@ -0,0 +1,918 @@
...
+			+ Clone
+			+ Default
+			+ TypeInfo
+			+ MaxEncodedLen;
+		type Balance: BalanceLike;
+
+		#[pallet::constant]
+		type NativeAssetId: Get<Self::AssetId>;
+
+		type ForeignTransactor: fungibles::Inspect<Self::AccountId, Balance = Self::Balance, AssetId = Self::AssetId>
```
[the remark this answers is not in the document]
PoisonPhang: Perhaps `FungiblesTransactor` and `MultiCurrencyTransactor` to be explicit. That could be helpful though.
+1

PoisonPhang (commented):
Notes from QA/FE/Integrations meeting

PoisonPhang:
Closed in favor of #2954

rfcs/0013-redesign-assets-id-system.md
```diff
@@ -0,0 +1,255 @@
...
+To create a local asset ID, we can form a full asset ID from the combination of
+the assets' namespace (pallet or protocol) and a nonce/key provided by that 
+namespace.
+
+## Foreign Asset ID Generation
+
+Foreign assets should always be created with a provided, and unique, relative 
+multi-location. A corresponding local asset ID that implements the `Copy` trait 
+is still required in addition to the multi-location. To ensure we have both of 
+theses, we can hash the provided multi-location to create a local asset ID.
```
dzmitry-lahoda, line 73: some protocols do hash, some not. where we ended up?

rfcs/0013-redesign-assets-id-system.md
```diff
@@ -0,0 +1,255 @@
...
+## Local Asset ID Generation
+
+Each asset within our consensus system should have its own local asset ID. 
+Currently, we have hard-coded these local IDs rather than dynamically generating 
+them as needed. While existing assets should maintain their current ID, we can 
+define a new way to derive asset IDs automatically for assets created in the 
+future.
+
+To create a local asset ID, we can form a full asset ID from the combination of
+the assets' namespace (pallet or protocol) and a nonce/key provided by that
```
dzmitry-lahoda, line 65: what way we ended up. how nonce/key is created/generated?

rfcs/0013-redesign-assets-id-system.md
```diff
@@ -0,0 +1,255 @@
...
+
+Given the scope of assets, the design will be broken down into several more 
+consumable sections.
+
+## Local Asset ID Generation
+
+Each asset within our consensus system should have its own local asset ID. 
+Currently, we have hard-coded these local IDs rather than dynamically generating 
+them as needed. While existing assets should maintain their current ID, we can 
+define a new way to derive asset IDs automatically for assets created in the
```
dzmitry-lahoda, line 61: what is decided derivation for asset ids?

rfcs/0013-redesign-assets-id-system.md
dzmitry-lahoda: i see. it defined later.

rfcs/0013-redesign-assets-id-system.md
dzmitry-lahoda: i see. it defined later.

rfcs/0013-redesign-assets-id-system.md
dzmitry-lahoda: i see. it defined later.

rfcs/0013-redesign-assets-id-system.md
```diff
@@ -0,0 +1,255 @@
...
+be the nonce provided by the source.
+
+To create a hash from the nonce, `sp_core::hashing::blake2_64` can be used
+
+**Steps**:
+
+* Provide assets registry with either our previously described local asset ID
+or a multi-location
+
+* If assets registry receives a multi-location, hash the multilocation to create
```
dzmitry-lahoda, line 196: IBC location of incoming assets from Composable to Picasso would be `transferPortOnComposable/channelOnComposable/<hash of XcmLocation(1, Here)>` That will be value provided by IBC to `transactor`

dzmitry-lahoda: but then, what asset id will be on Picasso? would it be `<hash of transferPortOnComposable/channelOnComposable/<hash of XcmLocation(1, Here)>` or will try to make XcmLocation(StringIdentifierOfIbcClient, StringIdentifierOfIbcConnection, StringIdentifierOfPort, StringIdentifierOfChannel, ThenPortAgain, ThenChannel, then incoming Hash)`?

PR #2883, opened 2023-01-09
#cu-86698p7yq removing tests cache to avoid missed tests runs

dzmitry-lahoda:
- avoid issue when tests fail because of host updates
- will add back as pure nix build after nix upgrade (nix 22.11 can run cgroups (containers) in build)
- fail fast on timeout
- depends on https://github.com/ComposableFi/composable/pull/2882/checks


dzmitry-lahoda:
closed for rebased PR to fix LFS

PR #2905, opened 2023-01-11
fixed banner via #ipfs

dzmitry-lahoda:


PoisonPhang (approved):

cor:
IMO this is no longer needed because we fixed LFS billing (tho lfs is ofc very cool)

PR #2907, opened 2023-01-11
[PICA-71, PICA-77] Staking Rewards updates

benluelo:
(no longer as small as originally intended 😛)

- cleans up benchmarks
- cleans up reward accumulation code
- other small improvements

~~I've tried to keep the commits clean on this branch, it should be possible to review each commit individually.~~ no longer true due to the mess that was https://app.clickup.com/t/20465559/PICA-77

### DIFF
--- a/code/parachain/frame/composable-traits/src/staking/mod.rs
+++ b/code/parachain/frame/composable-traits/src/staking/mod.rs
@@ -54,6 +54,19 @@ impl<Balance> RewardRate<Balance> {
 		Self { period: RewardRatePeriod::PerSecond, amount: amount.into() }
 	}
 }
+impl<Balance> RewardRate<Balance>
+where
+	Balance: Copy + Into<u128> + From<u128>,
+{
+	/// The amount that will be rewarded in one period.
+	// TODO(benluelo): Encode somehow that one period will always be rewardable. We can most likely
+	// use `Validated` for this, although we may encounter some issues with orphan rules.
+	pub fn amount_per_period(&self) -> Option<Balance> {
+		u128::from(self.period.as_secs().get())
+			.checked_mul(self.amount.into())
+			.map(Into::into)
+	}
+}
 
 #[derive(RuntimeDebug, PartialEq, Eq, Clone, MaxEncodedLen, Encode, Decode, TypeInfo)]
 pub enum RewardRatePeriod {
--- a/code/parachain/frame/staking-rewards/src/lib.rs
+++ b/code/parachain/frame/staking-rewards/src/lib.rs
@@ -49,6 +49,8 @@ mod validation;
 pub mod prelude;
 pub mod weights;
 
+use core::num::{NonZeroU128, NonZeroU64};
+
 use crate::prelude::*;
 
 use composable_support::math::safe::{SafeDiv, SafeMul, SafeSub};
@@ -213,7 +215,7 @@ pub mod pallet {
 	pub enum RewardAccumulationHookError {
 		BackToTheFuture,
 		RewardsPotEmpty,
-		ArithmeticError,
+		Overflow,
 	}
 
 	#[pallet::error]
@@ -737,18 +739,17 @@ pub mod pallet {
 					);
 
 					ensure!(
-						initial_reward_config.iter().all(|(_, reward_config)| {
+						initial_reward_config.iter().all(|(reward_asset_id, reward_config)| {
 							if reward_config.reward_rate.amount > T::Balance::zero() {
-								// If none zero reward, check that the slashed amount is greater
-								// than ED
+								// if the reward rate amount is non-zero, then ensure that the
+								// slashed amount is >= the exisistential deposit for this asset
 								lock.unlock_penalty
 									.left_from_one()
 									.mul(reward_config.reward_rate.amount) >=
-									existential_deposit
+									T::ExistentialDeposits::get(reward_asset_id)
 							} else {
-								// Else, return true so check passes
-								// NOTE(connor): This is a band-aid that some better type management
-								// would remove the need for, outside the scope of this PR
+								// otherwise, since there are no rewards, no need to check against
+								// the existential deposit
 								true
 							}
 						}),
@@ -1466,8 +1467,6 @@ pub mod pallet {
 			reward: &mut Reward<T::Balance>,
 			now_seconds: u64,
 		) {
-			use RewardAccumulationCalculationError::*;
-
 			let pool_account = Self::pool_account_id(&pool_id);
 
 			match do_reward_accumulation::<T>(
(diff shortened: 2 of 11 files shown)


code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -734,18 +739,17 @@ pub mod pallet {
...
 							if reward_config.reward_rate.amount > T::Balance::zero() {
-								// If none zero reward, check that the slashed amount is greater
-								// than ED
+								// if the reward rate amount is non-zero, then ensure that the
+								// slashed amount is >= the exisistential deposit for this asset
 								lock.unlock_penalty
 									.left_from_one()
 									.mul(reward_config.reward_rate.amount) >=
-									existential_deposit
+									T::ExistentialDeposits::get(reward_asset_id)
```
benluelo, line 749: @PoisonPhang I believe you wrote this originally - can you confirm that this change makes sense? also cc @vimukthi-git here

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1664,85 +1663,101 @@ fn update_rewards_pool<T: Config>(
...
+			.ok_or(RewardAccumulationCalculationError::RewardsPotEmpty)?;
+
+	let releasable_periods_surpassed =
+		cmp::min(maximum_releasable_periods, periods_surpassed.into());
+
+	// SAFETY: Usage of mul is safe here. See the following proof:
+	//
+	// ```plaintext
+	//                                    (  total_locked_rewards           elapsed_time        )
+	// releasable_periods_surpassed = min ( ---------------------- , -------------------------- )
```
vimukthi-git, line 1715: I would just define some symbols to represent each value when things become this long. Not easy to read as of now.

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1664,85 +1663,101 @@ fn update_rewards_pool<T: Config>(
...
+	// always infallible (the resulting value will always be <= the total_locked_rewards input due
+	// to integer division), the second part is also infallible if it is less than the first.
+	let newly_accumulated_rewards = releasable_periods_surpassed
+		.checked_mul(reward_rate_amount)
+		.expect("should not fail; see above for proof; qed;");
+
+	let new_total_rewards = newly_accumulated_rewards
+		.get()
+		.checked_add(reward.total_rewards.into())
+		// TODO(benluelo): Rename to overflow to be more specific
```
vimukthi-git, line 1739: ?

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1664,85 +1663,101 @@ fn update_rewards_pool<T: Config>(
...
-			T::Assets::release(
-				asset_id,
-				pool_account,
-				newly_accumulated_rewards.into(),
-				false, // not best effort, entire amount must be released
-			)
-			.expect("funds should be available to release based on previous check; qed;");
+	// if no periods have been surpassed, short-circuit
+	let Some(periods_surpassed) = NonZeroU64::new(periods_surpassed) else {
+		return Ok(())
```
vimukthi-git, line 1695: :+1:

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1664,85 +1663,101 @@ fn update_rewards_pool<T: Config>(
...
-		if !newly_accumulated_rewards.is_zero() {
-			T::Assets::release(
-				asset_id,
-				pool_account,
-				newly_accumulated_rewards.into(),
-				false, // not best effort, entire amount must be released
-			)
-			.expect("funds should be available to release based on previous check; qed;");
+	// if no periods have been surpassed, short-circuit
+	let Some(periods_surpassed) = NonZeroU64::new(periods_surpassed) else {
```
vimukthi-git, line 1694: Excessive verbosity to check if something is zero and return no?

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1664,85 +1663,101 @@ fn update_rewards_pool<T: Config>(
 }
 
 /// Calculates the update to the reward and unlocks the accumulated rewards from the pool account.
+///
+/// NOTE: `<uX as Div<NonZeroUX>>::div` is used throughout this function as a safe,
```
vimukthi-git, line 1667: :+1:

code/parachain/frame/staking-rewards/src/lib.rs
vimukthi-git: It makes sense, as discussed back then. Reward rates should allow always for the unlock penalty to be transferred if an early unlock happens.

vimukthi-git (approved):

vimukthi-git:
@benluelo We need a clickup task linked for any github PRs.

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: I was asking about the use of `T::ExistentialDeposits::get(reward_asset_id)` rather than the existential deposit of the pool asset. I believe the previous implementation was incorrect

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: forgot to remove this comment, thanks :+1:

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: This also carries the non-zero invariant through the rest of the code without having to track it ourselves

code/parachain/frame/staking-rewards/src/lib.rs
benluelo: I was going to define symbols, but I wanted to show explicitly what each part was, to more easily explain the proof. I agree that it is somewhat unwieldy, I'll see what I can do

code/parachain/frame/staking-rewards/src/lib.rs
PoisonPhang: Yes, this makes sense +1

PoisonPhang (approved):
LGTM

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1595,8 +1533,134 @@ pub mod pallet {
...
+	let pool_account = Pallet::<T>::pool_account_id(&pool_id);
+
+	log::info!("calculating rewards for pool {pool_id:?}, asset {reward_asset_id:?}");
+
+	// no need to calculate if the pot is empty, this will be unset if and when enough funds
+	// are added to the pot
+	if RewardsPotIsEmpty::<T>::contains_key(pool_id, reward_asset_id) {
+		log::info!(
+			"pot for pool {pool_id:?}, asset {reward_asset_id:?} is empty, not accumulating"
+		);
```
benluelo, line 1554: These logs are not permanent, i added them to debug the accumulation logic. i would like to keep some of them however, as i feel this would be useful information to have in case something goes wrong in production (perhaps using sp-tracing)

code/parachain/frame/composable-traits/src/staking/mod.rs
```diff
@@ -54,6 +54,19 @@ impl<Balance> RewardRate<Balance> {
 		Self { period: RewardRatePeriod::PerSecond, amount: amount.into() }
 	}
 }
+impl<Balance> RewardRate<Balance>
+where
+	Balance: Copy + Into<u128> + From<u128>,
+{
+	/// The amount that will be rewarded in one period.
+	// TODO(benluelo): Encode somehow that one period will always be rewardable. We can most likely
```
vimukthi-git, line 62: We need to consider not adding so many TODOs and may be use something like `IMPROVEMENT`. TODO implies a definite thing to do. Otherwise code looks incomplete to the reader, which is not the case here.

code/parachain/frame/staking-rewards/src/test/mod.rs
```diff
@@ -104,6 +103,34 @@ fn duration_presets_minimum_is_1() {
 	});
 }
 
+#[test]
+fn zero_length_duration_preset_works() {
```
vimukthi-git, line 107: :+1:

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -1595,8 +1533,134 @@ pub mod pallet {
...
+	pool_id: T::AssetId,
+	reward_asset_id: T::AssetId,
+	reward: &mut Reward<T::Balance>,
+	now_seconds: u64,
+) {
+	let pool_account = Pallet::<T>::pool_account_id(&pool_id);
+
+	log::info!("calculating rewards for pool {pool_id:?}, asset {reward_asset_id:?}");
+
+	// no need to calculate if the pot is empty, this will be unset if and when enough funds
```
vimukthi-git, line 1549: :+1:

vimukthi-git (approved):
LGTM

PR #2911, opened 2023-01-12
#cu-20465559 #ibc #cosmwasm module integration into #dali runtime

dzmitry-lahoda:
made coswasm to impl latest ibc traits and started to route pallet calls to cosmwasm on dali

### DIFF
--- a/code/parachain/frame/cosmwasm/src/ibc.rs
+++ b/code/parachain/frame/cosmwasm/src/ibc.rs
@@ -159,9 +159,15 @@ impl<T: Config> Pallet<T> {
 	}
 }
 
-#[derive(Default, RuntimeDebug, Eq, PartialEq, Clone)]
+#[derive(RuntimeDebug, Eq, PartialEq, Clone)]
 pub struct Router<T: Config> {
 	_marker: PhantomData<T>,
+}
+
+impl<T: Config> Default for Router<T> {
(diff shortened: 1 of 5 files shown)


code/parachain/frame/cosmwasm/src/ibc.rs
```diff
@@ -159,11 +159,17 @@ impl<T: Config> Pallet<T> {
 	}
 }
 
-#[derive(Default, RuntimeDebug, Eq, PartialEq, Clone)]
+#[derive(RuntimeDebug, Eq, PartialEq, Clone)]
 pub struct Router<T: Config> {
 	_marker: PhantomData<T>,
 }
 
+impl<T: Config> Default for Router<T> {
```
blasrodri, line 167: why not deriving Default?

blasrodri (approved):

code/parachain/frame/cosmwasm/src/ibc.rs
dzmitry-lahoda: as I recall runtme macro does not works with attributes.

obsessed-cake (approved):

dzmitry-lahoda:
@andyjsbell please take a look

andyjsbell (approved):
thanks @dzmitry-lahoda

PR #2916, opened 2023-01-13
PICA-70: Properly calculate unvested amount

easteregg:


frontend/apps/picasso/pages/crowdloan-rewards/claim.tsx
```diff
@@ -57,19 +56,49 @@ const ERROR_MESSAGES = {
   },
 };
 
+function getUnvestedAmount(
```
vimukthi-git, line 59: Ideally this should be a RPC, but I guess doesn't make sense to spend a lot of time on crowdloan pallet

vimukthi-git (approved):

PR #2922, opened 2023-01-13
zombienet runs devnet all and in devnet container well #cu-862j1xn5b

dzmitry-lahoda:
- removed karura only devnet (not run directly ever, only as part of ALL)
- DALI container and ALL ZN run well
- removed not used files (these are generated with ALL script now)
- prepared some container tools for reuse in bridge arion (moved them into tools)


inputs/paritytech/zombienet/flake-module.nix
```diff
@@ -0,0 +1,83 @@
...
+        rev = "b9f089eb55a7cb1d4c12575b8323cb9b9fab4a60";
+        hash = "sha256-EC6XKbcI+Is0RGlmC8WGjPqiFh9Ulf3bXDoVihtYqsU=";
+      };
+      paritytech-zombienet = pkgs.stdenv.mkDerivation {
+        name = "zombienet";
+        src = "${paritytech-zombienet-src}/javascript";
+        buildInputs = with pkgs; [ nodejs ];
+        nativeBuildInputs = with pkgs;
+          [
+            yarn
```
obsessed-cake, line 16: Is it exactly needed? You use `npm` to build `zombienet`.

obsessed-cake (approved):

inputs/paritytech/zombienet/flake-module.nix
dzmitry-lahoda: they have adhoc custom build with `so` deps (rust stuff), that just does not builds/runs in nix that simple. so need to try again later

dzmitry-lahoda: so answers to be clear-  i tried to run build, it failed with some nasty error. so need to try again as some point.

KaiserKarel (approved):
It's a bit above my nix level, but looks good

cor (approved):
LGTM! Nice work and I like how you made `zombienet` a directory. Seems like a good solution for when an input has "supporting files"

dzmitry-lahoda:
cc @andyjsbell

inputs/paritytech/zombienet/flake-module.nix
dzmitry-lahoda: https://app.clickup.com/t/20465559/NX-67

PR #2923, opened 2023-01-13
NX-23 - `nix run .\#help`

cor:
Content suggestions are welcome. Currently looks like this:

image

### DIFF
--- a/flake/help.nix
+++ b/flake/help.nix
@@ -1,14 +1,13 @@
 { self, ... }: {
-  perSystem = { pkgs, ... }:
-   {
-      packages = {
-        help = pkgs.writeShellApplication {
-          name = "help";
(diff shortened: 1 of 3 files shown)


flake/help.nix
```diff
@@ -0,0 +1,14 @@
+{ self, ... }: {
+  perSystem = { config, self', inputs', pkgs, system, ... }:
```
obsessed-cake, line 2:
```suggestion
{ ... }: {
  perSystem = { pkgs, ... }:
```

obsessed-cake (approved):

flake/help.nix
cor: done in different commit

andyjsbell (approved):

PR #2930, opened 2023-01-16
PBL-169: Add Alert when User doesn't have enough KSM on USDT Picasso=>Statemine

easteregg:


frontend/apps/picasso/pages/transfers/index.tsx
```diff
@@ -43,7 +46,22 @@ const Transfers: NextPage = () => {
...
+      to === "statemine" &&
+      ksmBalance.lt(
+        fromChainIdUnit(
+          PICASSO_STATEMINE_KSM_TRANSFER_FEE,
+          tokens.ksm.decimals.picasso
+        )
+      ) &&
+      selectedToken === "usdt",
+    [from, to, ksmBalance, tokens.ksm.decimals.picasso, selectedToken]
+  );
```
fl-y, line 64: Does useMemo make sense here? On a quick skim it doesn't seem so to me

reference: https://blog.logrocket.com/rethinking-hooks-memoization/

frontend/apps/picasso/pages/transfers/index.tsx
```diff
@@ -146,6 +164,11 @@ const Transfers: NextPage = () => {
         <Grid item {...gridItemStyle("1.5rem")}>
           <TransferRecipientDropdown />
         </Grid>
+        {isKSMAlertVisible ? (
+          <Grid item {...gridItemStyle("1.5rem")}>
+            <TransferKSMAlert />
+          </Grid>
+        ) : null}
```
fl-y, line 171:
```suggestion
        {isKSMAlertVisible && (
          <Grid item {...gridItemStyle("1.5rem")}>
            <TransferKSMAlert />
          </Grid> 
        )}
```
unecessary ternary

fl-y (approved):
Left comments that do not necessarily need to be addressed

frontend/apps/picasso/pages/transfers/index.tsx
easteregg: Conditional rendering with logical operators has a tendency to create issues later on if we don't be careful with it to make sure it is always boolean. There is always the danger of short-circuiting when the parent condition changes. 

That's why I have chosen the safest, but, more verbose route.

frontend/apps/picasso/pages/transfers/index.tsx
easteregg: You are right, I usually don't overuse memoization, but had a gut feeling this one needed it. Will remove, retest and post update here.

easteregg: You got it 100%. it doesn't add any value. removing it.

easteregg:
Closing in favor of https://github.com/ComposableFi/composable/pull/2963

PR #2933, opened 2023-01-16
[PICA-86] add test for claiming as per bug report

benluelo:
https://app.clickup.com/t/20465559/PICA-86


vimukthi-git (approved):

vimukthi-git:
This had errors which we missed when rebasing

PR #2935, opened 2023-01-16
Fix simnode check (kinda)

benluelo:


vimukthi-git:
Given Cost(high) vs benefit (~0) of simnode I think we should just get rid of it

KaiserKarel:
Since simnode is deprecated, this can be closed.

PR #2939, opened 2023-01-17
tagging, so hasura is database AND gql api

dzmitry-lahoda:


dzmitry-lahoda:
please repone if any need

PR #2940, opened 2023-01-17
migrate bridge stuff to zn 2 #Cu-867747xuf

dzmitry-lahoda:
based on https://github.com/ComposableFi/composable/pull/2922 for https://app.clickup.com/t/867747xuf

- removed not used bridge stuff
- centauri net works as it worked (removed mounting file toml as directory)
- added oneliner to follow logs of devnet a (not generalised, so it is clear how to proceed) 
- `find` to be in base container 
- enabled logs overrides and offchain indexer in devnet
- nix run .#fmt (seems people do not run fmt)


flake/zombienet.nix
```diff
@@ -97,11 +117,26 @@
...
         zombienet-picasso-complete = {
           type = "app";
           program = self'.packages.zombienet-picasso-complete;
         };
 
+        zombienet-log-follow = {
+          program = pkgs.writeShellApplication rec {
+            name = "zombienet-log-follow";
+            text = ''
+              docker exec -it composable-devnet-a-1   bash  -c 'LOG=$(find /tmp/ -name "zombie-*" | head --lines=1)/alice.log && tail --follow $LOG'
```
blasrodri, line 134: nice

blasrodri: why only Alice?

blasrodri (approved):

flake/zombienet.nix
dzmitry-lahoda: only alice.log node has well know port 9988, other nodes has random ports and will not be published outside

dzmitry-lahoda: so likely nothing will try to connect to bob.log

andyjsbell (approved):
Thanks @dzmitry-lahoda

dzmitry-lahoda:
moving to fork

PR #2906, opened 2023-01-11
[FE] switch rpc to picasso-rpc-lb.composablenodes.tech

fl-y:
all in title

PR #2941, opened 2023-01-17
PICA-90: Update endpoint

easteregg:


fl-y:
Duplicate of https://github.com/ComposableFi/composable/pull/2906 but will close that one instead

fl-y (changes requested):
actually you might have missed one other link?

easteregg:
@fl-y The other place is actually not used anywhere, we need to phase it out. but for clarify I've updated it to avoid confusion in the future.

PR #2942, opened 2023-01-18
CosmWasm Pallet IBC Integration tests #Cu-20465559

dzmitry-lahoda:
- split out module for ibc-rs <-> cosmwasm IBC types mapping 
- IBC router calls hook OR VM as with IBC rs calls (proper use of recursion)


dzmitry-lahoda:
moving to fork

PR #2943, opened 2023-01-18
(CU-862hy89h1) update legal disclaimers and disclosures in book

Martin-Kiss:


Martin-Kiss:
Reopened from a personal fork as #2961

PR #2175, opened 2022-11-02
[CU-#2t7r08p] Integration Tests: Staking Rewards

Dom-Roth:
~~This reverts commit b0b39205~~
This reverts a revert commit.

## Issue

The original staking rewards integration tests should have never been merged in the first place. Due to bugs causing failing tests.
Therefore the previous PR got reverted, and now the revert gets reverted, because we still need those tests.

## Additional Changes

~~~~

## Checklist

~~- [ ] Updated the rust/typescript docs.~~
~~- [ ] Updated the `docs/`.~~

PR #2950, opened 2023-01-19
Integration Tests: Staking Rewards

Dom-Roth:
Recreation of ~~#2175~~ #2889


josietyleung:
we recoped so closing

PR #2970, opened 2023-01-19
Use subwasm to export chain metadata #NX-30

obsessed-cake:


KaiserKarel:
Task linked: [NX-30 Use `subwasm` to export chain metadata](https://app.clickup.com/t/20465559/NX-30)

PR #2972, opened 2023-01-20
Switch to pull_request_target

KaiserKarel:
Normally, PRs from forks cannot access secrets. This allows them to do so. PRs that alter workflow files won't be able to access secrets still I think.

### DIFF
--- a/.github/workflows/build-test-deploy.yml
+++ b/.github/workflows/build-test-deploy.yml
@@ -1243,6 +1243,9 @@ jobs:
 
 
   nix-command-pr-comments:
+    permissions:
+      pull-requests: write
+      issues: write
     runs-on:
       - self-hosted
       - linux
--- a/.github/workflows/build-test-deploy.yml
+++ b/.github/workflows/build-test-deploy.yml
@@ -35,6 +35,9 @@ jobs:
       image: niteo/nixpkgs-nixos-22.05:316b762afdb9e142a803f29c49a88b4a47db80ee
     steps:
       - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          persist-credentials: false
       - uses: "./.github/templates/watch-exec"
         with:
           command: nix -- build .#devnet-initialize-script-picasso-persistent
@@ -56,6 +59,9 @@ jobs:
       image: niteo/nixpkgs-nixos-22.05:316b762afdb9e142a803f29c49a88b4a47db80ee
     steps:
       - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          persist-credentials: false
       - uses: "./.github/templates/watch-exec"
         with:
           command: nix -- build .#common-deps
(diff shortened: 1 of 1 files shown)


KaiserKarel:
Task linked: [SRE-534 PR from forks into Composable fail all jobs](https://app.clickup.com/t/20465559/SRE-534)

dzmitry-lahoda (approved):
thanks. so i guess secret part could be confirmed by sre or infra because if not people will have rotate keys.

KaiserKarel:
> thanks. so i guess secret part could be confirmed by sre or infra because if not people will have rotate case.

So in this case, secrets will work in PRs from forks but the workflow code that is run is using base, not fork.

dzmitry-lahoda:
ah, ok than.

KaiserKarel:
I've checked that it is valid syntax, so unsure why the runners aren't picking it up. @radupopa2010 can you have a look?

radupopa2010:
People suggest that  changing the name of the job requires a change in the config of "branch protection rules"

I see here se change the event name, so we might need to check the branch protection rules. 
I do not have admin access for this repo so I can't have a look; 
@KaiserKarel can you please check those?


https://github.com/orgs/community/discussions/26698#discussioncomment-3252954
_[Configuration of branch protection rules](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule) is completely separate from your workflows. I assume that at some point in the past you’ve marked the previous jobs as required checks (step seven in the documentation, or search for “required status checks”). Once enabled, those jobs are required by name._

_You’ve changed job names, so you need to update the branch protection rule: Remove the old jobs, so waiting for jobs that don’t exist any more doesn’t block merging. And ideally add your new job (the non-matrix one) to that gets checked._

KaiserKarel:
No that's not the case, the problem is that the workflow itself is not being spun up at all. If it was incorrect naming, we'd see it being spun up but the checks not being satisfied.

KaiserKarel:
Do the runners have some log somewhere saying: syntax error in workflow file?

radupopa2010 (approved):

vimukthi-git (approved):
LGTM

radupopa2010:
> Do the runners have some log somewhere saying: syntax error in workflow file?

We use google logs exlorer
https://console.cloud.google.com/logs/query;cursorTimestamp=2023-01-20T11:34:24.087359931Z?referrer=search&project=composable-ci

The logs view is aggregated and we need to filter by VM instance id. To get that I need to run `gcloud compute instances describe`  on the name of the runner

 We use ephemeral runners and we have a bunch of names like `ghr-2aa49e`.   When running a job one can pickup the name of the runner from the setup job , e.g.
`Runner name: 'ghr-7a252e'` ; 

Since VM is not running anymore it's difficult for me to pull that ID now.

KaiserKarel:
Luckily we figured out the issue, but indeed nice to be able to store those logs.

PR #2983, opened 2023-01-20
XCVM-282: Halborn audit fixes (HAL-01,HAL-02,HAL-03,HAL-04,HAL-05)

aeryz:
Upgrade Cosmwasm VM which includes halborn's audit fixes. See the details: https://docs.google.com/document/d/17yRM8JrzS6sppEAo0KsXeRbLEyvyuN0WYI0LMOWAM9k

To briefly explain what is done:
1. `HAL-01`: This is false positive. Because if you check the implementation of `UpdateAdmin`, you can see that we are already converting the address from `String` to the native address by using `TryInto` which means that the vm implementer must take care of this. The abstract VM can't validate the address since it doesn't know the format.
2. `HAL-02`: This was saying that the wasm contract validation is not done but this is actually being done in the pallet side. Now we get this validation logic to the VM and still call the validation method from the pallet. So we are good.
3. `HAL-03`: We weren't doing enough validation for the allocated regions, and we added the necessary validations to VM.
4. `HAL-04`: We were missing some pointer validation and we added the necessary checks to VM.
5. `HAL-05`: We were using an unsafe function without checking the size of the underlying data which could potentially panic, and we fixed that on the VM.

**NOTE**: The fixes are done in the abstract VM side. This is just for integrating that fixes. So, please also check the following PR to see more: https://github.com/ComposableFi/cosmwasm-vm/pull/10


KaiserKarel:
Task linked: [XCVM-282 Move exports validation to CosmWasm VM](https://app.clickup.com/t/20465559/XCVM-282)

hussein-aitlahcen (approved):

KaiserKarel:
Could you describe how the fixes were applied?

KaiserKarel:
As well as include the HAL identifier in the commit

aeryz:
> As well as include the HAL identifier in the commit

there are multiple identifiers so, should I make it like `HAL-01,HAL-02,HAL-03,HAL-04,HAL-05 fixes`

KaiserKarel:
> > As well as include the HAL identifier in the commit
> 
> there are multiple identifiers so, should I make it like `HAL-01,HAL-02,HAL-03,HAL-04,HAL-05 fixes`

Yeah that's fine, as long as we can grep for it later. Something like $REPORT_NAME:[$IDENTIFIERS]

KaiserKarel (approved):

KaiserKarel:
@mergifyio refresh

KaiserKarel:
@mergifyio refresh

PR #2934, opened 2023-01-16
[PICA-76] wrap LockConfig::duration_presets in an enum

benluelo:
https://app.clickup.com/t/20465559/PICA-76

PR #2999, opened 2023-01-25
Wrap `LockConfig::duration_presets` in an enum

benluelo:
See #2934 for original PR and previous reviews

### DIFF
--- a/code/parachain/frame/staking-rewards/src/benchmarking.rs
+++ b/code/parachain/frame/staking-rewards/src/benchmarking.rs
@@ -49,13 +49,17 @@ fn get_reward_pool<T: Config>(
 
 fn lock_config<T: Config>() -> LockConfig<T::MaxStakingDurationPresets> {
 	LockConfig {
-		duration_presets: [
-			(ONE_HOUR, FixedU64::from_rational(101, 100).try_into_validated().expect(">= 1")), /* 1% */
-			(ONE_MINUTE, FixedU64::from_rational(1_001, 1_000).try_into_validated().expect(">= 1")), /* 0.1% */
-		]
-		.into_iter()
-		.try_collect()
-		.unwrap(),
+		duration_multipliers: frame_support::bounded_btree_map! {
+			// 1%
+			ONE_HOUR => FixedU64::from_rational(101, 100)
+				.try_into_validated()
+				.expect(">= 1"),
+			// 0.1%
+			ONE_MINUTE => FixedU64::from_rational(1_001, 1_000)
+				.try_into_validated()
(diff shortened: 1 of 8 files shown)


KaiserKarel:
Task linked: [PICA-76 Wrap `LockConfig::duration_presets` in an enum](https://app.clickup.com/t/20465559/PICA-76)

code/parachain/frame/staking-rewards/src/benchmarking.rs
```diff
@@ -49,13 +51,20 @@ fn get_reward_pool<T: Config>(
...
+				(ONE_HOUR, FixedU64::from_rational(101, 100).try_into_validated().expect(">= 1")),
+				// 0.1%
+				(
+					ONE_MINUTE,
+					FixedU64::from_rational(1_001, 1_000).try_into_validated().expect(">= 1"),
+				),
+			]
+			.into_iter()
+			.try_collect()
+			.unwrap(),
```
kkast, line 66: maybe also use bounded_btree_map! as in test/mod.rs?

code/parachain/frame/staking-rewards/src/test/mod.rs
```diff
@@ -246,23 +236,17 @@ fn create_staking_reward_pool_should_fail_when_slashed_minimum_amount_is_less_th
...
-							), // 0.1%
-						]
-						.into_iter()
-						.try_collect()
-						.unwrap(),
+						duration_multipliers: bounded_btree_map! {
+							// 1%
+							ONE_HOUR => FixedU64::from_rational(101, 100)
+								.try_into_validated()
+								.expect(">= 1"),
```
kkast, line 243: is the old error message not valid anymore?

kkast (approved):

code/parachain/frame/staking-rewards/src/benchmarking.rs
benluelo: not possible since that macro isn't available in no_std (which benchmarks are)

code/parachain/frame/staking-rewards/src/test/mod.rs
benluelo: I read it as "expect <the value> to be >= 1", which is easier to grok than "expect <the value> to be a valid reward multiplier". just a semantic thing, doesn't really mean anything since it's a test anyways

vimukthi-git (approved):

cor (approved):

benluelo:
@mergify refresh

PoisonPhang (approved):

PR #3001, opened 2023-01-26
[PICA-102] add claimed_amount to Claimed event

kkast:
https://app.clickup.com/t/86775ywct

- add claimed_amount to Claimed event
- removed logic for slashing rewards for early unstakes
- fixed tests

### DIFF
--- a/code/parachain/frame/staking-rewards/src/lib.rs
+++ b/code/parachain/frame/staking-rewards/src/lib.rs
@@ -167,8 +167,7 @@ pub mod pallet {
 			fnft_collection_id: T::AssetId,
 			/// FNFT Instance Id
 			fnft_instance_id: T::FinancialNftInstanceId,
-			// Claimed amount map<reward id, amount>
-			claimed_amount: BTreeMap<T::AssetId, T::Balance>,
+			claimed_amounts: BTreeMap<T::AssetId, T::Balance>,
 		},
 		StakeAmountExtended {
 			/// FNFT Collection Id
@@ -1141,7 +1140,7 @@ pub mod pallet {
 				owner: who.clone(),
 				fnft_collection_id: *fnft_collection_id,
 				fnft_instance_id: *fnft_instance_id,
-				claimed_amount: claimed_amounts,
+				claimed_amounts,
 			});
 
 			Ok(())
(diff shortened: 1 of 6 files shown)


KaiserKarel:
Task linked: [PICA-102 Add claimed_amount to claimed event](https://app.clickup.com/t/20465559/PICA-102)

kkast:
@benluelo please take a look at collect_rewards function logic

vimukthi-git (approved):

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -167,6 +167,8 @@ pub mod pallet {
 			fnft_collection_id: T::AssetId,
 			/// FNFT Instance Id
 			fnft_instance_id: T::FinancialNftInstanceId,
+			// Claimed amount map<reward id, amount>
```
benluelo, line 170: This comment doesn't add anything, can be removed

code/parachain/frame/staking-rewards/src/lib.rs
```diff
@@ -167,6 +167,8 @@ pub mod pallet {
...
+			claimed_amount: BTreeMap<T::AssetId, T::Balance>,
```
benluelo, line 171: `claimed_amounts` makes more sense here, as there is more than one amount

code/parachain/frame/staking-rewards/src/test/mod.rs
```diff
@@ -1378,6 +1378,7 @@ fn claim_with_insufficient_pot_funds() {
 				owner: BOB,
 				fnft_collection_id: STAKING_FNFT_COLLECTION_ID,
 				fnft_instance_id: bob_stake_id,
+				claimed_amount: BTreeMap::from([(USDT::ID, 900000)]),
```
benluelo, line 1381: would be helpful to explain where these claimed amounts come from, or use a function to get the value (i.e. claim_of_stake or something similar)

code/parachain/frame/staking-rewards/src/test/test_update_reward_pools.rs
```diff
@@ -217,6 +218,7 @@ fn update_accumulates_properly() {
 				owner: CHARLIE,
 				fnft_collection_id: STAKING_FNFT_COLLECTION_ID,
 				fnft_instance_id: stake_id,
+				claimed_amount: BTreeMap::from([(USDT::ID, USDT::units(1) / 1000 * 6)]),
```
benluelo, line 221: prefer to use `[..].into_iter().collect()` over `BTreeMap::from([..])`

benluelo: There is also a `btree_map` helper function that can be used as well, but .iter.collect tends to format more nicely

benluelo (changes requested):

code/parachain/frame/staking-rewards/src/test/mod.rs
benluelo: also `900000` should be `900_000`

benluelo (approved):

PR #3048, opened 2023-01-30
chore: add pallet-ibc to Picasso

blasrodri:
As part of the DOT<>KSM release, this PR includes `pallet-ibc` into the `Picasso` runtime.


code/parachain/runtime/picasso/src/lib.rs
```diff
@@ -178,6 +191,88 @@ parameter_types! {
 	pub const SS58Prefix: u8 = 49;
 }
 
+type CosmwasmRouter = cosmwasm::ibc::Router<Runtime>;
+
+#[derive(Clone, Debug, Eq, PartialEq, Default)]
+pub struct Router {
+	pallet_ibc_ping: pallet_ibc_ping::IbcModule<Runtime>,
```
Wizdave97, line 198: We don't need this pallet here.

blasrodri: How do I create the router then?

Wizdave97 (commented):
pallet-ibc-rpc needs to be enabled.

code/parachain/runtime/picasso/src/lib.rs
Wizdave97: You can replace it with the cosmwasm router

vimukthi-git:
@blasrodri This should not be merged before release 3, I recommend maintaining your own branch or using dali for testing for now.

KaiserKarel:
> @blasrodri This should not be merged before release 3, I recommend maintaining your own branch or using dali for testing for now.

@vimukthi-git checking in. Because we'd like to get this merged to get QA started.

code/parachain/runtime/picasso/src/lib.rs
```diff
@@ -1230,3 +1464,27 @@ cumulus_pallet_parachain_system::register_validate_block!(
...
+	const LIGHT_CLIENT_PROTOCOL: pallet_ibc::LightClientProtocol =
+		pallet_ibc::LightClientProtocol::Grandpa;
+	type AccountIdConversion = ibc_primitives::IbcAccount<AccountId>;
+	type Fungibles = Assets;
+	type ExpectedBlockTime = ExpectedBlockTime;
+	type Router = Router;
+	type MinimumConnectionDelay = MinimumConnectionDelay;
+	type ParaId = parachain_info::Pallet<Runtime>;
+	type RelayChain = RelayChainId;
+	type WeightInfo = ();
```
KaiserKarel, line 1486: This will need to be generated and set.

code/parachain/runtime/picasso/src/lib.rs
```diff
@@ -1230,3 +1464,27 @@ cumulus_pallet_parachain_system::register_validate_block!(
...
+	type AdminOrigin = EnsureRoot<AccountId>;
```
KaiserKarel, line 1487: This should be RootOrHalfCouncil

code/parachain/runtime/picasso/src/lib.rs
```diff
@@ -1230,3 +1464,27 @@ cumulus_pallet_parachain_system::register_validate_block!(
...
+	type SentryOrigin = EnsureRoot<AccountId>;
```
KaiserKarel, line 1488: Does sentry need to be root?

KaiserKarel:
@blasrodri Does this include the whitelisted usage?

blasrodri:
> @blasrodri Does this include the whitelisted usage?

It doesn't. What assets/policy do we want? And at which level do we want to define this (runtime vs pallet-ibc)?
cc @Wizdave97

code/parachain/runtime/picasso/src/lib.rs
blasrodri: Is there a guide/ref on how to do it?

code/parachain/runtime/picasso/src/lib.rs
blasrodri: do you mean
`type AdminOrigin = EnsureRootOrHalfNativeCouncil;`

vimukthi-git:
> @vimukthi-git checking in. Because we'd like to get this merged to get QA started.

@KaiserKarel So this is now part of release 3? Generally you can do QA on dali as well as we do now for staking.

andyjsbell:
> It doesn't. What assets/policy do we want? And at which level do we want to define this (runtime vs pallet-ibc)?

@blasrodri @KaiserKarel looks like we should get this covered in the scope for the parent initiative before we move forward

KaiserKarel:
> > It doesn't. What assets/policy do we want? And at which level do we want to define this (runtime vs pallet-ibc)?
> 
> @blasrodri @KaiserKarel looks like we should get this covered in the scope for the parent initiative before we move forward

I think it is fine merging this; 3rd party devs will be at risk of bricking their tokens on LAYR but not much else, and in a subsequent PR add the address whitelist?

code/parachain/runtime/picasso/src/lib.rs
```diff
@@ -178,6 +196,78 @@ parameter_types! {
...
+	}
+}
+
+pub struct IbcDenomToAssetIdConversion;
+
+impl DenomToAssetId<Runtime> for IbcDenomToAssetIdConversion {
+	type Error = DispatchError;
+
+	fn from_denom_to_asset_id(denom: &String) -> Result<CurrencyId, Self::Error> {
+		let denom_bytes = denom.as_bytes().to_vec();
```
KaiserKarel, line 222: Is the to_vec necessary here ownership wise?

code/parachain/runtime/picasso/src/lib.rs
KaiserKarel: I think @vimukthi-git can perform this to you as part of the release flow (benching everything).

code/parachain/runtime/picasso/src/lib.rs
```diff
@@ -23,6 +23,8 @@ pub const WASM_BINARY_V2: Option<&[u8]> = Some(include_bytes!(env!("PICASSO_RUNT
 #[cfg(not(feature = "builtin-wasm"))]
 pub const WASM_BINARY_V2: Option<&[u8]> = None;
 
+extern crate alloc;
```
KaiserKarel, line 26: isn't extern crate deprecated, and we can just refer to alloc directly?

code/parachain/runtime/picasso/src/lib.rs
```diff
@@ -150,6 +165,9 @@ parameter_type_with_key! {
 }
 
 parameter_types! {
+	pub const RelayChainId: RelayChain = RelayChain::Rococo;
+	pub const SpamProtectionDeposit: Balance = 1_000_000_000_000;
+	pub const MinimumConnectionDelay: u64 = 0;
 	// how much block hashes to keep
 	pub const BlockHashCount: BlockNumber = 250;
```
KaiserKarel, line 172: nit: perhaps a power of 2 because computer science?

code/parachain/runtime/picasso/src/lib.rs
```diff
@@ -178,6 +196,78 @@ parameter_types! {
 	pub const SS58Prefix: u8 = 49;
```
KaiserKarel, line 196: This should be in common/constants, instead of being hardcoded (I think we already have a variable used by the RPC)

code/parachain/runtime/picasso/src/lib.rs
```diff
@@ -178,6 +196,78 @@ parameter_types! {
...
+	}
+
+	fn has_route(_module_id: &ModuleId) -> bool {
+		false
+	}
+
+	fn lookup_module_by_port(_port_id: &PortId) -> Option<ModuleId> {
+		None
+	}
+}
```
KaiserKarel, line 214: Perhaps move this to pallet-ibc and rename to NoneRouter/NeverRouter?

code/parachain/runtime/picasso/src/lib.rs
```diff
@@ -178,6 +196,78 @@ parameter_types! {
...
+	type Error = DispatchError;
+
+	fn from_denom_to_asset_id(denom: &String) -> Result<CurrencyId, Self::Error> {
+		let denom_bytes = denom.as_bytes().to_vec();
+		if let Some(id) = IbcDenoms::<Runtime>::get(&denom_bytes) {
+			return Ok(id)
+		}
+
+		let asset_id =
+			<currency_factory::Pallet<Runtime> as CurrencyFactoryT>::create(RangeId::IBC_ASSETS)?;
```
KaiserKarel, line 228: @benluelo this will be affected by assets rewrite

code/parachain/runtime/picasso/src/lib.rs
```diff
@@ -178,6 +196,78 @@ parameter_types! {
...
+
+		IbcDenoms::<Runtime>::insert(denom_bytes.clone(), asset_id);
+		IbcAssetIds::<Runtime>::insert(asset_id, denom_bytes);
+
+		let location = XcmAssetLocation::new(MultiLocation::new(
```
KaiserKarel, line 233: Isn't this just `HERE` @dzmitry-lahoda

code/parachain/runtime/picasso/src/lib.rs
```diff
@@ -178,6 +196,78 @@ parameter_types! {
...
+			location,
+			Rational64::one(),
+			Some(12),
+		)?;
+
+		Ok(asset_id)
+	}
+
+	fn from_asset_id_to_denom(id: CurrencyId) -> Option<String> {
+		IbcAssetIds::<Runtime>::get(id).and_then(|denom| String::from_utf8(denom).ok())
```
KaiserKarel, line 248: If we have an invalid denom in storage, this will silently say it is simply missing. Can we change the signature to either Option<Result<String>>, or add a log line so we know it fails because bad encoding slipped in?

code/parachain/runtime/picasso/src/lib.rs
```diff
@@ -178,6 +196,78 @@ parameter_types! {
...
+
+		let denoms = iterator.by_ref().take(limit as usize).map(|(_, denom)| denom).collect();
+		let maybe_currency_id = iterator.next().map(|(id, ..)| id);
+		IbcAssets {
+			denoms,
+			total_count: IbcAssetIds::<Runtime>::count() as u64,
+			next_id: maybe_currency_id,
+		}
+	}
+}
```
KaiserKarel, line 269: Can we use into() instead of `as` to avoid possible truncation? Perhaps not needed because fetching u32::MAX items from storage is not feasible either; but still

code/parachain/runtime/picasso/src/lib.rs
benluelo: CC @PoisonPhang

code/parachain/runtime/picasso/src/lib.rs
blasrodri: I thought the same, but

```
error[E0433]: failed to resolve: use of undeclared crate or module `alloc`
  --> parachain/runtime/picasso/src/lib.rs:37:5
   |
37 | use alloc::string::String;
   |     ^^^^^ use of undeclared crate or module `alloc`
   |
   = help: add `extern crate alloc` to use the `alloc` crate
   ```

File: code/parachain/runtime/picasso/src/lib.rs (hunk 8 above)
Comment (blasrodri) [reply code/parachain/runtime/picasso/src/lib.rs:196]: Where is that located? I've seen that we're hardcoding this number many times throughout the codebase.
Happy to reference from a constants crate

File: code/parachain/runtime/picasso/src/lib.rs (hunk 9 above)
Comment (blasrodri) [reply code/parachain/runtime/picasso/src/lib.rs:214]: https://github.com/ComposableFi/centauri/pull/195/files

File: code/parachain/runtime/picasso/src/lib.rs (hunk 11 above)
Comment (dzmitry-lahoda) [reply code/parachain/runtime/picasso/src/lib.rs:216]: i would perfer we do not register anything in assets registry until rewrite. imho. we could do it in separate pr also. 

there could be different encoding depending on if we use or not xcm v3 i  think.

on current xcm v2 code base, it would be (parent = 0, general index = asset id).  or better (parent = 0, pallet id = 190, general index).

Comment (dzmitry-lahoda) [reply code/parachain/runtime/picasso/src/lib.rs:216]: pallet id convention helps to identify protocol which manages tokens on this chain.

Comment (dzmitry-lahoda) [reply code/parachain/runtime/picasso/src/lib.rs:216]: yeah

Better hardcode information and do not have any registration. as it will register new token each transfer? 

we will handle it later.

File: code/parachain/runtime/picasso/src/lib.rs (hunk 10 above)
Comment (PoisonPhang) [reply code/parachain/runtime/picasso/src/lib.rs:211]: Yes, this process will changes when done through assets registry.
Currently planning on using the `denom_bytes` as a type of foreign asset to register and create an asset ID through assets registry.

File: code/parachain/runtime/picasso/src/lib.rs (hunk 11 above)
Comment (PoisonPhang) [reply code/parachain/runtime/picasso/src/lib.rs:216]: I'd rather us treat the IBC Denom and the XCM location as different types of foreign asset ID.

File: code/parachain/runtime/picasso/src/lib.rs [hunk 14]
```diff
@@ -178,6 +197,68 @@ parameter_types! {
...
+		IbcAssetIds::<Runtime>::get(id).and_then(|denom| {
+			String::from_utf8(denom)
+				// TODO: replace this for `inspect_err` when it's available as a stable feature?
+				.map_err(|e| {
+					log::error!("Invalid encoding for asset id {} - err = {:?}", id, e);
+					e
+				})
+				.ok()
+		})
+	}
```
PoisonPhang, line 240: This can be handled by pallet-assets-registry

code/parachain/runtime/picasso/src/lib.rs
KaiserKarel: I think I am completely derping as I cannot find it.

code/parachain/runtime/picasso/src/lib.rs
blasrodri: can you please add that interaction?

dzmitry-lahoda:
I will fix all on other PR. This PR did merge, not rebase and on other fork. Hard to work with. Will cherry pick. Will respolve comments as these fixed.

blasrodri:
Thanks @dzmitry-lahoda

PR #3056, opened 2023-01-31
Fix HTTP error 200 from nix #cu-20465559 #XCVM-288 

dzmitry-lahoda:
for https://app.clickup.com/t/20465559/XCVM-288

- retry never worked (if I read code right - tried to bash - seems never worked - was afraid || in bash works like && )
- if cache failed, build without (longer, but robust)
- changes in rust code just to trigger some invalidation (and sure - make our linter stricter:)
- same but from upstream to check more gates https://github.com/ComposableFi/composable/pull/3067

# notes

```
if [[ $STATUS_CODE -ne 0 ]] || [[ $STATUS_CODE -ne $STATUS_TRANSIENT_FAILURE ]] ; then
                break
            fi
```

so if status code not 0 or not transient, then break while loop. 

1 != 1 || 1 != 42 = true

### DIFF
--- a/.github/templates/watch-exec/action.yml
+++ b/.github/templates/watch-exec/action.yml
@@ -83,10 +83,11 @@ runs:
             if [[ $STATUS_CODE -eq 0 ]] ; then
               break
             fi
-            
-            if [[ $STATUS_CODE -ne 0 ]] || [[ $STATUS_CODE -ne $STATUS_TRANSIENT_FAILURE ]] ; then
+
+            if [[ $STATUS_CODE -ne 0 ]] && [[ $STATUS_CODE -ne $STATUS_TRANSIENT_FAILURE ]] ; then
                 break
             fi
+            
             ((try=try+1))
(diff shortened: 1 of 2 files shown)


KaiserKarel:
Task linked: [XCVM-288 Fix HTTP error 200 from nix](https://app.clickup.com/t/20465559/XCVM-288)

obsessed-cake (approved):

KaiserKarel:
@dzmitry-lahoda we have a problem where PRs from forks don't run the new workflow yamls. Can you make the base of the PR the monorepo (you have access), so we can test the correctness?

dzmitry-lahoda:
@KaiserKarel https://github.com/ComposableFi/composable/pull/3067

vimukthi-git (approved):
LGTM

dzmitry-lahoda:
as per admin request needed to handle this PR from upstream. so closing from fork.

PR #3075, opened 2023-01-31
ibc cw pallet test #Cu-20465559 #xcvm-277

dzmitry-lahoda:
for https://app.clickup.com/t/20465559/XCVM-277
- was able only do calls up to call back to router, as it goes to emulation of IBC pallet stuff, stopped here - emulation of any kind  could be discussed
- used proper type for return of open call (either () or open ibc result)
- did some chain of calls with events in hook


KaiserKarel:
Task linked: [XCVM-277 Write positive ping pong test](https://app.clickup.com/t/20465559/XCVM-277)

hussein-aitlahcen (approved):

aeryz (approved):

PR #3089, opened 2023-02-01
NX-96 - CI Revamp

cor:
- We're now running all `nix builds` in a single check, `build-packages`, on a new machine, `x64-monster`. This is significantly faster (one order of magnitude) and will get even faster once we implement disk caching.
- We are now also building all packages on `arm` runners, this should make `nix build`s much faster for people who are on `aarch64-linux` (such as M1 users)
- simnode is fully deprecated and deleted
- We've upgraded `nix` to `2.13.2`
- We've upgraded our `nixpkgs-channel` to `nixos-22.11`
- We're no longer using HTTP2 for nix builds, which should resolve the network errors we've been having
- GitHub actions templates have been simplified
- We now build more packages and images
- Docker images have been renamed for consistency, and we push more images
- Tests and packages that are not deterministic are all marked with `[FLAKY]`. Flaky tests are not part of `effects-gate` in order to speed up development
- Fixes incorrect concurrency groups
- Unblocks us from delivering ARM binaries in the releases
- Makes it easy to add more checks by just expanding the checks array

Note that initial arm builds will be slow as we have 0 cache for arm yet. However this cache should be filled automatically now, speeding up future builds


KaiserKarel:
Task linked: [NX-96 `nix build .#everything`](https://app.clickup.com/t/20465559/NX-96)

flake/check.nix
```diff
@@ -1,6 +1,59 @@
...
+          # price-feed # errored in ci
+          # devnet-container # errored in ci
+          # bridge-devnet-dali-container
+          devnet-initialize-script-picasso-persistent
+
+          # not needed as part of effects-gate
+          devnet-dali-complete
+          devnet-picasso-complete
+          devnet-dali
+          # devnet-centauri # disabled because this is not properly nixified
```
dzmitry-lahoda, line 41: please remove all commented lines.

flake/check.nix
```diff
@@ -1,6 +1,59 @@
...
+          cargo-fmt-check
+          cargo-clippy-check
+          cargo-deny-check
+          cargo-udeps-check
+          taplo-check
+          prettier-check
+          nixfmt-check
+          deadnix-check
+          spell-check
+          # docs-static # todo: deployment, temporarily disabled because of LFS
```
dzmitry-lahoda, line 19: please remove commented lines

flake/check.nix
```diff
@@ -1,6 +1,53 @@
 { self, ... }: {
   perSystem = { config, self', inputs', pkgs, system, ... }: {
     packages = {
+      all-ci-packages = pkgs.linkFarmFromDrvs "all-ci-packages"
```
dzmitry-lahoda, line 4: 🧸️https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines 

how happy CI users will find errors why something wrong? logs could be huge.

cor: Excellent idea to group them, not sure how I would cleanly inject it into the build process though. Any suggestions?

dzmitry-lahoda: wrap each package into https://nixos.org/manual/nix/stable/language/builtins.html#builtins-trace

cor: Sounds good, but how well will it work within the actual logs, since it prints output from each builts in parallel

dzmitry-lahoda: that is why i am🧸️.  is it million lines of logs intermixed? how people will use CI to see anything?

dzmitry-lahoda: I am personally OK. I can read any logs. But I have seen that most people do not.

cor: Yeah I understand, I guess we'll just need to teach people to read logs. So far from errors i've encountered it was pretty obvious from which check it was coming, so I think it is possible to teach them.

cor: Else we can look at if it's possible to dump them into a log file, and then group them by built derivation afterwards

dzmitry-lahoda:
@cor Do you have some non chached run job at hand? to look at

.github/workflows/build-test-deploy.yml
```diff
@@ -15,104 +15,29 @@ permissions:
...
   DOCKER_REGISTRY_NAME: composablefi
   # weird, next is implicit dependency of docker action
   DOCKER_USER_OPTION: '$UID:$GID'
 
 jobs:
-  check-privilege:
-    name: "Privilege Check"
+  privilege-check:
+    name: "Privilege check"
     if: ${{ github.event_name == 'push' || (github.event_name == 'pull_request_target' && github.event.pull_request.head.repo.id != 383289760) || (github.event_name == 'pull_request' && github.event.pull_request.head.repo.id == 383289760) }}
```
KaiserKarel, line 30: Maybe while we are cleaning up, move 383289760 to an env variable

.github/workflows/build-test-deploy.yml
```diff
@@ -137,906 +62,417 @@ jobs:
...
+    name: "Nix flake check"
     needs:
-      - check-privilege
-    continue-on-error: false
+      - privilege-check
     runs-on:
       - self-hosted
       - linux
       - x64
       - sre
```
KaiserKarel, line 73: Since this is a lightweight check; we can run this on github infra instead of our own runners.

KaiserKarel: Saves us money + means we need 1 less autoscale.

.github/workflows/build-test-deploy.yml
```diff
@@ -137,906 +62,417 @@ jobs:
...
         with:
+          lfs: true
           ref: ${{ github.event.pull_request.head.sha }}
           persist-credentials: false
-      - uses: "./.github/templates/watch-exec"
+      - name: Nix flake show
+        run: |
+          nix --version
+          nix show-config
+          nix flake show --allow-import-from-derivation  --show-trace --fallback --debug --print-build-logs --keep-failed
```
KaiserKarel, line 155: Since we are cleaning up; could we lessen the logs necessary? I'm not sure if `ACTIONS_STEP_DEBUG` is accessible in the job, but if that is the case, it would be nice to only add `--show-trace --debug --print-build-logs` conditionally. DEBUG mode can be enabled from the UI (Rerun with debug logs).

.github/templates/watch-exec/action.yml
```diff
@@ -5,46 +5,14 @@ inputs:
...
     required: false
-  nix-sandbox: 
-    description: "nix option sandbox value"
-    required: false
-    default: "relaxed"
 
 runs:
   using: "composite"
   steps:
     - run: |
```
KaiserKarel, line 15: See my other comment on conditional debugging. Nice to apply here as well.

.github/workflows/build-test-deploy.yml
```diff
@@ -137,906 +62,417 @@ jobs:
...
-          nix --version
-          nix show-config
-          nix flake show --allow-import-from-derivation  --show-trace --fallback --debug --print-build-logs --keep-failed
-
-  cargo-fmt-check:
-    name: "Cargo fmt check"
+          chmod +x ./.github/workflows/nix-flake-check.sh
+          source ./.github/workflows/nix-flake-check.sh
+  
+  build-packages:
```
KaiserKarel, line 108:
```suggestion
  all-packages:
```

.github/workflows/build-test-deploy.yml
```diff
@@ -137,906 +62,417 @@ jobs:
...
     needs: 
-      - check-privilege
+      - privilege-check
+      - lfs-check
+      - build-packages
     runs-on:
       - self-hosted
       - linux
       - x64
       - sre
```
KaiserKarel, line 389: Since the packages for integration tests will be built already, we might be able to run this on GA runners instead of our own.

.github/workflows/nix-comment.yml
```diff
@@ -0,0 +1,58 @@
...
+            ### View the docs:
+            ```bash
+            nix run ".#docs-server"
+            ```
+            ### Run this without Nix in Docker.
+            ```bash
+            docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v nix:/nix -it nixos/nix bash -c "nix-env -iA nixpkgs.cachix && cachix use composable-community && nix run github:ComposableFi/Composable/${{ github.event.pull_request.head.sha }}#devnet-dali -L --extra-experimental-features nix-command --extra-experimental-features flakes"
+            ```
+            _NOTE: You can swap `devnet-dali` in the command above with any Nix package_
+            _For more info on how to use Nix, check out our [Nix docs](https://docs.composable.finance/nix.html)_
```
KaiserKarel, line 55: Since composable is now in the flake registry; can we simplify the commands?

flake/check.nix
```diff
@@ -1,6 +1,56 @@
...
+          cargo-fmt-check
+          cargo-clippy-check
+          cargo-deny-check
+          cargo-udeps-check
+          taplo-check
+          prettier-check
+          nixfmt-check
+          deadnix-check
+          spell-check
+          docs-static # TODO(cor): deployment
```
KaiserKarel, line 15: Is this todo resolved?

KaiserKarel (changes requested):
I saw in the description that containers are renamed. Are they still pushed under the same name to docker hub so that SRE flows don't break?

.github/workflows/build-test-deploy.yml
cor: Fixed in e2053b0f1feea96a4486328248b1e23594393a0d

.github/workflows/build-test-deploy.yml
cor: Fixed in 308a340e31f8cfed26a7c7a49d0d50bb8ec75979

.github/workflows/build-test-deploy.yml
cor: Not sure how many resources running the actual tests will take, but I'm keen to try. Applied it in 045c82b121051118fbf08af2c28c595ed2132c97

.github/workflows/build-test-deploy.yml
cor: Done

.github/workflows/nix-comment.yml
cor: Actually we only want to use the flake registry entry for commands that run on branch `main`, since almost all PRs will be made from forks. In order to accoomodate forks, I've changed the bot to use `${{ github.event.pull_request.repository}}` instead

cor: We could use the `composable` registry entry for PRs that originate from the monorepo instead of from a fork, but I don't think that's worth it since it should happen very infrequently and it complicates this job

flake/check.nix
cor: Not yet, will add tomorrow/monday

cor:
> I saw in the description that containers are renamed. Are they still pushed under the same name to docker hub so that SRE flows don't break?

@KaiserKarel 
I renamed them and informed everyone in #devs, with regards to SRE flows, only `cmc-api` (previously `cmc-api-server`) needs to be changed which @bostercf is already aware of. As an extra check I'll confirm this monday that the flows are updated prior to merging. If it will take too much time to update the flow then we revert the namechange and merge it without it

flake/check.nix
cor: done

.github/templates/watch-exec/action.yml
cor: Done in 8d92f0c788b8b6a2b9093fdc5188eb2c5448ba9f

.github/workflows/build-test-deploy.yml
cor: Tried to do this in 8d92f0c788b8b6a2b9093fdc5188eb2c5448ba9f, but unsure if it works since the runners are down

cor: Tried to do this in 8d92f0c788b8b6a2b9093fdc5188eb2c5448ba9f, but unsure if it works since the runners are down

dzmitry-lahoda (approved):
ok for me, but

1. it got 2x stronger machine. usually bigger CPU get bigger networks -> less HTTP errors. +++++
2. it got fix of HTTP2 via config (nixos ingores, security?, parameters) -> less HTTP errors +++++
3. caching of deps, so there was alternative ways to do it instead of vertical scalability ++++
4. arm - good. so again. it could be done separate :)  ++++
5. likely other great fixes along the way ++++++


anyway, good it runs and it is overall better. so hope Picasso team approves that they can get use of this CI, and specifically of logs/jobs which lack structure of failure.

i can read such logs, so I am not the only consumer of this.

.github/workflows/build-test-deploy.yml
dzmitry-lahoda: I used it like this 
image

KaiserKarel (approved):

.github/workflows/build-test-deploy.yml
cor: Ah thanks for the exampke, @dzmitry-lahoda !

PoisonPhang (approved):

Dom-Roth (approved):

0xBrainjar2 (approved):

0xBrainjar2 (approved):

0xBrainjar2 (approved):

PR #3116, opened 2023-02-03
Fork-off allow custom rpc URL

vimukthi-git:


composablejs/packages/fork-off/src/index.ts
```diff
@@ -42,7 +42,7 @@ const polkadotPath = path.join(__dirname, "../data", fileNames.polkadot);
 const originalChain = "picasso";
 const forkChain = "picasso-dev";
 const parachainId = 2087;
-const endpoint = "wss://picasso-rpc.composable.finance";
+const endpoint = process.env.PICASSO_WS_RPC || "wss://picasso-rpc.composable.finance";
```
andresmechali, line 45: Should we use `wss://picasso-rpc-lb.composablenodes.tech` as default?

andresmechali (approved):

composablejs/packages/fork-off/src/index.ts
vimukthi-git: This is just a patch we can make modifications to default when we know the confirmed URL

vimukthi-git:
mergify seems to be confused. Merging manually.

PR #3124, opened 2023-02-06
staging persistent devnets runs out of zombienet instead of polkalaunch #CU-3u4b2mw

dzmitry-lahoda:
- replaced all references to Polkalauncher and running Zombienet


KaiserKarel:
Task linked: [CU-3u4b2mw staging persistent devnets runs out of zombienet instead of polkalaunch](https://app.clickup.com/t/3u4b2mw)

PR #3141, opened 2023-02-08
[PBL-394] Query asset prices from Coingecko

andresmechali:
https://app.clickup.com/t/20465559/PBL-394


KaiserKarel:
Task linked: [PBL-394 Query asset prices from Coingekko](https://app.clickup.com/t/20465559/PBL-394)

subsquid/src/server-extension/resolvers/pabloDaily.ts
```diff
@@ -77,13 +81,18 @@ export class PabloDailyResolver implements ResolverInterface<PabloDaily> {
       return acc;
     }, {});
 
-    const totalVolumes = Object.keys(swapsMap).map(
-      assetId =>
+    const totalVolumes: Array<PoolAmount> = [];
+
+    for (const assetId of Object.keys(swapsMap)) {
+      const price = await getOrCreateAssetPrice(manager, assetId, new Date().getTime());
```
fl-y, line 87: I think this should be processed s.t. 
1. each price is checked first
2. Get an array of timestamps that are missing
3. request all of the missing ones at once with [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) and apply the processing mechanism to all of them as they arrive

This will probably be much more efficient code

fl-y: however since we're on the clock let's get this merged and apply this going forward

fl-y (approved):

subsquid/src/utils.ts
```diff
@@ -44,3 +49,43 @@ export function divideBigInts(a: bigint, b: bigint): number {
...
+      coinId = "kusama";
+      break;
+    }
+    default: {
+      throw new Error("Invalid assetId");
+    }
+  }
+
+  const endpoint = `https://api.coingecko.com/api/v3/coins/${coinId}/history?date=${queryDate}&localization=en`;
+  const res = await fetch(endpoint);
```
fl-y, line 83: haven't checked personally but afaik you can probably query multiple dates at once for further optimisation

andresmechali: Yes this can be done but this function should be called once per asset every "day" while processing, so twice every 7,200 / 14,400 blocks. I can work on optimizing how we handle this after release.

subsquid/src/server-extension/resolvers/pabloDaily.ts
fl-y: ah makse sense - looks like I was also referencing the wrong code block lol.
Gotcha sir 🫡

subsquid/src/utils.ts
fl-y: makes sense sir 🫡

PoisonPhang (approved):

PR #3144, opened 2023-02-08
updated starting block for Picasso

andresmechali:


KaiserKarel (approved):

radupopa2010 (approved):

subsquid/src/config.ts
```diff
@@ -27,7 +27,7 @@ export const archive = (): string => {
 
 export const firstBlock = (): number => {
   const relayChain = chain();
-  if (relayChain === "wss://rpc.composablenodes.tech") {
+  if (relayChain.toLowerCase().includes("picasso")) {
```
KaiserKarel, line 30: I think for now this is fine, but why not set firstBlock with an environment variable?

KaiserKarel: With a default if not provided based on this approach?

andresmechali: Makes sense, done.

easteregg (approved):

PR #3148, opened 2023-02-09
XCVM-255: Pallet-Cosmwasm Tests

aeryz:
## Introduction

Several must-have tests were added which tests functionality of host functions, extrinsics and submessage handling mechanism. There should definitely more tests related to gas metering, IBC, smaller unit tests etc. but I think this is good for now. At least we can test most of the scenarios and fixed several severe bugs before the audit.

## Added tests
* `host_functions`: Tests the host function implementations of the VM as well as the gas usage.
* `extrinsics`: Tests all extrinsics.
* `submessages`: Tests submessage execution flow by creating a contract which returns the desired submessage from the `instantiate` function. This way, we can test the full execution flow of a submessage.

## Fixes found and applied during tests:
* When handling `Migrate`, `UpdateAdmin` and `ClearAdmin` submessages, we were considering the caller as the parent's caller, but it should be the contract who returned the submessage.
* `db_scan/db_next` mechanism was completely broken, `db_next` was always returning `None`.
* If the migrate extrinsic is called and `code_id` is the same as the `new_code_id`, `migrate` entrypoint of the contract was not getting called.
* `set_contract_meta` was charging a wrong amount of gas.
* `db_write` was charging a wrong amount of gas if a value is overwritten.
* Address was generated incorrectly in the `instantiate` extrinsic.

## Missing tests:
* `gas` metering related tests.
* IBC tests.
* Some unit tests.


KaiserKarel:
Task linked: [XCVM-255 Tests before audit](https://app.clickup.com/t/20465559/XCVM-255)

hussein-aitlahcen (approved):

hussein-aitlahcen (approved):

dzmitry-lahoda (approved):

PR #3157, opened 2023-02-09
subsquid - deployment - automate the steps required to trigger  re-processing

radupopa2010:


KaiserKarel:
Task linked: [PBL-386 subsquid - deployment - automate the steps required to trigger  re-processing](https://app.clickup.com/t/20465559/PBL-386)

subsquid/scripts/init.sh
```diff
@@ -1,7 +1,44 @@
 #!/bin/sh
 
+set -eou pipefail
+
+# main entry point for the script
+main() {
+  #  ${foo:="default value if unset"}
+  if ${TRIGGER_REPROCESSING:=false} ; then
```
andresmechali, line 8: Shouldn't this be `true`?

subsquid/scripts/init.sh
```diff
@@ -1,7 +1,44 @@
...
+  node -r dotenv/config lib/processor.js || log DIE "Starting the processor step failed"
+}
+
+run_steps_to_trigger_reprocessing() {
+  log INFO "Reprocessing triggered"
+  set_db_connection_details
+  psql -c "DROP SCHEMA public CASCADE ;
+           DROP SCHEMA squid_processor cascade ;
+           create SCHEMA public  ;
+           create schema squid_processor ;"
```
andresmechali, line 29: Not sure if required but are we missing the grants?

subsquid/scripts/init.sh
radupopa2010: please see the comment from above:
`  #  ${foo:="default value if unset"}`   this means set a default value if this env var is unset
if `TRIGGER_REPROCESSING` is set, then the set value if used

subsquid/scripts/init.sh
radupopa2010: I tested on stating and the grants are not needed ; for prod I will need add them, but it's just a one time thing because the `postgres` user was used

subsquid/scripts/init.sh
andresmechali: Perfect!

andresmechali (approved):

PoisonPhang (approved):

PR #3166, opened 2023-02-13
chore(cosmwasm): upgrade wasmi

aeryz:
- Upgraded `cosmwasm-vm` to use the latest wasmi.
- Updated benchmarks.


blasrodri (approved):

code/parachain/frame/cosmwasm/src/weights.rs
```diff
@@ -1,23 +1,22 @@
...
 // --chain=dali-dev
 // --execution=wasm
 // --wasm-execution=compiled
-// --wasm-instantiation-strategy=legacy-instance-reuse
 // --pallet=cosmwasm
 // --extrinsic=*
 // --steps=50
 // --repeat=10
-// --output
+// --output=
```
blasrodri, line 19: is this `=` needed?

hussein-aitlahcen (approved):

code/parachain/frame/cosmwasm/src/weights.rs
aeryz: don't think so but its auto-generated

PR #3199, opened 2023-02-16
ci: change subsquid approval to 1

fl-y:
PR to reduce required approvals to 1 from blockchain integrations

- Subsquid PRs take longer due to the requirement to have 2+ approvals and not enough incentives for other developers to review.
- Also subsquid gets tested vigorously by QA so one approval should be enough to get the tests rolling out

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -36,3 +36,18 @@ pull_request_rules:
       queue:
         priority: high
         name: default
+  # Reduce reviewers for subsquid since QA tests it vigorously anyway
+  - name: Automatic merge on approval (Subsquid)
+    conditions:
+      - and:
+          - "#approved-reviews-by>=1"
+          - approved-reviews-by=@ComposableFi/blockchain-integrations
+          - "#review-threads-unresolved=0"
+      - base=main
+      - files~=^subsquid\/
+      - branch-protection-review-decision=APPROVED
+
+    actions:
+      queue:
+        priority: high
+        name: default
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -31,8 +31,21 @@ pull_request_rules:
       - base=main
       - files~=^code\/parachain\/runtime\/picasso\/
       - branch-protection-review-decision=APPROVED
-
     actions:
       queue:
         priority: high
+        name: default
+
+  # Reduce reviewers for subsquid since QA tests it vigorously anyway
```
vimukthi-git, line 39: I don't entirely agree with this statement. Code review is not just to catch bugs. It involves many other aspects such as design review for new features as well.

vimukthi-git (approved):
LGTM, though I don't see the problem with current review process. Could you point to some examples how long it took to review?

.mergify.yml
fl-y: I agree with your statement however not with the reality of the validity of code reviews atm.
I spent multiple hours 2 days ago trying to figure out bugs on production and I've never seen any substantial changes suggested to subsquid related PRs so far.
If you believe there are aspects that trump whatever other priority you have to review, please go ahead, requesting changes should block the PR from being merged

KaiserKarel (approved):

.mergify.yml
vimukthi-git: >I spent multiple hours 2 days ago trying to figure out bugs on production and I've never seen any substantial changes suggested to subsquid related PRs so far.

This is not enough sample size to evaluate the review system for Subsquid I can tell you that.

josietyleung:
@fl-y is this still needed?

dzmitry-lahoda:
it is globally 1 now.

PR #3209, opened 2023-02-17
[PBL-508] fix lockable stake shares

kkast:
fix lockable stake shares https://app.clickup.com/t/20465559/PBL-508
test for https://app.clickup.com/t/20465559/PBL-507
added test ED to be 0 for XPICA

### DIFF
--- a/code/parachain/frame/staking-rewards/src/lib.rs
+++ b/code/parachain/frame/staking-rewards/src/lib.rs
@@ -66,11 +66,12 @@ use core::{
 };
 use frame_support::{
 	traits::{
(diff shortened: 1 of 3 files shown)


KaiserKarel:
Task linked: [PBL-508 Staking Rewards: After a user unstakes, if another user stakes to the pool with an amount lower than previously returned share tokens, the share tokens are not locked in proxy wallet](https://app.clickup.com/t/20465559/PBL-508)

vimukthi-git (approved):

dzmitry-lahoda:
ping

PR #3211, opened 2023-02-20
Upgrading version of dependencies of Dotsama repos #NX-106 #CU-20465559

dzmitry-lahoda:
for https://app.clickup.com/t/20465559/NX-106
- updated all dotsama 3rd party inputs version
- added more 3rd party inputs to all total build
- used fetchgit as it is fixed (can use that in nix configs, unlike fetchFromGithub)
- allow to see all errors without running many times


obsessed-cake (approved):

Dom-Roth (approved):
LGTM!

vimukthi-git (approved):

PR #3232, opened 2023-02-27
upgrade of centauri pallet and set IBC gates #CU-8669mcw9g

dzmitry-lahoda:
For https://app.clickup.com/t/8669mcw9g

Upgraded Centauri to new version. 

Setup release gates for IBC pallet.

### DIFF
--- a/code/parachain/runtime/picasso/src/ibc.rs
+++ b/code/parachain/runtime/picasso/src/ibc.rs
@@ -2,6 +2,7 @@ use ::ibc::core::{
 	ics24_host::identifier::PortId,
 	ics26_routing::context::{Module, ModuleId},
 };
+use frame_support::traits::EitherOf;
 use pallet_ibc::{
 	light_client_common::RelayChain, routing::ModuleRouter, DenomToAssetId, IbcAssetIds, IbcAssets,
 	IbcDenoms,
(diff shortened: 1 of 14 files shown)


obsessed-cake (approved):

code/parachain/runtime/picasso/src/ibc.rs
```diff
@@ -104,6 +105,28 @@ impl core::str::FromStr for MemoMessage {
...
+use pallet_ibc::ics20::Ics20RateLimiter;
+
+pub struct ConstantAny;
+
+impl Ics20RateLimiter for ConstantAny {
+	fn allow(
+		msg: &pallet_ibc::ics20::Ics20TransferMsg,
+		_flow_type: pallet_ibc::ics20::FlowType,
+	) -> Result<(), ()> {
+		// one DOT/PICA, so so for USDT not safe, but we do not yet do it
```
blasrodri, line 122:
```suggestion
		// one DOT/PICA, so for USDT not safe, but we do not yet do it
```

code/parachain/runtime/picasso/src/ibc.rs
```diff
@@ -104,6 +105,28 @@ impl core::str::FromStr for MemoMessage {
...
+		if msg.token.amount.as_u256() <= ::ibc::bigint::U256::from(10_u64.pow(12)) {
```
blasrodri, line 123: where's this limit coming from?

dzmitry-lahoda: 1 DOT/PICA

dzmitry-lahoda: 12 decimals.

sure, will need to implement real stuff when have some full design. or split constant per channel (needs decimals, better not to do until assets merged, so in theory can do).

dzmitry-lahoda: we can change any value. so would prefer we have some RL pallet.

blasrodri (approved):

blasrodri (approved):

blasrodri:
great job!!

PR #3235, opened 2023-03-01
composable runtime ibc config and governance #CEN-117

dzmitry-lahoda:
for https://app.clickup.com/t/20465559/CEN-117

I have to setup all gates and governance on Composable same as on Picasso. Security increased.

Setup hardcoded assets exactly as they should be when not hardcoded for DOT KSM transfer. So we can run without any ceremonies.

And finally IBC enabled in Composable.


code/Cargo.toml
```diff
@@ -93,8 +95,8 @@ frame-system = { git = "https://github.com/ComposableFi/substrate", branch = "po
 frame-system-benchmarking = { git = "https://github.com/ComposableFi/substrate", branch = "polkadot-v0.9.36" }
 frame-system-rpc-runtime-api = { git = "https://github.com/ComposableFi/substrate", branch = "polkadot-v0.9.36" }
 frame-try-runtime = { git = "https://github.com/ComposableFi/substrate", branch = "polkadot-v0.9.36" }
-pallet-assets = { git = "https://github.com/ComposableFi/substrate", branch = "polkadot-v0.9.36" }
-pallet-asset-tx-payment = { git = "https://github.com/ComposableFi/substrate", branch = "polkadot-v0.9.36" }
+# pallet-assets = { git = "https://github.com/ComposableFi/substrate", branch = "polkadot-v0.9.36" }
```
blasrodri, line 98: how come is this being commented?

code/parachain/runtime/composable/src/gates.rs
```diff
@@ -0,0 +1,114 @@
...
+
+impl InstanceFilter<RuntimeCall> for ProxyType {
+	fn filter(&self, c: &RuntimeCall) -> bool {
+		match self {
+			ProxyType::Any => true,
+			ProxyType::Governance => matches!(
+				c,
+				RuntimeCall::Democracy(..) |
+					RuntimeCall::Council(..) |
+					//RuntimeCall::TechnicalCommittee(..) |
```
blasrodri, line 22: why is the technical committee commented?

code/Cargo.toml
dzmitry-lahoda: i have remomoved unused dependency. ci linter failed with patching unused. i have commemted it out so when it will become used(when xcm tests enabled back).

code/parachain/runtime/composable/src/gates.rs
```diff
@@ -0,0 +1,114 @@
...
+					RuntimeCall::XcmpQueue(..) |
+					RuntimeCall::RelayerXcm(..)
+			),
+		}
+	}
+	fn is_superset(&self, o: &Self) -> bool {
+		match (self, o) {
+			(x, y) if x == y => true,
+			(ProxyType::Any, _) => true,
+			(_, ProxyType::Any) => false,
```
blasrodri, line 44:
```suggestion
```

dzmitry-lahoda: it is same in picasso. i did not changed semantics copying code from there. so will not.fix in this pr even if it is not right.

code/parachain/runtime/composable/src/gates.rs
dzmitry-lahoda: will fix

code/parachain/runtime/composable/src/gates.rs
dzmitry-lahoda: so imho we should recheck these value. so likely will do in other pr. for now that is 1 to 1 with picasso.

blasrodri (approved):

obsessed-cake (approved):

PR #3241, opened 2023-03-02
[PICA-234] Addition of chainspec for Picasso Rococo

andyjsbell:
Chainspec plain and raw for Picasso on Rococo with the following addresses:
sudo : `5EPf2srNf9AKWF84qrE6HyFLMPPtSV79gNtijfb2fuEPNkjF`
collators : `5EZDCyHLbXqPQyEEVetAVSUEjR5pr8wGQC8UyHGhBJaVTG5Z` `5F48mqyrRfif4DS7wdtZ7ratVXCdE1sSmszSPYAu549c59Wz` 

sudo is minted with 10 MPICA


KaiserKarel:
Task linked: [PICA-234 Generate chainspec and genesis artefacts](https://app.clickup.com/t/20465559/PICA-234)

dzmitry-lahoda:
Consider generating genesis on latest main.  I may fail to run because lack of some genesis config data from my recent merge. At least human readable version will not run. As for human readable version, thanks for sharing. But seem nobody pushes it into git on our repo and in partity (because it becames useless after soon). 

Also not sure if that is scope of task or not. To run collator, need to add that with other chains. Search how other files load via cmd switch. This would be `picasso-rococo` (like parity does).

vimukthi-git (approved):

PR #3246, opened 2023-03-03
added way to test upgrades on new data

dzmitry-lahoda:


obsessed-cake (approved):

josietyleung:
@dzmitry-lahoda what do we do with this?

PR #3247, opened 2023-03-03
[ SRE-587 ] Use RPC nodes as boot nodes for Composable -> main

bostercf:
Use RPC nodes as boot nodes for Composable


KaiserKarel:
Task linked: [SRE-587 See if using hosts file entries and firewall rules a node can be brought up without the listed bootnode](https://app.clickup.com/t/20465559/SRE-587)

obsessed-cake (approved):

PR #3257, opened 2023-03-07
Publish `hyperspace-picasso` and `hyperspace-composable` docker images

obsessed-cake:


KaiserKarel:
Task linked: [NX-108 Publish hyperspace-picasso and hyperspace-composable docker images](https://app.clickup.com/t/20465559/NX-108)

dzmitry-lahoda (commented):
can you reuse attributes via // merge. really easy

.github/workflows/build-test-deploy.yml
```diff
@@ -293,7 +293,22 @@ jobs:
           password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
           name: hyperspace-dali
           artifact: hyperspace-dali:latest
-      
+      - name: Publish hyperspace-picasso to docker hub
```
dzmitry-lahoda, line 296: may this work with gh job matrix?

blasrodri (commented):
Don’t you need to do this in Centauri? So that it compiles with different targets and builds the image right after?

.github/workflows/build-test-deploy.yml
obsessed-cake: yes, this may
but matrix strategy work on level `jobs.<job_id>`.
so CI will run the [job](https://github.com/ComposableFi/composable/blob/main/.github/workflows/build-test-deploy.yml#L219-L295) wholly for every runtime, i.e. 3 times.

obsessed-cake:
> Don’t you need to do this in Centauri? So that it compiles with different targets and builds the image right after?

We [pull cetauri repo](https://github.com/ComposableFi/composable/pull/3257/files#diff-951363b79004cf0a828b185b1b7e750db3b65d182ef7be83b95127f455e3b880L4-L9) to build `subxt-client` and `hyperspace`.
We use different feature flags to build `hyperspace` for different runtime (`--features dali` for dali).

.github/workflows/build-test-deploy.yml
obsessed-cake: I moved a building of Hyperspace docker image to a separate CI job.

PR #3260, opened 2023-03-08
vesting cli

dzmitry-lahoda:
Vesting script.

- First one was not finished nor merged
- Finished is for crowdloans, not vesting
- Polkadotjs not working (reported bug)

Did Collator vesting ok with it. And now Oracle


andresmechali (approved):

andyjsbell (approved):
Thanks @dzmitry-lahoda 

I know we are squashing commits before a merge but it could be helpful to have commits messages that make sense or are even squashed themselves for the PR 🙏

PR #3278, opened 2023-03-14
using faster http2 for nix cache as it was fixed and autoconfig cache for flake

dzmitry-lahoda:
removed obsolete lines from nix doc


andresmechali (approved):
🚀

obsessed-cake (approved):

PR #3284, opened 2023-03-15
change hasher for ed

kkast:
This hasher will allow us to retrieve asset ids that are in storage.


kkast:
@vimukthi-git should we change hashers for other assets-registry storage maps? rn they are using twox128 which wont expose asset ids of the assets that have AssetRatio set or the asset ids that have foreign location

PR #3326, opened 2023-03-28
fix and clean for shells

dzmitry-lahoda:
- used same base shell for all shells
- removed this repo packages from shells (we have cache autosetup now, so no need to bake it on start)
- fixed ROCKDB lib issue in default shell


andresmechali (approved):

kkast (approved):
LGTM

PR #3342, opened 2023-03-30
fix benchmark for asset registry

andyjsbell:
Benchmark is expecting `Default` for `ForeignAssetId` which isn't implemented.


josietyleung:
@kkast is this done thanks?

PR #3343, opened 2023-03-31
update paths for oracle setup instructions

andresmechali:
Some updates to the Oracle setup guide.


dzmitry-lahoda:
@kkast @bostercf

PR #3345, opened 2023-03-31
enable BYOG in Composable

dzmitry-lahoda:
- refactored fees and version into separate files,
- so it is easy to reuse and configure
- and BYOG enabled Composable runtime
- !!! BYOG and payments parameters are same as before. In Composable same as on Picasso


code/parachain/runtime/composable/src/lib.rs
```diff
@@ -665,6 +607,7 @@ construct_runtime!(
 		Sudo: sudo = 2,
 		RandomnessCollectiveFlip: randomness_collective_flip = 3,
 		TransactionPayment: transaction_payment = 4,
+		AssetTxPayment : asset_tx_payment  = 12,
```
blasrodri, line 610: Nit, but why not put it below if they are ordered?

blasrodri (approved):

kkast (approved):
changes look good

vimukthi-git (approved):

PR #3360, opened 2023-04-06
[PICA-319] fixed rewards based on staked amount

kkast:
fixed rewards based on staked amount + added test for rewards + fixed previously ignored tests

### DIFF
--- a/code/parachain/frame/oracle/src/lib.rs
+++ b/code/parachain/frame/oracle/src/lib.rs
@@ -836,26 +836,38 @@ pub mod pallet {
 			}
 			if let Some(mut reward_tracker) = Self::get_reward_tracker_if_enabled() {
 				// divide the per asset reward(by weight) by the number of oracles
-				let reward_amount_per_oracle: T::Balance = safe_multiply_by_rational(
+				let reward_amount_per_asset: T::Balance = safe_multiply_by_rational(
 					reward_tracker.current_block_reward.unique_saturated_into(),
 					asset_info.reward_weight.unique_saturated_into(),
 					reward_tracker.total_reward_weight.unique_saturated_into(),
 				)?
-				.checked_div(rewarded_oracles.len() as u128)
-				.ok_or(ArithmeticError::DivisionByZero)?
 				.into();
-				if !reward_amount_per_oracle.is_zero() {
-					for accounts in rewarded_oracles {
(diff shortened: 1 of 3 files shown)


KaiserKarel:
Task linked: [PICA-319 Update rewards logic](https://app.clickup.com/t/20465559/PICA-319)

andresmechali (approved):
LGTM

code/parachain/frame/oracle/src/lib.rs
```diff
@@ -836,26 +836,38 @@ pub mod pallet {
...
-							reward_amount_per_oracle,
-						)?;
-						// track the total being rewarded
-						reward_tracker.total_already_rewarded = reward_tracker
-							.total_already_rewarded
-							.saturating_add(reward_amount_per_oracle);
+				if !reward_amount_per_asset.is_zero() {
+					let rewarded_oracles_total_staked: T::Balance = rewarded_oracles
+						.iter()
+						.map(|account| OracleStake::<T>::get(&account.0).unwrap_or_default())
```
blasrodri, line 848: why is getting the default reasonable? can we add a comment?

code/parachain/frame/oracle/src/tests.rs
```diff
@@ -277,6 +280,10 @@ mod add_asset_and_info {
 				// and therefore this test should also be changed.
 				prop_assert_eq!(MaxAssetsCount::get(), 2_u32);
 
+
+				asset_info_1.reward_weight = asset_info_1.reward_weight.min(u128::MAX / 3);
```
blasrodri, line 284: can we test with different weights?

kkast: this is proptest, waits for 3 will be all different, i divided by 3 to make sure that they wont overflow. i can change it so that one is 1/12 of the max, another 1/6th of  the max , and third 3/4 of the max

code/parachain/frame/oracle/src/lib.rs
kkast: to get paid reward, user needs to submit price, to submit price user needs to have a stake.
the user can potentially submit the price and then remove stake in one block, and then rewards are distributed on_initialize of the next block. so in this case we should take 0 as its stake

PR #3369, opened 2023-04-08
picasso-rococo spec in node and removed some dali

dzmitry-lahoda:


blasrodri (approved):
Did you add the genesis?

obsessed-cake (approved):

PR #3375, opened 2023-04-10
fixes for nix version in codespace

dzmitry-lahoda:


fl-y (approved):
Have no idea what I'm looking at but approving to speed the process up.

.github/workflows/release-artifacts.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     name: "Build runtime"
     runs-on:
       - self-hosted
-      - x64-cute
+      - x64-monster
```
blasrodri, line 20: Why the change?

blasrodri (approved):

PR #3409, opened 2023-04-13
testfast

dzmitry-lahoda:
release fast test runtime and node. 

can normal node run testfast runtime ? YES. So need to use testfast to export chainspec. and run it with release node.

can testfast node run release runtimes? YES.

can testfast runtime be upgraded to release and back? YES

what is testfast = test is that Origin are relaxed , fast means configs for less time to wait (future will do 4 second block time instead of 12)

### DIFF
--- a/code/parachain/runtime/picasso/src/ibc.rs
+++ b/code/parachain/runtime/picasso/src/ibc.rs
@@ -142,19 +142,14 @@ impl pallet_ibc::Config for Runtime {
 	type IsReceiveEnabled = ConstBool<true>;
 	type IsSendEnabled = ConstBool<true>;
 
-	#[cfg(feature = "testnet")]
 	type AdminOrigin = EnsureRootOrOneThirdNativeTechnical;
-	#[cfg(feature = "testnet")]
 	type FreezeOrigin = EnsureRootOrOneThirdNativeTechnical;
+
 	#[cfg(feature = "testnet")]
 	type TransferOrigin = system::EnsureSigned<Self::IbcAccountId>;
 	#[cfg(feature = "testnet")]
 	type RelayerOrigin = system::EnsureSigned<Self::IbcAccountId>;
 
-	#[cfg(not(feature = "testnet"))]
-	type AdminOrigin = EnsureRootOrOneThirdNativeTechnical;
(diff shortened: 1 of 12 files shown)


obsessed-cake (dismissed):

code/parachain/runtime/picasso/src/ibc.rs
```diff
@@ -134,18 +134,32 @@ impl pallet_ibc::Config for Runtime {
...
+	type AdminOrigin = EnsureRootOrOneThirdNativeTechnical;
+	#[cfg(feature = "testnet")]
+	type FreezeOrigin = EnsureRootOrOneThirdNativeTechnical;
+	#[cfg(feature = "testnet")]
+	type TransferOrigin = system::EnsureSigned<Self::IbcAccountId>;
+	#[cfg(feature = "testnet")]
+	type RelayerOrigin = system::EnsureSigned<Self::IbcAccountId>;
+
+	#[cfg(not(feature = "testnet"))]
+	type AdminOrigin = EnsureRootOrOneThirdNativeTechnical;
```
andresmechali, line 155: This should be 2/3 no?

code/parachain/runtime/picasso/src/ibc.rs
```diff
@@ -134,18 +134,32 @@ impl pallet_ibc::Config for Runtime {
...
+	#[cfg(not(feature = "testnet"))]
+	type FreezeOrigin = EnsureRootOrOneThirdNativeTechnical;
```
andresmechali, line 157: Also this

andresmechali (approved):

code/parachain/runtime/picasso/src/ibc.rs
kkast: agreed

kkast (changes requested):

code/parachain/runtime/picasso/src/ibc.rs
kkast: seems like it is correct, but if with or without testfast feature has the the same AdminOrigin and FreezeOrigin, maybe we shouldnt put them separately

kkast (approved):

kkast (approved):

PR #3412, opened 2023-04-14
Update integration-tests

obsessed-cake:


KaiserKarel:
Task linked: [PICA-324 BUG - Malicious Oracle can manually submit price to change the price of asset on chain and not gets slashed](https://app.clickup.com/t/20465559/PICA-324)

PR #3425, opened 2023-04-17
all relay runtimes as deployed

dzmitry-lahoda:
need to generated subxt client per relay we are on too to build and run hyperspace. so added all runtimes to be awailable into build scripts.


blasrodri:
Nice! Could you also add `--pruning=archive` to both the {Relay, Para} chain for the dev net?

PR #3444, opened 2023-04-19
prelease clean up (pr for commit from which would do release)  #misc

dzmitry-lahoda:
- killed spam bot on pr
- update release version
- update release docs
- speed up some checks to start (deps review is zero second start instead of 3 minutes)
- bumped centauri version


kkast (approved):
lgtm

Issue #3386, opened 2023-04-11
fix list_assets rpc


itsbobbyzz:
Currently list_assets rpc still uses hardcoded list of CurrencyIds, 

- [ ] need to use keys from ExistentialDeposit storage map of AssetsRegistry: assets_registry::ExistentialDeposit::<Runtime>::iter_keys().collect::<Vec<CurrencyId>>()
- [ ] include all assets in the asset-registry

dzmitry-lahoda:
was it done?

dzmitry-lahoda:
assuming fixed or not used by FE

PR #3447, opened 2023-04-19
Include local assets in the RPC method `assets_listAssets` #3386

obsessed-cake:
See https://github.com/ComposableFi/composable/issues/3386 for details.


kkast (dismissed):
Good, but we need to rebase

obsessed-cake:
> Good, but we need to rebase

I did this.

PR #3458, opened 2023-04-20
hyperspace bridge imagination for prod

dzmitry-lahoda:
- image of prod hyper space in nix
- pruning archieve in devnet
- ksm in devnet in genesis
- moved some nix checks to nix to easy run during dev


blasrodri:
@dzmitry-lahoda can we add into devnet? Since we are stress testing, sometimes it breaks as seems like the requests get too large for the limit.
```
  --rpc-max-request-size=5
#  <size in MB> 
```

blasrodri (approved):

obsessed-cake (approved):

PR #3466, opened 2023-04-21
pallet-ibc: add multisig for fee accounts on Composable and Picasso

blasrodri:
In order to collect the fees charged by Centauri, we've set up two multisigs. 
This PR adds the functionality (paired with https://github.com/ComposableFi/centauri/pull/287) to enact this.

PR #3470, opened 2023-04-21
pre release 6.10017.0

dzmitry-lahoda:
- instead of ParaId(u32) used u32 directly as ParaId reexported by cumulus and polkadot and leads to weird conflicts (and XCM uses u32 directly anyway)
- fork of substrate with 2 wasm fixes (seems one is next release of parity)
- split out jobs only for hyperspace prod clients and run in ci (and some tiny improvements in ci to reduce fails)
- automation for release tags
- mac is not autoscaled, so run it only on merge to main or from repo prs, not externals


blasrodri:
We need to merge https://github.com/ComposableFi/composable/pull/3466

code/parachain/frame/assets-registry/src/lib.rs
```diff
@@ -117,7 +116,7 @@ pub mod pallet {
 	pub type MinFeeAmounts<T: Config> = StorageDoubleMap<
 		_,
 		Blake2_128Concat,
-		ParaId,
+		u32,
```
blasrodri, line 119: why this change?

kkast (approved):

PR #3552, opened 2023-05-03
Integrate Flat Fee into picasso runtime.

RustNinja:
This PR related to https://github.com/ComposableFi/centauri/issues/253

- add dependency to `pallet-ibc` into `pallet-pablo` to use `FlatFeeConvertor` trait
- reconfigure `pallet_ibc::ics20_fee::Config` for picasso and composable runtime
- implement `FlatFeeConvertor` for `pallet-pablo` and use it in `picasso` runtime
- use `NonFlatFeeConverter` from centauri repo that implement FlatFeeConvertor and always return None that means not a flat fee but use percentage.

Require runtime upgrade for picasso-testnet to test flat fee functionality there.

Notes. For now this changes related only for integration of flat fee to picasso and does not contains functionality for whitlist of channels that not supposed to charge any fee in ibc-pallet.
please give your opinion about it.

@blasrodri

TODO: 
- add RPC endpoints that will return charging flat fee amount for Frontend.

### DIFF
--- a/code/parachain/frame/pablo/Cargo.toml
+++ b/code/parachain/frame/pablo/Cargo.toml
@@ -33,6 +33,8 @@ sp-io = { default-features = false, workspace = true }
 sp-runtime = { default-features = false, workspace = true }
 sp-std = { default-features = false, workspace = true }
 
+pallet-ibc = { workspace = true, default-features = false }
+
 [dev-dependencies]
 composable-tests-helpers = { path = "../composable-tests-helpers" }
 frame-benchmarking = { default-features = false, workspace = true }
@@ -78,4 +80,5 @@ std = [
   "sp-io/std",
   "sp-runtime/std",
   "sp-std/std",
+  "pallet-ibc/std"
 ]
--- a/code/parachain/frame/pablo/src/lib.rs
+++ b/code/parachain/frame/pablo/src/lib.rs
@@ -65,7 +65,7 @@ pub mod pallet {
 	use composable_traits::{
 		assets::CreateAsset,
 		defi::{CurrencyPair, Rate},
-		dex::{Amm, BasicPoolInfo, Fee, FlatFeeConverter, PriceAggregate},
+		dex::{Amm, BasicPoolInfo, Fee, PriceAggregate},
 	};
 	use core::fmt::Debug;
 	use frame_support::{
@@ -77,6 +77,7 @@ pub mod pallet {
 		},
 		transactional, BoundedBTreeMap, PalletId, RuntimeDebug,
 	};
+	use pallet_ibc::ics20_fee::FlatFeeConverter;
 	use sp_arithmetic::FixedPointOperand;
 
 	use composable_maths::dex::{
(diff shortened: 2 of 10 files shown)


code/parachain/frame/pablo/Cargo.toml
```diff
@@ -33,6 +33,8 @@ sp-io = { default-features = false, workspace = true }
 sp-runtime = { default-features = false, workspace = true }
 sp-std = { default-features = false, workspace = true }
 
+pallet-ibc = { workspace = true, default-features = false }
```
blasrodri, line 36: Pallet ibc should be there already as ibc?

code/parachain/frame/pablo/src/lib.rs
```diff
@@ -252,7 +253,7 @@ pub mod pallet {
 			+ Ord;
 
 		/// Type representing the Balance of an account.
-		type Balance: BalanceLike + SafeSub + Zero + FixedPointOperand;
+		type Balance: BalanceLike + SafeSub + Zero + FixedPointOperand + Into<u128>;
```
blasrodri, line 256: I originally envisioned importing Pablo from pallet ibc. Or this can be done totally generic and wired up at the config level of the runtime instead

code/parachain/runtime/composable/src/lib.rs
```diff
@@ -603,6 +603,8 @@ parameter_types! {
 	pub const VaultMinimumDeposit: Balance = 10_000;
 	pub const VaultMinimumWithdrawal: Balance = 10_000;
 	pub const VaultPalletId: PalletId = PalletId(*b"cubic___");
+	pub AssetIdUSDT: CurrencyId = CurrencyId::INVALID;
+	pub FlatFeeUSDTAmount: Balance = 0 * CurrencyId::unit::<Balance>(); //None
```
blasrodri, line 607: Here we need to whitelist so that we don’t double charge except that the channel isn’t part of the list

kkast: I think this is for the next task. Here we don't double charge. We charge only on receive always

code/parachain/frame/pablo/src/lib.rs
kkast: @RustNinja this is why I added to composbale-traits package fee converter trait

blasrodri: Yeah this is something we discussed and I probably gave him the wrong directions

kkast: Tbh it also makes sense to define that trait in Centauri because it is where it is used

RustNinja: one more argument to declare FlatFeeConvertor trait to centauri is that

composable-trait uses some features from nightly channel.
i faced with  a lot of errors that it not possible anymore to build/test centauri palelt-ipc without +nightly channel if we will bring composable trait to centauri.
of course we can add toolchain channel
`channel = "nightly-2022-11-17"` 
it will works . but anyway it is one more dependency...

blasrodri: Cool then let’s keep it in Centauri

kkast (dismissed):

kkast (approved):

blasrodri (approved):

PR #3559, opened 2023-05-04
Update to polkadot-v0.9.39 #3523

obsessed-cake:
- [ ] PR title is my best effort to provide summary of changes and has clear text to be part of release notes 
- [ ] I marked PR by `misc` label if it should not be in release notes
- [ ] I have linked Zenhub/Github issue if one exists
- [ ] I was clear on what type of deployment required to release my changes (node, runtime, contract, indexer, on chain operation, frontend, infrastructure) if any in PR title or description
- [ ] I waited and did best effort for `Effect gate, automatically merged if passed`(effects-job) to finish with success(green check mark) with my changes
- [ ] I have added at least one reviewer in reviewers list
- [ ] I tagged(@) or used other form of notification of one person who I think can handle best review of this PR

PR #3579, opened 2023-05-09
Update to polkadot-v0.9.39

vmarkushin:
There were a few new associated types introduced in the frame pallets. I set them to the values I think should fit the runtimes, but feel free to change them anyway.

P.S. I can't push to the initial [PR](https://github.com/ComposableFi/composable/pull/3559), so I've opened this one.


blasrodri (approved):

dzmitry-lahoda:
cc @obsessed-cake

PR #3602, opened 2023-05-15
satemine and statemint in devnets for usdt, and improvement on some `look and feel` of CI (really no changes, just more green look)

dzmitry-lahoda:
- updated parity chains to latest version
- added to devnets by default
- clean up of some dead nix
- tuned jobs to be more green (hack around cancel = grey, not read in GH actions) - so that can optimize away more jobs
- do not publish docs from forks (as it fails with token issue)


.github/workflows/pr-workflow-check.yaml
```diff
@@ -20,6 +20,13 @@ jobs:
       github_event_pull_request_head_repo_id : ${{ github.event.pull_request.head.repo.id }}
       github_workflow: ${{ github.workflow }}
       github_event_pull_request_head_sha: ${{ github.event.pull_request.head.sha }}
+      flow: |
+        ${{ 
+          (github.event_name == 'push' && 'push') ||
+          (inputs.github_event_name == 'merge_group' && 'merge_queue_check') ||
+          (inputs.github_event_name == 'pull_request_target' && inputs.github_event_pull_request_head_repo_id != 383289760 && 'pr_from_fork') ||
```
blasrodri, line 27: What’s this id?

blasrodri (dismissed):

.github/workflows/pr-workflow-check.yaml
dzmitry-lahoda: this is hack around GH. GH runs fork and branch PR flows at same time and also it does not allows to make green job if want to skip its deps (so skip steps work well, but need shortcut for it).  fixing GH action (likely reason people migrate from GH instead of fix btw).

dzmitry-lahoda:
@fl-y I ignored docs from forks. so it should be green. but it not part of gate anyway. just noise.

dzmitry-lahoda:
If anybody interested I can show exactly why GH action suck, and why there is suck workaround :)

kollegian (dismissed):

dzmitry-lahoda:
also need to test devnet on darwin after ZN update.

PR #3633, opened 2023-05-21
5 times cheaper and 3 times faster ci

dzmitry-lahoda:
Current CI machines suck, they slow and we overpay 2x more minutes(they case when faster machines when per minute would be cheaper). And developers have to wait.

But more important, /nix/store is not cached among runs, which leads to increase of time thing run. 
GH runners are stateless when autoscaled out of box. Making store cache is hard.
Using /nix/store cache would make it way more faster too. See ultra fast M1 job.

Solution should not be complicated, as there are alternative solution which solve all of it fundamentally, likely should migrate on these as time goes (so none of them does M1 yet).

Cloud machines are costing 4-10X times more than machines hosted from dedicated servers providers per month. So if machine is used several hours a day, it is already cheaper than cloud.

We do not have team variations so fast like usual user facing stuff (e.g. games or chats or web sites), so really no need auto scale.

Also we need hard clear cap on number of concurrency of jobs, so we do not overpay. Yes - devs will wait longer when active stuff going on (BTW saw it 3x times faster, so most of time they will wait less), but that is the case - can add hardware if company will think dev times is higher cost than hardware. So super clear nice way to define values.

Given that machines mostly will check state, ok to have 2 types of machines, slow autoscaled(docs, fe build) and dedicated. So we kind of solving variability too.

So this PR replaced longest running jobs with dedicated machines. And after some tests will replace all remaining cute/monster machines from GCE.

Follow up of this PR is to add more machines and remove all cloud.

TLDR; 
Dedicated hardware machines in CI are faster, cheaper(in all senses) and fit better to this CI rather than auto scaled garbage (garbage in direct sense because we rent GCE garbage machines).

- other than that did minor code changes improves to trigger builds, also removed legacy dead nix, and used latest cachix on new machines, and fixed missing SHA commit in merge queue (when main was used instead of merge queue branch).


dzmitry-lahoda:
was killed by combination of linux and cachix which fails https://discourse.haskell.org/t/facing-mmap-4096-bytes-at-nil-cannot-allocate-memory-youre-not-alone/6259/15 ..

dzmitry-lahoda:
now 4 machines in run. 

need to review and run. so we can have experiment and switch in next 2 weeks.

@blasrodri

dzmitry-lahoda:
@RustNinja please find time to review

dzmitry-lahoda:
i am not sure what we are wating and continue to burn 14K month extra we can avoid.

PR #3675, opened 2023-06-08
xcvm: remove concept of bridge security

mina86:
As we’re heading towards IBC-based solution, the bridge security concept is no longer necessary.  Simplify code by removing all handling of it.

### DIFF
--- a/code/xcvm/cosmwasm/contracts/common/src/router.rs
+++ b/code/xcvm/cosmwasm/contracts/common/src/router.rs
@@ -2,7 +2,7 @@ use crate::shared::BridgeMsg;
 use cw_xcvm_utils::DefaultXCVMProgram;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
-use xcvm_core::{BridgeSecurity, CallOrigin, Displayed, Funds, InterpreterOrigin};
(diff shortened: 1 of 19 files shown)


blasrodri (commented):
If we remove this, and down the line want to add it, would there be an issue on backward compatibility?

mina86:
We use protobufs for everything so I think we should be able to add it back.

blasrodri (approved):

dzmitry-lahoda:
@blasrodri we do not use protobuf. adding and using protobuf is not aligned priority.

mina86:
I was under the impression that we were essentially settled on protobuf.  Either way, we need some way for versioning programs anyway so whatever encoding we settle on, we should be able to add new features to the programs.

dzmitry-lahoda:
eventually yes, protobuf. but we did not palled that in v0. so we not 'use protobuf', but 'we will use protobuf when in some point in future'. btw json versioned as good protobuf.

dzmitry-lahoda:
sure we can find blocker using v0 or change plans to do more stuff now then later, but that should be aligned in plan or board, because current implementation is json and work should be done.

mina86:
Sure, I’m not saying we have to start using protobuf right now.  The point is that if we ever want to add bridge security back, we will have a way to do it at that point.

PR #3684, opened 2023-06-12
XC Multihop proposal document transfer details and improvements 

dzmitry-lahoda:


rfcs/0015-one-signature-multihop-transfer-solution-ibc-xcm.md
```diff
@@ -150,11 +149,21 @@ Cannot express(securely) arbitrary routing and programs execution.
 
 To be handled in next RFC or implementation.
 
+## Error handling
```
blasrodri, line 152: Can you please provide an example where the unwinding happens, and it succeeds? And one where it doesn't?

blasrodri:
Please let's add a few examples, with some graphics (ascii art is also welcome)

RustNinja:
@dzmitry-lahoda @blasrodri @josietyleung 
Me as a Polkadot account who wants to initiate transfers from DOT to Osmosis/Banksy.
- i will create the xcm call reserve transfer assets and will specify multi-location.
- Without rollbackk/bad path when ibc will fail when deliver packet from composable to picasso then my token blocked on composable. I do not have any LAYR token to initiate tx and move DOT from composable back to polkadot or via IBC further to picasso via IBC.

current version of pallet ibc does not have ibc message forwarding with memo and does

blasrodri:
Seems like we need to implement https://github.com/strangelove-ventures/packet-forward-middleware but for both pallet-ibc and the relationship between ibc + xcm

rfcs/0015-one-signature-multihop-transfer-solution-ibc-xcm.md
dzmitry-lahoda: sure, easy. just have read reference module. see ok. will unwind as needed if to do in Substrate same as in Go. Plus final original tracking of sent xcm packet (basically unwind for xcm in hook).

dzmitry-lahoda:
> current version of pallet ibc does not have ibc message forwarding with memo and does

exactly, this should be implemted, ported from referenced go code

dzmitry-lahoda:
> Seems like we need to implement https://github.com/strangelove-ventures/packet-forward-middleware but for both pallet-ibc and the relationship between ibc + xcm

@blasrodri yes. that was idea. port some Go code to XCM.

RustNinja (approved):

dzmitry-lahoda:
@RustNinja may you summarize all quesitions remaining here so we align these in this PR?

```
Nikita  [3:43 PM]
I have a few question about Batch tx.
Who are gonna change the banksy to make a batch tx?
Does the batch tx should work for any chain?
Or just for our set of chains?
polkadot - composable - picasso - banksy - osmo? and reverse
Is it for any tokens? or is just for dot? all tokens
why do we need
```
What else? (on top of request how we handle XCM rollbacks

RustNinja:
> 



> @RustNinja may you summarize all quesitions remaining here so we align these in this PR?
> 
> ```
> Nikita  [3:43 PM]
> I have a few question about Batch tx.
> Who are gonna change the banksy to make a batch tx?
> Does the batch tx should work for any chain?
> Or just for our set of chains?
> polkadot - composable - picasso - banksy - osmo? and reverse
> Is it for any tokens? or is just for dot? all tokens
> why do we need
> ```
> 
> What else? (on top of request how we handle XCM rollbacks

thanks. they all are answered already.

yes. xcm rollbacks is very useful. Thank you so much.

dzmitry-lahoda:
Added some, will consistent little bit. Not sure about amount of details for making all IBC middleware XCM. We far behind official ibc-rs. Seems will need some glue to hook some callbacks (middleware actually) into pallet-ibc. I do not specify that. It is pure grunt eng work :)

RustNinja (approved):

dzmitry-lahoda:
https://github.com/strangelove-ventures/packet-forward-middleware/pull/87

dzmitry-lahoda:
no reaction, nice

blasrodri (approved):

RustNinja (approved):

dzmitry-lahoda:
execute can be only called by root, so cannot do arbitrary txes yet.

RustNinja:
> execute can be only called by root, so cannot do arbitrary txes yet.
Thanks.
Does it mean that we need to use `reserve_transfer_assets` with a custom multilocation (pallets and index) to move assets from polkadot/substrate chain?

RustNinja:
@dzmitry-lahoda 
another question is how to initiate the ibc/xcm call? 

there are 2 options:

- should it be in the same block when ibc/xcm instruction was executed? 
as a callback function in the same block?

- Does pallet xcm ibc pallet need just update the storage in callback and then 
next block on_finalize hook will trigger new xcm/ibc packet?

which option do you prefer?

dzmitry-lahoda:
in same block immediately. that is simpler. llwe did not do xcm benches, but can run and make bench depend on account to incroporate ibc. or can optimize like you said.

dzmitry-lahoda:
where to encode index staff, dest or beneficiary, what works.

RustNinja:
@dzmitry-lahoda 
https://github.com/ComposableFi/composable/blob/main/rfcs/0015-one-signature-multihop-transfer-solution-ibc-xcm.md#centauricosmos-sdk------composable-picassosubstrate---composable-composablesubstrate---bifrostsubstrate

i want to confirm:
```
{
  "forward": {
    "receiver": "composable-alice",
    "port": "transfer",
    "channel": "channel-123",
    "next": {
      "substrate" : {
          "parent" : "Polkadot",
          "parachain" : "Bifrost",
          "account" : "alice", 
      }
    }
  }
}
```

am i right that substrate hop can be only the last hop in memo?
Because other substrate chains do not support multi hop so in our example composable or picasso will able to forward message once directly to polkadot or kusama or bifrost... 
that's all

PR #3716, opened 2023-06-19
xcvm: introduce ExecuteProgramMsg to refactor messages slightly

mina86:

### DIFF
--- a/code/xcvm/cosmwasm/contracts/common/src/gateway.rs
+++ b/code/xcvm/cosmwasm/contracts/common/src/gateway.rs
@@ -1,6 +1,5 @@
+use cosmwasm_schema::cw_serde;
 use cw_xc_utils::DefaultXCVMProgram;
-use schemars::JsonSchema;
-use serde::{Deserialize, Serialize};
 use xc_core::{CallOrigin, Displayed, Funds, InterpreterOrigin, NetworkId};
 
 /// Prefix used for all events attached to gateway responses.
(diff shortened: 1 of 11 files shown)


code/xcvm/cosmwasm/contracts/common/src/gateway.rs
```diff
@@ -21,49 +20,46 @@ pub struct InstantiateMsg {
...
-		/// The program salt.
-		salt: Vec<u8>,
-		/// The program.
-		program: DefaultXCVMProgram,
-		/// Assets to fund the XCVM interpreter instance
-		/// The interpreter is funded prior to execution
-		assets: Funds<Displayed<u128>>,
-	},
+	/// Sent by the user to execute a program on their behalf.
+	ExecuteProgram(ExecuteProgramMsg),
```
dzmitry-lahoda, line 31: {
	execute_program: ExecuteProgramMsg,
},

dzmitry-lahoda (approved):

blasrodri (dismissed):

code/xcvm/cosmwasm/contracts/common/src/gateway.rs
mina86: IMO that only adds noise to the code.  ExecuteProgramMsg is the entirety of the ExecuteProgram message so it makes sense to use a singleton tuple the same way its used in Bridge below.

dzmitry-lahoda: generally i agree, bot not in this case so

mina86: Well, what’s different about this case?

dzmitry-lahoda: 1. There was case when it was {}, not (). So indicator it can happen again.
2. I have to write EPP { foo : }, I would like to write(or macro) EP { foo : } too. For consistency.
3. If EP will be not { foo: } BUT EP { foo : , tips_fees_time_configs_and_hints_from_end_user : } will not require refactor back. Oh 1 just happened in future.
4. Also refactoring to () was done, it it was not done it would be faster to refactor use {} in first place.

mina86: Done.

dzmitry-lahoda (approved):

PR #3734, opened 2023-06-22
xcvm: remove Network generic argument from xc_core::Instruction

mina86:
It’s rather unlikely that network in an XCVM Instruction type will ever be represented as something other than NetworkId.  Network is part of the interface which needs to be understood by all the chains in the same way and have the same representation.

Remove Network generic argument from xc_core::Instruction to simplify the code slightly.

### DIFF
--- a/code/xcvm/cosmwasm/contracts/utils/src/lib.rs
+++ b/code/xcvm/cosmwasm/contracts/utils/src/lib.rs
@@ -1,9 +1,7 @@
 use cosmwasm_std::CanonicalAddr;
 use std::collections::VecDeque;
-use xc_core::{Balance, Funds, NetworkId};
 
-pub type DefaultXCVMInstruction =
-	xc_core::Instruction<NetworkId, Vec<u8>, CanonicalAddr, Funds<Balance>>;
+pub type DefaultXCVMInstruction = xc_core::Instruction<Vec<u8>, CanonicalAddr, xc_core::Funds>;
 pub type DefaultXCVMProgram = xc_core::Program<VecDeque<DefaultXCVMInstruction>>;
 pub type DefaultXCVMPacket = xc_core::Packet<DefaultXCVMProgram>;
 pub type Salt = Vec<u8>;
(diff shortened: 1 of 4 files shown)


dzmitry-lahoda (approved):
just before deleting.

history is that there are 2 network ids. one is sure simple non feneric nimber to be used on the wire. other one was added to allow write typed programs, so when tou write for eth you can only use tx provider for eth, and than can use in spawn other type, and get other set of tx providers. compile time safety of program builder pattern. so may be some code not compiles anymore.

i have not strong opinion her, so.generally i am pro typing.

mina86:
I was wondering about that but I’m not clear how that would work.  Currently we have:

```rust
enum Instruction<Network, Payload, Account, Assets> {
	Transfer { to: Destination<Account>, assets: Assets },
	Call { bindings: Bindings, encoded: Payload },
	Spawn {
		network: Network,
		salt: Vec<u8>,
		assets: Assets,
		program: Program<VecDeque<Self>>,
	},
}
```

I can imagine that for example on Ethereum I could have:

```
type EthInstruction = Instruction<eth::Network, eth::Payload, Account, Assets>;
```

and then make `eth::Payload` would be a rich type which makes sure
that the payload is valid Ethereum call, but such type would only
allow spawning on Ethereum.

In `Spawn` instruction network can be anything so it cannot be limited
by type to just a single network.

dzmitry-lahoda:
i saw it was working as i recall. builder interface was typen safe.

dzmitry-lahoda:
but it generated non type safe tre in the end.

mina86:
The builder uses NetworkId as the argument:

```rust
pub struct ProgramBuilder<CurrentNetwork, Account, Assets> {
	pub tag: Vec<u8>,
	pub instructions: VecDeque<Instruction<NetworkId, Vec<u8>, Account, Assets>>,
	pub _marker: PhantomData<CurrentNetwork>,
}
```

The type safety comes from the method definitions, for example spawn
has a network type argument:

```rust
	pub fn spawn<SpawningNetwork, E, FinalNetwork, F>(
		self,
		tag: impl Into<Vec<u8>>,
		salt: impl Into<Vec<u8>>,
		assets: impl Into<Assets>,
		f: F,
	) -> Result<ProgramBuilder<FinalNetwork, Account, Assets>, E>
```

(Even though I’m confused why it has SpawningNetwork and FinalNetwork
as separate arguments.  As far as I can tell they will always be the
same)

PR #3759, opened 2023-07-03
Porting memo spec wasm termination from Osmosis for XCVM ICS-20 memo as program carrier

dzmitry-lahoda:


code/parachain/frame/composable-traits/src/ibc.rs
```diff
@@ -0,0 +1,145 @@
...
+#[cfg_attr(feature = "std", derive(JsonSchema))]
+pub enum SudoMsg {
+    #[serde(rename = "ibc_lifecycle_complete")]
+    IBCLifecycleComplete(IBCLifecycleComplete),
+}
+
+pub const SENDER_PREFIX : &str = "ibc-wasm-hook-intermediary";
+
+/// from Go code to make compliant wasm hook
+pub fn derive_intermediate_sender(channel: String, original_sender : String, bech32_prefix : String) -> Result<String, ()> {
```
mina86, line 54: It’s more idiomatic to take the arguments by reference:

```suggestion
pub fn derive_intermediate_sender(channel: &str, original_sender: &str, bech32_prefix: &str) -> Result<String, ()> {
```

This way caller is not forced to make an owned object which this
function discards anyway.

code/parachain/frame/composable-traits/src/ibc.rs
```diff
@@ -0,0 +1,145 @@
...
+
+impl MemoData {
+	pub fn forward(forward: Forward) -> Self {
+		Self::Forward(forward)
+	}
+
+	pub fn wasm(wasm: Wasm) -> Self {
+		Self::Wasm(wasm)
+	}
+}
```
mina86, line 145: I probably wouldn’t bother with those methods.  `MemoData::Forward(...)` is just as good as `MemoData::forward(...)`.

mina86 (approved):

code/parachain/frame/composable-traits/src/ibc.rs
dzmitry-lahoda: yeah, sorry. just direct cp paste from go. did not noticed

code/parachain/frame/composable-traits/src/ibc.rs
dzmitry-lahoda: will remove in next pr if needed

dzmitry-lahoda: or split per parameter

Issue #3403, opened 2023-04-13
CosmWasm VM: Address Halborn audit report 


itsbobbyzz:
This task relates to the Halborn audit report of the Cosmwasm pallet and specifically the medium issue Hal-01 per page 10 of the report. 

3.1 (HAL-01) INSTANTIATE CALL
MISSING ZERO CHECK - MEDIUM
Description:
The gas parameter in the instantiate function of the code lacks validation,
which means that it can accept a value of zero (0) for gas. This can
potentially lead to an undefined behavior, where an attacker can deploy a
contract with no gas limit, causing the contract to consume an excessive
amount of resources and exhaust the Virtual Machine resources, leading
to unexpected behavior inside the Virtual Machine.

josietyleung:
@dzmitry-lahoda do you know if this done already? thanks

dzmitry-lahoda:
not done

itsbobbyzz:
[Composable_Finance_CosmWasm_Pallet_Substrate_Pallet_Security_Audit (1) (1).pdf](https://github.com/ComposableFi/composable/files/12039079/Composable_Finance_CosmWasm_Pallet_Substrate_Pallet_Security_Audit.1.1.pdf)

dzmitry-lahoda:
done by document from Mina86. Reopen if needed

PR #3787, opened 2023-07-13
pallet-cosmwasm: make sure contract are called with non-zero gas

mina86:
Closes: https://github.com/ComposableFi/composable/issues/3403

- [X] PR title is my best effort to provide summary of changes and has clear text to be part of release notes
- [X] I marked PR by `misc` label if it should not be in release notes
- [X] I have linked Zenhub/Github or any other reference item if one exists
- [X] I was clear on what type of deployment required to release my changes (node, runtime, contract, indexer, on chain operation, frontend, infrastructure) if any in PR title or description
- [X] I have added at least one reviewer in reviewers list

### DIFF
--- a/code/parachain/frame/cosmwasm/src/lib.rs
+++ b/code/parachain/frame/cosmwasm/src/lib.rs
@@ -462,6 +462,7 @@ pub mod pallet {
 			gas: u64,
 			message: ContractMessageOf<T>,
 		) -> DispatchResultWithPostInfo {
+			ensure_minimum_gas::<T>(gas, T::WeightInfo::instantiate(funds.len() as u32))?;
 			T::ExecuteWasmOrigin::ensure_origin(origin.clone())?;
 			let who = ensure_signed(origin)?;
 			let mut shared = Self::do_create_vm_shared(gas, InitialStorageMutability::ReadWrite);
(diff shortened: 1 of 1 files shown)


code/parachain/frame/cosmwasm/src/lib.rs
```diff
@@ -462,6 +462,7 @@ pub mod pallet {
 			gas: u64,
 			message: ContractMessageOf<T>,
 		) -> DispatchResultWithPostInfo {
+			ensure_minimum_gas::<T>(gas, T::WeightInfo::instantiate(funds.len() as u32))?;
```
dzmitry-lahoda, line 465: i think it is next:
1. instantiate pallet overhead with dummy contract instance was tested
2. that is payed  by #[pallet::weight(T::WeightInfo::instantiate(funds.len()), plus gas taken at same time. 
3. so for any call here, user never ever gets T::WeightInfo::instantiate(funds.len()  back
4. so no reason to check it that is part of gas
5. instantiate ALWASY takes T::WeightInfo::instantiate(funds.len()  and never returns - in case error too.

dzmitry-lahoda: so tbh review from Halborn is garbage

dzmitry-lahoda: ensure_minimum_gas::<T>(gas, T::WeightInfo::instantiate(funds.len() as u32))?; 

may be
ensure!(gas >= T::WeightInfo::instantiate(funds.len() as u32))?;

PR #3800, opened 2023-07-17
xcvm over ICS20 assets

dzmitry-lahoda:
- replaced transport with ics 20 memo
- so had to replace assets transfer
- so added more storage items to handle ics 20 routes
- made security according wasm hook with verification of sender
- needed to use precompile as cw std lacks memo, so unified usage of cw things in core crate (so can use in precompiles in substrate)
- moved reusable parts all in core (like cw std, or osmosis std, ibc rs do)
- made PICA routed via other chains is not PICA when get back to Picasso (without path reverse)
- improved type system of xcvm, made it more integrated
- removed some obsolete not used and not usable TS codes
- update of some relevant docs
- as contracts are part of general Cargo.toml, fixed many lintings (like Clone cases)
- greened CW builds as part of CI
- https://github.com/ComposableFi/composable/actions/runs/5614667327/job/15213335028?pr=3800
- made spit of anonymous vs not programs

- [x] I have linked Zenhub/Github or any other reference item if one exists https://github.com/ComposableFi/Private-Issues-/issues/77


code/xcvm/cosmwasm/contracts/gateway/src/error.rs
```diff
@@ -38,4 +38,15 @@ pub enum ContractError {
 	UnknownChannel,
 	#[error("The asset is already registered.")]
 	AlreadyRegistered,
+	#[error("Route not found.")]
+	RouteNotFound,
+	#[error("{0:?}")]
+	Bech32(bech32_no_std::Error),
```
mina86, line 44:
```suggestion
	Bech32(#[from] bech32_no_std::Error),
```
and than you won’t need to make the From implementation yourself.

dzmitry-lahoda:
i am not sure if should remove open channel, it still may be usefull for data propagation without assets and in future be part of batch (like ics20+ics27 batches into one call)

code/Cargo.toml
```diff
@@ -343,21 +349,42 @@ wasmi-validation = { version = "0.5", default-features = false }
...
+num = { version = "0.4.0", default-features = false }
+half = { version = "2.2.1", default-features = false }
+
+
 thiserror = { version = "1.0.38", package = "thiserror-core", default-features = false }
 
+[patch.crates-io]
+
+
 #######################################
```
mina86, line 388:
```suggestion
#######################################
```

code/xcvm/cosmwasm/contracts/gateway/src/error.rs
```diff
@@ -38,4 +38,16 @@ pub enum ContractError {
...
+	#[error("{0:?}")]
+	Serde(#[from] serde_json_wasm::ser::Error),
+}
+
+impl From<bech32_no_std::Error> for ContractError {
+	fn from(value: bech32_no_std::Error) -> Self {
+		Self::Bech32(value)
+	}
 }
```
mina86, line 53:
```suggestion
	#[error("{0}")]
	Bech32(#[from] bech32_no_std::Error),
	#[error("{0}")]
	Serde(#[from] serde_json_wasm::ser::Error),
}
```

code/xcvm/lib/core/README.md
```diff
@@ -0,0 +1,6 @@
+# Overview
+
+Staff to do cross chains apps.
```
mina86, line 3:
```suggestion
Stuff to do cross chains apps. 
```

code/xcvm/lib/core/src/gateway.rs
```diff
@@ -49,17 +59,47 @@ pub enum ExecuteMsg {
...
 	/// Message sent by an admin to remove an asset from registry.
 	UnregisterAsset {
 		asset_id: AssetId,
 	},
+
+	Wasm(Ics20MessageHook),
+}
+
+/// when message is sent to other side, we should identify receiver of some kind
+#[derive(Serialize, Deserialize, JsonSchema, Clone, PartialEq, Debug)]
```
mina86, line 73:
```suggestion
#[cw_serde]
#[derive(Debug)]
```

code/xcvm/lib/core/src/gateway.rs
```diff
@@ -49,17 +59,47 @@ pub enum ExecuteMsg {
...
+pub enum GatewayId {
+	CosmWasm(Addr),
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+
```
mina86, line 79:
```suggestion
#[cw_serde]
#[derive(Debug)]
```

code/xcvm/lib/core/src/gateway.rs
```diff
@@ -49,17 +59,47 @@ pub enum ExecuteMsg {
...
+	pub bridged: Option<BridgeAsset>,
+}
+
+impl Asset {
+	pub fn denom(&self) -> String {
+		self.local.denom()
+	}
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
```
mina86, line 92:
```suggestion
#[cw_serde]
#[derive(Debug)]
```

mina86 (dismissed):
This really should be split into multiple smaller PRs.

code/xcvm/cosmwasm/contracts/gateway/src/error.rs
dzmitry-lahoda: hm, for some reason bech32 does not work like that

code/xcvm/lib/core/README.md
dzmitry-lahoda: oh, i disabled spell checking as too many false positives. so will fix this

code/xcvm/lib/core/src/gateway.rs
dzmitry-lahoda: i cannot do it. because will not be able to compile for substrate. this lib is for cross consumption of these things

code/xcvm/lib/core/src/gateway.rs
dzmitry-lahoda: same. until cw no srd merged.

code/xcvm/lib/core/src/gateway.rs
dzmitry-lahoda: i did not actually forked cw serde for no std, so will not work. so may other reason to upgrade forks and fix macro later

code/xcvm/lib/core/src/gateway.rs
mina86: Sorry, didn’t notice it was in lib/core.

dzmitry-lahoda:
Fixing final steps 

```
warning: /home/dz/github.com/ComposableFi/composable/code/parachain/runtime/picasso/Cargo.toml: unused manifest key: dependencies.frame-system.package
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on package cache
    Blocking waiting for file lock on package cache
   Compiling getrandom v0.2.10
error: the wasm*-unknown-unknown targets are not supported by default, you may need to enable the "js" feature. For more information see: https://docs.rs/getrandom/#webassembly-support
   --> /home/dz/.cargo/registry/src/index.crates.io-6f17d22bba15001f/getrandom-0.2.10/src/lib.rs:285:9
    |
285 | /         compile_error!("the wasm*-unknown-unknown targets are not supported by \
286 | |                         default, you may need to enable the \"js\" feature. \
287 | |                         For more information see: \
288 | |                         https://docs.rs/getrandom/#webassembly-support");
    | |________________________________________________________________________^

error[E0433]: failed to resolve: use of undeclared crate or module `imp`
   --> /home/dz/.cargo/registry/src/index.crates.io-6f17d22bba15001f/getrandom-0.2.10/src/lib.rs:341:9
    |
341 |         imp::getrandom_inner(dest)?;
    |         ^^^ use of undeclared crate or module `imp`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `getrandom` (lib) due to 2 previous errors
````
of no_std

File: code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs [hunk 8]
```diff
@@ -97,7 +97,7 @@ pub enum GasOutcome {
 
 impl Gas {
 	pub fn new(max_frames: u32, initial_value: u64) -> Self {
-		let max_frames = usize::try_from(max_frames).unwrap();
+		let max_frames = max_frames as usize;
```
mina86, line 100: But why?  `as` is a tricky way to cast since it may loose precision.  With `try_from` we’re sure we’ll never loose data and on platforms where usize is 32-bit or 64-bit wide both compile to the same code.

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
```diff
@@ -126,7 +126,7 @@ impl Gas {
 	}
 
 	pub fn pop(&mut self) {
-		let child = self.checkpoints.pop().unwrap();
+		let child = self.checkpoints.pop().expect("non empty");
```
mina86, line 129: That’s just noise IMO.

code/xcvm/lib/core/src/gateway.rs
```diff
@@ -131,20 +151,23 @@ impl AssetReference {
...
+			AssetReference::Virtual { cw20_address } => ["cw20:", cw20_address.as_str()].concat(),
 		}
 	}
 }
 
-#[cw_serde]
-#[derive(QueryResponses)]
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+#[serde(rename_all = "snake_case")]
+#[cfg_attr(feature = "std", derive(schemars::JsonSchema, QueryResponses))]
```
mina86, line 161: Maybe we can just drop QueryResponses? To be honest I don’t understand what it gives us.

code/xcvm/lib/core/src/ibc/mod.rs
```diff
@@ -49,21 +49,21 @@ pub struct IbcRoute {
...
+		IbcIcs20Sender::SubstratePrecompile(addr) => Ok(WasmMsg::Execute {
 			contract_addr: addr.into(),
-			msg: to_binary(&transfer).unwrap(),
+			msg: to_binary(&transfer)?,
 			funds: <_>::default(),
 		}
-		.into(),
-		_ => todo!(),
+		.into()),
+		_ => unreachable!("until we implement"),
```
mina86, line 67:
```suggestion
		_ => unimplemented!(),
```

code/xcvm/lib/core/src/network.rs
```diff
@@ -14,7 +14,7 @@ use serde::{Deserialize, Serialize};
 pub struct InterpreterOrigin {
 	pub user_origin: UserOrigin,
 	#[serde(with = "hex")]
-	#[schemars(with = "String")]
+	#[cfg_attr(feature = "std", schemars(with = "String"))]
```
mina86, line 17: Does this actually break otherwise? Interesting…

mina86 (approved):

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
dzmitry-lahoda: max_frames is u32. so you saying for case when usize is u16? or we going to increase u32 to u64? but in this case would not it mean usize sure is 16. also why we should potentially brick and panic runtime, while loosing precision we just execute without unwrap linter? we have unwrap linter, i am fixing. what is best fix? unwraps sometimes bad sometime good. but would you prefer suppress linter? or just execute?

i prefer exactly `as usize` code as it will just work without linter noise and susipicous unwrap

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
dzmitry-lahoda: we have linter, now contracts are part for common CI. so i fixed linter.  other option fix by supperess (ugly) or disable all unwrap?

code/xcvm/lib/core/src/gateway.rs
dzmitry-lahoda: see `returns` . It allows to generate nice json schema so that people can generate nice TypeScript code for FE.

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
mina86: I’d much rather we got rid of the lint check.

code/xcvm/lib/core/src/network.rs
dzmitry-lahoda: if you will return it without attr, it will fail to compile as part of Substrate runtime 100%. But I assume that all which is in core almost certainly wil be used by it. Example? Polkadot -> Compoble -> sudo XCVM  on behalf some origin directly instead of contract. other case sure. Gov precompiles may be. But idea is that at least half of core reusable by substrate/rpc/etc. So it is cross compiles for std and not. 

It will gov way in September as CW will support no_std

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
dzmitry-lahoda: btw, `gas_checkpoint_pop` allows updates and bugs to CW brick runtime easy :) it could just return error empty checkpoint.

dzmitry-lahoda: bad code

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
mina86: I’m saying, always prefer `try_from` over `as` so that the code is consistent everywhere and we don’t silently loose precision.  In other places we may have `u64` argument cast to `usize` and there the check is necessary.

> also why we should potentially brick and panic runtime

Because if the value doesn’t fit the type we’re casting to, that’s a bug in code and the best thing to do is abort the program.

> unwraps sometimes bad sometime good

Unwraps are bad for input validation or error handling.  They are good in situations where triggering the unwrap would indicate a bug in the code.  E.g. in situations like her where programmer just assumes that they can cast `max_frames` to `usize`.

I find the lint check counterproductive.  `unwrap` ends up being replaced by `expect` with a meaningless message or by a worse code.  Furthermore, it has too many false negatives.  Indexing for example is essentially a hidden unwrap call.

code/xcvm/lib/core/src/network.rs
mina86: > It will go away in September as CW will support no_std

Last comment I got a couple days ago is that cosmwasm 2.0 will be ready when it’s ready without them committing to any timeline.  We may need to prepare for continuing to use a fork.

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
mina86: Returning error works for me.  I actually took a cursory look at whether we could have some kind of scoping here so rather than push and pop operations have push which returns a guard object which pops as RAII.  Didn’t see an obvious way to do it though.

mina86: https://github.com/ComposableFi/composable/pull/3815

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
dzmitry-lahoda: unwrap must not happen in parachains. never. if you know that it never happens - please write down exactly why - expect. or loose precision. I prefer loose precision than paying parity 20K USD to unbrick chain.

dzmitry-lahoda: this is default guide line for Substrate and linter in ecosystem. btw in CW there is no stacktrackes, so without expect you will not see anything at all. what really failed.

so not sure why you are so on unwrap when it is terrible...

dzmitry-lahoda: > Furthermore, it has too many false negatives. Indexing for example is essentially a hidden unwrap call.

We have linter forbidding indexing, you are right. We have it.

dzmitry-lahoda: > expect with a meaningless message or by a worse code

why it is meaningless? for me it at least something. i wrote before about it.

dzmitry-lahoda: btw in this case real fix is to use usize or u16 for frame size. so we do not even have option to panic here or do unwraps.

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
dzmitry-lahoda: real fix in CW repo side i guess.  so it it is like big work.

code/xcvm/cosmwasm/contracts/gateway/src/assets.rs
```diff
@@ -1,25 +1,27 @@
 use crate::{
-	common,
+	auth,
 	error::{ContractError, ContractResult},
-	msg, state,
+	events::make_event,
+	msg,
+	state::{self},
```
mina86, line 6:
```suggestion
	state,
```

code/xcvm/cosmwasm/contracts/interpreter/src/contract.rs
```diff
@@ -323,14 +318,9 @@ pub fn interpret_spawn(
 
 	let mut response = Response::default();
 	for (asset_id, balance) in assets.0 {
-		if balance.amount.is_zero() {
-			// We ignore zero amounts
-			continue
-		}
```
mina86, line 329: Why not keep it?

code/xcvm/lib/core/src/cosmwasm.rs
```diff
@@ -20,9 +20,9 @@
 //! Let's assume that we wan't to send `WasmMsg::Execute`:
 //! ```json
 //! {
-//! 	"contract_addr": "",
-//! 	"msg": "SOME PAYLOAD",
-//! 	"funds": []
+//!    "contract_addr": "",
+//!    "msg": "SOME PAYLOAD",
+//!    "funds": []
```
mina86, line 25:
```suggestion
//!     "contract_addr": "",
//!     "msg": "SOME PAYLOAD",
//!     "funds": []
```

code/xcvm/lib/core/src/cosmwasm.rs
```diff
@@ -50,12 +50,12 @@
 //! let payload_bindings: OrderedBindings = [(13, BindingValue::Register(Register::This))].into();
 //! let cw20_transfer_msg = Cw20ExecuteMsg::Transfer {
 //!     // Make sure to leave fields that uses late-bindings empty
-//! 	recipient: String::new(),
+//!    recipient: String::new(),
```
mina86, line 53:
```suggestion
//!     recipient: String::new(),
```

code/xcvm/lib/core/src/cosmwasm.rs
```diff
@@ -50,12 +50,12 @@
...
 //!     amount: 10000,
 //! };
 //!
 //! let payload = LateCall::wasm_execute(
-//! 	StaticBinding::Some(BindingValue::Asset(1)),
+//!    StaticBinding::Some(BindingValue::Asset(1)),
```
mina86, line 58:
```suggestion
//!     StaticBinding::Some(BindingValue::Asset(1)),
```

mina86 (dismissed):

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
```diff
@@ -0,0 +1,141 @@
...
+use crate::{
+	auth,
+	contract::EXEC_PROGRAM_REPLY_ID,
+	error::{ContractError, ContractResult},
+	events::make_event,
+	state,
+	state::{NetworkItem, OtherNetworkItem},
+};
+
+pub(crate) fn handle_bridge_forward(
```
blasrodri, line 28: what does handle bridge forward means?

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
```diff
@@ -0,0 +1,141 @@
...
+	_: auth::Interpreter,
+	deps: DepsMut,
+	info: MessageInfo,
+	msg: xc_core::gateway::BridgeMsg,
+) -> ContractResult<Response> {
+	let channel_id = state::IBC_NETWORK_CHANNEL
```
blasrodri, line 34: unused

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
```diff
@@ -0,0 +1,141 @@
...
+		>,
+	> = XcPacket {
+		interpreter: String::from(info.sender).into_bytes(),
+		user_origin: msg.interpreter_origin.user_origin,
+		salt: msg.msg.salt,
+		program: msg.msg.program,
+		assets: msg.msg.assets,
+	};
+
+	let (local_asset, amount) = packet.assets.0.get(0).expect("verified at outer boundaries");
```
blasrodri, line 52: how come this safe and it won't panic

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
```diff
@@ -0,0 +1,141 @@
...
+			"assets",
+			serde_json_wasm::to_string(&packet.assets)
+				.map_err(|_| ContractError::FailedToSerialize)?,
+		)
+		.add_attribute(
+			"program",
+			serde_json_wasm::to_string(&packet.program)
+				.map_err(|_| ContractError::FailedToSerialize)?,
+		);
+	if !packet.salt.is_empty() {
```
blasrodri, line 67: what happens if it's empty?

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
```diff
@@ -0,0 +1,141 @@
...
+	let call_origin = CallOrigin::Remote { user_origin: packet.user_origin };
+	let execute_program =
+		ExecuteProgramMsg { salt: packet.salt, program: packet.program, assets: packet.assets };
+	let msg =
+		ExecuteMsg::ExecuteProgramPrivileged { call_origin, execute_program, tip: info.sender };
+	let msg = wasm_execute(env.contract.address, &msg, Default::default())?;
+	Ok(Response::new().add_submessage(SubMsg::reply_always(msg, EXEC_PROGRAM_REPLY_ID)))
+}
+
+fn ensure_anonymous(program: &DefaultXCVMProgram) -> ContractResult<()> {
```
blasrodri, line 132: given the fact that this is recursive, don't we want to cap the depth?

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
```diff
@@ -0,0 +1,141 @@
...
+
+pub(crate) fn ics20_message_hook(
+	_: auth::WasmHook,
+	msg: Ics20MessageHook,
+	env: Env,
+	info: MessageInfo,
+) -> Result<Response, ContractError> {
+	let packet: XcPacket = decode_packet(&msg.data).map_err(ContractError::Protobuf)?;
+
+	ensure_anonymous(&packet.program)?;
```
blasrodri, line 122: I don't understand why the name of this function. Seems like you're checking that an instruction is a Transfer or spawn, right?

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
dzmitry-lahoda: but is not sent too, because there is deeper method which sends

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
dzmitry-lahoda: it uses default empty slat interpreter

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
dzmitry-lahoda: or Buy, we decided to hold Buy. see responses about security issue you raised. we cannot rely on sender authorization, so really we do not know who called this, so really it is anonymous

dzmitry-lahoda: no all IX can be anonymous, but some are, ike Buy/Lend/Swap/Borrow

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
dzmitry-lahoda: it is already capped by 2 layers above

code/xcvm/cosmwasm/contracts/interpreter/src/contract.rs
dzmitry-lahoda: waste of time and resources and code space. i am fine sending zero, also it is per spec.

dzmitry-lahoda: any comments my it is needed? i do not see value, only waste

RustNinja (approved):

PR #3802, opened 2023-07-18
xcvm: use hex to encode binary data in JSON messages

mina86:
hex is more compact JSON representation than an array of bytes.
Prefer it for UserId and program origin salt.

We’re not using cosmwasm_std::HexBinary because it doesn’t play nice
with scale.  It’s most strightforward to just use serde(with) option.

### DIFF
--- a/code/xcvm/lib/core/src/network.rs
+++ b/code/xcvm/lib/core/src/network.rs
@@ -13,7 +13,17 @@ use serde::{Deserialize, Serialize};
 )]
 pub struct InterpreterOrigin {
 	pub user_origin: UserOrigin,
+	#[serde(with = "hex")]
+	#[schemars(with = "String")]
 	pub salt: Vec<u8>,
+}
+
+impl Display for InterpreterOrigin {
+	#[inline]
+	fn fmt(&self, fmtr: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+		let salt = hex::encode(self.salt.as_slice());
+		core::write!(fmtr, "{}-{salt}", self.user_origin)
+	}
 }
 
 /// The origin of a user, which consist of the composite, origin network and origin network user id.
@@ -26,6 +36,13 @@ pub struct UserOrigin {
 	pub user_id: UserId,
 }
 
+impl Display for UserOrigin {
+	#[inline]
+	fn fmt(&self, fmtr: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+		core::write!(fmtr, "{}-{}", self.network_id, self.user_id)
+	}
+}
+
 /// Arbitrary `User` type that represent the identity of a user on a given network, usually a public
 /// key.
 #[cfg_attr(feature = "std", derive(schemars::JsonSchema))]
@@ -33,7 +50,17 @@ pub struct UserOrigin {
 	Clone, PartialEq, Eq, PartialOrd, Ord, Debug, Encode, Decode, TypeInfo, Serialize, Deserialize,
 )]
 #[repr(transparent)]
-pub struct UserId(pub Vec<u8>);
+pub struct UserId(
+	#[serde(with = "hex")]
+	#[schemars(with = "String")]
+	pub Vec<u8>);
+
+impl Display for UserId {
+	#[inline]
+	fn fmt(&self, fmtr: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+		hex::encode(self.0.as_slice()).fmt(fmtr)
+	}
+}
 
 impl From<Vec<u8>> for UserId {
 	fn from(x: Vec<u8>) -> Self {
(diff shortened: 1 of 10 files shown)


mina86:
@dzmitry-lahoda, is there any reason we were using array representation? Will this change break anything?

dzmitry-lahoda (approved):

dzmitry-lahoda:
i not aware od any reasons.  soni would perfer either 0x hex encoding or base58 :)

mina86:
Anything but base58.  It’s the pinnacle of terrible encodings with O(n²) complexity.

Noticed that salt also uses array so changed that as well.  And had to add JsonScheme implementation which is a bit of a shame.

However, I’ve noticed that currently our code doesn’t compile
(abridged for brevity):

```
$ g co up/main
HEAD is now at 673db18f2 fix typo
$ c t -p xc-core --features std
error[E0277]: the trait bound `Addr: JsonSchema` is not satisfied
   --> lib/core/src/bridge.rs:21:46
    |
21  |     Remote { protocol: BridgeProtocol, relayer: Addr, user_origin: UserOrigin },
    |                                                 ^^^^ the trait `JsonSchema` is not implemented for `Addr`
    |
note: required by a bound in `SchemaGenerator::subschema_for`
   --> /home/mpn/.cargo/registry/src/index.crates.io-6f17d22bba15001f/schemars-0.8.12/src/gen.rs:218:38
    |
218 |     pub fn subschema_for<T: ?Sized + JsonSchema>(&mut self) -> Schema {
    |                                      ^^^^^^^^^^ required by this bound in `SchemaGenerator::subschema_for`

error[E0277]: the trait bound `Addr: JsonSchema` is not satisfied
  --> lib/core/src/bridge.rs:18:36
   |
18 | #[cfg_attr(feature = "std", derive(schemars::JsonSchema))]
   |                                    ^^^^^^^^^^^^^^^^^^^^ the trait `JsonSchema` is not implemented for `Addr`
   |
   = note: this error originates in the derive macro `schemars::JsonSchema` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: the trait bound `Addr: JsonSchema` is not satisfied
   --> lib/core/src/bridge.rs:22:16
    |
22  |     Local { user: Addr },
    |                   ^^^^ the trait `JsonSchema` is not implemented for `Addr`
    |
note: required by a bound in `SchemaGenerator::subschema_for`
   --> /home/mpn/.cargo/registry/src/index.crates.io-6f17d22bba15001f/schemars-0.8.12/src/gen.rs:218:38
    |
218 |     pub fn subschema_for<T: ?Sized + JsonSchema>(&mut self) -> Schema {
    |                                      ^^^^^^^^^^ required by this bound in `SchemaGenerator::subschema_for`
```

code/xcvm/lib/core/src/network.rs
```diff
@@ -5,17 +5,26 @@ use alloc::vec::Vec;
 use parity_scale_codec::{Decode, Encode};
 use scale_info::TypeInfo;
 use serde::{Deserialize, Serialize};
+use serde_with::{base64::Base64, serde_as};
 
 /// The interpreter origin, composite of a user origin and a salt.
-#[cfg_attr(feature = "std", derive(schemars::JsonSchema))]
```
dzmitry-lahoda, line 10: json schema is used to provide cw interface json schema for people to be generated. if this type is part of public contract interface, it should be json schema

code/xcvm/lib/core/src/network.rs
```diff
@@ -26,14 +35,28 @@ pub struct UserOrigin {
...
+	fn fmt(&self, fmtr: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+		core::write!(fmtr, "{}-{}", self.network_id, self.user_id)
+	}
+}
+
 /// Arbitrary `User` type that represent the identity of a user on a given network, usually a public
 /// key.
-#[cfg_attr(feature = "std", derive(schemars::JsonSchema))]
+#[serde_as]
 #[derive(
```
dzmitry-lahoda, line 48: same here about json schema in public api

code/xcvm/lib/core/src/network.rs
mina86: It’s implemented below on line 198.  The reason for lack of derive is because the derive(JsonSchema) doesn’t support serde_as.  I’m looking into some better ways of handling this without the need for all the explicit implementation.

code/xcvm/lib/core/src/network.rs
mina86: Implemented below on line 176.

code/xcvm/lib/core/src/network.rs
mina86: This turned out to be simpler than I thought. Done.

dzmitry-lahoda:
cw is hed one for many things https://github.com/CosmWasm/cosmwasm/blob/aca2de10d2332609a822bd1a8d5e37159fd6abc8/packages/std/src/hex_binary.rs#L226

also pd.js is good with hex accounts.

@bengalmozzi @josietyleung what you prefer for account encoding in cosmos, hex or base64?

bengalmozzi:
@dzmitry-lahoda Isn't our address encoded in base54? if it's not possible then I prefer hex

mina86:
HexBinary didn’t quite work out due to conflict with SCALE but managed to get it working with serde(with).

PR #3803, opened 2023-07-19
xcvm: minor refactoring and cleanups

mina86:
* Use ‘Result` type alias consistently in all contracts.  By using
  default generic arguments this makes the code more concise.
* Cleanup unused dependencies.

- [X] PR title is my best effort to provide summary of changes and has clear text to be part of release notes
- [X] I marked PR by `misc` label if it should not be in release notes
- [X] I have linked Zenhub/Github or any other reference item if one exists
- [X] I was clear on what type of deployment required to release my changes (node, runtime, contract, indexer, on chain operation, frontend, infrastructure) if any in PR title or description

### DIFF
--- a/code/xcvm/cosmwasm/contracts/gateway/src/state.rs
+++ b/code/xcvm/cosmwasm/contracts/gateway/src/state.rs
@@ -1,17 +1,10 @@
 use crate::msg;
 
 use cosmwasm_std::{Addr, IbcEndpoint, IbcTimeout, StdResult, Storage};
-use cw_storage_plus::{Item, Map, PrimaryKey, UniqueIndex};
-use ibc_rs_scale::{
-	applications::transfer::TracePrefix,
-	core::ics24_host::identifier::{ChannelId, ConnectionId},
-};
+use cw_storage_plus::{Item, Map};
+use ibc_rs_scale::core::ics24_host::identifier::ChannelId;
 use serde::{Deserialize, Serialize};
-use xc_core::{
-	gateway::{Asset, GatewayId},
-	location::ForeignAssetId,
-	AssetId, Funds, IbcIcs20Sender, InterpreterOrigin, NetworkId,
-};
+use xc_core::{AssetId, IbcIcs20Sender, InterpreterOrigin, NetworkId};
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 #[cfg_attr(feature = "std", derive(schemars::JsonSchema))]
@@ -68,7 +61,7 @@ pub struct OtherNetworkItem {
 pub struct NetworkItem {
 	/// something which will be receiver on other side
 	/// case of network has XCVM deployed as contract, account address is stored here
-	pub gateway_to_send_to: Option<GatewayId>,
+	pub gateway_to_send_to: Option<msg::GatewayId>,
 	/// Cosmos bech32 prefix per network,
 	/// if there is prefix chain accounts are Cosmos SDK compatible chain
 	pub cosmos_prefix: Option<String>,
--- a/code/Cargo.lock
+++ b/code/Cargo.lock
@@ -3293,7 +3293,6 @@ name = "cw-xc-gateway"
 version = "0.1.0"
 dependencies = [
  "bech32-no_std",
- "cosmwasm-schema 1.2.5",
  "cosmwasm-std",
  "cw-storage-plus",
  "cw-utils",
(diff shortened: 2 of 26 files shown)


code/xcvm/cosmwasm/contracts/gateway/src/state.rs
```diff
@@ -1,14 +1,13 @@
...
 use cosmwasm_std::{Addr, IbcEndpoint, StdResult, Storage};
 use cw_storage_plus::{Item, Map};
-use schemars::JsonSchema;
-use serde::{Deserialize, Serialize};
 use xc_core::{AssetId, InterpreterOrigin, NetworkId};
 
 pub(crate) type ChannelId = String;
 
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[cw_serde]
```
dzmitry-lahoda, line 10: what is problem with this code. 
as soon as will use Substrate(sp) for doing arbitrary dispatch calls to runtime, or depending on XCM, which is not Scheam (but burried some in Polkadot repo). I wouill not be able to use these types with Schema in same crate. 

Our Polkadot grand tells need for that dep.  Also it hurdens mobing shared code to core. Sure this exact line is not an issue so.

Also, does Schema increase wasm side or it compiled away? Is that metadata stored with code?

mina86: We’re using forked cosmwasm-schema with no_std support though so compatibility with no_std environments shouldn’t be an issue?

mina86: @dzmitry-lahoda, so I’ve spent some time trying to check this but at the moment the contracts don’t compile in no_std environment (I’ve tried building for thumbv6m-none-eabi target).  I’ve looked into this but unless we need to be able to compile the contracts for no_std right now, I’d say let’s go with this PR and then figure out no_std later.  For example, I have some PRs for cosmwasm to support no_std (a bit different approach to your fork) so this should be doable in the future.  For JsonSchema I’m pretty sure that the solution is to have cosmwasm-std not derive it if building no_std.

dzmitry-lahoda: tbh whater no std works consiatently accross cw and ibc and substrate, good.

without std feature and wasm32 target work on my pr.
sure more strict version of embeeded target not.

dzmitry-lahoda: i make all jsonschema conditional in all places and it works consistently ok.

mina86: > without std feature and wasm32 target work on my pr.

But that doesn’t really means the code works without std.  If our core is no_std but a dependency is not than wasm32 build will work fine but it doesn’t mean the code is no_std.  This PR compiles for wasm32 as well.

> i make all jsonschema conditional in all places and it works consistently ok.

So with cw_serde, this just needs to happen in no_std fork of cosmwasm-std.

Also, what’s the status of your PR? ’Cuz it looks like I’m kind of blocked on it?

dzmitry-lahoda: did you had time to review it? if you are blocked please review. what exactly you arr blocked on?

dzmitry-lahoda: i mean it was 100 green, but because of some changes  needed to rebase. i guess blocking parts where here.

mina86: > did you had time to review it? if you are blocked please review. what exactly you arr blocked on?

I have approved it already a while back.  I’m actually not clear on the blocking status, but if your PR improves no_std situation than I guess this PR would benefit from that? Then again, this PR does compile for wasm32 so maybe that’s not an issue.

> i mean it was 100 green, but because of some changes needed to rebase.

And this is another reason to prefer small PRs.  It’s easier to rebase them and merge them quickly. :]

dzmitry-lahoda: i replaced xcvm internal, size of pr depends on size of replacement in grrat degree.

mina86: OK, so what’s the next step with those PRs? As far as I see, this PR doesn’t need any further changes and can be merged.

mina86:
@dzmitry-lahoda, PTAL

dzmitry-lahoda (dismissed):

mina86:
@dzmitry-lahoda, PTAL, had to fix fmt

PR #3804, opened 2023-07-19
Release v8.10033.0

kkast:
release-v8.10033 is currently the same as release 10032 release-v1.10031.0

release-33-changes branch is current main without commit enabling flat fee


dzmitry-lahoda:
nice. do we need it? is it WIP or DRAFT?

PR #3822, opened 2023-07-21
pallet-cosmwasm: remove unnecessary `as uXX` casts on literals

mina86:
In most cases Rust can figure out the type of an integer literal so an explicit `as uXX` cast is not necessary.  In a couple of places where the type is needed to disambiguate types, using typed integer literal is more concise.


dzmitry-lahoda:
that is generated i guess. on regeneration it will be again here.

mina86:
D’oh, you’re right. I somehow missed the giant message at the top.

Issue #3958, opened 2023-07-26
Provide interface for IBC ICS 30 Middleware and describe how it used by PFM, Hooks, Wasm,  (and may be other) to work well


dzmitry-lahoda:
**Motivation**

Need a way to handle PFM/WASM in IBC on Picasso/Composable. Also need way to ensure that XCM integration works as robust as PFM/WASM on Centauri. For this Middleware is used in Cosmos. 

**Suggested Solution**

Update RFC for multihop with desired interface and document usage.

**Alternatives**

Attempted alternative to share several references and also high level description in XCM IBC multi hop is not helfull.

dzmitry-lahoda:
Waiting for feedback and comentary from bridge team, not urgent

dzmitry-lahoda:
assumed provided

PR #3963, opened 2023-07-27
Create 0017-ibc-memo-handling

dzmitry-lahoda:
Required for merge:

Makes review faster:
- [x] Linked Zenhub/Github/Slack/etc reference if one exists  https://github.com/ComposableFi/composable/issues/3958#issue-1822878073
- Clicking all checkboxes 


blasrodri (commented):
If there's a packet to be executed on the CW vm, and it fails, then IBC should roll back that operation. 
As long as the interface allows to roll it back properly, IBC and ICS20 can return an error. Which will be processed by the source chain, and roll back the tokens properly. 
Is this enough?

dzmitry-lahoda:
When ICS20 packet arrives, it has memo.

required :
1. If memo has wasm instruction, and it fails to parse wasm instruction or execution of cw contracts fails, error ack must be returned, assets sent are remaining on sender.   

^^ that is what i read from reading middlewares, i did not looked into go hosts. and this level of control is required for XCVM

what other do as i know, things will generally fail, but would be easy to tell it is not XCVM failure:
2. if there is some not know `next` deep in memo, which is not what we handle and not current level of handling, it should not be parsed nor  that part of memo should lead to memo handling failure, but that part should be propagated futher. it is safe to assume memo is just some json.  
3. if somebody sends pfm via  picasso/composable, `next` thing, it must be handled as async packet. without these XCM will not be PFM compliant (tokens stuck on Picasso/Composable without rollback). given   length of hops in our setup, i would say it needed. 

2 and 3 not must for XCVM, failure will happend in IBC without even reaching XCVM

dzmitry-lahoda:
> If there's a packet to be executed on the CW vm, and it fails, then IBC should roll back that operation.
As long as the interface allows to roll it back properly, IBC and ICS20 can return an error. Which will be processed by the source chain, and roll back the tokens properly.
Is this enough?

I cannot tell you exactly as I not precisely know what you mean. Here is my more rewrite

 If there's a packet to be executed with memo, and memo exectuion fails, then IBC should roll back that operation.
As long as the host allows to roll it tokens and changed from memo transaction as unit, IBC and ICS20 and memo can return an error. Which will be processed by the source chain, and roll back the tokens properly. While on receiver chain there is no changes from memo transaction execution nor tokens.

This is enough to be clear why XCVM failed and prevent partial failure on receiver (so it is clear why XCVM failed.

dzmitry-lahoda:
I will update RFC. @blasrodri anything is unclear yet?

blasrodri:
@dzmitry-lahoda let's focus on (1). As we currently do not **need** to support PFM on our chain, right? But (1) seems like a sensible thing to support

dzmitry-lahoda:
There is no block which will prevent people to send pfm to our chains, so that it more of question to PO what to do that, Asynx xonfirmations. Same decision for PO do not fail arbitrary memos, as it is now. 

I xcvm there is flag if chain supports pfm or cw or both, so any flag set will change deciaoon for xcvm programs. That will be po decision. So xcvm requires arbitrary memo format preferably just arbitrary json, hope i can write around current memo handling in custom middleware, hope.

Transaction support of memo and ics, same as above. Required for valid xcvm programs in case of errors, PO to decide, hope can be coded around, but that harder to do and may need some rewrite.

I am not like talking about now and that, more of clarity about what is IBC cosmos compaible chain people and xcvm programs expect.

blasrodri:
I think that right now we just need to build the basic infra that XCVM requires

blasrodri:
Would (1) be enough for v0 of XCVM work?

dzmitry-lahoda:
I do not understand what you mean by `(1)`

blasrodri:
> I do not understand what you mean by `(1)`

required :

If memo has wasm instruction, and it fails to parse wasm instruction or execution of cw contracts fails, error ack must be returned, assets sent are remaining on sender.

dzmitry-lahoda:
yes, this is needed for consistency, without not sure at all what todo.

so we do not have pfm enabled on composable than. right?

for compliant memo i can do myself, i will rewrite memo things from scratch as need to make it compliant with wasm termination handling and  not failing when it should not.

dzmitry-lahoda:
@blasrodri technically i guess, not validated, transaction can be written in middlware using transact attribute. so in theory do not need to change ics 20 transfer, and memo handling. but for these both it would be full rewrite of all memo handling and middlewares. so it will not be true imlementation on host level context in centauri, but will work same i think.

for async packets, which are essentce of PFM, XCVM may consider 3 options
1. no composable nor picasso nor any parachain are PFM enabled(incuding substrate pfm), so cannot pass through these.
2. only picasso CW enabled, so only here can reach with programs, not PFM, but Polkadto/Moonbeam would be unresachable.
3. enable CW in Composable, than can reach parachains with pure XCVM prograsm
4. consider PFM enabled (including Substrae), and send packets, but consider it will fail without rollbacks.
5. true PFM can only be implemented (or verified it, but I doubt its) only in Centauri code of host/channel handling.

PR #3989, opened 2023-07-31
xcvm: change XCVMAck into an enum; and swap values

mina86:
XCVMAck can hold limited set of values (success or failure) so rather
than writing it as a newtype around u8 change it into an enum.  This
simplifies any code that needs to match on an XCVMAck value since it
doesn’t need to check values greater than 1.

Furthermore, rename the variants to Ok and Fail to better fit Rust
naming and be more distinguishable; and swap the value so that Ok is
zero and Fail is one.

Required for merge:

Makes review faster:
- Clicking all checkboxes


dzmitry-lahoda (approved):
i did that in my PR to btw

PR #4002, opened 2023-08-01
xcvm: actually accept deposits in escrow contract

mina86:
Implement handling of deposits to the Virtual Wallet Escrow contract.
Aspects of receiving funds (native and CW20) and refunding them are
implemented.  IBC communication with the accounts contract is still to
be done.

Required for merge:

Makes review faster:
- Clicking all checkboxes

### DIFF
--- a/code/xcvm/cosmwasm/contracts/escrow/Cargo.toml
+++ b/code/xcvm/cosmwasm/contracts/escrow/Cargo.toml
@@ -18,8 +18,12 @@ cosmwasm-std = { workspace = true, features = ["ibc3"] }
 cw-storage-plus.workspace = true
 cw-utils = { workspace = true, features = ["std"] }
 cw2.workspace = true
+cw20.workspace = true
+derive_more.workspace = true
 parity-scale-codec.workspace = true
+serde.workspace = true
+serde-json-wasm.workspace = true
 strum.workspace = true
 thiserror.workspace = true
 
-xc-core = { path = "../../../lib/core", features = ["std"] }
+xc-core = { path = "../../../lib/core", features = ["std", "cw20"] }
--- a/code/xcvm/cosmwasm/contracts/escrow/src/auth.rs
+++ b/code/xcvm/cosmwasm/contracts/escrow/src/auth.rs
@@ -1,6 +1,7 @@
 //! Module with authorisation checks.
 
 use crate::{
+	assets,
 	error::{ContractError, Result},
 	state,
 };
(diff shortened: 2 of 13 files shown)


code/xcvm/cosmwasm/contracts/escrow/Cargo.toml
```diff
@@ -18,8 +18,10 @@ cosmwasm-std = { workspace = true, features = ["ibc3"] }
...
 cw-utils = { workspace = true, features = ["std"] }
 cw2.workspace = true
+cw20.workspace = true
 parity-scale-codec.workspace = true
+serde-json-wasm.workspace = true
 strum.workspace = true
 thiserror.workspace = true
 
-xc-core = { path = "../../../lib/core", features = ["std"] }
+xc-core = { path = "../../../lib/core", features = ["std", "cw20"] }
```
dzmitry-lahoda, line 27: i see that contacts outside do not make it feature, it just cw20 and native all the time.

code/xcvm/lib/core/src/escrow.rs
```diff
@@ -51,51 +58,18 @@ pub enum ExecuteMsg {
...
-	/// Amount of the asset.
-	pub amount: u128,
-}
-
-/// Local asset identifier.  XXX TODO
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
-#[serde(rename_all = "snake_case")]
-#[cfg_attr(feature = "std", derive(schemars::JsonSchema))]
-pub enum LocalAssetId {
-	Native { denom: String },
```
dzmitry-lahoda, line 89: asset is always native. always. i mean has local denomination. but also it  has external path if it was bridges. more like struct

code/xcvm/cosmwasm/contracts/escrow/src/auth.rs
```diff
@@ -126,6 +150,31 @@ pub(crate) fn handle_break_glass(_: Admin, deps: DepsMut, env: Env, _info: Messa
...
+		asset_id: xc_core::AssetId,
+	}
+
+	impl Cw20Contract {
+		/// Creates a new `Cw20Contract` if the address corresponds to a known
+		/// asset.
+		///
+		/// If it doesn’t, returns [`ContractError::UnknownAsset`] error.
+		pub fn new(_storage: &dyn Storage, _address: Addr) -> Result<Self> {
+			todo!()
```
dzmitry-lahoda, line 173: not sure, how to create contract on address ahead? afaik it is always unpredictalbe

code/xcvm/cosmwasm/contracts/escrow/Cargo.toml
mina86: The reason for separate function is that to have cw20 definitions in xc-core it needs to depend on cw20 crate which is not no_std.  I suppose I could just enable cw20 when std is selected.  Having cw20 as separate feature just thought cleaner I guess?  Thoughts?

code/xcvm/lib/core/src/escrow.rs
mina86: There can be either funds attached to the message and those have denomination or there can be CW20 assets which have contract addresses.  Is that not the case?

code/xcvm/cosmwasm/contracts/escrow/src/auth.rs
mina86: As far as I understand, for CW20 we get messages from CW20 contract and need to be able to understand that contract’s address.  The point of asset registry (which the code is currently lacking) is to translate that address to asset id.  So that’s what this method will eventually do.

code/xcvm/cosmwasm/contracts/escrow/Cargo.toml
dzmitry-lahoda: https://github.com/CosmWasm/cw-plus/pull/874

code/xcvm/lib/core/src/escrow.rs
dzmitry-lahoda: you are right.

dzmitry-lahoda: in terms of cw20 assets to be moved. sure. but for remote assets not sure.

code/xcvm/cosmwasm/contracts/escrow/Cargo.toml
mina86: Right, but that’s cw-plus. cw20 still pulls std in. We really probably should have a `cosmwasm` feature on `xc-core` crate since it should work on any chain. Let’s leave `cw20` for now and see how it goes.

dzmitry-lahoda: ah, yes. features are good, right.

Issue #3952, opened 2023-07-25
cw xc interpreter contracts fails to upload because of cw-vm issue


dzmitry-lahoda:
**Description**

```
Error:           Error: Command failed with exit code 1: bash -c /nix/store/zmg0pdcwqqmdlpa2h5q7zh7kwsszrn74-composable/bin/composable build-spec --chain /tmp/zombie-85c126a71a3702bb9c4f5b7bf88042db_-597335-T8xssAor9VJt/picasso-dev-2087-rococo-local-plain.json --disable-default-bootnode  --raw > /tmp/zombie-85c126a71a3702bb9c4f5b7bf88042db_-597335-T8xssAor9VJt/picasso-dev-2087-rococo-local-raw.json
2023-07-25 10:13:36 Building chain spec    
2023-07-25 10:13:36 assembling new collators for new session 0 at #0    
2023-07-25 10:13:36 assembling new collators for new session 1 at #0    

====================

Version: 8.10032.0-0000000000000000000000000000000000000000

   0: sp_panic_handler::set::{{closure}}
   1: std::panicking::rust_panic_with_hook
   2: std::panicking::begin_panic_handler::{{closure}}
   3: std::sys_common::backtrace::__rust_end_short_backtrace
   4: rust_begin_unwind
   5: core::panicking::panic_fmt
   6: core::result::unwrap_failed
   7: <pallet_cosmwasm::pallet::GenesisConfig<T> as frame_support::traits::hooks::GenesisBuild<T>>::build
   8: std::thread::local::LocalKey<T>::with
   9: sp_state_machine::basic::BasicExternalities::execute_with_storage
  10: <picasso_runtime::GenesisConfig as sp_runtime::BuildStorage>::assimilate_storage
  11: sp_runtime::BuildStorage::build_storage
  12: <sc_chain_spec::chain_spec::ChainSpec<G,E> as sc_chain_spec::ChainSpec>::as_json
  13: sc_service::chain_ops::import_blocks::build_spec
  14: sc_cli::commands::build_spec_cmd::BuildSpecCmd::run
  15: sc_cli::runner::Runner<C>::sync_run
  16: composable_node::command::run
  17: composable::main
  18: std::sys_common::backtrace::__rust_begin_short_backtrace
  19: std::rt::lang_start::{{closure}}
  20: std::rt::lang_start_internal
  21: main
  22: __libc_start_call_main
  23: __libc_start_main@@GLIBC_2.34
  24: _start


[...]

PR #4003, opened 2023-08-01
fixed f32/f64 issue in cosmwasm contracts validation

dzmitry-lahoda:
Some explanation is here https://github.com/cosmos/ibc-rs/pull/803

So wasmd validation in cosmos will not find reference to code (which never executes) it fails like floats anymore and will success upload. 
Any attempt to strip float code failed (likely it anticipates input which may contain floats in some case, which is not case to handle at all).

Also had hard times with std because of centauri repo, added CI checks to ensure we always good.

May be fixes https://github.com/ComposableFi/composable/issues/3952 too.

Required for merge:

Makes review faster:
- Clicking all checkboxes 


code/xcvm/cosmwasm/contracts/gateway/clippy.toml
```diff
@@ -0,0 +1,11 @@
+disallowed-types = [
+    # { path = "usize", reason = "variable size" }, # cannot on now, because serde, even if there is no usize in type
```
mina86, line 2: I don’t even see how usize could be disallowed since that’s what Rust uses for lengths.

mina86 (approved):

code/xcvm/cosmwasm/contracts/gateway/clippy.toml
dzmitry-lahoda: yes, but seems when serde and serde_json operate, they have some switch, which touches usize. 

so not usize bad for serde/json, but usize in api.

dzmitry-lahoda: so tbh half of cases when usize is used can use more concrete size

dzmitry-lahoda: just left comment for people

dzmitry-lahoda: it will forbid usage of usize to cast or declare it, but .len() will work. it is stupid macro on ident declaration, not really parsing type tree seem

code/parachain/node/src/chain_spec/picasso.rs
```diff
@@ -31,10 +31,7 @@ pub fn genesis_config(
 		contracts.push(contract);
 	}
 	if let Some(_contract) = option_env!("CW_XC_INTERPRETER_WASM_PATH") {
```
mina86, line 33:
```suggestion
	if let Some(contract) = option_env!("CW_XC_INTERPRETER_WASM_PATH") {
```

mina86 (dismissed):

dzmitry-lahoda:
because wasmd dislikes f32/f64 and serde is stupid to do these, had do use alt serde , and alt serde are no-std fork because need to share same compilation with dotsama and cw, but because some of deps I named in centauri use same alt serde as transient dep as std, so do not compile it, but because that thing enables std, it enabled it in alt serde (features are global). soooo. 

doing branch of centauri, with fixed (i posted issue to vlad 3 month ago - and now it hits), plus extra check this not to be merged ever again. cc @blasrodri

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
```diff
@@ -109,7 +109,7 @@ impl Gas {
 	}
 
 	fn current_mut(&mut self) -> &mut u64 {
-		self.checkpoints.last_mut().unwrap()
+		self.checkpoints.last_mut().expect("account by external caller")
```
mina86, line 112: This description doesn’t make sense to me.  This never fails because there’s no code path in which checkpoints can be empty when current_mut is called.

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
```diff
@@ -148,7 +148,7 @@ impl Gas {
 		if self.checkpoints.len() < 2 {
 			return Err(NoCheckpointToPop)
 		}
-		let child = self.checkpoints.pop().unwrap();
+		let child = self.checkpoints.pop().expect("account by external caller");
```
mina86, line 151: Ditto.  This never fails because we’ve just checked that length is at least two.

mina86 (approved):

RustNinja (approved):

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
dzmitry-lahoda: if let skip 1 pop

functional handling is awesome

i just fixed linter, i fully agree with liner. 

sure better fix is doing skip 1 pop map ok or err. code was bad. 8 do not fix code

dzmitry-lahoda: now. just ensured linter runs on ci. i am for fixing it in proper skip pop map or or err, but that was not point now. point was to get linter.

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
dzmitry-lahoda: yes, that what i tried to write. there is method. somebody calls it. we assume he knows what he is doing

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
dzmitry-lahoda: i will send functional approach in next pr.

dzmitry-lahoda: it will be fun with mute. my point is that whole plhandling is broken. it is easy to write provable code. just issue is that there are tons of shit code to handle...

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
dzmitry-lahoda: will send prorper handing in next pr. so now unwraps or expects. guide from parity is no unwraps, i am for it.

dzmitry-lahoda: they write expect most of time. btw do not recall saw unwrap

dzmitry-lahoda:
ahahaha
```rust
	fn current_mut(&mut self) -> &mut u64 {
		self.checkpoints.last_mut().expect("account by external caller")
	}


	/// Pops the last gas checkpoint.
	///
	/// Any gas limit remaining in the checkpoint is added back to the parent
	/// checkpoint.  Returns an error if function tries to pop the final
	/// checkpoint.
	pub fn pop(&mut self) -> Result<(), NoCheckpointToPop> {
		if self.checkpoints.len() < 2 {
			return Err(NoCheckpointToPop)
		}
		let child = self.checkpoints.pop().expect("account by external caller");
		let parent = self.current_mut();
		*parent += child;
		Ok(())
	}
```

PR #4004, opened 2023-08-01
functional typed safe correct bounded vectors handling

dzmitry-lahoda:
- improved safety of code of gas checkpoint handling in cosmwasm pallet using non empty vector
- improved const generic bi bounded vec to be more usable for some bounded scenarios
- changed cw pallet config for usage of const so that it can be paired with const traits/structs, like bi bounded vec

Required for merge:

Makes review faster:
- Clicking all checkboxes 


code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
```diff
@@ -86,7 +87,7 @@ pub struct Gas {
 	/// When a new checkpoint is created, gas from the current one is moved to
 	/// the new one (see [`Self::push`]) such that total remaining gas is the
 	/// sum of gas on all checkpoints.
-	checkpoints: Vec<u64>,
+	checkpoints: Vec1<u64>,
```
mina86, line 90: `Vec1`? Not `NonEmptyVec`? It looks like all we need is `pop` method in BiBoundedVec and we don’t need `vec1` dependency.

code/parachain/frame/cosmwasm/src/mock.rs
```diff
@@ -496,10 +496,10 @@ impl PalletHook<Test> for MockHook {
 }
 
 impl Config for Test {
+	const MAX_FRAMES: u16 = 64;
 	type RuntimeEvent = RuntimeEvent;
 	type AccountIdExtended = AccountId;
 	type PalletId = CosmwasmPalletId;
-	type MaxFrames = MaxFrames;
```
mina86, line 502: Why are we changing this as well?

code/parachain/frame/cosmwasm/src/lib.rs
```diff
@@ -208,6 +208,9 @@ pub mod pallet {
 	pub trait Config:
 		frame_system::Config<AccountId = AccountIdOf<Self>> + Send + Sync + Debug
 	{
+		/// Max number of frames a contract is able to push, a.k.a recursive calls.
+		const MAX_FRAMES: u16;
```
blasrodri, line 212: Why not u8?

dzmitry-lahoda: it was u32 some time ago.

but i guess contract gas stack can be in theory u16, but i am alos pretty sure it will never reach u8.

code/parachain/frame/cosmwasm/src/mock.rs
dzmitry-lahoda: because that is improvable for const expressions usage and also this is super tech parameter no worth noising it in api.

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
dzmitry-lahoda: from non several non empty libs, most popular are bounded vec and v1. v1 has small vec integration and has more downloads. bv author abandoned somewhat project and does  ot keeps it updated with stable and nighly features(behind flag), and pop is just one of items, need more on bv.

i already posted request into v1 to be more like bv without loosing ergonomics

dzmitry-lahoda: if non empty to be used further, v1 has more complete api tbh. bv is toy.

code/parachain/frame/composable-support/src/collections/vec/bounded/bi_bounded_vec.rs
```diff
@@ -55,6 +55,16 @@ pub enum BiBoundedVecOutOfBounds {
 	},
 }
 
+impl<T, const U: usize> BiBoundedVec<T, 1, U> {
```
dzmitry-lahoda, line 58: this guy needs > 1. and it gives a lot. by author does no evolves it (  . until comparisons this approach is not yet great.

dzmitry-lahoda: with const vector

code/parachain/frame/cosmwasm/src/mock.rs
dzmitry-lahoda: 8 modify it towards using const bool operators for safety and code simplicity. no reason to use parity thing.

limit will never be dynamic i guess

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
mina86: OK, but I’m just confused about this PR. Why is BiBoundedVec being changed if we end up using vec1? Either change BiBoundedVec to have pop and use that or leave it alone and use vec1 here instead.

code/parachain/frame/composable-support/src/collections/vec/bounded/bi_bounded_vec.rs
```diff
@@ -55,7 +75,65 @@ pub enum BiBoundedVecOutOfBounds {
...
+	///
+	/// # Example
+	/// ```rust
+	/// use composable_support::collections::vec::bounded::BiBoundedVec;
+	/// use sp_std::convert::TryInto;
+	///
+	/// let data: BiBoundedVec<_, 2, 8> = vec![1u8, 2].try_into().unwrap();
+	/// assert_eq!(data.first(), &1);
+	/// ```
+	pub fn first(&self) -> &T {
```
kkast, line 114: why to remove option? wouldnt this crash? dont we have that uses the previous version of first function?

dzmitry-lahoda: we care abound bounded case. for bounde case it is it, fomr 0 bound it is option. see 0 and > impla of first

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
dzmitry-lahoda: I started to change BBV, added pop and stuff. But hit limit that need to change a lot of code to propagate T parameter to Gas/VM structs.

dzmitry-lahoda: that why used V1, but BV is changes are good anyway, just was not able to use it now.

> Why is BiBoundedVec being changed if we end up using vec1?
because changes are good.

>  Either change BiBoundedVec to have pop and use that or leave it alone and use vec1 here instead.
i will not leave it, i improved it. i will not remove improve. i can remove it but that is stupid.

mina86: OK, then BV changes should be moved to a separate PR since they have no baring on the checkpoints.

dzmitry-lahoda: what is the purpose of split?

mina86: Smaller PRs which are easier to understand and ability to easily cherry pick or revert unrelated changes when needed.

dzmitry-lahoda: are you going to revert something from this PR?

mina86: I don’t have a crystal ball so I don’t know what I’m going to be doing in the future.

RustNinja (approved):

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
dzmitry-lahoda: you have. so we can think of:
1. if any functional in that area developed at same time making merge harder
2. if changes are risky
3. if changes are not clear
4. if changes on top of positive things has negative things
5. if changes are really big so not clear what they are doing
6. are changes controversial and opinionated
7. is there alignment among team they are ok
8. to we have CI which changes some unit tests and compilation and at least start up of processes to check changes
9. what process we are working  - i think we deliver directly to prod, how it influences?
10. how size of company influnces?

... i miss some other things in list, you may add.
 
so we have crystal ball in this case. sure we have no in other cases.  but this case has this one.

kkast (approved):

code/parachain/frame/cosmwasm/src/runtimes/abstraction.rs
dzmitry-lahoda: what do you mean changes are not clear?

dzmitry-lahoda: so these simple rust changes are not clear? what you mean too large for what it is doing? what pr is doing? is not it doing what is described? it 10 files btw of simple rust. all changes  are in same area. 

i think even that multiple prs will be less clear as it would be not clear where changes are going.

for me costs are high doing multiple prs time would be spent in  relation to many other places were time can be spent.

mina86: What’s not clear is why all of those changes are group together and how they relate to each other.

Checking size of the PR through CI is not viable since there is no threshold in which PR is too large.  Some big refactoring or automated changes may be very large but easy to understand.  At the same time, some comparatively small PR which has multiple unrelated changes in too big.

With appropriately-sized PRs including a proper commit message, it’s clear that all changes are directly related to each other and what they are trying to achieve. I don’t know what you mean by not knowing where changes are going.

For the time, it take would take maybe five minutes to split. At the same time, self-contained PRs would take much less time to review.

PR #4012, opened 2023-08-03
feat(xcvm): assets handling - docs, events, devnet setup

dzmitry-lahoda:
image

- documented assets id formation, and did relevant code for devnet (and future asset id gen)
- improve naming of assets enum and added draft for adr handling (for multi hop case)
- init assets in osmosis devnet (with generated ids)
- finalized schema with add of XCM to location
- regenerated JSONs with schema (https://github.com/ComposableFi/composable/issues/4014)

Required for merge:

Makes review faster:
- [x] Linked Zenhub/Github/Slack/etc reference if one exists https://github.com/ComposableFi/research/issues/244
- Clicking all checkboxes 


code/xcvm/README.md
```diff
@@ -69,4 +85,14 @@ For each chain and protocol it makes pragmatics hacks to use existing liquidity
...
+
+### Asset id encoding
+
+Each asset id is 128 bit number with 4 first bytes are network id, means that numbers never overlap.
+
+So it will not be the case that on one chain 123213 means PEPA and on other chain 123213 means SHIB.
+
+Prefix allows to find network to look at for asset info.
+
+For security reasons when assets sent from chain to chain, asset id changes.
```
mina86, line 98: IS this for security reasons? It’s more because it becomes a different asset thus needs a different id and different entry in asset registry.

code/xcvm/cosmwasm/contracts/gateway/README.md
```diff
@@ -7,3 +7,6 @@ Hides custom entry points and messaging  under unified security, routing, assets
 
 Is Entrypoint and exit for Programs. Also handles interactions with [OTP](otp.md).
 Book keeps per-chain state. Holds the address mappings for cross-chain accounts and instantiates new interpreters.
+
+
+## Handling cases when there is no contract
```
mina86, line 12:
```suggestion
```

code/xcvm/lib/core/src/asset.rs
```diff
@@ -402,3 +402,37 @@ impl<T> From<Funds<T>> for Vec<(u128, T)> {
...
+	[
+		network_id.0.to_be_bytes().to_vec(),
+		protocol_id.to_be_bytes().to_vec(),
+		nonce.to_be_bytes().to_vec(),
+	]
+	.concat()
+	.try_into()
+	.map(u128::from_be_bytes)
+	.expect("[u8; 8] + bytes(u64) = [u8; 16]")
+	.into()
```
mina86, line 416: There’s no need to allocate a Vector or for all the try_into and
except.  Just do simple arithmetic:
```suggestion
	AssetId::from((u128::from(network_id) << 96) |
		(u128::from(protocol_id) << 64) |
		(u128::from(nonce))
```

code/xcvm/lib/core/src/asset.rs
```diff
@@ -402,3 +402,37 @@ impl<T> From<Funds<T>> for Vec<(u128, T)> {
 			.collect()
 	}
 }
+
+pub fn generate_asset_id(network_id: NetworkId, protocol_id: u32, nonce: u64) -> AssetId {
```
mina86, line 406: This needs doc-comment. What’s protocol_id and nonce?

code/xcvm/README.md
dzmitry-lahoda: read ics20 spec, it becomes different for sole reason of escrow protection of one chain against other. if it would be teleported(zk may be) it would same asset. there is no other reason except security.

dzmitry-lahoda: osmosis does not trust PICA from juno to be PICA from Centauri. if they would it would be same PICA on all chains. no need to 2 ids.

mina86: I think this is worded very strangely.  When you transfer assets between chains, you get a completely different asset on a completely different chain and that’s why asset id must change.

mina86 (approved):

code/xcvm/README.md
dzmitry-lahoda: okey, asset on new chain, why new ID? Centauri has PICA 42. Now sent to Osmosis has 42 too. So from osmosis it can got to JUNO -> Centauir. It would be 42 on each hop.  Why would somebody make it diffenet? it is same assets

100 amount of PICA `Centaur -> Osmosis -> Juno -> Centauri` = `Centauri -> Osmosis -> Stride` = `Centauri -> Osmosis -> Centauri` . All this PICAs are same and can be interchanged (put into same AMM pool for example). Why change id?

mina86: Well then that’s a question to you to explain in the document because at this point I don’t know either.

blasrodri: I think what he means is that on IBC the denom is path dependent.

mina86: Right, so because it’s a different path it’s a different asset and thus has different asset id.  That’s why I think saying that it’s for security reasons that the asset id changes is misleading.  Asset id maps one-to-one with specific denomination/contract/path on specific chain.  Even if we didn’t care about security we would still need to use different asset ids.

dzmitry-lahoda: I explained, because of security reasons :)

dzmitry-lahoda: Path depenance why? I already pointed to ICS20 and exampled.

dzmitry-lahoda: > . Even if we didn’t care about security we would still need to use different asset ids.

that is not true

mina86: In that case I come back to my original confusion and implore you to explain in the document when does the asset id change and for what reasons.  As far as I understand, if I have `port-1/channel-1/denom` and `port-2/channel-2/denom` i have to treat and track those separately.

dzmitry-lahoda: will improve that part in next prs.

dzmitry-lahoda: security is shortest correct explanation. security(until ZK/proofs/trust teleport) leads to path dependence. i guess we can make it clear other way.

mina86: Security and design in ICS20 leads to path dependency and path dependency leads to us having to use different asset ids.  Thus a more straightforward way to describe the situation is that because asset id is dependent on chain where it resides and its full path than sending assets between chains changes their asset id.

If paths didn’t matter in ICS20 than we wouldn’t have different asset ids for the same source asset going through different routes.  Am I mistaken here?

dzmitry-lahoda: feel free to improve docs, sure it is better. but i was thinking more simple.

security  -> asset id per chain
rather than
security  -> path -> asset id per chain

same work for XCM btw. 

XCM has teleport, in this case there is no need to be path dependant no need to have asset id to change. it can be asset id chanerated on source chain and move all along teleports (sure when it is moved one hop path dependant magic property lots).

so for me it is assets changes its identity when it it propagates overnetwork.

there is no other reason except security to make assets different ids.

when secuirty mitigated (ZK, proofs, teleports of trust, stables) -> no path - no escrows -> id is same in trust boudaries.

so feel free to improve as others (FE/Indexers/Dev) may like your explanatin

mina86: But ‘security -> asset id per chain’ isn’t simpler.  It’s more confusing because it obscures important design behind asset ids, that there’s bijection between them and supported (network, denom/route) pairs.  So because transferring assets changes their (network, denom/route) that forces a change of asset id.

> XCM has teleport, in this case there is no need to be path dependant no need to have asset id to change. it can be asset id chanerated on source chain and move all along teleports (sure when it is moved one hop path dependant magic property lots).

But still, if you transfer an asset, it’s on a different network thus its asset id changes, doesn’t it?

> so for me it is assets changes its identity when it it propagates over network.
>
> there is no other reason except security to make assets different ids.

The reason for different id is because assets change their paths and networks when they are transferred.

dzmitry-lahoda: hope you will find time to document our discussions for others.

> But still, if you transfer an asset, it’s on a different network thus its asset id changes, doesn’t it?

PICA on picasso 0000000000_000000000001
PICA on Composable 0000000001_000000000001
PICA on Centauri 0000000002_000000000001
PICA on Osmosis 0000000003_000000000001

If we would make them teleport within Composable topology (full trust or ZK proofs)
PICA on picasso 0000000000_000000000001
PICA on Composable 0000000000_000000000001
PICA on Centauri 0000000000_000000000001
PICA on Osmosis 0000000003_000000000001

We could peek any network as source btw.

dzmitry-lahoda: > The reason for different id is because assets change their paths and networks when they are transferred.

So in case of teleport you can use just source asset as ID. 
But if if assets leaves secured topology, it gets new ID.

dzmitry-lahoda: > If paths didn’t matter in ICS20 than we wouldn’t have different asset ids for the same source asset going through different routes. Am I mistaken here?

So it this case all networks would trust each other (or ZK) about how much each network has how much asset, because path will not bind asses  to escrow/mint, but work on burn/mint.r 

So root cause is security, solution is ESCROW+ Path Dependency.

dzmitry-lahoda: > The reason for different id is because assets change their paths and networks when they are transferred.

you can also look into it as on  derived assets, all derived assets has different ID, bridges are just derivations and pools.

dzmitry-lahoda: > The reason for different id is because assets change their paths and networks when they are transferred.
Reason is that security mechanism is path+escrow. 

Changing network is not reason.

ZK burn would not lead to asset id change.

PR #4028, opened 2023-08-07
feat(ibc-cw): correct transfer programs encoded and step closer to final execution 

dzmitry-lahoda:
So here is one step closer attempt in script to get XCVM demo transfer. 
Script now forms correct program, transfers funds into it, so yet failing on IBC decoding. 
Fixing in next PR.

some chores:
- added command to kill all devnet fast
- added some debug improvements
- improved naming for some code in xcvm
- rerun schema generator
- simpler XCVM programs in JSON (hex salt/tag in all places)

Required for merge:

Makes review faster:
- Clicking all checkboxes 


code/xcvm/lib/core/src/instruction.rs
```diff
@@ -77,6 +77,7 @@ pub enum Instruction<Payload, Account, Assets> {
 	/// The salt is used to track the program when events are dispatched in the network.
 	Spawn {
 		network: crate::network::NetworkId,
+		#[serde(serialize_with = "hex::serialize", deserialize_with = "hex::deserialize")]
 		salt: Vec<u8>,
```
mina86, line 81: This is missing schemars `with` annotation.

code/xcvm/lib/core/src/proto.rs
```diff
@@ -650,9 +650,9 @@ where
 	}
 }
 
-// TODO(aeryz): This can be a helper function in SDK so that users won't
-// necessarily need to know how the ratio is handled in our SDK
-// Calculates `x` in the following equation: nom / denom = x / max
+/// This can be a helper function in SDK so that users won't
+/// necessarily need to know how the ratio is handled in our SDK
+/// Calculates `x` in the following equation: nom / denom = x / max
```
mina86, line 655: But why remove TODO?

code/xcvm/lib/core/src/prelude.rs
```diff
@@ -5,8 +5,8 @@ pub use alloc::{
 	vec,
 	vec::Vec,
 };
-pub use core::str::FromStr;
-pub use cosmwasm_std::{Addr, Binary, Coin, Uint128};
+pub use core::{fmt::Display, str::FromStr};
+pub use cosmwasm_std::{Addr, Binary as Base64Binary, Coin, HexBinary, Uint128};
```
mina86, line 9: No.  Having prelude is bad enough.  Having prelude where a type is renamed is definitely a NAK.

code/xcvm/lib/core/src/instruction.rs
dzmitry-lahoda: may you show example?

code/xcvm/lib/core/src/prelude.rs
dzmitry-lahoda: what nak means?

dzmitry-lahoda: it is type alias. so we should not use type aliases?

dzmitry-lahoda: if we use any existing type alias in prelude than we should not use that too?

why?

mina86: NAK = Negative Acknowledgement.

We should not be using preludes or wildcard imports at all since they make reading code harder.  But since you insist on using preludes, we definitely should not invent new names in a prelude.

code/xcvm/lib/core/src/instruction.rs
mina86:
```
$ git grep schemars.\*with
xcvm/lib/core/src/network.rs:   #[cfg_attr(feature = "std", schemars(with = "String"))]
xcvm/lib/core/src/network.rs:   #[cfg_attr(feature = "std", schemars(with = "String"))]
xcvm/lib/core/src/packet.rs:    #[cfg_attr(feature = "std", schemars(with = "String"))]
xcvm/lib/core/src/packet.rs:    #[cfg_attr(feature = "std", schemars(with = "String"))]
```

code/xcvm/lib/core/src/prelude.rs
dzmitry-lahoda: so we should not use type aliases in prelude?

dzmitry-lahoda: Let align with others on the case

code/xcvm/lib/core/src/instruction.rs
dzmitry-lahoda: thanks, seems this is no documented on schemars. used what documented.

dzmitry-lahoda: equivalent

mina86 (approved):

PR #4041, opened 2023-08-08
feat(cw-xc): faster cosmos devnet with fix for interpreter instantiation

dzmitry-lahoda:
So I tuned some bash configs  of XCVM, improve startup time of cosmos net for dev, and continue to run transfer.
Transfer failed because of  https://github.com/CosmWasm/cosmwasm/issues/1810 . So fix was to make short name for interpreter. Hash may fail too, so used counter id.

I think in next iter  should use network prefixed counter. So that can have global interpreter ids, but no must.

Tested by running cosmos devnet oneliner to setup xcvm. And forced create interpreter for user. 

Also I split XCVM admin key and other admin key, so these processes can do tx in parallel, as it visible in devnet init graph.

Also started doing some logging for sanity
```
{"level":"info","process":"centauri","replica":1,"message":"{\"force_instantiate\":{\"user_origin\":\"centauri1e0nh3davd6d3yh73sldwndww7zchetwsae8m9rz9x6hlg7w4ysrqhmaqju\"}}"}
{"level":"info","process":"centauri","replica":1,"message":"xcvm:: {\"id\":0,\"result\":{\"ok\":{\"events\":[{\"type\":\"instantiate\",\"attributes\":[{\"key\":\"_contract_address\",\"value\":\"centauri1tlcw3llaxzs3jam7yheqxhr3k0n4y5sfrnah72n6aqt0rlp4xq6qpjy5cc\"},{\"key\":\"code_id\",\"value\":\"2\"}]},{\"type\":\"wasm-xcvm.interpreter\",\"attributes\":[{\"key\":\"_contract_address\",\"value\":\"centauri1tlcw3llaxzs3jam7yheqxhr3k0n4y5sfrnah72n6aqt0rlp4xq6qpjy5cc\"},{\"key\":\"data\",\"value\":\"eyJ1c2VyX29yaWdpbiI6eyJuZXR3b3JrX2lkIjoyLCJ1c2VyX2lkIjoiNjM2NTZlNzQ2MTc1NzI2OTMxNjUzMDZlNjgzMzY0NjE3NjY0MzY2NDMzNzk2ODM3MzM3MzZjNjQ3NzZlNjQ3Nzc3Mzc3YTYzNjg2NTc0Nzc3MzYxNjUzODZkMzk3MjdhMzk3ODM2Njg2YzY3Mzc3NzM0Nzk3MzcyNzE2ODZkNjE3MTZhNzUifSwic2FsdCI6IjY0NjU2NjYxNzU2Yzc0In0=\"}]}],\"data\":\"CkNjZW50YXVyaTF0bGN3M2xsYXh6czNqYW03eWhlcXhocjNrMG40eTVzZnJuYWg3Mm42YXF0MHJscDR4cTZxcGp5NWNj\"}}}"}
{"level":"info","process":"centauri","replica":1,"message":"xcvm:: saved interpreter"}
````

Required for merge:

Makes review faster:
- Clicking all checkboxes 

### CODE REVIEW

File: code/xcvm/cosmwasm/contracts/gateway/src/interpreter.rs [hunk 1]
```diff
@@ -0,0 +1,114 @@
...
+) -> Result {
+	let config = load_this(deps.storage)?;
+	let interpreter_code_id = match config.gateway.expect("expected setup") {
+		GatewayId::CosmWasm { interpreter_code_id, .. } => interpreter_code_id,
+	};
+
+	let call_origin = CallOrigin::Local { user: user_origin };
+	let interpreter_origin = InterpreterOrigin {
+		user_origin: call_origin.user(config.network_id),
+		salt: b"default".to_vec(),
```
mina86, line 32: Should salt be argument instead? (And if we don’t care about salt here maybe just set it to empty value?)

code/xcvm/cosmwasm/contracts/gateway/src/interpreter.rs
```diff
@@ -0,0 +1,114 @@
...
+	let next_interpreter_id: u128 = state::interpreter::INTERPRETERS_COUNT.load(deps.storage)? + 1;
+	let instantiate_msg = WasmMsg::Instantiate2 {
+		admin: Some(admin.clone().into_string()),
+		code_id: interpreter_code_id,
+		msg: to_binary(&cw_xc_interpreter::msg::InstantiateMsg {
+			gateway_address: admin.clone().into_string(),
+			interpreter_origin: interpreter_origin.clone(),
+		})?,
+		funds: vec![],
+		label: ["xcvm_interpreter", &next_interpreter_id.to_string()].join("_"),
```
mina86: Since you’re using `to_string` anyway, this is probably better as format:

```suggestion
		label: format!("xcvm_interpreter_{}", next_interpreter_id),
```

code/xcvm/cosmwasm/contracts/gateway/src/interpreter.rs
```diff
@@ -0,0 +1,114 @@
...
+		salt: to_binary(&interpreter_origin.to_string().as_bytes())?,
```
mina86: Why are we serialising salt? Shouldn’t this be:

```suggestion
		salt: interpreter_origin.to_string().into_bytes().into(),
```

code/xcvm/lib/core/src/gateway/mod.rs
```diff
@@ -53,6 +53,19 @@ pub enum ExecuteMsg {
 	MessageHook(XcMessageData),
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+#[serde(rename_all = "snake_case")]
+#[cfg_attr(feature = "std", derive(schemars::JsonSchema))]
+pub enum TestSubMsg {
```
mina86, line 59:
```suggestion
#[cfg(test)]
pub enum TestSubMsg {
```
?

dzmitry-lahoda: oh, will delete, used for tests

code/xcvm/cosmwasm/contracts/gateway/src/interpreter.rs
dzmitry-lahoda: cw is allocating strings like a hell overall, it is too annoying to think when or when not and do such fixes, as long as cw is json, it is premature optimization to optimize format.

mina86: That’s true but in this instance format is also shorter and more readable.

code/xcvm/cosmwasm/contracts/gateway/src/interpreter.rs
```diff
@@ -29,9 +30,15 @@ pub(crate) fn force_instantiate(
 	let call_origin = CallOrigin::Local { user: user_origin };
 	let interpreter_origin = InterpreterOrigin {
 		user_origin: call_origin.user(config.network_id),
-		salt: b"default".to_vec(),
+		salt: salt.clone().map(|x| x.into_bytes()).unwrap_or_default(),
```
mina86, line 33: How about

```
let salt = salt.map_or(Vec::new(), |salt| salt.into_bytes());
```

at the beginning of the function and then you don’t have to repeat
this map operation in two places (on line 33 and 40).

code/xcvm/lib/core/src/gateway/config.rs
```diff
@@ -163,7 +163,8 @@ pub enum ConfigSubMsg {
 	ForceRemoveAsset { asset_id: AssetId },
 
 	/// instantiates default interpreter on behalf of user
-	ForceInstantiate { user_origin: Addr },
+	/// `salt` - human string, converted to hex or base64 depending on implementation
+	ForceInstantiate { user_origin: Addr, salt: Option<String> },
```
mina86, line 167: Perhaps:

```suggestion
	ForceInstantiate {
		user_origin: Addr,
		#[serde(skip_serializing_if = "String::if_empty", default)]
		salt: String,
	},
```

This makes the code simpler since we don’t need to worry about `None`
values and the end effect is the same, i.e. lack of salt is
interpreted as empty value.

mina86 (approved):

code/xcvm/lib/core/src/gateway/config.rs
mina86: Also, should salt be `Binary` instead?

code/xcvm/cosmwasm/contracts/gateway/src/interpreter.rs
dzmitry-lahoda: fixed in current PR i am on.

Issue #4052, opened 2023-08-10
feat(devnet): improve cosmos devnet startup time by baking cw contracts into genesis 


dzmitry-lahoda:
**Motivation**

I need to debug a lot complex CW contracts setup.

**Suggested Solution**

To speed up setup, either use CW batch interface or Cosmos batch interfaces.

**Alternatives**

Start/fix unit/simulator testing.
 
Fix redeploy script (so can deploy any times and get new ids all the time), so no need to restart network.

**Additional Information**

dzmitry-lahoda:
so plan is, to use batch from pr. merge contracts into genesis (export json, see how it is -> dasel). it will make start up  in moment. 

genesis in theory can be repeated for relay later, but this is not part of this work (relays are more subtle things)

PR #4070, opened 2023-08-14
feat(xcvm): batch config msg + release of wasm files

dzmitry-lahoda:
Required for merge:

Makes review faster:
- [x] Linked Zenhub/Github/Slack/etc reference if one exists https://github.com/CosmWasm/cosmwasm/discussions/1814 https://github.com/ComposableFi/composable/issues/4052
- Clicking all checkboxes 

### DIFF
--- a/code/xcvm/cosmwasm/contracts/gateway/src/contract/execute.rs
+++ b/code/xcvm/cosmwasm/contracts/gateway/src/contract/execute.rs
@@ -1,5 +1,6 @@
 use crate::{
 	assets, auth,
+	batch::BatchResponse,
 	error::{ContractError, Result},
 	events::make_event,
 	interpreter, msg,
@@ -8,8 +9,8 @@ use crate::{
 };
 
 use cosmwasm_std::{
-	ensure, entry_point, wasm_execute, Addr, BankMsg, Coin, CosmosMsg, Deps, DepsMut, Env,
-	MessageInfo, Response,
+	entry_point, wasm_execute, Addr, BankMsg, Coin, CosmosMsg, Deps, DepsMut, Env, MessageInfo,
+	Response,
 };
 use cw20::{Cw20Contract, Cw20ExecuteMsg};
 
@@ -28,7 +29,7 @@ pub fn execute(deps: DepsMut, env: Env, info: MessageInfo, msg: msg::ExecuteMsg)
 	match msg {
 		ExecuteMsg::Config(msg) => {
 			let auth = auth::Admin::authorise(deps.as_ref(), &info)?;
-			handle_config_msg(auth, deps, msg, &env)
+			handle_config_msg(auth, deps, msg, &env).map(Into::into)
 		},
 
 		msg::ExecuteMsg::ExecuteProgram { execute_program, tip } =>
@@ -60,7 +61,12 @@ pub fn execute(deps: DepsMut, env: Env, info: MessageInfo, msg: msg::ExecuteMsg)
 	}
 }
 
-fn handle_config_msg(auth: auth::Admin, mut deps: DepsMut, msg: ConfigSubMsg, env: &Env) -> Result {
+fn handle_config_msg(
+	auth: auth::Admin,
+	mut deps: DepsMut,
+	msg: ConfigSubMsg,
+	env: &Env,
+) -> Result<BatchResponse> {
 	deps.api.debug(serde_json_wasm::to_string(&msg)?.as_str());
 	match msg {
 		ConfigSubMsg::ForceNetworkToNetwork(msg) =>
@@ -79,15 +85,10 @@ fn handle_config_msg(auth: auth::Admin, mut deps: DepsMut, msg: ConfigSubMsg, en
 			salt,
 		),
 		ConfigSubMsg::Force(msgs) => {
-			let mut aggregated = Response::new();
+			let mut aggregated = BatchResponse::new();
 			for msg in msgs {
 				let response = handle_config_msg(auth, deps.branch(), msg, env)?;
-				ensure!(
-					response.attributes.is_empty() && response.data.is_none(),
-					ContractError::BatchedCallsCannotReturnData
-				);
-				aggregated =
-					aggregated.add_submessages(response.messages).add_events(response.events);
+				aggregated.merge(response);
 			}
 			Ok(aggregated)
 		},
--- a/code/xcvm/cosmwasm/contracts/gateway/src/contract/execute.rs
+++ b/code/xcvm/cosmwasm/contracts/gateway/src/contract/execute.rs
@@ -16,7 +16,7 @@ use cw20::{Cw20Contract, Cw20ExecuteMsg};
 use xc_core::{gateway::ConfigSubMsg, CallOrigin, Displayed, Funds, InterpreterOrigin};
 
 #[cfg_attr(not(feature = "library"), entry_point)]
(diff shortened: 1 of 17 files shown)


code/xcvm/cosmwasm/contracts/gateway/src/interpreter.rs
```diff
@@ -18,7 +18,7 @@ use crate::{auth, prelude::*};
 pub(crate) fn force_instantiate(
 	_: auth::Admin,
 	gateway: Addr,
-	deps: DepsMut,
+	deps: &mut DepsMut,
```
blasrodri, line 21: Is this the standard?

dzmitry-lahoda: you mean? what part of it? i just did config batch simplest possible way from config messages so can config and reconfig whole xcvm using one json tx

code/xcvm/cosmwasm/contracts/gateway/src/contract/execute.rs
```diff
@@ -60,7 +60,12 @@ pub fn execute(deps: DepsMut, env: Env, info: MessageInfo, msg: msg::ExecuteMsg)
 	}
 }
 
-fn handle_config_msg(auth: auth::Admin, deps: DepsMut, msg: ConfigSubMsg, env: Env) -> Result {
+fn handle_config_msg(
+	auth: auth::Admin,
+	deps: &mut DepsMut,
```
mina86, line 65: This should still work with plain `mut deps: DepsMut`, no? This will leave more code unchanged.

code/xcvm/cosmwasm/contracts/gateway/src/interpreter.rs
mina86: Alternative would be to use `deps.fork()` which clones the DepsMut.  That’s probably the intended way to deal with situations like this but really there’s probably very little difference.

code/xcvm/cosmwasm/contracts/gateway/src/contract/execute.rs
```diff
@@ -71,8 +71,26 @@ fn handle_config_msg(auth: auth::Admin, deps: DepsMut, msg: ConfigSubMsg, env: E
...
+			salt,
+		),
+		ConfigSubMsg::Force(msgs) => {
+			let mut aggregated = Response::new();
+			for msg in msgs {
+				let response = handle_config_msg(auth, deps.branch(), msg, env)?;
+				ensure!(
+					response.attributes.is_empty() && response.data.is_none(),
+					ContractError::BatchedCallsCannotReturnData
+				);
```
mina86, line 88: The correct way to deal with this not being allowed is making it impossible to express in code for example by introducing:

```rust
struct ConfigMsgResponse {
    pub messages: Vec<SubMsg>,
    pub events: Vec<Event>,
}

impl ConfigMsgResponse {
    pub fn add_message(mut self, msg: impl Into<CosmosMsg>) -> Self {
        self.messages.push(SubMsg::new(msg))
        self
    }

    pub fn add_submessage(mut self, msg: SubMsg<T>) -> Self {
        self.messages.push(msg);
        self
    }

    pub fn add_event(mut self, event: Event) -> Self {
        self.events.push(event);
        self
    }

    pub fn merge(&mut self, other: Self) {
        self.messages.extend(other.messages.drain(..));
        self.events.extend(other.events.drain(..));
    }
}

impl From<ConfigMsgResponse> for Response {
    fn from(resp: ConfigMsgResponse) -> Self {
        let mut result = Self::new();
        result.messages = resp.messages;
        result.events = resp.events;
        result
    }
}
```rust

and then, after changing all methods to return `Result<ConfigMsgResponse>`, this method will be something along the lines of:

```rust
fn handle_config_msg(auth: auth::Admin, mut deps: DepsMut, msg: ConfigSubMsg, env: &Env) -> Result {
	match msg {
		ConfigSubMsg::ForceAsset(msg) => assets::force_asset(auth, deps, msg),
		...,
		ConfigSubMsg::Force(msgs) => {
			let mut aggregated = ConfigMsgResponse::new();
			for msg in msgs {
				aggregated.merge(handle_config_msg(auth, deps.branch(), msg, env)?);
			}
			OK(aggregated)
		},
	}.map(Response::from)
}
```

dzmitry-lahoda: i can do it in next prs if patter will repeat. for me current approach is minimally intrusive and aligns with cw coding practices. 

i agree  that special object is eventually better, but i am pretty sure this code is half complete for seamless replacement of buildin response with custom, and in the end can end up that why use response at all, we could use custom non reponse object to instance only topology, not instanti which has message. so in theory no need messages too. 

so i agree with approach in general, but i do not think it should be done now.

mina86: The current approach is less code but it exposes and internal bug as an external contract error.  It introduces failure condition which can easily be avoided.  I don’t see why this shouldn’t be done now.  If messages aren’t needed either that’s even better since messages handling can be removed as well.

code/xcvm/cosmwasm/contracts/gateway/src/interpreter.rs
blasrodri: I'm more asking about the `&mut` on DeptsMut
https://github.com/deus-labs/cw-contracts/blob/main/contracts/cw-quadratic-funding/src/contract.rs#L18

dzmitry-lahoda: i fixed, used `fork` as suggested michal, i guess that is standard

code/xcvm/cosmwasm/contracts/gateway/src/contract/execute.rs
dzmitry-lahoda: message is needed for  creating contracts, interpreter. admin creates interpreter. in theory can refactor out that method. because i do not really need it in deploy, but that is other think. so to have batch need to change many things - a lot . to make it typesafe. which is not CW coding practice.

if CW will provide batch, it may be as I coded.

dzmitry-lahoda: btw,
```rust
struct BatchMsgResponse {
    inner: Response,
}
```
could be even better

mina86: You don’t need to change a lot of things to have this be type-safe.  You need `ConfigMsgResponse` as in my previous comment and then change `force_` methods to return it rather than `Response`.  That’s not such a high price to pay when it eliminates an error-prone construct which this PR exposes as part of public API.

code/xcvm/cosmwasm/contracts/gateway/src/error.rs
```diff
@@ -76,6 +76,9 @@ pub enum ContractError {
 	GatewayForNetworkNotFound(NetworkId),
 	#[error("Anonymous calls can do only limitet set of actions")]
 	AnonymousCallsCanDoOnlyLimitedSetOfActions,
+	/// use events attributes to return data
+	#[error("Batched calls cannot return data")]
+	BatchedCallsCannotReturnData,
```
mina86, line 81: This is no longer used

mina86 (approved):

code/xcvm/cosmwasm/contracts/gateway/src/error.rs
dzmitry-lahoda: fixed in next pr

PR #4094, opened 2023-08-21
chore(ci): remove duplicate dependencies warning noise

mina86:
At the moment build-all-misc-packages CI produces *a lot* of warnings
about duplicate dependencies.  Over 8k lines in fact.  Those warnings
aren’t really useful and only make it harder to view through the logs
when the CI step fails.

This should remove those noisy warnings.

So far no one cared about the duplicate crates.  AS such, removing the
warnings shouldn’t have any negative impact.  If will to fix the
situation ever materialises, those warnings should be addressed and
then turned into errors.

Required for merge:

Makes review faster:
- Clicking all checkboxes

### DIFF
--- a/code/deny.toml
+++ b/code/deny.toml
@@ -148,8 +148,10 @@ registries = [
 # More documentation about the 'bans' section can be found here:
 # https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html
 [bans]
-# Lint level for when multiple versions of the same crate are detected
-multiple-versions = "warn"
+# Lint level for when multiple versions of the same crate are detected.  This
+# warning is too noisy to be useful at this point.  As of August 2023 it outputs
+# 8k lines and only pollutes CI.  Because of that, allow multiple versions.
+multiple-versions = "allow"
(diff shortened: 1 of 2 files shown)


dzmitry-lahoda:
no sure. we have issues aleady. we should fix it, not hide. and make it hard error.

dzmitry-lahoda:
you can run locally as dev any ci checks to get, you can use quiet. no need to change ci which shows all issues.

dzmitry-lahoda (changes requested):
do not hide issue, as it is lead to 2 issues already mentioned in slacks channels. 

either fix issue or run as dev locally, as we dicussed there is no reason why CI must be same as local dev.

tbh i think CI must show all issues and tech debts we have by default.

dzmitry-lahoda:
another option to suppress separate crates. 

please do not start disabling CI warnings as @vimukthi-git did  which

dzmitry-lahoda:
https://composablefinance.slack.com/archives/C02MSA16DCP/p1690552515773059

https://composablefinance.slack.com/archives/C053UJMGT1R/p1692609031644119

PR #4119, opened 2023-09-07
feat(xcvm): simplify apply_bindings function

mina86:
Rather than dealing with pre-initialised buffer and having to track
indexes when setting data in the output payload, change the
apply_bindings function to append data into a vector which does index
tracking by itself.  This removes offset tracking from the function
simplifying it.  With that, change the function so it no longer takes
output vector as argument but rather allocates vector internally.

While changing the apply_bindings function in xc_core also change how
it’s used in the interpreter contract refactoring the code
slightly. Part of it is a consequence of changing the signature of the
apply_binding function but partially it’s just refactoring splitting
functions into smaller, more manageable chunks.

Required for merge:

Makes review faster:
- Clicking all checkboxes

### DIFF
--- a/code/Cargo.lock
+++ b/code/Cargo.lock
@@ -6632,9 +6632,6 @@ name = "lazy_static"
 version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e2abad23fbc42b3700f2f279844dc832adb2b2eb069b2df918f455c4e18cc646"
-dependencies = [
- "spin 0.5.2",
-]
 
 [[package]]
 name = "lazycell"
@@ -12050,7 +12047,6 @@ dependencies = [
  "cosmwasm-std",
  "frame-support",
  "ibc 0.43.1",
- "lazy_static",
  "parity-scale-codec",
  "scale-info",
  "serde",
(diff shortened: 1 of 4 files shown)


dzmitry-lahoda:
Analogous, in semantics, approach work in Solidty EVM? As I recall design of feature (i do not know coding) was deterministic low on resource usage as require by Ethereum.

mina86:
I don’t see why this wouldn’t work for other blockchains. Fundamentally this does the same thing as the previous code.  There’s just less code for tracking indexes and offsets and the vector resizes dynamically if needed. There could be some option to optimise it by using vectored output (so instead of collecting everything to single buffer have a `Vec<Cow<[u8]>>`) but whether that would require more support from other parts of the infrastructure.

dzmitry-lahoda:
Let me write program which checks binding work(or not work), and test out same program on this one.

mina86:
@dzmitry-lahoda, friendly ping

dzmitry-lahoda:
yeah, it is in the queue. need to test on devnet. to avoid breaking it (sure it may not work now, but I saw it working - so need to test before and after). i  guess when write doc.

dzmitry-lahoda:
will test this week

dzmitry-lahoda (approved):

PR #4145, opened 2023-09-17
feat(cvm): error handling of stuck funds

dzmitry-lahoda:
So when packet is sent, it stores mapping for interpreter <-> packet. 
In case of packet timeout/error, error with funds on gateway send back to interpreter.  

Cleanup is not implemented now. Clean up must solve clean up of good state too.

Required for merge:

Makes review faster:
- Clicking all checkboxes 


code/xcvm/cosmwasm/contracts/gateway/src/contract/sudo.rs
```diff
@@ -1,19 +1,42 @@
...
+
+/// return funds to interpreter and sets final error
+fn handle_transport_failure(
+	deps: DepsMut,
+	_env: Env,
+	channel: ChannelId,
+	sequence: u64,
+	reason: String,
+) -> Result<cosmwasm_std::Response, ContractError> {
+	let msg = cw_xc_interpreter::msg::ExecuteMsg::SetErr { reason };
```
blasrodri, line 35: can we have a log here?

blasrodri (commented):
Can you have some tests were there's timeout, or another error and we track that the state has been properly stored?

dzmitry-lahoda:
@blasrodri what kind of tests you are suggesting?

dzmitry-lahoda:
i test only manually once, usually when write docs or deploy to mainnet.

blasrodri:
> i test only manually once, usually when write docs or deploy to mainnet.

For now, something simple such as:
- received a packet with success==false -> tracks state
- received timeout -> tracks state

Mostly will serve as a guard to catch regressions

code/xcvm/cosmwasm/contracts/gateway/src/contract/sudo.rs
```diff
@@ -1,19 +1,42 @@
...
+			} else {
+				Ok(Response::new())
+			},
+		SudoMsg::IBCLifecycleComplete(IBCLifecycleComplete::IBCTimeout { channel, sequence }) =>
+			handle_transport_failure(deps, env, channel, sequence, "timeout".to_string()),
 	}
 }
+
+/// return funds to interpreter and sets final error
+fn handle_transport_failure(
```
blasrodri, line 28: why not pass the `msg` as argument instead?

code/xcvm/cosmwasm/contracts/gateway/src/contract/sudo.rs
blasrodri: This would avoid the need of passing the string `timeout`

code/xcvm/lib/core/src/transport/ibc/mod.rs
```diff
@@ -84,19 +96,30 @@ pub fn to_cw_message<T>(
...
-				contract_addr: addr.into(),
-				msg: to_binary(&transfer)?,
-				funds: <_>::default(),
-			}
-			.into())
+			Ok((
+				WasmMsg::Execute {
+					contract_addr: addr.into(),
+					msg: to_binary(&transfer)?,
+					funds: <_>::default(),
```
blasrodri, line 103: why is it correct to pass a default here?

code/xcvm/lib/core/src/transport/ibc/mod.rs
```diff
@@ -84,19 +96,30 @@ pub fn to_cw_message<T>(
...
+				}
+				.into(),
+				TransportTrackerId::Ibc {
+					channel_id: route.channel_to_send_over.clone(),
+					sequence: 42,
```
blasrodri, line 108: same here: why hardcode?

code/xcvm/cosmwasm/contracts/gateway/src/contract/sudo.rs
dzmitry-lahoda: i added log. i still send "timeout". because i send some description of error to interpreter with assets, so it sets error register to final error whatever, which is string.

code/xcvm/lib/core/src/transport/ibc/mod.rs
dzmitry-lahoda: commented.

code/xcvm/lib/core/src/transport/ibc/mod.rs
dzmitry-lahoda: because precompile impl maps coing to pallet-assets and transfers from sender. no need to send 2 times. btw same with cosmos

blasrodri (dismissed):

code/xcvm/cosmwasm/contracts/gateway/src/contract/sudo.rs
dzmitry-lahoda: because message is both success and failure and has different interface and SUDO can be other then IBC.

so for handling just IBC failrues there is only specific data taken from context.

dzmitry-lahoda:
```
For now, something simple such as:

received a packet with success==false -> tracks state
received timeout -> tracks state
Mostly will serve as a guard to catch regressions
```

@blasrodri you mean automated integration on DevNet, automated unit on contract (and VM level), or some manual steps documented?

blasrodri:
> ```
> For now, something simple such as:
> 
> received a packet with success==false -> tracks state
> received timeout -> tracks state
> Mostly will serve as a guard to catch regressions
> ```
> 
> 
>      
>   
> 
> @blasrodri you mean automated integration on DevNet, automated unit on contract (and VM level), or some manual steps documented?

Integration on DevNet would be nice! But if it's too much of a hassle, at least a unit test.

dzmitry-lahoda:
> Integration on DevNet would be nice! But if it's too much of a hassle, at least a unit test.

Tbh both are hassle. 

Unit test does not makes sence because it is pure integration (depends on all API work as understood),
nad Integraiton is hard because Centauri and Osmosis chains has different API(because osmosis uses old SDK), any test will not support client to Picasso, and hard to pick go vs rust as test lang.

So I use example for tests. Will add to document. So one can configure small timeout, run program, and observer logs, manually so.

PR #4155, opened 2023-09-25
chore(cvm): EVM contract is built on CI via foundry

dzmitry-lahoda:
Required for merge:

Makes review faster:
- Clicking all checkboxes 


code/xcvm/evm/.gitignore
```diff
@@ -0,0 +1,14 @@
...
+cache/
+out/
+
+# Ignores development broadcast logs
+!/broadcast
+/broadcast/*/31337/
+/broadcast/**/dry-run/
+
+# Docs
+docs/
```
mina86, line 11: Why are we ignoring docs?

mina86 (approved):

code/xcvm/evm/.gitignore
dzmitry-lahoda: it is default generated by foundry. whole ignore. i do not know why they do it. will not dig into their reasons why. may be because of how they handle deps.

dzmitry-lahoda: i agree thay may be they overshoot, but will  not touch any default until they are blocker for now.

PR #4164, opened 2023-09-29
test(runtime): Runtime Tests Framework and Ibc Tests 

kollegian:
Required for merge:

Makes review faster:
- [X ] I marked PR by `misc` label if it should not be in release notes
- [X ] I tagged(`@`) or used other form of notification of one person who I think can handle best review of this PR
- Clicking all checkboxes 


dzmitry-lahoda (approved):
if any code was generated i would prefer it is not pushed to repo.

blasrodri:
Regarding IBC tests
- add some multihop transfers (usgin XCM as well)
- add some failing transfers that have an incorrect `memo` field and see that they fail, and the user get the funds back
- are we testing timeouts?

dzmitry-lahoda:
would be awesome to have one simplest test run on ci. i can help to run it in ci. so if it will be unstable , test should be fixed. after we have one e2e simplest. we can add all others must to be added. we need first to get 80% of test value.

kollegian:
> Regarding IBC tests
> 
> * add some multihop transfers (usgin XCM as well)
> * add some failing transfers that have an incorrect `memo` field and see that they fail, and the user get the funds back
> * are we testing timeouts?

@blasrodri multihop will be in a separate test and I was planning to add memo validation there, and timeouts is a good point, will add here.

PR #4182, opened 2023-10-09
feat(cvm): initial eth support (wires until final step transfer shortcut , eth abi message, erc20)

dzmitry-lahoda:
Required for merge:

Makes review faster:
- Clicking all checkboxes 

### DIFF
--- a/code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
+++ b/code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
@@ -57,7 +57,7 @@ pub(crate) fn handle_bridge_forward(
 			.msg
 			.assets
 			.0
-			.get(0)
+			.first()
 			.map(|(_, amount)| (route.on_remote_asset, *amount))
 			.expect("not empty");
 
@@ -154,7 +154,7 @@ pub fn ibc_ics_20_transfer_shortcut(
 						.ics20
 						.expect("ics20")
 						.sender,
-				})
+				});
 			}
 		}
 	}
@@ -182,8 +182,9 @@ pub fn get_this_route(
 
 	let sender_gateway = match this.gateway.expect("we execute here") {
 		GatewayId::CosmWasm { contract, .. } => contract,
-		GatewayId::Evm { .. } =>
-			Err(ContractError::BadlyConfiguredRouteBecauseThisChainCanSendOnlyFromCosmwasm)?,
+		GatewayId::Evm { .. } => {
+			Err(ContractError::BadlyConfiguredRouteBecauseThisChainCanSendOnlyFromCosmwasm)?
+		},
 	};
 
 	let channel = other.connection.ics_20.ok_or(ContractError::ICS20NotFound)?.source;
--- a/code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
+++ b/code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
@@ -22,6 +22,9 @@ use crate::{
 	state,
 };
 
+// 1. if there is know short cat multi hop path it is used up to point in cannot be used anymore (in this case CVM Executor call is propagated)
+// 2. if there is no solved multiprop route, only single hope route checked amid 2 networks and if can do shortcut
+// 3. else full CVM Executor call is propagated
 pub(crate) fn handle_bridge_forward(
 	_: auth::Interpreter,
 	deps: DepsMut,
@@ -35,54 +38,56 @@ pub(crate) fn handle_bridge_forward(
 	));
 
 	ensure_eq!(msg.msg.assets.0.len(), 1, ContractError::ProgramCannotBeHandledByDestination);
-	// algorithm to handle for multihop
-	// 1. recurse on program until can with memo
-	// 2. as soon as see no Spawn/Transfer, stop memo and do Wasm call with remaining Packet
-
 	let (local_asset, amount) = msg.msg.assets.0.get(0).expect("proved above");
 
-	let route: IbcIcs20Route = get_route(deps.storage, msg.to, *local_asset)?;
-	state::tracking::bridge_lock(deps.storage, (msg.clone(), route.clone()))?;
+	let (msg, event) = if let Some(transfer_shortcut) = get_direct() {
+	} else {
+		let route: IbcIcs20Route = get_this_route(deps.storage, msg.to, *local_asset)?;
+		state::tracking::bridge_lock(deps.storage, (msg.clone(), route.clone()))?;
 
-	let asset = msg
-		.msg
-		.assets
-		.0
-		.get(0)
-		.map(|(_, amount)| (route.on_remote_asset, *amount))
-		.expect("not empty");
+		let asset = msg
+			.msg
+			.assets
+			.0
+			.get(0)
+			.map(|(_, amount)| (route.on_remote_asset, *amount))
+			.expect("not empty");
 
-	let packet = XcPacket {
-		interpreter: String::from(info.sender).into_bytes(),
-		user_origin: msg.interpreter_origin.user_origin.clone(),
-		salt: msg.msg.salt,
-		program: msg.msg.program,
-		assets: vec![asset].into(),
+		let packet = XcPacket {
+			interpreter: String::from(info.sender).into_bytes(),
+			user_origin: msg.interpreter_origin.user_origin.clone(),
+			salt: msg.msg.salt,
+			program: msg.msg.program,
+			assets: vec![asset].into(),
+		};
+
+		deps.api.debug(&format!(
+			"cvm::gateway::ibc::ics20 route {}",
+			&serde_json_wasm::to_string(&route)?
+		));
+
+		let mut event = make_event("bridge")
+			.add_attribute("to_network_id", msg.to.to_string())
+			.add_attribute(
+				"assets",
+				serde_json_wasm::to_string(&packet.assets)
+					.map_err(|_| ContractError::FailedToSerialize)?,
+			)
+			.add_attribute(
+				"program",
+				serde_json_wasm::to_string(&packet.program)
+					.map_err(|_| ContractError::FailedToSerialize)?,
+			);
+		if !packet.salt.is_empty() {
+			let salt_attr = Binary::from(packet.salt.as_slice()).to_string();
+			event = event.add_attribute("salt", salt_attr);
+		}
+
+		let coin = Coin::new(amount.0, route.local_native_denom.clone());
+
+		let msg = to_cw_message(deps.as_ref(), deps.api, coin, route, packet, block)?;
+		(msg, event)
 	};
-
-	deps.api
-		.debug(&format!("cvm::gateway::ibc::ics20 route {}", &serde_json_wasm::to_string(&route)?));
-
-	let mut event = make_event("bridge")
-		.add_attribute("to_network_id", msg.to.to_string())
-		.add_attribute(
-			"assets",
-			serde_json_wasm::to_string(&packet.assets)
-				.map_err(|_| ContractError::FailedToSerialize)?,
-		)
-		.add_attribute(
-			"program",
-			serde_json_wasm::to_string(&packet.program)
-				.map_err(|_| ContractError::FailedToSerialize)?,
-		);
-	if !packet.salt.is_empty() {
-		let salt_attr = Binary::from(packet.salt.as_slice()).to_string();
-		event = event.add_attribute("salt", salt_attr);
-	}
-
-	let coin = Coin::new(amount.0, route.local_native_denom.clone());
-
-	let msg = to_cw_message(deps.as_ref(), deps.api, coin, route, packet, block)?;
 
 	Ok(Response::default()
 		.add_event(event)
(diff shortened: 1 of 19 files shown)


code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
```diff
@@ -35,64 +39,130 @@ pub(crate) fn handle_bridge_forward(
...
 
 	Ok(Response::default()
 		.add_event(event)
 		.add_submessage(SubMsg::reply_on_success(msg, ReplyId::TransportSent.into())))
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "std", derive(schemars::JsonSchema))]
+#[serde(rename_all = "snake_case")]
+pub struct Shortcut {
```
kkast, line 122: can  you please add a comment what shortcut is for?

kkast (dismissed):

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
dzmitry-lahoda: Will add. In Short, shortcut is `instead of sending full program to target network send just simple remote execution request. for example many platforms support transfer and exchange, so it is cheaper to send just native instuction than full program.`

docs/docs/technology/cvm.md
```diff
@@ -1,13 +1,14 @@
 # Composable Virtual Machine (CVM)
 
-Previously known as XCVM, the Composable Virtual Machine (CVM) allows is language and execution runtime for cross chain programs. The CVM is a orchestration language on top of standards for cross-chain communication, specifically over [IBC](./composable-ibc.md).
+Previously known as XCVM, the Composable Virtual Machine (CVM) is language and execution runtime for cross chain programs and intents settlement.
```
mina86, line 3: Do we need to keep the XCVM name alive? Perhaps
```suggestion
The Composable Virtual Machine (CVM) is language and execution runtime for cross chain program execution and intents settlement. 
```

code/xcvm/lib/core/src/gateway/config.rs
```diff
@@ -260,23 +281,26 @@ pub struct BridgeAsset {
...
 }
 
 impl AssetReference {
 	pub fn denom(&self) -> String {
 		match self {
-			// AssetReference::ADR001 { sha256 } => ["ibc/",
-			// &sha256.to_string()].concat().to_string(),
 			AssetReference::Native { denom } => denom.clone(),
 			AssetReference::Cw20 { contract } => ["cw20:", contract.as_str()].concat(),
+			AssetReference::Erc20 { contract } => ["erc20:", &contract.to_string()].concat(),
```
mina86, line 303:
```suggestion
			AssetReference::Erc20 { contract } => format!("erc20:{contract}"),
```

code/xcvm/cosmwasm/contracts/gateway/src/error.rs
```diff
@@ -80,6 +80,10 @@ pub enum ContractError {
 	GatewayForNetworkNotFound(NetworkId),
 	#[error("Anonymous calls can do only limitet set of actions")]
 	AnonymousCallsCanDoOnlyLimitedSetOfActions,
+	#[error("Not supported runtime on this network")]
+	NotSupportedRuntimeOnThisNetwork,
```
mina86, line 84:
```suggestion
	#[error("RuntimeUnsupportedOnNetwork")]
	RuntimeUnsupportedOnNetwork,
```

This sounds better IMO.

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
```diff
@@ -35,68 +39,137 @@ pub(crate) fn handle_bridge_forward(
...
+						.expect("ibc")
+						.channels
+						.expect("channels")
+						.ics20
+						.expect("ics20")
+						.sender,
+				})
+			}
+		}
+	}
```
mina86, line 160: It should be possible to flatten a bit:

```suggestion
	if let Some(ics20) = other.connection.ics_20 {
		if other.connection.use_shortcut == Some(true) {
			let sending = this
				.ibc
				.expect("ibc")
				.channels
				.expect("channels")
				.ics20
				.expect("ics20")
				.sender;
			return Ok(IbcIcs20TransferShortcutRoute {
				source: ics20.source,
				denom: asset.local.denom(),
				sending,
			})
		}
	}
```

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
```diff
@@ -35,68 +39,137 @@ pub(crate) fn handle_bridge_forward(
...
+					coin,
+					route,
+					packet,
+					block,
+					contract,
+				)?;
+				(msg, event)
+			},
+			GatewayId::Evm { .. } => Err(ContractError::NotImplemented)?,
+		}
```
mina86, line 111: The body of the if makes sense as a separate function. It would also make the diff easier to read.

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
```diff
@@ -35,68 +39,137 @@ pub(crate) fn handle_bridge_forward(
...
+	pub source: ChannelId,
+	pub denom: String,
+	pub sending: xc_core::transport::ibc::IbcIcs20Sender,
+}
+
+/// this method return route in case program can be just transfer
+pub fn ibc_ics_20_transfer_shortcut(
+	deps: Deps,
+	msg: &xc_core::gateway::BridgeForwardMsg,
+) -> Result<IbcIcs20TransferShortcutRoute, ContractError> {
```
mina86, line 135: This sounds like `-> Result<Option<IbcIcs20TransferShortcutRoute>, ContractError>` to me and `Ok(None)` returned if use_shortcut is false or unset I guess.

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
```diff
@@ -35,68 +39,137 @@ pub(crate) fn handle_bridge_forward(
...
-				.map_err(|_| ContractError::FailedToSerialize)?,
-		);
-	if !packet.salt.is_empty() {
-		let salt_attr = Binary::from(packet.salt.as_slice()).to_string();
-		event = event.add_attribute("salt", salt_attr);
-	}
-
-	let coin = Coin::new(amount.0, route.local_native_denom.clone());
+	let (msg, event) = if let Ok(transfer_shortcut) =
+		ibc_ics_20_transfer_shortcut(deps.as_ref(), &msg)
```
mina86, line 45: Why are we ignoring error? As commented on the function, it seems to me that this function should return `Result<Option<_>, _>` and then this would be `if let Some(shortcut) = ibc_ics_20_transfer_shortcut(deps.as_ref(), &msg)?`.

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
```diff
@@ -35,68 +39,137 @@ pub(crate) fn handle_bridge_forward(
...
+		unimplemented!("add tracking lock for funds return usual cosmos message to transfer as defined in {:?}", transfer_shortcut);
+	} else {
+		let route: IbcIcs20ProgramRoute = get_this_route(deps.storage, msg.to, *local_asset)?;
+		state::tracking::bridge_lock(deps.storage, (msg.clone(), route.clone()))?;
+
+		let asset = msg
+			.msg
+			.assets
+			.0
+			.get(0)
```
mina86, line 60: Perhaps
```suggestion
			.front()
```

mina86 (approved):

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
dzmitry-lahoda: will be in next PR to act according shortcut. now no chains configured to use it.

```
unimplemented!("add tracking lock for funds return usual cosmos message to transfer as defined in {:?}", transfer_shortcut);
```

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
dzmitry-lahoda: agree that part to be refactored, will do when will handle impl of `shortcut` and `GatewayId::Evm { .. } => Err(ContractError::NotImplemented)?,` 

this PR is more of refactoring so that can handle these both places.

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
dzmitry-lahoda: option with/vs result opinionated. 

my opinion if there is result there is no need option, given rust made results as ergonomic as options, do not see reason to use option. 

also soon rust will have keyword generics around result (at least they say), even more less need to use option.

also in case of error like no asset/connection - neither option/result observable as it tries to invoke general program, so it does not improves debugging.

also some languages ditch options, like Roc, as lead to more onsistentencies, Option<T>, is just Result<T,()>

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
dzmitry-lahoda: sure, will try in next PR.  it is possible to flatten even more tbh. 

just not sure what you typed compiles on version of Rust we use.

code/xcvm/lib/core/src/gateway/config.rs
dzmitry-lahoda: i believe more likely owners will intorduce as_str will introduce sooner or later, when more ibc. so will retain with as easier to refactor to as_str

dzmitry-lahoda:
@mina86 will improve that big `use routes to send message` method in next PR where will use all routes.

I will not use result<option< , see my comment.

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
dzmitry-lahoda: there is no front, only first.

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
mina86: If the error this function returns doesn’t matter and is never inspected than this should be `Option<IbcIcs20TransferShortcutRoute>` then.  If the function returns an error and the error is ignored, that raises questions. My understanding of this function is that it can return:
- an error which should be propagated,
- `None` if shortcut is disabled or
- `Some(shortcut)` if shortcut is enabled.

Thus my suggestion is `Result<Option<_>, _>`. This isn’t opinionated. It better documents what the function returns.

mina86 (approved):

kkast (approved):

code/xcvm/cosmwasm/contracts/gateway/src/contract/ibc/ics20.rs
dzmitry-lahoda: > if the function returns an error and the error is ignored, that raises questions. 

why than not raises question ignoring errors inside function if it returns Option? and error is not ignored, it just tells that there is no shortcut for some reason. later I can log it just to see traces from production on routes resolutions.  so i better return error and have logs higher level, than make option and prevent higher level to debug.

dzmitry-lahoda: > an error which should be propagated,
> None if shortcut is disabled or
> Some(shortcut) if shortcut is enabled.
> Thus my suggestion is Result<Option<_>, _>. This isn’t opinionated. It better documents what the function returns.

I understand these words, but thing is i do not see how it helps ergonomics/debugging, it may help some readers, but that is opinionated.

dzmitry-lahoda: If I would return Option<T>, but one of methods called inside returns error. Would `?` make Result in case of error become None? So I do not have to call something `.map_error_to_option()`.

dzmitry-lahoda: I would be happy to document failing to solve shortcut as one of enum variants.

Method tries to get shortcut route, it cannot get, it errors. Why? Because either no connection or shortcut disabled, or other reason. For me option is just another case. Specifically direct IBC ICS20 connection not found. But later for sure will have non direct solutions by unrolling prefix path (need to build index for that so). In this case no direct connection means nothing. So would no direction be error or option? I would prefer error so I can log it for tracing.

mina86: > If I would return Option, but one of methods called inside returns error. Would ? make Result in case of error become None? So I do not have to call something .map_error_to_option().

If one of the calls return an error, why isn’t it propagated? (And for function returning Option, dealing with Result is just `.ok()?`).

> Method tries to get shortcut route, it cannot get, it errors. Why? Because either no connection or shortcut disabled, or other reason. 

But shortcut being disabled is not an error. Unless I misunderstand something, there are three possible options: shortcut, no shortcut or error. If shortcut is enabled and resolved correctly, caller uses the shortcut; if shortcut is disabled, caller uses full path; and if some error while figuring out error happens, caller propagates that error.  This is best modelled by `Result<Option<_>, _>`.  With plain `Result<_>` the caller must now match on the error type to decide whether error needs to be propagated or ‘no shortcut’ path chosen.

PR #4215, opened 2023-10-18
feat(mantis): demo order contract and fixes for no-std support

dzmitry-lahoda:
Required for merge:

Makes review faster:
- Clicking all checkboxes 


code/xcvm/cosmwasm/contracts/order/src/lib.rs
```diff
@@ -1,27 +1,26 @@
...
 
-use crate::error::ContractError;
-
-pub type Amount = Displayed<u128>;
-pub type OrderId = Displayed<u128>;
+// fix core in std in contract
+// use xc_core::{service::dex::ExchangeId, shared::Displayed, NetworkId};
+pub type ExchangeId = Uint128;
+pub type Amount = u128;
+pub type OrderId = Uint128;
```
mina86, line 22: I find this confusing. Why is ExchangeId set to Uint128 while Amount is u128? It seems to me that those aliases should all either be plain Rust type or Uint128 and then in structs we should accordingly either use Uint128 or those aliases.

code/xcvm/cosmwasm/contracts/order/src/lib.rs
```diff
@@ -124,15 +124,18 @@ pub struct SolvedOrder {
 impl SolvedOrder {
 	pub fn new(order: OrderItem, solution: Cow) -> StdResult<Self> {
 		ensure!(
-			order.msg.wants.amount.u128() >= solution.given.0,
-			StdError::GenericErr { msg: "user limit was not satisfied".to_string() }
+			order.msg.wants.amount.u128() <= solution.given.u128(),
```
mina86, line 127: Wouldn’t the following work?

```suggestion
			order.msg.wants.amount <= solution.given,
```

code/xcvm/cosmwasm/contracts/order/src/lib.rs
```diff
@@ -124,15 +124,18 @@ pub struct SolvedOrder {
...
+			StdError::generic_err(format!(
+				"user limit was not satisfied {:?} {:?} ",
+				&order, &solution
```
mina86, line 130:
```suggestion
				"user limit was not satisfied {order:?} {solution:?} ",
```

mina86 (approved):

code/xcvm/cosmwasm/contracts/order/src/lib.rs
dzmitry-lahoda: Sylvia does not compile messages with u128, only Uint128.

code/xcvm/cosmwasm/contracts/order/src/lib.rs
dzmitry-lahoda: some time ago it was not working for sure, as they did no impl traits, did not checked now. would it?

code/xcvm/cosmwasm/contracts/order/src/lib.rs
dzmitry-lahoda: it uses references or clones? would it move objects into log. I am just not sure it compiles

code/xcvm/cosmwasm/contracts/order/src/lib.rs
dzmitry-lahoda: will try with displayed after making it compile in wasm std

mina86: Sure, the messages should use Uint128 but my point is that either consistently use Uint128 in the messages or declare those aliases to be Uint128 and use them. The way here some are Uint128 and some are u128 is confusing and misleading.

code/xcvm/cosmwasm/contracts/order/src/lib.rs
mina86: Clones are never done implicitly and `format` always works on references. If you use `&order` you’re effectively using a double reference. It compiles to the same thing of course but just using it in format message is more concise.

mina86 (dismissed):

PR #4222, opened 2023-10-18
feat(ibc): upgrade composable-ibc to latest

kkast:
Required for merge:

Makes review faster:
- Clicking all checkboxes 


dzmitry-lahoda (changes requested):
does transfer works?

code/parachain/runtime/common/src/ibc.rs
```diff
@@ -31,3 +38,127 @@ where
 }
 
 pub type MinimumConnectionDelaySeconds = ConstU64<10>;
+
+#[derive(Deserialize, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum MemoMiddlewareNamespaceChain {
+	Forward { next: Option<Box<Self>> },
+	Wasm { next: Option<Box<Self>> },
```
dzmitry-lahoda, line 46: prevents forwarding evm memo

dzmitry-lahoda: any non recognizable attribute is ok.

dzmitry-lahoda: memo can be empty json and other fields. and even  on json

https://github.com/cosmos/ibc/tree/main/spec/app/ics-020-fungible-token-transfer#using-the-memo-field

and still valid.

dzmitry-lahoda:
transfer should work as before this pr.
memo, we either fix my comments, or any cvm shortcut integration is blocked.

dzmitry-lahoda:
do not vidate  memo like that.

code/parachain/runtime/common/src/ibc.rs
vmarkushin: In memo validation function just return if the memo is empty

vmarkushin: If you need to forward evm memo, just add it here

dzmitry-lahoda: i do no see test proving that non optional enum will return will no not error on empy json. also there is parser code on ibc repo which may fail. and also whole cvm, cw works with memo inside xc-core, which may be used by pallet xcm ibc hops. so this memo must be removed and be just string. anyway, cw and cvm was cancelled or on hold. i second time repeated my suggestion. so if not listening, pleae fill free whatever way it will work according memo spec in ics20 at least. which by spec supports any attributes, i mentioned it 3 times already

dzmitry-lahoda: i do no see test proving that non optional enum will return will no not error on empy json. also there is parser code on ibc repo which may fail. and also whole cvm, cw works with memo inside xc-core, which may be used by pallet xcm ibc hops. so this memo must be removed and be just string. anyway, cw and cvm was cancelled or on hold. i second time repeated my suggestion. so if not listening, pleae fill free whatever way it will work according memo spec in ics20 at least. which by spec supports any attributes, i mentioned it 3 times already

blasrodri:
@dzmitry-lahoda can you please clarify your concerns, and provide a solution in pseudocode? I'm not clear on what your suggestion fully refers to.

dzmitry-lahoda (dismissed):

dzmitry-lahoda:
I need know how to add new field to memo. Specifically how to add `evm` call for hook. What would be my steps? 
I see 6 places:
1. centaur-private memo parser
2. composable-ibc memo parser
3. composable memo parser as done by Kostya in PR
4. composable-private memo parser as done by Kostya in PR
5. composable cvm and mulithop memo parser/generator as used in these
6. composable-private 

After I updated all 6 repositories. What order to update pallet-ibc/subxt clients/hyperpace refernces to to make it integrated.

I do not know how.

blasrodri (dismissed):

RustNinja (approved):

PR #4240, opened 2023-10-21
add mock runtime for xcm-helper pallet. WIP.

RustNinja:
Introduce mock runtime for test in xcm-helper pallet.
Work in progress.


dzmitry-lahoda (commented):
hm, tbh you may consider using real chain runtime with simulator

PR #4257, opened 2023-10-27
feat(cvm): intercept(percentage amounts and slippage) usage in exchange

dzmitry-lahoda:
So can use any amount, max from virtual wallet will be taken. No slippage support for now.

Required for merge:

Makes review faster:
- Clicking all checkboxes 


code/xcvm/cosmwasm/contracts/interpreter/src/contract.rs
```diff
@@ -182,20 +182,34 @@ fn interpret_exchange(
...
-		amount: want.0[0].1.amount.intercept.to_string(),
-	};
+
+	if want.1.amount.is_both() {
+		return Err(ContractError::CannotDefineBothSlippageAndLimitAtSameTime)
+	}
+
+	if want.1.amount.is_ratio() {
+		return Err(ContractError::ExchangeDoesNotSupportSlippage)
+	}
```
mina86, line 208:
```suggestion
	if want.1.amount.is_ratio() {
		return Err(if want.1.amount.is_absolute() {
			ContractError::CannotDefineBothSlippageAndLimitAtSameTime
		} else {
			ContractError::ExchangeDoesNotSupportSlippage
		})
	}
```

And then get rid of `is_both` method because it has a *terrible* name.

code/xcvm/lib/core/src/asset.rs
```diff
@@ -165,6 +165,10 @@ impl Amount {
 		self.slope.0 == 0
 	}
 
+	pub const fn is_both(&self) -> bool {
```
mina86, line 168: This is a terrible name for the method.

dzmitry-lahoda: like is complex, is dual, is function, is linear, is absolute and ratio?

mina86: Another option is to just get rid of it and explicitly test both. I think this type is convoluted enough that maybe it’s not worth making convenience methods whose meaning isn’t immediately obvious.

PR #4269, opened 2023-11-01
feat(cvm): using shared code in wasm std target of order contract and sample solutio to cvm traverse

dzmitry-lahoda:
need https://github.com/CosmWasm/cosmwasm/pull/1931 . because we depend on git commit of no_std which cannot be ^ compiled with version upper of std exporting functions (until patching std not to export, hacks)

Required for merge:

Makes review faster:
- Clicking all checkboxes 


code/xcvm/cosmwasm/contracts/order/src/lib.rs
```diff
@@ -263,17 +266,70 @@ impl OrderContract<'_> {
...
+					assets: <_>::default(), // map spawn.carry to CVM assets
+					program: XcProgram {
+						tag: b"solution".to_vec(),
+						instructions: vec![], // we really just do final transfer
+					},
+				}
+			};
+			result.push(spawn);
+		}
+		result
```
mina86, line 324: The Spawn instruction differ in only program so the `if` statement can
be shorter with less duplication.  Also, this could be collect:

```suggestion
		spawns.into_iter().map(|spawn| {
			let program = if let Some(execute) = spawn.execute {
				Self::traverse_route(execute)
			} else {
				XcProgram {
					tag: b"solution".to_vec(),
					instructions: vec![], // we really just do final transfer
				}
			};
			XcInstruction::Spawn {
				network_id: spawn.to_chain.into(),
				salt: b"solution".to_vec(),
				assets: <_>::default(), // map spawn.carry to CVM assets
				program,
			}
		}).collect()
```

mina86 (approved):

PR #4445, opened 2024-02-07
feat(defaultfee): Specify default fee for network

kkast:
added 2 fields to config: 

NativeAssetId to allow setting native token via set_payment_asset
DefaultFeeAsset which is used when PaymentAssets in asset-tx-payment is None

This way people who already set Dot to be fee asset will be able to use it, and new users wont need to set Dot to be default asset

Required for merge:
- Other rules GitHub shows you, or can be read in [configuration](github.com/ComposableFi/env/terraform/github.com/branches.tf) 

Makes review faster:
- Clicking all checkboxes 

### DIFF
--- a/code/parachain/runtime/composable/src/fees.rs
+++ b/code/parachain/runtime/composable/src/fees.rs
@@ -67,6 +67,8 @@ impl asset_tx_payment::HandleCredit<AccountId, Tokens> for TransferToTreasuryOrD
 
 parameter_types! {
 	pub AssetConfigHoldIdentifier: TemporalHoldIdentifier = ();
+	pub NativeAssetId: CurrencyId = CurrencyId::PICA;
+	pub DefaultFeeAsset: CurrencyId = CurrencyId::DOT;
 }
 
 impl asset_tx_payment::Config for Runtime {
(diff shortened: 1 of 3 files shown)


dzmitry-lahoda (approved):

code/parachain/runtime/composable/src/fees.rs
```diff
@@ -67,6 +67,8 @@ impl asset_tx_payment::HandleCredit<AccountId, Tokens> for TransferToTreasuryOrD
 
 parameter_types! {
 	pub AssetConfigHoldIdentifier: TemporalHoldIdentifier = ();
+	pub NativeAssetId: CurrencyId = CurrencyId::PICA;
+	pub DefaultFeeAsset: CurrencyId = CurrencyId::DOT;
```
dzmitry-lahoda, line 71: just to be sure DOT id is corrent, it is dot on Composable and will work right.

kkast: you are correct, this isnt composable Dot

dzmitry-lahoda (approved):

PR #4450, opened 2024-02-15
feat: Add HashedDescription XCM converter and remove TinkernetMultisig configs

arrudagates:
This PR adds the common XCM location converter HashedDescription, which is used across the ecosystem, to Picasso.
Since the current Polkadot SDK version in use does not contain these configs, they are temporarily pulled locally into `parachains/runtime/common` and added to the Picasso runtime from the `common` local crate.

For reference, chains using HashedDescription include the [relay chain](https://github.com/polkadot-fellows/runtimes/blob/main/relay/kusama/src/xcm_config.rs#L76) and system parachains like the [Asset Hub](https://github.com/polkadot-fellows/runtimes/blob/main/system-parachains/asset-hubs/asset-hub-kusama/src/xcm_config.rs#L93).

Another chain also using HashedDescription but pulling it from a later Polkadot SDK version is [Astar](https://github.com/AstarNetwork/Astar/blob/master/primitives/src/xcm/mod.rs#L277-L390)

Required for merge:
- Other rules GitHub shows you, or can be read in [configuration](github.com/ComposableFi/env/terraform/github.com/branches.tf) 

Makes review faster:
- Clicking all checkboxes 

### DIFF
--- a/code/parachain/runtime/picasso/src/xcmp.rs
+++ b/code/parachain/runtime/picasso/src/xcmp.rs
@@ -93,11 +93,7 @@ pub type Barrier = (
 	AllowSubscriptionsFrom<ParentOrSiblings>,
 	AllowTopLevelPaidExecutionFrom<Everything>,
 	TakeWeightCredit,
-	WithComputedOrigin<
-		AllowTopLevelPaidExecutionFrom<orml_xcm_builder_kusama::TinkernetMultisigMultiLocation>,
-		UniversalLocation,
-		ConstU32<8>,
-	>,
+	WithComputedOrigin<AllowTopLevelPaidExecutionFrom<Everything>, UniversalLocation, ConstU32<8>>,
 );
 
 pub type LocalOriginToLocation = SignedToAccountId32<RuntimeOrigin, AccountId, RelayNetwork>;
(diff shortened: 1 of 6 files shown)


dzmitry-lahoda:
@arrudagates thank you

code/parachain/runtime/picasso/src/xcmp.rs
```diff
@@ -93,11 +93,7 @@ pub type Barrier = (
 	AllowSubscriptionsFrom<ParentOrSiblings>,
 	AllowTopLevelPaidExecutionFrom<Everything>,
 	TakeWeightCredit,
-	WithComputedOrigin<
-		AllowTopLevelPaidExecutionFrom<orml_xcm_builder_kusama::TinkernetMultisigMultiLocation>,
-		UniversalLocation,
-		ConstU32<8>,
-	>,
+	WithComputedOrigin<AllowTopLevelPaidExecutionFrom<Everything>, UniversalLocation, ConstU32<8>>,
```
RustNinja, line 96: I concerned that prev it allowed the TopLevelPaidExecutionFrom only from TinkernetMultisigMultiLocation
but now you replaced to Everything,
Seems a big changes..

arrudagates: I changed to `Everything` because it is considered safe to allow paid execution, as it requires the execution to be paid for in XCM fees + transaction fees. The [relay chain](https://github.com/polkadot-fellows/runtimes/blob/main/relay/kusama/src/xcm_config.rs#L167) and the [system chains](https://github.com/polkadot-fellows/runtimes/blob/main/system-parachains/asset-hubs/asset-hub-kusama/src/xcm_config.rs#L452) both use the same barrier.

Would you prefer I keep this constrained to Tinkernet origins?

arrudagates:
@RustNinja Any blockers on this?

dzmitry-lahoda (approved):

